PR #36, opened 2025-08-08
Modify validate and formatting workspace URL in login logic

mingyu-00:
Now alpacon-cli supports two workspace url patterns(example):
1. `alpacon login https://alpacax.dev.alpacon.io`
2. `alpacon login https://dev.alpacon.io/alpacax`

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"github.com/alpacax/alpacon-cli/api/auth"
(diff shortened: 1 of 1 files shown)


mingyu-00:
> Does this also work for just `alpacon.io`? Because the application url `alpacon.io/alpacax` will be mapped to `alpacax.ap1.alpacon.io`. Maybe it need to be statically mapped in this case.

That's a great question. The login works regardless of the region, whether it's dev, ap1, or any other. For example, to log in to the ap1 environment, both of the following methods are possible:

1.  `alpacon login https://alpacax.ap1.alpacon.io`
2.  `alpacon login https://ap1.alpacon.io/alpacax`

PR #49, opened 2025-09-08
48 command warning for websh

mingyu-00:
We will prevent unintended consequences that can occur when the `websh` command is executed without using quotes.

For example, when running `alpacon websh test ls -al && mkdir test_dir`, the command after `&&` is executed on the local machine, not the remote server.

Following the SSH command convention, we will not block unquoted command execution. Instead, we will add a warning message and a confirmation prompt (y/n) to ask the user if they want to proceed with the command.

When using `alpacon-cli` in a CI/CD environment, it's not possible to provide a `y/n` response. Therefore, we'll add a **`-y` option** to skip the warning and proceed with execution.

### DIFF
--- a/cmd/websh/websh.go
+++ b/cmd/websh/websh.go
@@ -149,7 +149,7 @@ var WebshCmd = &cobra.Command{
 		} else if len(commandArgs) > 0 {
 			if len(commandArgs) > 1 && !skipQuestion {
 				utils.CliWarning("Command without quotes may cause unexpected behavior. Consider wrapping the command in quotes.")
-				confirm := utils.ConfirmModal("Do you want to continue executing the command?")
+				confirm := utils.CommandConfirmModal()
 				if !confirm {
 					return
 				}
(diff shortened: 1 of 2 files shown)


eunyoung14:
I think prompting for y/n in an interactive terminal to prevent unintended command execution is a good idea. However, I’m not sure if adding a `-y` flag intuitively maps to that functionality. Wouldn’t it make more sense to simply proceed without prompting when the terminal is detected as non-interactive? I believe there should be a way to check whether the terminal is interactive.

mingyu-00:
> I think prompting for y/n in an interactive terminal to prevent unintended command execution is a good idea. However, I’m not sure if adding a `-y` flag intuitively maps to that functionality. Wouldn’t it make more sense to simply proceed without prompting when the terminal is detected as non-interactive? I believe there should be a way to check whether the terminal is interactive.

Thanks for the great feedback! I've taken your advice and decided to stop using the `-y` flag. Instead, I've modified the code to determine whether the shell is interactive or not internally, and based on that, decide whether to display a command confirmation question.

PR #51, opened 2025-09-11
Resolve issue that logout failure in CLI

mingyu-00:
There was a past issue where logout was failing in our Auth0 environment. I've now resolved this by sending an API request to a blacklist endpoint to invalidate the access token and also revoking the refresh token, ensuring a secure logout.

### DIFF
--- a/api/auth/auth.go
+++ b/api/auth/auth.go
@@ -18,11 +18,10 @@ import (
 )
 
 const (
-	loginURL     = "/api/auth/login/"
-	logoutURL    = "/api/auth/logout/"
-	blacklistURL = "/api/auth0/token/blacklist/"
-	tokenURL     = "/api/auth/tokens/"
-	statusURL    = "/api/status/"
+	loginURL  = "/api/auth/login/"
+	logoutURL = "/api/auth/logout/"
+	tokenURL  = "/api/auth/tokens/"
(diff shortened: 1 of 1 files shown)


api/auth/auth.go
```diff
@@ -184,11 +186,29 @@ func DeleteAPIToken(ac *client.AlpaconClient, tokenID string) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to revoke token: %v", err)
+		}
+	}
+	err := config.DeleteConfig()
 	if err != nil {
 		return err
 	}
+	return nil
+}
```
eunyoung14, line 205: This logic should be implemented in `auth0.go` for modularity. The function name can be just `Logout` because the file name is already designating which auth provider is in use.

api/auth/auth.go
```diff
@@ -184,11 +186,29 @@ func DeleteAPIToken(ac *client.AlpaconClient, tokenID string) error {
...
+
+func LogOutWithOnPremise(ac *client.AlpaconClient) error {
```
eunyoung14, line 207: I also would keep the prior funciton name as it describes how the logout logic works for the local auth. (We do not call it on-premise login.) Or it just could be `Logout`.

eunyoung14 (approved):

PR #60, opened 2025-12-09
Add initializing username logic when required username error occured

mingyu-00:
When `user_username_required` error code occured in `cp`, `websh`, `exec` command, add logic for setting username dynamically.

### DIFF
--- a/utils/error_handler.go
+++ b/utils/error_handler.go
@@ -5,6 +5,11 @@ import (
 	"time"
 )
 
+const (
+	maxRetryDuration = 1 * time.Minute
+	retryInterval    = 5 * time.Second
+)
+
 // ErrorHandlerCallbacks defines callback functions for handling different error types
 type ErrorHandlerCallbacks struct {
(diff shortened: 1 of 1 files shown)


utils/error.go
```diff
@@ -7,6 +7,7 @@ import (
 
 const (
 	CodeAuthMFARequired = "auth_mfa_required"
+	UsernameRequired    = "user_username_required"
```
geunwoonoh: How about prefixing variable names with $\text{Code}$ for consistency?

utils/error_handler.go
```diff
@@ -35,10 +35,18 @@ func HandleCommonErrors(err error, serverName string, callbacks ErrorHandlerCall
 			CliErrorWithExit("MFA authentication failed: %s", err)
 		}
 
+		const maxRetryDuration = 1 * time.Minute
```
geunwoonoh: For better readability, please define the constants separately outside of the function.

PR #67, opened 2026-01-08
feat(ftp): add spinner and transfer status polling for cp command

mingyu-00:
## Summary

- Add spinner animation to provide visual feedback during file upload/download operations
- Implement transfer status polling to track upload/download completion
- Refactor spinner cleanup using `defer` for single-point resource management
- Handle polling timeout gracefully instead of treating it as an error

### New Features
- `utils/spinner.go`: New spinner utility with animated dots and frames
- `api/ftp/types.go`: Add `TransferStatusResponse` type for status polling

### Improvements
- Wrap transfer operations in anonymous functions with `defer spinner.Stop()` to ensure cleanup at a single point
- On polling timeout, return success with informational message instead of error (transfer may still complete on server)

## Test plan

- [x] Test file upload with `alpacon cp local_file server:/path`
- [x] Test folder upload with `alpacon cp -r local_folder server:/path`
- [x] Test file download with `alpacon cp server:/path/file local_path`
- [x] Test folder download with `alpacon cp -r server:/path/folder local_path`
- [x] Verify spinner displays and stops correctly on success/failure

### DIFF
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"mime/multipart"
 	"net/http"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


cmd/ftp/cp.go
geunwoonoh: Is there a specific reason why `utils.CliErrorWithExit()` and `utils.CliError()` are used interchangeably here? 
It seems like only `utils.CliErrorWithExit()` is used everywhere else

mingyu-00: `utils.CliError()` is used in case that don't want to perform exit logic with `os.Exit(1)` and it used now in `cp.go`

geunwoonoh: I see. Thanks for the explanation!

geunwoonoh (approved):

PR #131, opened 2026-03-20
feat(cp): add bulk transfer APIs and single/bulk upload routing

### LINKED ISSUE #127: fix(cp/webftp): improve alpacon cp
## Background

alpacon cp currently sends individual requests for each file when transferring multiple files, and provides no way to control overwrite behavior.

---

## Scope

Bulk transfer improvement

- Upload: replace the current per-file request loop with a bulk API call to handle multiple files in a single request
- Download: replace the current per-file request loop with a bulk API call to receive multiple files as a single zip archive

Add --no-overwrite flag

- The allow_overwrite field is currently hardcoded to true, making it impossible to prevent overwriting existing files
- When the flag is not specified, behavior remains true for backward compatibility

---

## Related files (CLI)

- cmd/ftp/
- api/ftp/

jisung-02:
Closes #127

## Summary

Improve `cp` file transfer by routing single files through the single API and multiple files through the bulk API for both uploads and downloads. Add `--no-overwrite` flag and fix trailing slash handling for folder uploads.

### Upload
- **Before (main):** Always called single API (`/api/webftp/uploads/`) in a for loop regardless of file count. `AllowOverwrite` was hardcoded `"true"`, client generated UUID (`github.com/google/uuid` dependency)
- **After:** 1 file → single API, 2+ files → bulk API (`/api/webftp/uploads/bulk/`). `UploadRequest.AllowOverwrite`/`AllowUnzip` changed from `string` to `bool` (matches server serializer). Removed `ID` field (server generates it). Removed `uuid` dependency

### Download
- **Before (main):** Called single API (`/api/webftp/downloads/`) in a for loop
- **After:** 1 file → single API, 2+ files → bulk API (`/api/webftp/downloads/bulk/`) — receives a single zip archive and extracts locally. Added `BulkDownloadResponse` type (server returns `path` as `[]string`)

### Other
- Added `--no-overwrite` (`-n`) flag — prevents overwriting existing files on the server (default: allow overwrite)
- Auto-append trailing slash to `remotePath` for folder uploads so the server recognizes it as a directory
- Changed `downloadSingleFile` from `utils.CliErrorWithExit` to `fmt.Errorf` — returns errors to caller instead of exiting directly
- Added 535 lines of webftp tests (12 test cases)

## etc

- Bulk download requires `zip` binary installed on the target server (alpamon zips files server-side before transfer). There is a dependency on the alpacon-server side.
- Some alpacon-server changes are required for correct operation.

## Checklist

- [x] Error handling is complete
- [x] Tests include edge cases
- [x] No hardcoded secrets or credentials
- [x] Self-reviewed the code

## Test plan

- [x] `go test -race ./api/ftp/...` — 12 tests pass
- [x] `go build` — build succeeds
- [x] Single file upload (`alpacon cp test.md server:/path/`)
- [x] Multi-file bulk upload (`alpacon cp test.md test2.md server:/path/`)
- [x] Single folder upload (`alpacon cp -r testdir server:/path`)
- [x] Single file download — current dir, rename, `/tmp/`, `/tmp` all work
- [x] Bulk download — E2E not testable (server missing `zip`), unit tests pass
- [x] Recursive folder download (`-r`) — server-side unzip issue, code unchanged from main

Closes #127

### DIFF
--- a/api/iam/iam.go
+++ b/api/iam/iam.go
@@ -15,6 +15,7 @@ const (
 	groupURL      = "/api/iam/groups/"
 	membershipURL = "/api/iam/memberships/"
 	usernameURL   = "/api/iam/username/"
+	inviteUserURL = "/api/workspaces/users/invite/"
 )
 
 func GetUserList(ac *client.AlpaconClient) ([]UserAttributes, error) {
@@ -76,6 +77,11 @@ func GetGroupDetail(ac *client.AlpaconClient, groupId string) ([]byte, error) {
 	}
 
 	return responseBody, nil
+}
+
+func InviteUser(ac *client.AlpaconClient, request UserInviteRequest) error {
+	_, err := ac.SendPostRequest(inviteUserURL, request)
+	return err
 }
 
 func CreateUser(ac *client.AlpaconClient, userRequest UserCreateRequest) error {
--- a/api/iam/iam_test.go
+++ b/api/iam/iam_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/alpacax/alpacon-cli/api"
 	"github.com/alpacax/alpacon-cli/client"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestGetUserList_Pagination(t *testing.T) {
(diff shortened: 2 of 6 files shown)


jisung-02:
All review items have been checked. 9 out of 10 are already resolved, and 1 is intentionally kept as-is.

1. Temp zip not cleaned up on Unzip failure — Resolved
   Added a deferred DeleteFile call so the temp zip is removed regardless of whether Unzip succeeds or fails.

2. Comment says "1 minute" but actual timeout is 30s — Resolved
   Replaced the fixed maxPollRetries constant with a dynamic timeout calculated from file count and size (calcPollTimeout), which also removed the outdated comment.

3. strings.Fields breaks paths with spaces — Resolved
   DownloadFile now accepts []string directly instead of parsing a single string with strings.Fields, so paths containing spaces are handled correctly.

4. All file contents loaded into memory (OOM risk) — Resolved
   This was a pre-existing issue that was addressed in this branch. Replaced ReadFileFromPath (which loaded entire files as []byte) with os.Open and io.Reader streaming, eliminating OOM risk for large files.

5. fetchFromURL retries 100 times on 4xx — Resolved
   4xx client error responses now fail immediately without retrying. Only 5xx and network errors are retried.

6. ExpiresAt type inconsistency (string vs time.Time) — Resolved
   Unified ExpiresAt to string across both UploadResponse and DownloadResponse, and removed the unused time import.

7. Partial failure in executeBulkUpload abandons remaining files — Resolved
   Changed the polling phase from early return to failure collection. All files are polled to completion and any failures are reported together at the end.

8. Missing test coverage (mismatch, timeout, 4xx) — Resolved
   Added TestExecuteBulkUpload_MismatchedResponseCount, TestPollTransferStatus_Timeout, and TestFetchFromURL_ClientErrorNoRetry.

9. uploadUrl naming — Resolved
   Renamed to uploadURL to follow Go naming conventions for acronyms.

10. Unsafe append in validatePaths — Resolved
    Replaced the in-place append with make and explicit appends to avoid mutating the underlying sources slice.

PR #150, opened 2026-04-16
feat(cmd/server): add server token create/ls/delete commands

### LINKED ISSUE #146: feat(server): extend server token management commands
## Background

Installing alpamon on a server currently requires a human to log in, manually copy a token, and paste it into the installation command. This blocks fully automated provisioning workflows (Terraform, Ansible, CloudInit, CI/CD pipelines).

The server supports **Registration Tokens** — a dedicated token type scoped to \`server:register\` only. A registration token pre-configures which groups and tags are automatically applied when alpamon uses it to self-register. Once created, the token key can be embedded in provisioning scripts so servers register themselves without any manual step.

Registration tokens are distinct from regular API tokens:
- They are scoped to \`server:register\` only (not \`*\`)
- They carry \`allowed_groups\` — the groups a server will be assigned to on registration
- They are managed via \`/api/servers/registration-tokens/\`

## What to implement

Extend the \`alpacon server token\` subcommand group (already has \`regenerate\`):

### \`server token create\`

\`\`\`
alpacon server token create --name prod-token
alpacon server token create --name prod-token --groups dev,ops
alpacon server token create --name prod-token --expires-in-days 90
\`\`\`

- Prompts interactively if \`--name\` is not provided
- \`--groups\`: comma-separated group names or UUIDs to assign on registration
- \`--expires-in-days\`: token expiration (optional, no expiry if omitted)
- Prints the token key on success — **displayed once only**, with a warning

### \`server token ls\`

\`\`\`
alpacon server token ls
alpacon server token list
\`\`\`

Displays a table with: token name, allowed groups, expiry, enabled status.

### \`server token delete\`

\`\`\`
alpacon server token delete my-token
alpacon server token rm my-token
alpacon server token delete my-token -y
\`\`\`

- Accepts token name or UUID
- Prompts for confirmation unless \`-y\` is given

## Typical automation flow

\`\`\`
# 1. Create a registration token
alpacon server token create \
  --name k8s-nodes \
  --groups infra \
  --expires-in-days 365

# 2. Embed the printed key in cloud-init / Terraform / Ansible
# 3. alpamon self-registers on boot using the key — no human interaction needed
\`\`\`

## API reference

| Method | Endpoint | Description |
|--------|----------|-------------|
| \`POST\` | \`/api/servers/registration-tokens/\` | Create registration token |
| \`GET\` | \`/api/servers/registration-tokens/\` | List registration tokens |
| \`DELETE\` | \`/api/servers/registration-tokens/{id}/\` | Delete registration token |

## Acceptance criteria

- [ ] \`server token create --name x --groups g1,g2\` creates a token and prints the key once
- [ ] \`server token create\` interactive mode prompts for name and groups
- [ ] \`server token ls\` shows all registration tokens in a table
- [ ] \`server token delete NAME\` resolves name to UUID and deletes with confirmation
- [ ] \`server token delete NAME -y\` skips confirmation

jisung-02:
## Summary

Installing alpamon previously required manually copying a registration token from
the web UI, making fully automated provisioning with Terraform/Ansible/CloudInit
impossible. This PR adds the `alpacon server token` subcommand group—create, ls,
and delete—so tokens can be managed entirely from the CLI.

Closes #146

## Changes

- Add `server token create` — generates a registration token; key is printed once and cannot be retrieved again; supports `--name`, `--groups`, `--expires-in-days`, `--output json`
- Add `server token ls` — lists all tokens with group UUIDs resolved to names; supports `--output json`
- Add `server token delete TOKEN` — deletes a token by name; `-y` skips confirmation
- Extend `RegistrationToken*` structs with `ExpiresAt`; add `RegistrationTokenAttributes` for table projection
- Add `DeleteRegistrationToken`, `GetRegistrationTokenAttributes`, `buildGroupUUIDToNameMap` to `api/server`
- Add 8 API-layer tests and 4 unit tests for `resolveGroupIDs`
- Fix: reject negative `--expires-in-days` values instead of silently creating non-expiring tokens
- Refactor: move `groupSummary` type before functions (Go declaration order); extract magic string to `iamGroupURL` constant

### Universal
- [ ] Self-reviewed the code
- [ ] No hardcoded secrets or credentials
- [ ] Updated relevant documentation

### Go
- [ ] Error handling is complete
- [ ] Tests include edge cases
- [ ] `go mod tidy` executed

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -174,6 +174,7 @@ type groupSummary struct {
 // buildGroupUUIDToNameMap fetches all IAM groups and returns a UUID→name map.
 // On failure it returns an empty map so callers never block on a group-lookup error.
 func buildGroupUUIDToNameMap(ac *client.AlpaconClient) map[string]string {
+	// "/api/iam/groups/" is intentionally duplicated from api/iam to avoid an import cycle.
 	groups, err := api.FetchAllPages[groupSummary](ac, "/api/iam/groups/", nil)
 	if err != nil {
 		return map[string]string{}
@@ -204,7 +205,17 @@ func GetRegistrationTokenAttributes(ac *client.AlpaconClient) ([]RegistrationTok
 		return nil, err
 	}
 
-	groupMap := buildGroupUUIDToNameMap(ac)
+	needsGroups := false
+	for _, t := range tokens {
+		if len(t.AllowedGroups) > 0 {
+			needsGroups = true
+			break
+		}
+	}
+	groupMap := map[string]string{}
+	if needsGroups {
+		groupMap = buildGroupUUIDToNameMap(ac)
+	}
 
 	var out []RegistrationTokenAttributes
 	for _, t := range tokens {
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/alpacax/alpacon-cli/api"
 	"github.com/alpacax/alpacon-cli/client"
@@ -161,6 +162,72 @@ func GetRegistrationTokenByName(ac *client.AlpaconClient, name string) (Registra
 
 func ListRegistrationTokens(ac *client.AlpaconClient) ([]RegistrationTokenDetails, error) {
 	return api.FetchAllPages[RegistrationTokenDetails](ac, registrationTokenURL, nil)
+}
+
+// groupSummary is a minimal projection used to build a UUID→name map for group display.
+// Keeping it local avoids importing api/iam and pulling in its heavier GroupResponse type.
+type groupSummary struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+// buildGroupUUIDToNameMap fetches all IAM groups and returns a UUID→name map.
+// On failure it returns an empty map so callers never block on a group-lookup error.
+func buildGroupUUIDToNameMap(ac *client.AlpaconClient) map[string]string {
+	groups, err := api.FetchAllPages[groupSummary](ac, "/api/iam/groups/", nil)
+	if err != nil {
+		return map[string]string{}
+	}
+	m := make(map[string]string, len(groups))
+	for _, g := range groups {
+		m[g.ID] = g.Name
+	}
+	return m
+}
+
+// DeleteRegistrationToken resolves a token name to its ID and sends a DELETE request.
+func DeleteRegistrationToken(ac *client.AlpaconClient, tokenName string) error {
+	token, err := GetRegistrationTokenByName(ac, tokenName)
+	if err != nil {
+		return err
+	}
+	_, err = ac.SendDeleteRequest(utils.BuildURL(registrationTokenURL, token.ID, nil))
+	return err
+}
+
+// GetRegistrationTokenAttributes returns all registration tokens projected for table/JSON display.
+// Group UUIDs are resolved to group names using a single batched lookup; on lookup failure,
+// UUIDs are shown as-is and the overall list is not blocked.
+func GetRegistrationTokenAttributes(ac *client.AlpaconClient) ([]RegistrationTokenAttributes, error) {
+	tokens, err := ListRegistrationTokens(ac)
+	if err != nil {
+		return nil, err
+	}
+
+	groupMap := buildGroupUUIDToNameMap(ac)
+
+	var out []RegistrationTokenAttributes
+	for _, t := range tokens {
+		names := make([]string, 0, len(t.AllowedGroups))
+		for _, id := range t.AllowedGroups {
+			if n, ok := groupMap[id]; ok {
+				names = append(names, n)
+			} else {
+				names = append(names, id)
+			}
+		}
+		expires := "never"
+		if t.ExpiresAt != nil {
+			expires = *t.ExpiresAt
+		}
+		out = append(out, RegistrationTokenAttributes{
+			Name:          t.Name,
+			AllowedGroups: strings.Join(names, ", "),
+			ExpiresAt:     expires,
+			Enabled:       t.Enabled,
+		})
+	}
+	return out, nil
 }
 
 func GetRegistrationGuideJSON(ac *client.AlpaconClient, platform, serverName, tokenID string) (RegistrationMethodGuideJsonResponse, error) {
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -15,7 +15,8 @@ const (
 	serverURL            = "/api/servers/servers/"
 	registrationTokenURL = "/api/servers/registration-tokens/"
 	registrationGuideURL = "/api/servers/registration-methods/token-install/guide/"
-	// iamGroupURL is intentionally duplicated from api/iam to avoid an import cycle.
+	// iamGroupURL is duplicated from api/iam so this package can build a lean
+	// UUID→name projection without pulling in the full iam.GroupResponse type.
 	iamGroupURL = "/api/iam/groups/"
 )
 
@@ -174,10 +175,12 @@ func ListRegistrationTokens(ac *client.AlpaconClient) ([]RegistrationTokenDetail
 }
 
 // buildGroupUUIDToNameMap fetches all IAM groups and returns a UUID→name map.
-// On failure it returns an empty map so callers never block on a group-lookup error.
+// On failure it emits a one-time warning and returns an empty map so callers
+// never block on a group-lookup error; the list still renders with raw UUIDs.
 func buildGroupUUIDToNameMap(ac *client.AlpaconClient) map[string]string {
 	groups, err := api.FetchAllPages[groupSummary](ac, iamGroupURL, nil)
 	if err != nil {
+		utils.CliWarning("Could not resolve group names; showing UUIDs instead: %s", err)
 		return map[string]string{}
 	}
 	m := make(map[string]string, len(groups))
@@ -218,7 +221,7 @@ func GetRegistrationTokenAttributes(ac *client.AlpaconClient) ([]RegistrationTok
 		groupMap = buildGroupUUIDToNameMap(ac)
 	}
 
-	var out []RegistrationTokenAttributes
+	out := make([]RegistrationTokenAttributes, 0, len(tokens))
 	for _, t := range tokens {
 		names := make([]string, 0, len(t.AllowedGroups))
 		for _, id := range t.AllowedGroups {
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -15,10 +15,19 @@ const (
 	serverURL            = "/api/servers/servers/"
 	registrationTokenURL = "/api/servers/registration-tokens/"
 	registrationGuideURL = "/api/servers/registration-methods/token-install/guide/"
+	// iamGroupURL is intentionally duplicated from api/iam to avoid an import cycle.
+	iamGroupURL = "/api/iam/groups/"
 )
 
 // ErrRegistrationTokenNotFound is returned when no registration token matches the given name.
 var ErrRegistrationTokenNotFound = errors.New("no registration token found with the given name")
+
+// groupSummary is a minimal projection used to build a UUID→name map for group display.
+// Keeping it local avoids importing api/iam and pulling in its heavier GroupResponse type.
+type groupSummary struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
 
 func GetServerList(ac *client.AlpaconClient) ([]ServerAttributes, error) {
 	servers, err := api.FetchAllPages[ServerDetails](ac, serverURL, nil)
@@ -164,18 +173,10 @@ func ListRegistrationTokens(ac *client.AlpaconClient) ([]RegistrationTokenDetail
 	return api.FetchAllPages[RegistrationTokenDetails](ac, registrationTokenURL, nil)
 }
 
-// groupSummary is a minimal projection used to build a UUID→name map for group display.
-// Keeping it local avoids importing api/iam and pulling in its heavier GroupResponse type.
-type groupSummary struct {
-	ID   string `json:"id"`
-	Name string `json:"name"`
-}
-
 // buildGroupUUIDToNameMap fetches all IAM groups and returns a UUID→name map.
 // On failure it returns an empty map so callers never block on a group-lookup error.
 func buildGroupUUIDToNameMap(ac *client.AlpaconClient) map[string]string {
-	// "/api/iam/groups/" is intentionally duplicated from api/iam to avoid an import cycle.
-	groups, err := api.FetchAllPages[groupSummary](ac, "/api/iam/groups/", nil)
+	groups, err := api.FetchAllPages[groupSummary](ac, iamGroupURL, nil)
 	if err != nil {
 		return map[string]string{}
 	}
--- a/cmd/server/token.go
+++ b/cmd/server/token.go
@@ -9,7 +9,7 @@ import (
 var tokenCmd = &cobra.Command{
 	Use:   "token",
 	Short: "Manage server registration tokens",
-	Long:  "Create, list, and delete registration tokens used by servers to self-register via alpamon.",
+	Long:  "Create, list, and delete registration tokens used by servers to self-register via Alpamon.",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if err := cmd.Help(); err != nil {
 			return err
--- a/cmd/server/token.go
+++ b/cmd/server/token.go
@@ -1,0 +1,25 @@
+package server
+
+import (
+	"errors"
+
+	"github.com/spf13/cobra"
+)
+
+var tokenCmd = &cobra.Command{
+	Use:   "token",
+	Short: "Manage server registration tokens",
+	Long:  "Create, list, and delete registration tokens used by servers to self-register via alpamon.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := cmd.Help(); err != nil {
+			return err
+		}
+		return errors.New("a subcommand is required. Use 'alpacon server token create', 'alpacon server token ls', or 'alpacon server token delete'. Run 'alpacon server token --help' for more information")
+	},
+}
+
+func init() {
+	tokenCmd.AddCommand(tokenCreateCmd)
+	tokenCmd.AddCommand(tokenListCmd)
+	tokenCmd.AddCommand(tokenDeleteCmd)
+}
--- a/cmd/server/token_create.go
+++ b/cmd/server/token_create.go
@@ -3,6 +3,7 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	iamAPI "github.com/alpacax/alpacon-cli/api/iam"
 	serverAPI "github.com/alpacax/alpacon-cli/api/server"
@@ -87,6 +88,7 @@ func init() {
 func resolveGroupIDs(ac *client.AlpaconClient, entries []string) ([]string, error) {
 	ids := make([]string, 0, len(entries))
 	for _, entry := range entries {
+		entry = strings.TrimSpace(entry)
 		if entry == "" {
 			continue
 		}
--- a/cmd/server/token_create.go
+++ b/cmd/server/token_create.go
@@ -35,12 +35,12 @@ The token key is displayed once at creation time and cannot be retrieved again.`
 
 		alpaconClient, err := client.NewAlpaconAPIClient()
 		if err != nil {
-			utils.CliErrorWithExit("connection to Alpacon API failed: %s. consider re-logging.", err)
+			utils.CliErrorWithExit("Connection to Alpacon API failed: %s. Consider re-logging.", err)
 		}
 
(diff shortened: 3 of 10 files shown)


cmd/server/token_create.go
```diff
@@ -0,0 +1,104 @@
...
+				utils.CliErrorWithExit("Failed to marshal response: %s.", err)
+			}
+			utils.PrintJson(data)
+		}
+
+		utils.CliSuccess("Registration token created: %s", resp.Name)
+		utils.CliWarning("Save this key now—it will not be shown again: %s", utils.Green(resp.Key))
+		if resp.ExpiresAt != nil {
+			utils.CliInfo("Token expires at: %s", *resp.ExpiresAt)
+		}
```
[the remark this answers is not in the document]
jisung-02: Intentional design. stdout and stderr are separate streams so the JSON is not corrupted. The warning on stderr is kept even in JSON mode because CI pipelines often redirect stderr to operator logs—stripping the key from the warning would silently remove the one-time-only notice from those logs. Omitting the key from the warning when --output json is set would mean any pipeline that swallows stderr loses the security reminder entirely. The current behavior is the safest trade-off.

cmd/server/token_create.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, entry := range entries {
+		if entry == "" {
+			continue
+		}
+		if utils.IsUUID(entry) {
+			ids = append(ids, entry)
+		} else {
+			id, err := iamAPI.GetGroupIDByName(ac, entry)
+			if err != nil {
+				return nil, fmt.Errorf("group %q not found: %w", entry, err)
```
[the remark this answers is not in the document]
jisung-02: Fixed in a32864f. Changed to 'failed to resolve group %q: %w' so network and HTTP errors are not misread as not-found.

cmd/server/token.go
```diff
@@ -0,0 +1,25 @@
...
+var tokenCmd = &cobra.Command{
+	Use:   "token",
+	Short: "Manage server registration tokens",
+	Long:  "Create, list, and delete registration tokens used by servers to self-register via alpamon.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := cmd.Help(); err != nil {
+			return err
+		}
+		return errors.New("a subcommand is required. Use 'alpacon server token create', 'alpacon server token ls', or 'alpacon server token delete'. Run 'alpacon server token --help' for more information")
+	},
```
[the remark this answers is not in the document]
jisung-02: Fixed in a32864f.

cmd/server/token_list.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+var tokenListCmd = &cobra.Command{
+	Use:     "ls",
+	Aliases: []string{"list"},
+	Short:   "List server registration tokens",
+	Long: `Display all server registration tokens.
+Tokens are used by alpamon to self-register servers without manual intervention.`,
```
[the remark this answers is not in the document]
jisung-02: Fixed in a32864f.

cmd/server/token_delete.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/spf13/cobra"
+)
+
+var tokenDeleteCmd = &cobra.Command{
+	Use:     "delete TOKEN",
+	Aliases: []string{"rm"},
+	Short:   "Delete a server registration token",
+	Long: `Delete a server registration token by name.
+The token is resolved by name—UUID input is not supported.`,
+	Example: `
```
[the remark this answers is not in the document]
jisung-02: Intentional design decision documented in the implementation plan. The delete command accepts only token names—UUID input is explicitly out of scope. The Long description is accurate. If UUID deletion is needed in a follow-up, it can be added by detecting UUID format (utils.IsUUID) and calling SendDeleteRequest directly, without going through GetRegistrationTokenByName.

api/server/server.go
```diff
@@ -163,6 +173,75 @@ func ListRegistrationTokens(ac *client.AlpaconClient) ([]RegistrationTokenDetail
...
+
+// DeleteRegistrationToken resolves a token name to its ID and sends a DELETE request.
+func DeleteRegistrationToken(ac *client.AlpaconClient, tokenName string) error {
+	token, err := GetRegistrationTokenByName(ac, tokenName)
+	if err != nil {
+		return err
+	}
+	_, err = ac.SendDeleteRequest(utils.BuildURL(registrationTokenURL, token.ID, nil))
+	return err
+}
```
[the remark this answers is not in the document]
jisung-02: By design—delete-by-name is the intentional behavior for this command. UUID-based deletion is not in the acceptance criteria for this issue. If added later, the fix would be to check utils.IsUUID(tokenName) and skip the GetRegistrationTokenByName lookup.

api/server/server.go
```diff
@@ -163,6 +173,75 @@ func ListRegistrationTokens(ac *client.AlpaconClient) ([]RegistrationTokenDetail
...
+func buildGroupUUIDToNameMap(ac *client.AlpaconClient) map[string]string {
+	groups, err := api.FetchAllPages[groupSummary](ac, iamGroupURL, nil)
+	if err != nil {
+		return map[string]string{}
+	}
+	m := make(map[string]string, len(groups))
+	for _, g := range groups {
+		m[g.ID] = g.Name
+	}
+	return m
```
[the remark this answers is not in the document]
jisung-02: The silent fallback to UUIDs is intentional per the design doc: group name resolution failure must not block the token list. The behavior (UUIDs shown as-is) is documented in the GetRegistrationTokenAttributes comment. Adding a warning could be done in a follow-up by changing buildGroupUUIDToNameMap to return (map[string]string, bool) and emitting utils.CliWarning from the cmd layer.

cmd/server/token_create.go
```diff
@@ -0,0 +1,104 @@
...
+	ids := make([]string, 0, len(entries))
+	for _, entry := range entries {
+		if entry == "" {
+			continue
+		}
+		if utils.IsUUID(entry) {
+			ids = append(ids, entry)
+		} else {
+			id, err := iamAPI.GetGroupIDByName(ac, entry)
+			if err != nil {
```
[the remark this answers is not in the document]
jisung-02: Fixed in 934f5ef. Added strings.TrimSpace(entry) before the empty/UUID/name checks in resolveGroupIDs.

cmd/server/token_list.go
```diff
@@ -0,0 +1,33 @@
...
+		alpaconClient, err := client.NewAlpaconAPIClient()
+		if err != nil {
+			utils.CliErrorWithExit("Connection to Alpacon API failed: %s. Consider re-logging.", err)
+		}
+
+		tokens, err := serverAPI.GetRegistrationTokenAttributes(alpaconClient)
+		if err != nil {
+			utils.CliErrorWithExit("Failed to retrieve the registration tokens: %s.", err)
+		}
+
```
geunwoonoh, line 30: Critical: `--output json` is advertised in the `Example` block (and in the PR description) but never implemented — `Run` only calls `PrintTable`, so JSON mode silently falls back to a table. Compare `token_create.go:63-69`, which branches on `utils.OutputFormat`.

```suggestion
		tokens, err := serverAPI.GetRegistrationTokenAttributes(alpaconClient)
		if err != nil {
			utils.CliErrorWithExit("Failed to retrieve the registration tokens: %s.", err)
		}

		if utils.OutputFormat == utils.OutputFormatJSON {
			data, err := json.Marshal(tokens)
			if err != nil {
				utils.CliErrorWithExit("Failed to marshal response: %s.", err)
			}
			utils.PrintJson(data)
			return
		}

		utils.PrintTable(tokens)
	},
```
(Remember to add the `encoding/json` import.)

cmd/server/token_create.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			utils.CliErrorWithExit("Failed to create registration token: %s.", err)
+		}
+
+		if utils.OutputFormat == utils.OutputFormatJSON {
+			data, err := json.Marshal(resp)
+			if err != nil {
+				utils.CliErrorWithExit("Failed to marshal response: %s.", err)
+			}
+			utils.PrintJson(data)
```
geunwoonoh, line 69: Warning: missing `return` after `PrintJson`. In `--output json` mode, control falls through to the `CliSuccess`/`CliWarning`/`CliInfo` calls below, which dump human-readable text to stderr. Callers scripting against the JSON output almost certainly don't want that — and the JSON already contains both `key` and `expires_at`.

```suggestion
		if utils.OutputFormat == utils.OutputFormatJSON {
			data, err := json.Marshal(resp)
			if err != nil {
				utils.CliErrorWithExit("Failed to marshal response: %s.", err)
			}
			utils.PrintJson(data)
			return
		}
```

cmd/server/token_create.go
```diff
@@ -0,0 +1,106 @@
...
+		groupIDs, err := resolveGroupIDs(alpaconClient, groups)
+		if err != nil {
+			utils.CliErrorWithExit("Failed to resolve groups: %s.", err)
+		}
+
+		req := serverAPI.RegistrationTokenRequest{
+			Name:          name,
+			AllowedGroups: groupIDs,
+		}
+		if expiresInDays < 0 {
```
geunwoonoh: Warning: input-validation ordering. `expiresInDays < 0` is checked *after* `resolveGroupIDs` has already hit the IAM API. A user typo like `--expires-in-days -1` wastes a network round-trip, and if `--groups` is also invalid the group-resolution error masks the real problem. Move the guard above `resolveGroupIDs` so invalid flags fail fast and offline.

```suggestion
		if expiresInDays < 0 {
			utils.CliErrorWithExit("--expires-in-days must be 0 (no expiry) or a positive number, got %d.", expiresInDays)
		}

		alpaconClient, err := client.NewAlpaconAPIClient()
		if err != nil {
			utils.CliErrorWithExit("Connection to Alpacon API failed: %s. Consider re-logging.", err)
		}

		groupIDs, err := resolveGroupIDs(alpaconClient, groups)
		if err != nil {
			utils.CliErrorWithExit("Failed to resolve groups: %s.", err)
		}

		req := serverAPI.RegistrationTokenRequest{
			Name:          name,
			AllowedGroups: groupIDs,
		}
		if expiresInDays > 0 {
			req.ExpiresAt = utils.TimeFormat(expiresInDays)
		}
```

api/server/server.go
```diff
@@ -14,11 +15,20 @@ const (
 	serverURL            = "/api/servers/servers/"
 	registrationTokenURL = "/api/servers/registration-tokens/"
 	registrationGuideURL = "/api/servers/registration-methods/token-install/guide/"
+	// iamGroupURL is intentionally duplicated from api/iam to avoid an import cycle.
+	iamGroupURL = "/api/iam/groups/"
```
geunwoonoh, line 20: Warning: the "avoid an import cycle" rationale is inaccurate. `api/iam` does not import `api/server`, and `api/event` already imports both — so `api/server` importing `api/iam` would not create a cycle. Either correct the comment to the real reason (e.g., "keep `api/server` from pulling in the full `GroupResponse` surface") or just import `iamAPI` and reuse `groupURL` with a trimmed response type.

api/server/server.go
```diff
@@ -163,6 +173,75 @@ func ListRegistrationTokens(ac *client.AlpaconClient) ([]RegistrationTokenDetail
 	return api.FetchAllPages[RegistrationTokenDetails](ac, registrationTokenURL, nil)
 }
 
+// buildGroupUUIDToNameMap fetches all IAM groups and returns a UUID→name map.
+// On failure it returns an empty map so callers never block on a group-lookup error.
+func buildGroupUUIDToNameMap(ac *client.AlpaconClient) map[string]string {
```
geunwoonoh, line 180: Suggestion: `buildGroupUUIDToNameMap` silently swallows every error (expired auth, 500, network). Users then see raw UUIDs in the table with no hint as to why. Consider emitting a one-time warning so the fallback is observable:

```suggestion
func buildGroupUUIDToNameMap(ac *client.AlpaconClient) map[string]string {
	groups, err := api.FetchAllPages[groupSummary](ac, iamGroupURL, nil)
	if err != nil {
		utils.CliWarning("Could not resolve group names; showing UUIDs instead: %s", err)
		return map[string]string{}
	}
	m := make(map[string]string, len(groups))
	for _, g := range groups {
		m[g.ID] = g.Name
	}
	return m
}
```

api/server/server.go
```diff
@@ -163,6 +173,75 @@ func ListRegistrationTokens(ac *client.AlpaconClient) ([]RegistrationTokenDetail
...
+			needsGroups = true
+			break
+		}
+	}
+	groupMap := map[string]string{}
+	if needsGroups {
+		groupMap = buildGroupUUIDToNameMap(ac)
+	}
+
+	var out []RegistrationTokenAttributes
```
geunwoonoh: Suggestion: pre-size the output slice to avoid repeated re-allocations — small win, also more idiomatic.

```suggestion
	out := make([]RegistrationTokenAttributes, 0, len(tokens))
```

cmd/server/token_create.go
```diff
@@ -0,0 +1,106 @@
...
+func init() {
+	tokenCreateCmd.Flags().StringP("name", "n", "", "A name to identify the token.")
+	tokenCreateCmd.Flags().StringSliceP("groups", "g", []string{}, "Group names or UUIDs to assign on registration (comma-separated).")
+	tokenCreateCmd.Flags().Int("expires-in-days", 0, "Number of days until the token expires (0 = no expiry).")
+}
+
+// resolveGroupIDs converts a list of group names or UUIDs to UUIDs.
+// Entries that are already UUIDs pass through unchanged.
+// Entries that look like names are resolved via the IAM API.
+func resolveGroupIDs(ac *client.AlpaconClient, entries []string) ([]string, error) {
```
geunwoonoh, line 90: Suggestion: `resolveGroupIDs` issues one sequential HTTP call per name. Fine for 1–3 groups (typical usage), but O(N) round-trips for larger inputs. If bulk usage is expected, consider a single `FetchAllPages[groupSummary]` call and a client-side name→ID lookup, mirroring `buildGroupUUIDToNameMap`. Not blocking.

cmd/server/token_create.go
```diff
@@ -0,0 +1,106 @@
...
+
+var tokenCreateCmd = &cobra.Command{
+	Use:   "create",
+	Short: "Create a new server registration token",
+	Long: `Generates a new server registration token.
+The token key is displayed once at creation time and cannot be retrieved again.`,
+	Example: `
+    alpacon server token create --name k8s-nodes --groups infra --expires-in-days 365
+    alpacon server token create  # interactive mode`,
+	Run: func(cmd *cobra.Command, args []string) {
```
geunwoonoh, line 23: Suggestion: the three new subcommands (`create`, `ls`, `delete`) use `Run` with `utils.CliErrorWithExit`, while the sibling `ServerCmd` uses `RunE` + `return err`. Not blocking, but `RunE` is generally preferred — it's easier to unit-test and plays better with Cobra's error hook. Something to align on going forward.

cmd/server/token_create_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/alpacax/alpacon-cli/api"
+	"github.com/alpacax/alpacon-cli/client"
+)
+
+// groupListResponse is a minimal helper type for mock IAM group responses.
+type groupListItem struct {
```
geunwoonoh, line 14: Suggestion: `{ID string; Name string}` appears as `groupSummary` (prod), `groupListItem` (here), and two anonymous structs in `server_test.go`. Consolidating into a single unexported test helper — or just reusing the prod `groupSummary` — would DRY things up.

geunwoonoh (changes requested):
## Summary

Adds `alpacon server token {create,ls,delete}` so registration tokens can be managed fully from the CLI (enabling Terraform/Ansible/CloudInit provisioning). Also adds `ExpiresAt` to the token types, an IAM-group UUID→name lookup for display, and 12 new tests.

---

### Critical issues (1)

1. **`token ls --output json` is documented but not implemented** — the subcommand only calls `PrintTable`.

### Warnings (3)

1. `token_create` falls through to `CliSuccess`/`CliWarning`/`CliInfo` after `PrintJson`, polluting stderr in JSON mode.
2. `expiresInDays < 0` validation runs *after* `resolveGroupIDs`, wasting an API round-trip on invalid input.
3. The "avoid import cycle" comment in `api/server/server.go` is inaccurate — no cycle exists today (`api/event` already imports both `api/iam` and `api/server`).

### Suggestions (5)

1. `buildGroupUUIDToNameMap` silently swallows errors; consider a one-time `CliWarning` so users aren't confused by raw UUIDs.
2. Pre-size `out` with `make([]RegistrationTokenAttributes, 0, len(tokens))`.
3. `resolveGroupIDs` makes one HTTP call per name; acceptable today, flag only if bulk usage is expected.
4. Subcommands use `Run` while sibling `ServerCmd` uses `RunE` — minor inconsistency.
5. `{ID,Name}` is duplicated across `groupSummary`, `groupListItem`, and anonymous structs in tests; a shared test helper would DRY this up.

### Highlights

- Good batched IAM lookup with the `needsGroups` short-circuit — avoids an N+1.
- Correct use of `api.FetchAllPages[T]` with a lean local projection type.
- `DeleteRegistrationToken` reuses `GetRegistrationTokenByName` so `ErrRegistrationTokenNotFound` behavior stays consistent.
- Solid test coverage: create (with/without expiry), delete (found/not found), attributes (group mapping / UUID fallback / never-expires / empty), plus 4 `resolveGroupIDs` cases.
- `CLAUDE.md` "Go declaration order" addition is reinforced by lifting `groupSummary` into the type block.
- Negative-days validation is a genuine UX improvement — prior behavior silently created non-expiring tokens.

---

**Verdict**: CHANGES REQUESTED — the `token ls --output json` gap is a real functional regression versus the documented contract; the rest are polish.

### Checklist

- [x] Tests included (12 new)
- [ ] `--output json` honored where documented (ls missing)

cmd/server/token_list.go
jisung-02: Thanks for the careful read. After digging in, I don't think this is actually broken — utils.PrintTable already branches on utils.OutputFormat at utils/output.go:34-45, emits json.MarshalIndent output for lists, and correctly yields [] for nil/empty slices. So PrintTable(tokens) alone honors --output json today, and the Example block matches actual behavior.

cmd/server/token_create.go
jisung-02: Fixed in 11e32c2. Added the missing return so --output json mode exits cleanly after PrintJson(data) without leaking CliSuccess/CliWarning/CliInfo text to stderr.

cmd/server/token_create.go
jisung-02: Fixed in 11e32c2. Moved the expiresInDays < 0 guard to immediately after flag parsing, so invalid flags fail fast offline before either client.NewAlpaconAPIClient or resolveGroupIDs runs. This also stops a --groups error from masking the real negative-days mistake.

api/server/server.go
jisung-02: Fixed in f4b4307. You're right — no cycle exists; api/event already imports both api/iam and api/server. Corrected the rationale to "keep api/server from pulling in the full iam.GroupResponse surface." I kept the local groupSummary projection (ID + Name) rather than importing iamAPI — the full GroupResponse carries fields this package doesn't need, so the duplicated constant is a deliberate trade for a leaner type surface.

api/server/server.go
jisung-02: Applied in f4b4307. Added `utils.CliWarning("Could not resolve group names; showing UUIDs instead: %s", err)` on the error path. Still non-blocking — the token list renders with raw UUIDs — but operators now see why names are missing (expired auth / 5xx / network) instead of silently getting UUIDs.

api/server/server.go
jisung-02: Applied in f4b4307. Now `out := make([]RegistrationTokenAttributes, 0, len(tokens))`.

cmd/server/token_create.go
jisung-02: Deferring as you marked this non-blocking. Typical usage is 1–3 groups per token where sequential round-trips are negligible. If bulk provisioning patterns emerge, happy to open an issue and migrate to `FetchAllPages[groupSummary]` + a client-side name→ID map (mirroring `buildGroupUUIDToNameMap`).

cmd/server/token_create.go
jisung-02: Agreed that `RunE` is preferable long-term, but this feels broader than this PR — a consistent pass over `cmd/server` (or `cmd/*`) in a dedicated cleanup PR would be cleaner than flipping only the three new subcommands here. Happy to take that on as a follow-up.

cmd/server/token_create_test.go
jisung-02: Agreed on the DRY point. Since `api/server/server_test.go` lives in the same package as the prod `groupSummary`, the fix is a mechanical `groupListItem` → `groupSummary` replace. Leaving it for a follow-up polish PR to keep this one focused on change-request feedback.

geunwoonoh (approved):

PR #176, opened 2026-05-15
refactor(client): lazy-load current user and remove preflight auth checks

jisung-02:
## Summary

- **Remove preflight auth calls from `NewAlpaconAPIClient`**: The constructor no longer makes two blocking HTTP round-trips (`/api/auth/is-authenticated/` + `/api/iam/users/-`) on every CLI invocation (~1.57 s overhead eliminated)
- **Add `LoadCurrentUser` with `sync.Once` caching**: Privilege and username data are fetched at most once per process lifetime, lazily, only in commands that actually need them
- **Extract `checkAuthStatus` helper**: Deduplicates 401/403 error handling that was repeated across `sendRequest`, `SendMultipartRequest`, and `SendGetRequestForDownload`
- **Fix `SendMultipartRequest` to accept HTTP 200 and 201**: The server returns 200 for single-file upload and 201 for bulk upload; previously only 201 was accepted, silently treating 200 as an error
- **Fix `SendGetRequestForDownload`**: Close response body before returning auth errors to prevent resource leaks
- **Code quality**: Improve declaration order (`LoadCurrentUser` before unexported helpers), separate exported/unexported struct fields with a blank line

## Performance

Measured on a live workspace (10 samples each):

| Step | Before | After |
|------|--------|-------|
| `checkAuth` preflight | ~818 ms | removed |
| `checkPrivileges` preflight | ~756 ms | removed |
| Total preflight overhead | ~1,574 ms | 0 ms |
| `alpacon server list` wall time | ~2,552 ms | ~978 ms |
| **Reduction** | | **~62%** |

## Test plan

- [ ] `go test -race ./client/...` — all 13 tests pass
- [ ] `go test -race ./...` — full suite passes
- [ ] `alpacon login` completes and reports logged-in user
- [ ] Commands gated on staff/superuser privilege (`user create`, `group create`, `group delete`) correctly reject general users
- [ ] `alpacon server list` (or any non-privilege-gated command) no longer waits for a privilege preflight call

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -120,6 +120,24 @@ func getUserPrivileges(isStaff, isSuperuser bool) string {
 	return "general"
 }
 
+func (ac *AlpaconClient) LoadCurrentUser() error {
+	if ac.userLoaded {
+		return nil
+	}
+	body, err := ac.SendGetRequest(checkPrivilegesURL)
+	if err != nil {
+		return err
+	}
+	var resp CheckPrivilegesResponse
+	if err := json.Unmarshal(body, &resp); err != nil {
+		return err
+	}
+	ac.Privileges = getUserPrivileges(resp.IsStaff, resp.IsSuperuser)
+	ac.Username = strings.TrimSpace(resp.Username)
+	ac.userLoaded = true
+	return nil
+}
+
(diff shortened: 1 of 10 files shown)


eunyoung14 (approved):
Nice catch!

Issue #19, opened 2025-05-13
cp command overwrite


parkm2ngyu00:
Fix issue where files were not properly overwritten due to missing overwrite flag in the `cp` command.

PR #192, opened 2026-05-27
feat(worksession): declare sudo bypass policies via CLI for agent flows

eunyoung14:
## Summary
Lets a non-interactive execute-command caller (e.g. an AI agent) run sudo without interactive MFA by pre-declaring sudo command patterns on a work session.

- `work-session create --sudo "cmd,cmd"` (repeatable, `--sudo-reason`) attaches MFA-bypass policies and implies the `sudo` scope. `users` is omitted so the server binds each policy to the session assignee.
- `work-session update [SESSION_ID] --sudo` adds patterns to an existing session. It sends the **full desired policy set** (existing policies echoed back by `id` + the additions) so the PUT-style modify endpoint never drops current policies, and guards against `pending` sessions (whose serializer ignores `sudo_policies`).
- `exec` prints an actionable hint pointing at `work-session update` when command output carries the `sudo_no_worksession_policy` denial code (paired with the alpacon-server PR).

## Flow
```
alpacon work-session create --purpose "nginx fix" --server web-01 --expires-in 2h \
  --sudo "systemctl restart nginx,systemctl reload nginx" --wait
alpacon work-session use <id>
alpacon exec web-01 "sudo systemctl restart nginx"   # no MFA / no password
# denied command -> hint -> alpacon work-session update --sudo "<cmd>"
```

## Depends on
- alpacon-server: `SUDO_NO_WORKSESSION_POLICY` error code (companion PR).

## Test plan
- [x] `go build ./...`, `go vet`, gofmt clean on changed files.
- [x] `cmd/worksession`, `cmd/exec`, `api/worksession` tests pass; added unit tests for `buildSudoPolicies` and `sudoDenialHint`.
- [ ] Manual: create a session with `--sudo`, run a covered sudo via `exec` (bypasses MFA); run an uncovered one (hint shown); `work-session update --sudo` then re-run.

### DIFF
--- a/api/worksession/types.go
+++ b/api/worksession/types.go
@@ -65,9 +65,15 @@ type SudoPolicyInline struct {
 }
 
 // WorkSessionUpdateRequest carries the FULL desired sudo policy set (PUT-style):
-// the server deletes any existing policy absent from the list. SudoPolicies is
-// sent even when empty so an explicit empty set clears all policies; callers
-// that only mean to add must echo the existing policies back.
+// the server deletes any existing policy absent from the list. Callers that
+// only mean to add must echo the existing policies back.
+//
+// Note on Go JSON encoding: a nil slice marshals to JSON `null` (treated by
+// the server as "field not provided" → no change). To explicitly clear all
+// policies, pass an explicitly empty slice (`[]SudoPolicyInline{}`), which
+// marshals to `[]`. The current CLI never sends nil here because the update
+// command guards `len(newPolicies) == 0` upstream and always builds the
+// desired slice via `make(...)`.
 type WorkSessionUpdateRequest struct {
 	SudoPolicies []SudoPolicyInline `json:"sudo_policies"`
 }
--- a/api/worksession/types.go
+++ b/api/worksession/types.go
@@ -20,6 +20,7 @@ type WorkSession struct {
 	CompletedAt   *time.Time            `json:"completed_at"`
 	AddedAt       time.Time             `json:"added_at"`
 	UpdatedAt     time.Time             `json:"updated_at"`
+	SudoPolicies  []SudoPolicyInline    `json:"sudo_policies"`
 }
 
 type WorkSessionAttributes struct {
@@ -35,11 +36,36 @@ type WorkSessionAttributes struct {
 }
 
 type WorkSessionCreateRequest struct {
-	Description   string   `json:"description"`
-	RequesterType string   `json:"requester_type"`
-	Scopes        []string `json:"scopes"`
-	Servers       []string `json:"servers"`
-	ExpiresAt     string   `json:"expires_at"`
+	Description   string             `json:"description"`
+	RequesterType string             `json:"requester_type"`
+	Scopes        []string           `json:"scopes"`
+	Servers       []string           `json:"servers"`
+	ExpiresAt     string             `json:"expires_at"`
+	SudoPolicies  []SudoPolicyInline `json:"sudo_policies,omitempty"`
+}
+
+// SudoPolicyInline is a sudo policy bound to a work session. It serves three
+// roles: read (when returned on a WorkSession), create (attached at session
+// create time), and modify (sent on update). Commands are the allowed patterns
+// (wildcards permitted); AllowBypassMFA lets matching sudo run without
+// interactive MFA, which is the only way a non-interactive execute-command
+// caller (e.g. an AI agent) can sudo. Users is intentionally omitted — the
+// server binds the policy to the session assignee automatically so it never
+// applies to other workspace users.
+//
+// ID is set only on read and on modify: an update PATCH sends the FULL desired
+// set, so existing policies must be echoed back with their ID (modify in place)
+// alongside new entries without an ID (additions). Omitting an existing policy
+// from the set deletes it, so callers must preserve the current entries.
+type SudoPolicyInline struct {
+	ID             string   `json:"id,omitempty"`
+	Commands       []string `json:"commands"`
+	Reason         string   `json:"reason,omitempty"`
+	AllowBypassMFA bool     `json:"allow_bypass_mfa"`
+}
+
+type WorkSessionUpdateRequest struct {
+	SudoPolicies []SudoPolicyInline `json:"sudo_policies,omitempty"`
 }
 
 type WorkSessionExtendRequest struct {
@@ -60,21 +86,21 @@ type TimelineItem struct {
 	ServerID  *string `json:"server_id"`
 
 	// shared across session-like types (websh_session, tunnel_session, ftp_session)
-	Username  string  `json:"username"`
-	Groupname string  `json:"groupname"`
-	ClosedAt  *string `json:"closed_at"`
-	ClientType string `json:"client_type"`
+	Username   string  `json:"username"`
+	Groupname  string  `json:"groupname"`
+	ClosedAt   *string `json:"closed_at"`
+	ClientType string  `json:"client_type"`
 
 	// command
-	Shell       string   `json:"shell"`
-	Line        string   `json:"line"`
-	Success     *bool    `json:"success"`
-	Denied      bool     `json:"denied"`
-	ElapsedTime float64  `json:"elapsed_time"`
+	Shell       string  `json:"shell"`
+	Line        string  `json:"line"`
+	Success     *bool   `json:"success"`
+	Denied      bool    `json:"denied"`
+	ElapsedTime float64 `json:"elapsed_time"`
 
 	// tunnel_session
-	IsTunnel   bool  `json:"is_tunnel"`
-	TargetPort *int  `json:"target_port"`
+	IsTunnel   bool `json:"is_tunnel"`
+	TargetPort *int `json:"target_port"`
 
 	// file_upload / file_download
 	Name string `json:"name"`
--- a/api/worksession/types.go
+++ b/api/worksession/types.go
@@ -64,8 +64,12 @@ type SudoPolicyInline struct {
 	AllowBypassMFA bool     `json:"allow_bypass_mfa"`
 }
 
+// WorkSessionUpdateRequest carries the FULL desired sudo policy set (PUT-style):
+// the server deletes any existing policy absent from the list. SudoPolicies is
+// sent even when empty so an explicit empty set clears all policies; callers
+// that only mean to add must echo the existing policies back.
 type WorkSessionUpdateRequest struct {
-	SudoPolicies []SudoPolicyInline `json:"sudo_policies,omitempty"`
+	SudoPolicies []SudoPolicyInline `json:"sudo_policies"`
 }
 
 type WorkSessionExtendRequest struct {
--- a/api/worksession/types.go
+++ b/api/worksession/types.go
@@ -48,8 +48,8 @@ type WorkSessionCreateRequest struct {
 // roles: read (when returned on a WorkSession), create (attached at session
 // create time), and modify (sent on update). Commands are the allowed patterns
 // (wildcards permitted); AllowBypassMFA lets matching sudo run without
-// interactive MFA, which is the only way a non-interactive execute-command
-// caller (e.g. an AI agent) can sudo. Users is intentionally omitted — the
+// interactive MFA, which is the only way a non-interactive caller (e.g. an AI
+// agent running 'exec') can sudo. Users is intentionally omitted — the
 // server binds the policy to the session assignee automatically so it never
 // applies to other workspace users.
 //
--- a/api/worksession/worksession.go
+++ b/api/worksession/worksession.go
@@ -63,6 +63,22 @@ func CreateWorkSession(ac *client.AlpaconClient, req WorkSessionCreateRequest) (
 	return &session, nil
 }
 
+// UpdateWorkSession PATCHes a work session. Used to attach sudo policies to an
+// existing session (e.g. after an execute-command sudo was denied). The server
+// may queue the change for approval, in which case it takes effect only once
+// approved.
+func UpdateWorkSession(ac *client.AlpaconClient, id string, req WorkSessionUpdateRequest) (*WorkSession, error) {
+	body, err := ac.SendPatchRequest(utils.BuildURL(workSessionURL, id, nil), req)
+	if err != nil {
+		return nil, err
+	}
+	var session WorkSession
+	if err = json.Unmarshal(body, &session); err != nil {
+		return nil, err
+	}
+	return &session, nil
+}
+
 func GetWorkSession(ac *client.AlpaconClient, id string) (*WorkSession, error) {
 	body, err := ac.SendGetRequest(utils.BuildURL(workSessionURL, id, nil))
 	if err != nil {
--- a/api/worksession/worksession.go
+++ b/api/worksession/worksession.go
@@ -64,7 +64,7 @@ func CreateWorkSession(ac *client.AlpaconClient, req WorkSessionCreateRequest) (
 }
 
 // UpdateWorkSession PATCHes a work session. Used to attach sudo policies to an
-// existing session (e.g. after an execute-command sudo was denied). The server
+// existing session (e.g. after an 'exec' sudo was denied). The server
 // may queue the change for approval, in which case it takes effect only once
 // approved.
 func UpdateWorkSession(ac *client.AlpaconClient, id string, req WorkSessionUpdateRequest) (*WorkSession, error) {
--- a/api/worksession/worksession_test.go
+++ b/api/worksession/worksession_test.go
@@ -290,7 +290,7 @@ func TestUpdateWorkSession(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		assert.Equal(t, http.MethodPatch, r.Method)
 		assert.Equal(t, "/api/work-sessions/sessions/ses-abc/", r.URL.Path)
-		_ = json.NewDecoder(r.Body).Decode(&gotBody)
+		assert.NoError(t, json.NewDecoder(r.Body).Decode(&gotBody))
 		w.Header().Set("Content-Type", "application/json")
 		_ = json.NewEncoder(w).Encode(WorkSession{ID: "ses-abc", Status: "active"})
 	}))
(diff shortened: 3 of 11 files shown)


cmd/worksession/worksession_update.go
```diff
@@ -0,0 +1,85 @@
...
+	Long: `Update an existing work session.
+
+Use --sudo to add MFA-bypass sudo command patterns to the session. This is the
+recovery path when an execute-command sudo was denied: add the command and re-run.
+Each --sudo value is a comma-separated pattern list (wildcards allowed) forming one
+policy. The session's existing policies are preserved; the additions may require
+approval before they take effect.
+
+If SESSION_ID is omitted, the workspace's active work session is used (set via
+'alpacon work-session use').`,
```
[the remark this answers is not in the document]
eunyoung14: Fixed. Help text now states resolution order: ALPACON_WORK_SESSION env var, then the workspace active session.

api/worksession/types.go
```diff
@@ -35,11 +36,36 @@ type WorkSessionAttributes struct {
...
+type SudoPolicyInline struct {
+	ID             string   `json:"id,omitempty"`
+	Commands       []string `json:"commands"`
+	Reason         string   `json:"reason,omitempty"`
+	AllowBypassMFA bool     `json:"allow_bypass_mfa"`
+}
+
+type WorkSessionUpdateRequest struct {
+	SudoPolicies []SudoPolicyInline `json:"sudo_policies,omitempty"`
 }
```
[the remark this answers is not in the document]
eunyoung14: Fixed. Dropped `omitempty` so the PUT-style field is always sent (an explicit empty set clears all policies).

cmd/exec/run.go
```diff
@@ -12,6 +13,27 @@ import (
...
+func sudoDenialHint(output string) string {
+	if !strings.Contains(output, sudoNoWorksessionPolicyCode) {
+		return ""
+	}
+	return fmt.Sprintf(
+		"%s sudo was denied: this command is not covered by an MFA-bypass policy in your work session.\n"+
+			"Add it and re-run:\n"+
+			"  alpacon work-session update --sudo \"<command>\"\n",
+		utils.Yellow("Hint:"),
+	)
```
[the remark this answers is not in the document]
eunyoung14: Fixed. Hint now shows `work-session update [SESSION_ID] --sudo "<command>"` and notes the optional ID for sessions without an active workspace session.

cmd/worksession/worksession_create.go
```diff
@@ -44,11 +46,20 @@ var workSessionCreateCmd = &cobra.Command{
...
+active state.
+
+If the work needs sudo, pre-declare the command patterns with --sudo. This attaches
+MFA-bypass sudo policies to the session so a non-interactive execute-command caller
+(e.g. an AI agent) can run those exact sudo commands without an interactive MFA
+prompt. The 'sudo' scope is added automatically, and the policies are submitted for
+approval together with the session. If a sudo command is later denied, add it to the
+session with 'alpacon work-session update <id> --sudo "<command>"'.`,
 	Example: `  alpacon work-session create --purpose "nginx fix" --scope command,websh --server web-01 --expires-in 2h
   alpacon work-session create --purpose "deploy" --scope command --server web-01,db-01 --expires-at 2026-05-09T10:00:00Z --wait
```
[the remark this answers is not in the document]
eunyoung14: Fixed. Updated the `--expires-at` example to a future date.

api/worksession/worksession.go
```diff
@@ -63,6 +63,22 @@ func CreateWorkSession(ac *client.AlpaconClient, req WorkSessionCreateRequest) (
 	return &session, nil
 }
 
+// UpdateWorkSession PATCHes a work session. Used to attach sudo policies to an
+// existing session (e.g. after an execute-command sudo was denied). The server
+// may queue the change for approval, in which case it takes effect only once
+// approved.
+func UpdateWorkSession(ac *client.AlpaconClient, id string, req WorkSessionUpdateRequest) (*WorkSession, error) {
```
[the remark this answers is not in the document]
eunyoung14: Fixed. Added an httptest for `UpdateWorkSession` asserting the PATCH method/path and the full-desired-set body shape (existing policy echoed by id + addition without one).

cmd/worksession/worksession_update.go
```diff
@@ -0,0 +1,85 @@
...
+	Args: cobra.MaximumNArgs(1),
+	Example: `  alpacon work-session update ses-abc123 --sudo "systemctl restart nginx"
+  alpacon work-session update --sudo "tail -f /var/log/nginx/*.log"`,
+	Run: func(cmd *cobra.Command, args []string) {
+		var sessionID string
+		if len(args) == 1 {
+			sessionID = args[0]
+		} else {
```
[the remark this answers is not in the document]
eunyoung14: The command is thin orchestration over now-tested pieces: `UpdateWorkSession` has an httptest (PATCH path + body shape) and `buildSudoPolicies` has a unit test. The remaining logic is a GET + slice append + a pending-status guard. Happy to add a full cobra-level test if you prefer, but it would mostly duplicate the API-helper coverage.

api/worksession/worksession_test.go
```diff
@@ -285,4 +285,29 @@ func TestGetWorkSessionTimeline_ExcludeRecords(t *testing.T) {
 	assert.Empty(t, result)
 }
 
+func TestUpdateWorkSession(t *testing.T) {
+	var gotBody WorkSessionUpdateRequest
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodPatch, r.Method)
+		assert.Equal(t, "/api/work-sessions/sessions/ses-abc/", r.URL.Path)
+		_ = json.NewDecoder(r.Body).Decode(&gotBody)
+		w.Header().Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
eunyoung14: Fixed. The test now asserts the JSON decode succeeds so a malformed body fails the test.

cmd/worksession/worksession_create.go
```diff
@@ -108,6 +119,17 @@ active state.`,
 				scopeList = append(scopeList, s)
 			}
 		}
+		// Build sudo bypass policies from --sudo. Each --sudo value is a
+		// comma-separated list of command patterns forming one policy; the
+		// policies are MFA-bypass (the only way a non-interactive
+		// execute-command caller can sudo). The server binds each policy to
+		// the session assignee automatically, so they never apply to other
+		// users. ``sudo`` scope is required server-side, so add it implicitly.
```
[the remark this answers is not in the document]
eunyoung14: Fixed. Replaced the RST-style double backticks with plain quotes ('sudo').

cmd/worksession/worksession.go
jisung-02: `work-session update` is missing from this list — it was added in this PR but not included here.

cmd/exec/run.go
```diff
@@ -12,6 +13,27 @@ import (
 	"github.com/alpacax/alpacon-cli/utils"
 )
 
+// sudoNoWorksessionPolicyCode is the server error code surfaced in command
+// output when a non-interactive sudo is denied for lack of a matching
+// MFA-bypass policy in the work session. Kept in sync with
+// alpacon-server utils/error_codes.py ErrorCode.SUDO_NO_WORKSESSION_POLICY.
+const sudoNoWorksessionPolicyCode = "sudo_no_worksession_policy"
```
jisung-02: sudoDenialHint is currently inoperative. The server sends sudo_no_worksession_policy as the reason field only, but alpacon_approval.c embeds error_code into *errstr (command output) — reason goes to syslog only. As a result, the hint condition can never be true. A follow-up PR is needed to add the error_code field to send_sudo_approval_response on the server side and add sudo_no_worksession_policy to the PAM allowlist.

cmd/worksession/worksession_update.go
```diff
@@ -0,0 +1,86 @@
...
+
+		// The modify endpoint takes the FULL desired set of policies (PUT-style):
+		// any existing policy missing from the request is deleted. Read the
+		// current set and echo it back (with IDs) alongside the additions so
+		// nothing is dropped.
+		session, err := wsapi.GetWorkSession(ac, sessionID)
+		if err != nil {
+			utils.CliErrorWithExit("Failed to load work session %s: %s.", sessionID, err)
+		}
+		if session.Status == pendingWorkSessionStatus {
```
jisung-02: create auto-adds the sudo scope when --sudo is used (worksession_create.go:129), but update can't do the same since WorkSessionUpdateRequest has no scopes field. Adding --sudo policies to a session created without the sudo scope makes the server reject with work_session_sudo_policy_without_scope. Like the pending guard right above (L64), consider a pre-flight check before building desired: if !slices.Contains(session.Scopes, "sudo"), fail with a clear message (e.g. "this session has no sudo scope, so sudo policies can't be added — create a new session with --scope sudo") instead of relying on the server-side rejection.

cmd/worksession/worksession.go
eunyoung14: Fixed. Added `work-session update` to the parent command help subcommand list.

cmd/exec/run.go
eunyoung14: Good catch — the lowercase match was indeed inoperative. Two companion changes landed for the end-to-end signal:

- alpacon-server #2153: adds the `error_code` field (UPPERCASE) alongside `reason` in `send_sudo_approval_response`, so `alpacon_approval.c` can surface it.
- alpamon-pam #19: adds `SUDO_NO_WORKSESSION_POLICY` to the auth-phase hard-deny allowlist (safety net for the same code).

This PR now matches `SUDO_NO_WORKSESSION_POLICY` (uppercase) in command output — what actually appears once the two companion PRs deploy.

cmd/worksession/worksession_update.go
eunyoung14: Fixed. The update command now pre-flights `session.Scopes` and fails with a clear message ("session does not include the \sudo\ scope; sudo policies cannot be added. Create a new session with --scope sudo and --sudo.") before building the desired set.

cmd/worksession/worksession_create.go
```diff
@@ -351,5 +393,7 @@ func init() {
 	workSessionCreateCmd.Flags().StringVar(&requesterType, "requester-type", "user", "Requester type: user or agent")
 	workSessionCreateCmd.Flags().BoolVar(&waitApproval, "wait", false, "Poll until the session is approved, then exit (does not set as active; combine with --use to attach automatically)")
 	workSessionCreateCmd.Flags().BoolVar(&useAfterCreate, "use", false, "Set the created session as the workspace's active session (requires status to reach 'active'; combine with --wait when approval is needed)")
+	workSessionCreateCmd.Flags().StringArrayVar(&createSudo, "sudo", nil, "Pre-declare sudo command patterns to run without interactive MFA (repeatable; each value is a comma-separated pattern list forming one policy, wildcards allowed). Required for non-interactive sudo via 'exec' (e.g. AI agents). Implies the 'sudo' scope. Patterns are submitted for approval with the session.")
```
junho226: Suggestion: `splitCSV` splits on every `,` and the flag help advertises wildcards but not the comma limitation. A pattern that legitimately contains a literal comma (e.g. a sudo command with `--opt=a,b`) gets silently split into two policies. Not a blocker for v1 since sudoers patterns rarely need literal commas, but the help text could call it out so users don't get confused later:

```suggestion
	workSessionCreateCmd.Flags().StringArrayVar(&createSudo, "sudo", nil, "Pre-declare sudo command patterns to run without interactive MFA (repeatable; each value is a comma-separated pattern list forming one policy, wildcards allowed — literal commas inside a pattern are not supported, use multiple --sudo flags instead). Required for non-interactive sudo via 'exec' (e.g. AI agents). Implies the 'sudo' scope. Patterns are submitted for approval with the session.")
```

api/worksession/types.go
```diff
@@ -35,11 +36,40 @@ type WorkSessionAttributes struct {
...
+	Reason         string   `json:"reason,omitempty"`
+	AllowBypassMFA bool     `json:"allow_bypass_mfa"`
+}
+
+// WorkSessionUpdateRequest carries the FULL desired sudo policy set (PUT-style):
+// the server deletes any existing policy absent from the list. SudoPolicies is
+// sent even when empty so an explicit empty set clears all policies; callers
+// that only mean to add must echo the existing policies back.
+type WorkSessionUpdateRequest struct {
+	SudoPolicies []SudoPolicyInline `json:"sudo_policies"`
```
junho226, line 79: Suggestion: the docstring says "SudoPolicies is sent even when empty so an explicit empty set clears all policies" — that's only true if the caller constructs `[]SudoPolicyInline{}` explicitly. A zero-value `WorkSessionUpdateRequest{}` marshals `SudoPolicies` as `null` (Go marshals nil slices to JSON null), not `[]`. The current `update` command always builds the slice via `make(...)` and guards `len(newPolicies) == 0` upstream so this never bites in practice, but a future caller could trip on it. Two low-cost options:

- Tighten the doc to say "pass an explicitly empty slice (`[]SudoPolicyInline{}`) to clear", or
- Add a small constructor helper (`NewWorkSessionUpdateRequest()`) that initializes the slice to empty rather than nil.

Not a blocker either way.

cmd/worksession/worksession_update.go
```diff
@@ -0,0 +1,112 @@
...
+		// scopes field, so we cannot add it here — fail early with guidance.
+		if err := validateSessionForSudoUpdate(session); err != nil {
+			utils.CliErrorWithExit("%s", err)
+		}
+
+		desired := make([]wsapi.SudoPolicyInline, 0, len(session.SudoPolicies)+len(newPolicies))
+		desired = append(desired, session.SudoPolicies...)
+		desired = append(desired, newPolicies...)
+
+		req := wsapi.WorkSessionUpdateRequest{SudoPolicies: desired}
```
junho226: Suggestion: I noticed the prior Copilot thread about a cobra-level test for this `Run` body and your reply that the pieces are individually covered. Agreed that `UpdateWorkSession`, `buildSudoPolicies`, and `validateSessionForSudoUpdate` are tested in isolation — the remaining unguarded behavior is the GET → echo-existing → append-new → PATCH **wiring** (the place a future refactor is most likely to regress the "don't drop existing policies" invariant).

If you want a cheap safety net, a single httptest with two handlers (GET returns a session with one existing `SudoPolicyInline`, PATCH asserts the body contains both the existing one by `id` and the new one without) would lock in the wiring. Skip-able if you'd rather move on; just noting where the test gap is so the next person doesn't have to re-derive it.

junho226 (dismissed):
## Summary

Adds a non-interactive sudo path for work sessions so callers like AI agents running `exec` can sudo without an interactive MFA prompt. Three pieces:

1. `work-session create --sudo "<cmd,cmd>"` (repeatable, optional `--sudo-reason`) attaches MFA-bypass `SudoPolicyInline` entries and implicitly adds the `sudo` scope.
2. New `work-session update [SESSION_ID] --sudo "<cmd>"` subcommand sends the **full desired set** (existing policies echoed back by `id` + additions) to the PUT-style modify endpoint so nothing is dropped. Guards against `pending` status and missing `sudo` scope with actionable messages.
3. `exec` prints a yellow hint pointing at `work-session update` when command output contains the `SUDO_NO_WORKSESSION_POLICY` denial code.

**Stack**: golang | **Changes**: +391 / -29 across 11 files

### Highlights

- Excellent intent-bearing doc comments — particularly `SudoPolicyInline` (read/create/modify roles, PUT semantics, why `Users` is omitted) and `sudoNoWorksessionPolicyCode` (the UPPERCASE rationale tied to `alpacon_approval.c`'s `[A-Z0-9_]` sanitizer).
- Pre-flight guards in `validateSessionForSudoUpdate` catch the two failure modes (`pending`, missing `sudo` scope) **before** the GET → append → PATCH cycle, surfacing a clear actionable error rather than a server-side `work_session_sudo_policy_without_scope` rejection.
- `--sudo` on create auto-adds the `sudo` scope (worksession_create.go:128-131), reducing flag verbosity.
- Targeted tests for each new pure function (`buildSudoPolicies`, `sudoDenialHint`, `validateSessionForSudoUpdate`) plus an httptest covering the PATCH path and full-desired-set body shape.
- Cross-PR alignment is documented inline (alpacon-server `error_code` field, alpamon-pam allowlist) — context that's easy to lose otherwise.
- Prior review threads from Copilot and @jisung-02 are all addressed (env-var resolution wording, `omitempty` removal on update request, future-dated example, sudo-scope pre-flight, etc.).

### Critical Issues (0)

None.

### Warnings (0)

None.

### Suggestions (3)

Minor, non-blocking. See inline comments.

---

**Verdict**: APPROVE

**Checklist**:
- [x] Code follows stack conventions (gofmt, idiomatic Go, slices/strings stdlib)
- [x] Tests included for new logic (units + httptest)
- [x] No security vulnerabilities (sudo bypass is server-gated; CLI only declares intent)
- [x] Documentation in command help and Go doc comments

---
🤖 Auto-review by junho-bot during junho226's leave. If this review looks off, please disregard — junho will follow up upon return.

cmd/worksession/worksession_create.go
eunyoung14: Fixed. Updated the --sudo flag help text to call out the comma-split limitation: literal commas inside a pattern are not supported, and users should pass the flag again for each policy that needs them.

api/worksession/types.go
eunyoung14: Good catch — tightened the doc to spell out the nil-vs-empty distinction explicitly: nil marshals to JSON `null` and the server reads it as "field not provided" (no change); only `[]SudoPolicyInline{}` clears all policies. Also noted that the current update command never sends nil because it guards `len(newPolicies) == 0` upstream and always builds the slice via `make(...)`.

cmd/worksession/worksession_update.go
eunyoung14: Agreed — extracted the GET → echo-existing → PATCH wiring into `attachSudoPoliciesToSession` and added a `TestAttachSudoPoliciesToSession_PreservesExistingPolicies` httptest covering exactly the regression bait you described (GET returns a session with one existing `SudoPolicyInline`, PATCH asserts the body has existing-with-ID + new-without-ID). Plus a `_PendingSessionAbortsBeforePATCH` test so the validator running before the PATCH is also locked down.

jisung-02:
I'll approve this once the merge conflict is resolved.

jisung-02 (approved):

PR #197, opened 2026-05-28
perf(cp,packages): cut transfer memory usage up to ~65x via streaming I/O

jisung-02:
## Summary

`alpacon cp` and `alpacon package upload/download` used to buffer entire payloads in memory. This PR converts them to streaming I/O so transfer memory no longer scales with file size, and adds atomic downloads + parallel bulk uploads.

## Peak memory (measured)

| Scenario | main | This PR | Reduction |
|---|---|---|---|
| 100 MB single file download | 240 MB | 6.3 MB | ~38x |
| 500 MB single file download | 1,159 MB | 7.2 MB | ~65x |
| 100 MB folder upload (10×10) | 329 MB | 7.5 MB | ~44x |
| 300 MB folder upload (10×30) | 681 MB | 7.9 MB | ~37x |

100 MB rows measured live against the `web-editor` dev server; 500/300 MB rows from local. Tool: `/usr/bin/time -l`. MD5 verified end-to-end.

## Per-path behavior

| Path | Before | After |
|---|---|---|
| `cp` single file download | `io.ReadAll` → `SaveFile` | `SaveStreamAtomic` direct to disk |
| `cp` recursive download | `io.ReadAll` whole zip | Stream zip to disk → unzip |
| `cp` folder upload | `utils.Zip()` in-memory zip | `SpoolToTempFile` + `ZipToWriter` |
| `cp` bulk multi-file upload | Sequential PUTs | 4-way concurrent PUTs |
| `package` upload | `bytes.Buffer` envelope | `SpoolToTempFile` + `SendMultipartStreamRequest` |
| `package` download | `io.ReadAll` → `SaveFile` | `SaveStreamAtomic` |

Single file upload was already streaming → unchanged.

## New utilities

- `utils.SaveStream`, `utils.SaveStreamAtomic` (temp file + atomic rename, symlink resolution, mode preservation)
- `utils.ZipToWriter`, `utils.SpoolToTempFile`, `utils.CleanupTempFile`
- `client.SendMultipartStreamRequest` (streaming multipart with `GetBody` for 307/308 redirect replay)
- Platform shims: `utils/file_replace_{unix,windows}.go` (Windows uses `windows.Rename`)
- `collectConcurrentFailures` helper, `bulkUploadConcurrency = 4`

## Compatibility & risks

- No public API changes.
- More temp file use (`os.TempDir`) — `SpoolToTempFile` surfaces clear errors when full.
- Folder upload adds disk I/O: net **faster** at 100 MB (-3.8 s), net **+14 s** at 300 MB. May grow on slow disks.
- `golang.org/x/sys` is now a direct dependency (used by `file_replace_windows.go`).
- Concurrency hardcoded to 4 — can be flag-exposed later.

## Test plan

- [x] `go test -race ./utils/... ./client/... ./api/ftp/... ./api/packages/...`
- [x] `golangci-lint run ./...`
- [x] `go build ./...`
- [x] Live measurement on `web-editor` for 100 MB download + folder upload (MD5 verified)
- [x] Multipart redirect replay covered by `TestSendMultipartStreamRequest_ReplaysFileBodyOnTemporaryRedirect`
- [x] Atomic write regressions: existing file preserved on read error, mode preserved, symlinks followed (incl. dangling)

## Out of scope

- Folder upload server-side `tee` error on missing remote dir (pre-existing, unrelated).
- `alpacon package` not measured live (production risk); identical swap pattern.
- >500 MB files / Windows targets not exercised.


utils/file_save.go
```diff
@@ -0,0 +1,141 @@
...
+
+func SaveStreamAtomic(fileName string, r io.Reader) (int64, error) {
+	targetName, err := resolveWritePath(fileName)
+	if err != nil {
+		return 0, err
+	}
+
+	dir := filepath.Dir(targetName)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return 0, fmt.Errorf("failed to create directories: %v", err)
```
geunwoonoh: Warning: error wrapping uses `%v` instead of `%w` here (and in `SaveStream`, `resolveWritePath`, `createReplacementTempFile`). `api/ftp/ftp.go` and `api/packages/packages.go` in this PR use `%w` consistently. Wrapping with `%v` breaks the unwrap chain, so `errors.Is(err, syscall.ENOSPC)` (disk-full detection, which the PR description calls out as a benefit) won't work. Recommend `%w`.

utils/file_save.go
```diff
@@ -0,0 +1,141 @@
...
+			nextName = filepath.Join(filepath.Dir(targetName), nextName)
+		}
+		targetName = nextName
+	}
+	return "", errors.New("too many symlinks while resolving write path")
+}
+
+func createReplacementTempFile(dir string, perm os.FileMode) (*os.File, error) {
+	for i := 0; i < 100; i++ {
+		name := filepath.Join(dir, fmt.Sprintf(".alpacon-%d-%d-%d.tmp", os.Getpid(), time.Now().UnixNano(), i))
```
geunwoonoh, line 130: Suggestion: this manual PID+UnixNano+counter loop reimplements `os.CreateTemp`. `os.CreateTemp(dir, ".alpacon-*.tmp")` keeps the temp file in the same directory (so the rename stays on one filesystem), and perm is already enforced afterward via `Chmod(existingPerm)` — so behavior is unchanged but simpler.

```suggestion
func createReplacementTempFile(dir string, perm os.FileMode) (*os.File, error) {
	file, err := os.CreateTemp(dir, ".alpacon-*.tmp")
	if err != nil {
		return nil, fmt.Errorf("failed to create temp file: %w", err)
	}
	if err := file.Chmod(perm); err != nil {
		_ = file.Close()
		_ = os.Remove(file.Name())
		return nil, fmt.Errorf("failed to set temp file permissions: %w", err)
	}
	return file, nil
}
```

api/ftp/ftp.go
```diff
@@ -31,6 +31,9 @@ const (
 	basePollTimeout    = 30 * time.Second
 	perFilePollTimeout = 10 * time.Second
 	perMBPollTimeout   = 5 * time.Second
+
+	bulkUploadConcurrency = 4
+	bulkPollConcurrency   = 8
```
geunwoonoh: Suggestion: the PR description says "Concurrency hardcoded to 4" but only mentions upload concurrency. This poll concurrency (8) isn't documented — a one-line comment about the intent would help future readers.

geunwoonoh (dismissed):
## Review summary

This PR converts the full-payload buffering in `alpacon cp` and `package upload/download` to streaming I/O. The core design is solid, and the test coverage is excellent (concurrency, atomic writes, symlinks, redirect replay, preserving the existing file on error). No blocking issues.

**Verified locally**: `go build ./...` ✅ / `go vet ./api/ftp/... ./api/packages/... ./client/... ./utils/...` ✅

---

### Warnings (3)

**1. `utils/utils.go:118` — `ReadFileFromPath` is now dead code**
This PR switched `UploadPackage` from `utils.ReadFileFromPath` to `os.Open`, removing its only caller. There are now zero callers across the codebase (tests included). Recommend removing it as part of this change.

**2. `utils/file_save.go` — error wrapping uses `%v` instead of `%w` (breaks the unwrap chain)**
`SaveStream` / `SaveStreamAtomic` / `resolveWritePath` / `createReplacementTempFile` all use `fmt.Errorf("...: %v", err)`, whereas `api/ftp/ftp.go` and `api/packages/packages.go` in this same PR consistently use `%w`. The PR description highlights "surfaces clear errors when full (disk full)" as a benefit, but wrapping with `%v` makes `errors.Is(err, syscall.ENOSPC)` impossible. Recommend standardizing on `%w` for consistency and Go convention.

**3. `api/ftp/ftp.go:82` `uploadToS3` — no `GetBody` on the S3 PUT path (no replay on redirect/retry)**
The multipart package upload (`client.SendMultipartStreamRequest`) carefully sets `GetBody` for `*os.File`, but the direct S3 PUT hides `io.Seeker` via `readOnly{file}`, so `http` cannot synthesize `GetBody`. The previous single-file upload used `bytes.NewReader` (implements Seeker), so `http` set `GetBody` automatically and 307/308 replay worked; the single-folder upload path (`readOnly{zipFile}`) can no longer replay. Real-world risk is low since presigned S3 URLs rarely redirect, but it is a behavior change worth noting. If desired, set `req.GetBody = func() (io.ReadCloser, error) { return os.Open(f.Name()) }` for `*os.File` in `uploadToS3` to keep it consistent.

---

### Suggestions (5)

**1. `utils/file_save.go:128` `createReplacementTempFile` — reimplements `os.CreateTemp`**
The manual PID+UnixNano+counter loop (100 retries) can be replaced with `os.CreateTemp(dir, ".alpacon-*.tmp")` (same directory still guarantees a same-filesystem rename). The perm is already enforced afterward via `Chmod(existingPerm)`, so behavior is unchanged. Simpler and uses the standard library.

**2. `api/ftp/ftp.go` `saveDownloadedURL` — redundant intermediate variable**
The non-recursive branch's `written, err := ...; if err != nil { return written, err }; return written, nil` can collapse to `return fetchFromURLToFile(...)`.

**3. `api/ftp/ftp.go:36` `bulkPollConcurrency = 8`**
The PR description says "Concurrency hardcoded to 4" but only mentions upload concurrency (4); poll concurrency (8) is not documented. A one-line note in the description/comment about the intent would help.

**4. `api/ftp/ftp.go` `fetchFromURLToFile` — `resp.Body` leak on retry paths (pre-existing, but this function was refactored)**
The 4xx immediate-return and 5xx retry (`time.Sleep` then continue) paths don't close the prior `resp.Body`; the `defer` only covers the final response after the loop. Since the function is already being touched, closing the body for non-OK responses inside the loop would be a good cleanup.

**5. `utils/file_save.go:18` `SaveStream` — exported but no external callers**
It's only used by `SaveFile` and tests. If there's no intended external use, it could be unexported (`saveStream`). Not required.

---

### Highlights 👍

- **Preserves the existing file on error**: the temp-file → atomic-rename design ensures a failed download never corrupts the existing file, clearly verified by tests (`...ReadErrorKeepsExistingFile`, `...NoContentDoesNotOverwrite...`).
- **`reserveDownloadArchiveTempPath`**: reserves a hidden temp path so it never collides with user-visible files the server returns (e.g. `download.zip`) — nice detail.
- **`collectConcurrentFailures`**: writes into per-index slots so output ordering is deterministic, and bounds concurrency cleanly with a semaphore. No data races.
- **`client.go` refactor**: flattening the `if/else` chain into named booleans (`statusIsError` / `postUnexpected` / `deleteUnexpected`) keeps the behavior identical while improving readability.
- **Redirect replay test** (`TestSendMultipartStreamRequest_ReplaysFileBodyOnTemporaryRedirect`) plus symlink / dangling-symlink cases are covered.
- **`server_create.go`**: changing `default` to an explicit `case "token-install"` plus an unknown-method error makes it more robust (slightly out of scope for streaming — a one-line mention in the description would be nice).

---

**Verdict**: APPROVED (nits only)

- [x] Code follows stack conventions — only the `%w` standardization to polish
- [x] Tests included and passing — excellent coverage
- [x] No security vulnerabilities — no hardcoded secrets / injection
- [x] No critical performance regression — the folder-upload disk I/O increase is documented in the description

jisung-02:
Thanks for the review. Here is how each item was addressed.

**Applied**
- **`ReadFileFromPath` dead code** — removed; it had no callers.
- **`file_save.go` error wrapping** — switched `%v` to `%w` across all 13 sites to preserve the unwrap chain. Left `destination is a directory: %s` as-is since it is a path string, not an error.
- **`uploadToS3` missing GetBody** — restored redirect replay that was lost in the streaming switch. Added an `osFileFrom` helper that unwraps `readOnly` to recover the `*os.File` and sets `req.GetBody` (same pattern as the multipart upload path).
- **`saveDownloadedURL` intermediate variable** — collapsed the non-recursive branch to `return fetchFromURLToFile(...)`.
- **`bulkPollConcurrency` intent** — added a one-line note: uploads (4) transfer payload bytes so they are kept low, while status polls (8) are lightweight so more are allowed.
- **`SaveStream` unexport** — lowered to `saveStream` since it has no external callers. Kept `SaveFile` exported as it is the public API.

**Not changed**
- **`fetchFromURLToFile` body leak** — no leak: at `ftp.go:455` non-OK responses are closed before the 4xx return and the retry branch, and the 200 OK case is handled by the defer.
- **`createReplacementTempFile` -> `os.CreateTemp`** — kept the current loop. `os.CreateTemp` always creates with 0600, but the current code gives new files the umask-applied default (typically 0644). Preserving that behavior would require reading the umask and calling Chmod, and `syscall.Umask` is unavailable on Windows, so it would need platform-split files. The cost outweighs the simplification, so the manual loop stays.

Build and `go test -race ./...` pass.

geunwoonoh (approved):

Issue #181, opened 2026-05-19
Surface the server-side WorkSession gate as first-class CLI signals


jisung-02:
## Background

`alpacon-server` requires a **WorkSession** for access-path operations called with interactive authentication (Session, Auth0 JWT, `APIToken` with `source='login'`). `ServiceToken` and `APIToken` with `source != 'login'` (PATs, server-registration tokens) bypass the requirement.

The gate covers four operations:

| Operation | Serializer |
|---|---|
| `websh`   | `websh/api/serializers.py` (`work_session_operation = 'websh'`) |
| `webftp`  | `webftp/api/serializers.py` (`work_session_operation = 'webftp'`) |
| `command` | `events/api/serializers.py` (`work_session_operation = 'command'`) |
| `sudo`    | `sudo/api/serializers.py` (`work_session_operation = 'sudo'`) |

Error codes the server may return (`ErrorCode` members in `alpacon-server/utils/error_codes.py`, raised from `work_sessions/services.py`):

- `WORK_SESSION_REQUIRED` — interactive auth without a session
- `WORK_SESSION_NOT_USABLE` — binding op against an unusable status
- `WORK_SESSION_NOT_ACTIVE` — runtime op while session is not yet active (or `starts_at > now`)
- `WORK_SESSION_EXPIRED` — `expires_at <= now`
- `WORK_SESSION_SCOPE_NOT_ALLOWED` — operation not in `scopes` (or agent + `websh`)
- `WORK_SESSION_SERVER_NOT_ALLOWED` — target server not in `servers`
- `WORK_SESSION_ASSIGNEE_MISMATCH` — auth principal does not match the assignee

## Status quo in the CLI

- The `alpacon worksession` group is already implemented (`create`, `use`, `current`, `list`, `approve`, `extend`, `revoke`, `timeline`, `complete`, `reject`, `describe`, `recording`, `activate`).
- The `--output table|json` persistent flag is wired via `utils.OutputFormat`.
- All error exits go through `utils.CliError(WithExit)` — red "Error: <msg>" on stderr followed by `os.Exit(1)`. Every failure uses exit code `1` (the one exception is `cmd/tunnel/run.go:205`, which propagates a remote child process exit code).
- `utils.ParseErrorResponse` already knows how to extract `code` / `source` from server responses formatted either as JSON or as `code: ...; source: ...`.
[...]

PR #216, opened 2026-06-04
feat(work-session): emit JSON error envelope on --output json error paths

jisung-02:
## Summary

Closes the remaining gap in #181: with `--output json`, work-session command failures now emit a machine-readable JSON error envelope on stderr instead of plain colored text. Also adds a README section documenting exit codes and the envelope schema.

### JSON error envelope (`utils/error_envelope.go`)

Two shared helpers, applied to all 61 error exits across every work-session subcommand:

- `CliErrorEnvelopeWithExit(operation, err, format, ...)`—server/runtime errors; extracts the server `code` via `ParseErrorResponse`
- `CliUsageErrorEnvelopeWithExit(operation, format, ...)`—local validation failures; fixed `error_code: "usage_error"`

Envelope shape (stderr, exit 1):

```json
{
  "ok": false,
  "exit_code": 1,
  "error_code": "work_session_not_usable",
  "message": "Failed to extend work session: ...",
  "context": { "operation": "extend" }
}
```

- `error_code` is omitted when no server code is available (connection/local-state failures)
- Table mode (default) is byte-identical to the previous output; exit codes unchanged
- Operation identifiers are shared constants (`opActivate`…`opUse`), so success `operation` and error `context.operation` always agree

### README

New "When a command is denied" section: exit code table, envelope schema, work-session denial codes, and branching examples for scripts/AI agents.

## Out of scope

- `cmd/worksession/resolve.go`—only reached via access commands (cp/websh/tunnel/exec); those paths keep their existing error handling
- `Refreshing access token...` stderr notice can precede the JSON envelope during token refresh—pre-existing behavior, not introduced here

## Test plan

- [x] `go test -race ./...` pass
- [x] New contract tests (`worksession_error_json_test.go`): server-code / usage-error / no-code shapes verified via subprocess against `httptest` server
- [x] Manual verification against a live workspace: all three envelope shapes + table-mode regression
- [x] `golangci-lint run` 0 issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ What each refusal code means and what to do next:
 | `work_session_assignee_mismatch` | session assigned to another principal | `work-session use <ID>` with your own session |
 | `work_session_not_usable` | session is no longer usable | `work-session create --use` |
 
-`work-session` subcommand failures (`create`, `use`, `extend`, ...) also emit a JSON error envelope under `--output json`, with exit code `1` and `error_code` carrying the server code when available (`usage_error` for local flag/argument errors). Run `alpacon whoami` to check upfront whether a work session is required for your auth.
+`work-session` subcommand failures (`create`, `use`, `extend`, ...) also emit a JSON error envelope under `--output json`, with exit code `1` and `error_code` carrying the server code when available (`usage_error` for local flag/argument errors). These envelopes may share an `error_code` with the gate-denial envelopes above—distinguish a subcommand failure (`exit_code: 1`) from a gate denial (`exit_code: 3`) via `exit_code`, not `error_code` alone. Run `alpacon whoami` to check upfront whether a work session is required for your auth.
 
 ## Exit codes
 
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ Run `alpacon --help` for the full list, or `alpacon <command> --help` for detail
 
 ## When a command is denied
 
-Under interactive auth (browser login), `websh`, `exec`, `cp`, and `tunnel` require an active work session. Without one, the command is refused with a diagnostic and exit code `3`:
+Under interactive auth (browser login), `websh`, `exec`, `cp`, `edit`, and `tunnel` require an active work session. Without one, the command is refused with a diagnostic and exit code `3`:
 
 ```
 Error: the command operation requires an active WorkSession on this authentication.
--- a/README.md
+++ b/README.md
@@ -224,6 +224,63 @@ $ alpacon audit <filters>                        # workspace audit log
 
 Run `alpacon --help` for the full list, or `alpacon <command> --help` for details on any command.
 
+## When a command is denied
+
+Under interactive auth (browser login), `websh`, `exec`, `cp`, and `tunnel` require an active work session. Without one, the command is refused with a diagnostic and exit code `3`:
+
+```
+Error: the command operation requires an active WorkSession on this authentication.
+
+  auth          : Browser login (interactive)
+  reason        : no WorkSession selected for this shell
+  required scope: command
+  target server : prod-1
+
+Next:
+  alpacon work-session create --scope command --server prod-1 --expires-in 1h --purpose "<intent>" --use  # create a new session and attach it
+  alpacon work-session ls --status active  # or reuse an existing active session
+  alpacon work-session use <ID>
+
+Note: Tokens issued by Alpacon (service or personal API token) bypass this check.
+```
+
+With `--output json`, the same refusal is a structured envelope on stderr—scripts and AI agents branch on `error_code` and run `next_actions` verbatim:
+
+```json
+{
+  "ok": false,
+  "exit_code": 3,
+  "error_code": "work_session_required",
+  "message": "the command operation requires an active WorkSession on this authentication.",
+  "reason": "no WorkSession selected for this shell",
+  "context": {
+    "auth_method": "Browser login",
+    "required_scope": "command",
+    "target_servers": ["prod-1"],
+    "current_worksession": null
+  },
+  "next_actions": [
+    "alpacon work-session create --scope command --server prod-1 --expires-in 1h --purpose \"<intent>\" --use  # create a new session and attach it",
+    "alpacon work-session ls --status active  # or reuse an existing active session",
+    "alpacon work-session use <ID>"
+  ]
+}
+```
+
+What each refusal code means and what to do next:
+
+| `error_code` | Meaning | Next |
+|---|---|---|
+| `work_session_required` | no session selected for this shell | `work-session create --use` or `work-session use <ID>` |
+| `work_session_not_active` | session not yet active (check `starts_at`) | wait, then `work-session current` |
+| `work_session_expired` | session has expired | `work-session extend <ID>` or create a new one |
+| `work_session_scope_not_allowed` | operation not in session scopes | create a session with the right `--scope` |
+| `work_session_server_not_allowed` | target server not in session | create a session with the right `--server` |
+| `work_session_assignee_mismatch` | session assigned to another principal | `work-session use <ID>` with your own session |
+| `work_session_not_usable` | session is no longer usable | `work-session create --use` |
+
+`work-session` subcommand failures (`create`, `use`, `extend`, ...) also emit a JSON error envelope under `--output json`, with exit code `1` and `error_code` carrying the server code when available (`usage_error` for local flag/argument errors). Run `alpacon whoami` to check upfront whether a work session is required for your auth.
+
 ## Exit codes
 
 | Code | Meaning |
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ What each refusal code means and what to do next:
 | `1`  | General error (network failure, server error, etc.) |
 | `2`  | Usage error (invalid flags or arguments) |
 | `3`  | WorkSession gate denied—the active session does not authorize this action |
+| `4`  | Pending human approval—the action is awaiting an out-of-band approve/reject in the Alpacon console (web/Slack), not refused. For `exec`, re-run the command after approval (or pass `--wait` on the original command to block); for `work-session create` the session already exists—after approval attach it with `alpacon work-session use <id>` (or pass `--wait` on the original create to block). Under `--output json`, a `{"status":"pending_approval", ...}` object is emitted. Returned by `exec` on a `SUDO_APPROVAL_REQUIRED` sudo denial and by `work-session create` when the session lands pending |
 
 ## Contributing
 
--- a/cmd/worksession/worksession_create.go
+++ b/cmd/worksession/worksession_create.go
@@ -76,19 +76,19 @@ session with 'alpacon work-session update <id> --sudo "<command>"'.`,
 		purpose = strings.TrimSpace(purpose)
 		if purpose == "" {
 			if !utils.IsInteractiveShell() {
-				utils.CliErrorWithExit("Non-interactive mode requires --purpose.")
+				utils.CliUsageErrorEnvelopeWithExit(opCreate, "Non-interactive mode requires --purpose.")
 			}
 			purpose = utils.PromptForRequiredInput("Purpose: ")
 		}
 		if len(createScopes) == 0 {
 			if !utils.IsInteractiveShell() {
-				utils.CliErrorWithExit("Non-interactive mode requires --scope.")
+				utils.CliUsageErrorEnvelopeWithExit(opCreate, "Non-interactive mode requires --scope.")
 			}
 			createScopes = splitCSV(utils.PromptForRequiredInput("Scopes (comma-separated, e.g. command,websh): "))
 		}
 		if len(createServers) == 0 {
 			if !utils.IsInteractiveShell() {
-				utils.CliErrorWithExit("Non-interactive mode requires --server.")
+				utils.CliUsageErrorEnvelopeWithExit(opCreate, "Non-interactive mode requires --server.")
 			}
 			createServers = splitCSV(utils.PromptForRequiredInput("Servers (comma-separated server names): "))
 		}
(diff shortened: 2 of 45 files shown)


cmd/worksession/worksession_use.go
```diff
@@ -32,15 +32,15 @@ Pass --unset (with no SESSION_ID) to clear the active work-session.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if unsetActiveWorkSession {
 			if len(args) > 0 {
-				utils.CliErrorWithExit("--unset cannot be combined with a SESSION_ID argument")
+				utils.CliUsageErrorEnvelopeWithExit(opUse, "--unset cannot be combined with a SESSION_ID argument")
 			}
```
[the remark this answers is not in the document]
jisung-02: Fixed. The `--unset` path now reports `context.operation` as `unset`, matching the success outputs.

cmd/worksession/worksession_use.go
```diff
@@ -50,12 +50,12 @@ Pass --unset (with no SESSION_ID) to clear the active work-session.`,
 				return
 			}
 			if err := RunUnset(); err != nil {
-				utils.CliErrorWithExit("%s", err)
+				utils.CliErrorEnvelopeWithExit(opUse, err, "%s", err)
 			}
```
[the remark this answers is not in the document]
jisung-02: Fixed. `RunUnset()` failures now report `context.operation` as `unset` as well.

utils/error_envelope_test.go
```diff
@@ -0,0 +1,54 @@
...
+	rendered, err := FormatJSON(env)
+	assert.NoError(t, err)
+	assert.NotContains(t, rendered, "error_code")
+	assert.Contains(t, rendered, `"operation": "use"`)
+}
+
+func TestBuildCliErrorEnvelopeFromErr_ExtractsServerCode(t *testing.T) {
+	// ParseErrorResponse parses the "msg; code: X; source: Y" form produced by parseAPIError.
+	err := errors.New("WorkSession is not usable; code: work_session_not_usable; source: work_session")
+	env := buildCliErrorEnvelopeFromErr("extend", err, "Failed to extend work session: refused.")
```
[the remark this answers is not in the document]
jisung-02: Fixed. Reworded the comment: the string form is the legacy fallback parsed by `ParseErrorResponse`, not what `parseAPIError` produces today.

README.md
```diff
@@ -224,6 +224,63 @@ $ alpacon audit <filters>                        # workspace audit log
...
+|---|---|---|
+| `work_session_required` | no session selected for this shell | `work-session create --use` or `work-session use <ID>` |
+| `work_session_not_active` | session not yet active (check `starts_at`) | wait, then `work-session current` |
+| `work_session_expired` | session has expired | `work-session extend <ID>` or create a new one |
+| `work_session_scope_not_allowed` | operation not in session scopes | create a session with the right `--scope` |
+| `work_session_server_not_allowed` | target server not in session | create a session with the right `--server` |
+| `work_session_assignee_mismatch` | session assigned to another principal | `work-session use <ID>` with your own session |
+| `work_session_not_usable` | session is no longer usable | `work-session create --use` |
+
+`work-session` subcommand failures (`create`, `use`, `extend`, ...) also emit a JSON error envelope under `--output json`, with exit code `1` and `error_code` carrying the server code when available (`usage_error` for local flag/argument errors). Run `alpacon whoami` to check upfront whether a work session is required for your auth.
```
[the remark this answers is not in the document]
jisung-02: The exit code table predates this PR (added in fa29159) and documents the reserved exit-code contract; this PR does not change exit-code behavior.

- Work-session local validation errors already exited `1` before this PR (via `CliErrorWithExit`); changing them to `2` would be a behavior change for scripts that branch on exit codes, which is out of scope here.
- The new sentence documents actual current behavior (`exit_code: 1` + `error_code: usage_error`), and the JSON `error_code` field is the reliable way to distinguish usage errors under `--output json`.
- Reconciling the table with runtime behavior for invalid flags/arguments is a pre-existing discrepancy that should be addressed separately.

cmd/worksession/worksession_output.go
```diff
@@ -72,6 +72,7 @@ func activeWorkSessionSetMessage(successPrefix, id, desc string) string {
 
 func printWorkSessionMutationJSON(output workSessionMutationOutput) {
 	if err := utils.PrintJSONValue(os.Stdout, output); err != nil {
+		// Plain text on purpose: the envelope uses the same serializer that just failed.
 		utils.CliErrorWithExit("Failed to marshal work-session result: %s", err)
 	}
```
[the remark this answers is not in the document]
jisung-02: Fixed. The marshal-failure path now emits the JSON error envelope via `CliErrorEnvelopeWithExit`, relying on `PrintJSONError`'s minimal-JSON fallback if marshalling fails again.

README.md
```diff
@@ -224,6 +224,63 @@ $ alpacon audit <filters>                        # workspace audit log
 
 Run `alpacon --help` for the full list, or `alpacon <command> --help` for details on any command.
 
+## When a command is denied
+
+Under interactive auth (browser login), `websh`, `exec`, `cp`, and `tunnel` require an active work session. Without one, the command is refused with a diagnostic and exit code `3`:
```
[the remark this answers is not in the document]
jisung-02: Fixed. Added `edit` to the list of work-session-gated commands.

utils/error_envelope.go
```diff
@@ -0,0 +1,53 @@
...
+// cliErrorCtx mirrors the "operation" key of the mutation success outputs.
+type cliErrorCtx struct {
+	Operation string `json:"operation"`
+}
+
+// buildCliErrorEnvelope assembles the error envelope; an empty errorCode is omitted from the output.
+func buildCliErrorEnvelope(operation, errorCode, message string) JSONErrorEnvelope[cliErrorCtx] {
+	return JSONErrorEnvelope[cliErrorCtx]{
+		OK:        false,
+		ExitCode:  1,
```
geunwoonoh, line 20: Warning: `exit_code` is hard-coded to `1` here. The same server `error_code` (e.g. `work_session_not_usable`) is emitted with `exit_code: 3` on the access-command gate path (`utils/worksession_error.go`), where the README documents it as a gate-denial code. Consumers that branch on `error_code` alone — which the README encourages — could conflate a subcommand failure (exit 1) with a gate denial (exit 3). Consider a one-line README note clarifying that these envelopes are always exit 1 and must be distinguished via `exit_code`.

cmd/worksession/worksession_create.go
```diff
@@ -191,48 +191,48 @@ session with 'alpacon work-session update <id> --sudo "<command>"'.`,
 			// Re-fetch so the serialized JSON matches the --wait --use path.
 			activeSession, err := RunUseSession(ac, session.ID)
 			if err != nil {
-				utils.CliErrorWithExit("Work session created (%s) but failed to set as active: %s. Run 'alpacon work-session use %s' to retry.", session.ID, err, session.ID)
+				utils.CliErrorEnvelopeWithExit(opCreate, err, "Work session created (%s) but failed to set as active: %s. Run 'alpacon work-session use %s' to retry.", session.ID, err, session.ID)
```
geunwoonoh: Warning: this is a partial-success path — the session *was* created, only the active-set step failed. Passing `err` for code extraction means that if `err` carries a server work_session code, the envelope's `error_code` will look like a gate denial despite the successful create. `nil` (or a usage-class code) may describe the outcome more accurately for this and the analogous path below (`RunUse` failure).

cmd/worksession/worksession_extend.go
```diff
@@ -23,26 +23,26 @@ var workSessionExtendCmd = &cobra.Command{
...
 		ac, err := client.NewAlpaconAPIClient()
 		if err != nil {
-			utils.CliErrorWithExit("Connection to Alpacon API failed: %s. Consider re-logging.", err)
+			utils.CliErrorEnvelopeWithExit(opExtend, err, "Connection to Alpacon API failed: %s. Consider re-logging.", err)
 		}
 
 		req := wsapi.WorkSessionExtendRequest{ExpiresAt: expiresAtVal}
 		if err := wsapi.ExtendWorkSession(ac, args[0], req); err != nil {
-			utils.CliErrorWithExit("Failed to extend work session: %s.", err)
+			utils.CliErrorEnvelopeWithExit(opExtend, err, "Failed to extend work session: %s.", err)
```
geunwoonoh, line 45: Suggestion: `err` is passed twice — once as the code-extraction argument and once as a format arg. This is intentional and documented, but verbose at call sites. If this helper's call frequency grows, consider an overload that defaults the message to `err.Error()`.

cmd/worksession/worksession_recording.go
```diff
@@ -44,12 +44,12 @@ var workSessionRecordingCmd = &cobra.Command{
 		}
 
 		if len(recordings) == 0 {
-			utils.CliErrorWithExit("No recordings found for session %s.", sessionID)
+			utils.CliErrorEnvelopeWithExit(opRecording, nil, "No recordings found for session %s.", sessionID)
```
geunwoonoh, line 48: Suggestion: classification inconsistency — "No recordings found" emits no `error_code` (nil err), while "index out of range" just below emits `usage_error`. Both are local conditions; the not-found vs. bad-flag distinction is defensible, but a one-line comment on the intended split would aid future readers.

geunwoonoh (commented):
## Review: #216 - feat(work-session): emit JSON error envelope on --output json error paths

**Verification**: `go vet` clean / `go test ./utils/... ./cmd/worksession/...` passing (verified locally)

### Summary

The design is solid and preserves existing behavior well (byte-identical table mode, unchanged exit codes). Sharing the operation identifiers as constants and delegating to `CliErrorWithExit` in non-JSON mode are both clean choices. No merge-blocking issues.

### Warnings (2)

**1. The same `error_code` can be emitted with different `exit_code` values — possible script-branching confusion.**
`buildCliErrorEnvelope` hard-codes `exit_code: 1`, and the new test asserts that a server code like `work_session_not_usable` is emitted with `exit_code: 1` on an `extend` failure. But the same `work_session_not_usable` code is emitted with `exit_code: 3` (`ExitCodeWorkSessionDenied`) on the access-command gate path (`utils/worksession_error.go`), and the README documents it as a gate-denial (exit 3) code. Since the README also advises consumers to "branch on `error_code`", an agent/script keying on `error_code` alone could conflate a subcommand failure (exit 1) with an access-gate denial (exit 3). It is not a true bug — `exit_code` and the `context` shape differ — but a one-line README note would make the distinction explicit (e.g. "subcommand-failure envelopes are always exit 1 and may share an `error_code` with gate-denial envelopes — distinguish via `exit_code`").

**2. Misleading `error_code` on partial-success paths.**
The "session created but failed to set active" paths in `worksession_create.go` extract the code from `err` even though the session was actually created. If that `err` carries a server work_session code, the envelope's `error_code` would look like a gate denial despite the partial success. `nil` (or a usage-class code) may be more accurate for these two paths.

### Suggestions (3)

**1.** The helper signature passes `err` twice (once for code extraction, once as a format arg). Intentional and well-documented, but the call sites are a little verbose; an overload that defaults the message to `err.Error()` could be considered if call frequency grows.

**2.** Classification inconsistency in `worksession_recording.go`: "No recordings found" emits no `error_code` (nil err), while "index out of range" emits `usage_error`. Both are local conditions; the not-found vs. bad-flag distinction is defensible, but a one-line comment on the intent would help.

**3.** `JSONErrorEnvelope.Context` uses `json:"context,omitempty"`, but `omitempty` has no effect on a non-pointer struct — `context` always renders. Harmless today since `operation` is always set, but worth knowing if the intent was to omit an empty context.

### Highlights

- Operation-identifier constants guarantee success `operation` and error `context.operation` always agree.
- Table mode stays byte-identical by delegating to `CliErrorWithExit`; exit codes unchanged.
- Excellent subprocess contract tests asserting real exit code + stderr JSON shape across server-code / usage-error / no-code forms.
- `error_code` is omitted when no server code is available, cleanly separating local vs. server failures.
- Out-of-scope `resolve.go` is correctly left alone and documented.

### Go conventions

`const → type → func` order respected, errcheck patterns satisfied, lowercase error strings, English comments. No violations. Only one `CliErrorWithExit` remains in the package (`resolve.go`), matching the stated scope.

**Verdict**: Approvable — no critical issues. Recommend only the README note for Warning #1; the rest are optional.

utils/error_envelope.go
jisung-02: Fixed in a83db4b. Added a sentence to the README's work-session subcommand envelope note: these envelopes may share an `error_code` with the gate-denial envelopes, so consumers should distinguish a subcommand failure (`exit_code: 1`) from a gate denial (`exit_code: 3`) via `exit_code`, not `error_code` alone.

cmd/worksession/worksession_create.go
jisung-02: Fixed in 2cb2a6b. Both partial-success paths (created-but-use-failed and approved-but-use-failed) now pass `nil` for code extraction, so the envelope omits `error_code` instead of reading like a create/gate failure. The retry guidance in the message is unchanged.

cmd/worksession/worksession_extend.go
jisung-02: Deferring this for now. Most call sites wrap the error with contextual prose (e.g. "Failed to extend work session: %s."), so a bare `err.Error()` default would rarely apply, and adding an overload today would mean two helpers to choose between. Will introduce it if call frequency grows, as you suggested.

cmd/worksession/worksession_recording.go
jisung-02: Fixed in a284e67. Added a one-line comment above the branch documenting the split: no recordings is a not-found state (no `error_code`), while a bad `--index` is a `usage_error`.

jisung-02:
Addressed @geunwoonoh's review (see inline replies). Also applied Suggestion 3 in 66d20cd: `omitempty` has no effect on a non-pointer struct field, so the `Context` tag is now `json:"context"`—output is unchanged and the tag matches actual behavior.

geunwoonoh (dismissed):
Please resolve the merge conflicts.

geunwoonoh (approved):

PR #237, opened 2026-06-18
refactor: remove redundant wrappers and consolidate stdlib helper

jisung-02:
## Overview

While auditing the codebase for over-engineering, I found a few single-caller wrappers and a duplicated helper. This PR cleans them up. No behavior change; net 35 lines removed.

### 1. Remove `downloadSingleFile` (`api/ftp/ftp.go`)

A pass-through that discarded the result and delegated straight to `downloadSingleFileWithResult`. The sole caller now calls `downloadSingleFileWithResult` directly.

### 2. Replace `splitCSV` with `utils.SplitAndTrim` (`cmd/worksession/`)

`splitCSV` was a duplicate of the existing `utils.SplitAndTrim(s, sep)`. Replaced all three call sites with `utils.SplitAndTrim(s, ",")` and deleted `splitCSV`. The only difference is the return value on empty input (`[]string{}` vs `nil`), which is irrelevant here since every caller only uses `len()`/`range`. Moved `TestSplitCSV` to `TestSplitAndTrim` in the `utils` package to preserve coverage.

### 3. Collapse tunnel signal wrappers (`cmd/tunnel/run.go`)

`interruptProcess` was a one-line delegation to `signalProcess(cmd, os.Interrupt)`; inlined at both call sites. `signalCommandProcess`/`killCommandProcess` duplicated the nil check already performed by their parents `signalProcess`/`killProcess`, making the inner check a dead branch — folded into the parents.

## Verification

`go build`, `go vet`, and `go test -race` all pass for `utils/`, `cmd/tunnel/`, `cmd/worksession/`, and `api/ftp/`.

## Intentionally kept

`truncateBody` and `createFolderZipTempFile` each have two callers; inlining them would duplicate code, so they stay.

### DIFF
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -578,11 +578,6 @@ func saveDownloadedURL(httpClient *http.Client, url, dest, remotePath string, re
 	return filePath, written, err
 }
 
-func downloadSingleFile(ac *client.AlpaconClient, remotePath, dest, serverID, username, groupname, resourceType, workSessionID string, recursive bool) error {
-	_, err := downloadSingleFileWithResult(ac, remotePath, dest, serverID, username, groupname, resourceType, workSessionID, recursive)
-	return err
-}
-
 func downloadSingleFileWithResult(ac *client.AlpaconClient, remotePath, dest, serverID, username, groupname, resourceType, workSessionID string, recursive bool) (DownloadedFile, error) {
 	downloadRequest := &DownloadRequest{
 		Path:         remotePath,
@@ -737,7 +732,8 @@ func DownloadFile(ac *client.AlpaconClient, sources []string, dest, username, gr
 		resourceType = "folder"
 	}
 
-	return downloadSingleFile(ac, remotePaths[0], dest, serverID, username, groupname, resourceType, workSessionID, recursive)
+	_, err = downloadSingleFileWithResult(ac, remotePaths[0], dest, serverID, username, groupname, resourceType, workSessionID, recursive)
+	return err
 }
 
 func DownloadFileToPath(ac *client.AlpaconClient, serverName, remotePath, localPath, username, groupname, workSessionID string) (DownloadedFile, error) {
--- a/cmd/tunnel/run.go
+++ b/cmd/tunnel/run.go
@@ -117,7 +117,7 @@ func monitorLocalCommand(runtime tunnelCommandRuntime, localCmd *exec.Cmd) (int,
 			}
 
 			utils.CliWarning("Tunnel closed: %s", cause)
-			interruptProcess(localCmd)
+			signalProcess(localCmd, os.Interrupt)
 			err := waitForProcessExit(waitCh, localCmd, remoteCloseGracePeriod)
 			exitCode := exitCodeFromProcessError(err)
 			if exitCode == 0 {
@@ -130,7 +130,7 @@ func monitorLocalCommand(runtime tunnelCommandRuntime, localCmd *exec.Cmd) (int,
 			if interruptCount == 1 {
 				if sig == os.Interrupt {
 					utils.CliInfo("Interrupt received. Stopping local command... (Press Ctrl+C again to force stop)")
-					interruptProcess(localCmd)
+					signalProcess(localCmd, os.Interrupt)
 				} else {
 					utils.CliInfo("Termination signal received. Stopping local command...")
 					signalProcess(localCmd, sig)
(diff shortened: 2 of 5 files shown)


api/ftp/ftp.go
```diff
@@ -578,11 +578,6 @@ func saveDownloadedURL(httpClient *http.Client, url, dest, remotePath string, re
 	return filePath, written, err
 }
 
-func downloadSingleFile(ac *client.AlpaconClient, remotePath, dest, serverID, username, groupname, resourceType, workSessionID string, recursive bool) error {
-	_, err := downloadSingleFileWithResult(ac, remotePath, dest, serverID, username, groupname, resourceType, workSessionID, recursive)
-	return err
-}
-
 func downloadSingleFileWithResult(ac *client.AlpaconClient, remotePath, dest, serverID, username, groupname, resourceType, workSessionID string, recursive bool) (DownloadedFile, error) {
```
geunwoonoh, line 581: Suggestion (optional): Now that its sibling `downloadSingleFile` is gone, the `WithResult` suffix no longer contrasts with anything and reads a bit awkwardly. Renaming this to `downloadSingleFile` would be cleaner. That said, it has two callers and this rename is orthogonal to the wrapper-removal goal of this PR, so leaving it out of scope is equally reasonable.

geunwoonoh (approved):
## Summary

A refactoring PR that cleans up over-engineering in the codebase. It removes three single-caller wrapper functions and one duplicated helper.

1. **Remove `downloadSingleFile`** (`api/ftp/ftp.go`) — a pass-through wrapper that discarded the result and delegated straight to `downloadSingleFileWithResult`. The sole caller now calls it directly.
2. **Replace `splitCSV` with `utils.SplitAndTrim`** (`cmd/worksession/worksession_create.go`) — a helper that was effectively a duplicate of `utils.SplitAndTrim`. All three call sites replaced and the helper deleted; the test was moved into the `utils` package.
3. **Collapse tunnel signal wrappers** (`cmd/tunnel/run.go`) — `interruptProcess` inlined; the redundant nil checks in `signalCommandProcess`/`killCommandProcess` (a dead branch, since the parents already guard) folded into the parents.

## Verification (reproduced locally)

Ran everything directly and confirmed all green:

- `go build ./...` -> pass
- `go vet` (4 changed packages) -> pass
- `go test` (4 changed packages) -> all `ok`
- Orphan-reference scan: no remaining references to `interruptProcess` / `signalCommandProcess` / `killCommandProcess` / `downloadSingleFile` / `splitCSV`
- Import hygiene: `errors` and `os` in `cmd/tunnel/run.go`, and `strings` in `worksession_create.go`, are all still used -> no unused imports

## Critical Issues (0)

None.

## Warnings (0)

None. This is a pure refactor with no behavior change, and every potential risk point was checked and found safe.

- **`SplitAndTrim` empty-input return difference (`nil` vs `[]string{}`)** — deliberately reviewed and safe. All three call sites use the result only via `len()`/`range` (`worksession_create.go:134`) or `utils.CompactStrings` (`worksession_create.go:165`), so a nil slice does not change behavior. Confirmed it does not flow into a JSON serialization path. The test was correctly updated to expect `nil`.
- **`err` reuse in `DownloadFile`** (`ftp.go:735`) — `err` is already declared at `ftp.go:721`, so the `_, err =` (not `:=`) reassignment is correct. The discarded `DownloadedFile` is unnecessary since this public API returns only an error.

## Suggestions (1, optional)

1. **[api/ftp/ftp.go:581] Tidy up the `downloadSingleFileWithResult` name (optional)** — with its sibling `downloadSingleFile` gone, the `WithResult` suffix has nothing to contrast with and reads a bit awkwardly. Renaming to `downloadSingleFile` would be cleaner. It has two callers though, and this rename is orthogonal to this PR's wrapper-removal goal, so keeping it out of scope is equally reasonable.

## Highlights

- **Go conventions followed**: the errcheck pattern (ignoring `errors.Is(err, os.ErrProcessDone)` on signal/kill failures) is preserved, and the nil guards are correctly retained in the parent functions, keeping behavior identical.
- **Dead branch correctly identified**: spotted that the inner nil checks in `signalCommandProcess`/`killCommandProcess` were unreachable due to the parents' (`signalProcess`/`killProcess`) guards, and folded them safely.
- **Test coverage preserved**: rather than dropping the test when deleting `splitCSV`, it was migrated to `utils.SplitAndTrim` and updated to reflect the return-value difference.
- **Performance**: only a negligible micro-improvement (removing one level of call indirection); no performance regressions.
- **PR hygiene**: net -35 lines, no behavior change, and cases like `truncateBody`/`createFolderZipTempFile` that have two callers (where inlining would duplicate code) are intentionally kept — the judgment is consistent.

---

**Verdict**: APPROVED (ready to merge)

A clean, low-risk refactor. Build / vet / test all pass, with no orphan references or unused imports. Suggestion #1 (the rename) is purely optional and fine to skip in this PR.

**Checklist**:
- [x] Follows stack conventions (Go declaration order, errcheck, English comments)
- [x] Tests included and passing (coverage preserved)
- [x] No security vulnerabilities (pure refactor, N/A)
- [x] Documentation update not needed (no user-facing behavior change)

api/ftp/ftp.go
jisung-02: I’ll leave this out of scope for this PR and handle it separately.

PR #245, opened 2026-06-30
feat: integrate /api/auth/whoami/ for login verification and identity display

### LINKED ISSUE #231: [feat] Adopt the principal-agnostic identity endpoint (`/api/auth/whoami/`) for login verification and `whoami`
## Background

The CLI verifies a login and renders `alpacon whoami` by loading the current **user** from `GET /api/iam/users/-`. That endpoint is user-only, so a **service token** (an application principal) gets 401 there. As a result:

- Login verification relies on a workaround that sniffs the `alpst-` token prefix and treats the profile-load failure as non-fatal.
- `alpacon whoami` cannot show *which application* a service token belongs to.

Once the server exposes a principal-agnostic identity endpoint, the CLI should consume it so both flows work for any principal without prefix sniffing.

## Goal

Use `GET /api/auth/whoami/` to:
1. verify a login for any principal (user, personal API token, service token), and
2. render `alpacon whoami`, including application identity (name, service account, scopes) for service tokens.

## API contract (defined by the server issue)

```jsonc
{
  "principal_type": "user" | "application",
  "auth": { "method": "service_token" | "api_token" | "session", "scopes": ["server:read", "command:create"] },
  "user": { "id": "...", "username": "...", "email": "...", "role": "superuser|admin|general" },
  "application": { "id": "...", "name": "...", "service_type": "...", "username": "<service account or null>" }
}
```

## Proposed solution

- Add a client call for `GET /api/auth/whoami/` and a typed response that branches on `principal_type`.
- **Login verification:** call `whoami` after authenticating; success verifies the credential for any principal (removes the prefix-based workaround). For a service token, the success message can name the application.
- **`whoami` rendering:** populate user identity for a `user` principal and application identity (name, service account, `service_type`, scopes) for an `application` principal.
- **Status-code visibility (prerequisite):** the CLI must distinguish 404 (endpoint absent on an old server → fall back) from 401 (genuine auth failure → fail). The client error type currently discards the HTTP status code; surface it so callers can branch.
- **404 fallback:** against servers without the endpoint, fall back to the existing `/api/iam/users/-` path and the current prefix-based login handling.

## Backward compatibility & rollout

- The server endpoint must be deployed **before** the CLI relies on it. Because the CLI falls back on 404, the CLI can still ship independently and degrade gracefully against older servers.
- Keep the fallback path for at least one release cycle to cover server/CLI version skew on self-hosted workspaces.

## Acceptance criteria

- [ ] Against a new server, `alpacon login -t alpst-…` verifies via `whoami` and succeeds; `alpacon whoami` shows the application name and scopes.
- [ ] Against an old server (404), login and `whoami` fall back to the previous behavior without error.
- [ ] An invalid token still fails login (401 is not swallowed as a fallback).
- [ ] The client error type exposes the HTTP status code so 404 and 401 are handled distinctly.
- [ ] For a `user` principal, `whoami` output is unchanged.

## Test plan

- Table-driven + httptest: parse `whoami` for both principal types; login on 200 / 404 / 401; `whoami` rendering for an application principal and the 404 fallback.
- Regression for the new status-code accessor on the client error type.
- Manual: CLI end-to-end against a new server (service token → application identity shown) and an old server (graceful fallback).

## Out of scope (separate decisions)

- Making user-administration commands (`user`/`group`, `approval list`, `webhook create`) principal-aware so a suitably scoped service token can use them — a policy decision about whether service tokens should perform user administration at all.

jisung-02:
## Background

The server now exposes a `GET /api/auth/whoami/` endpoint that distinguishes a credential's principal between user and application (service token) and lets us query that identity directly. The existing CLI inferred login verification from the token prefix, and the `whoami` command could only show a user profile. This PR wires the whoami endpoint into login verification and identity display.

## Changes

Propagate the HTTP status code through the API error chain. `withStatus` tags each error read from a response with its HTTP status: it sets `statusCode` on an `*apiError` in place, and wraps any other error (empty-body read failure, unexpected non-JSON Content-Type) in a `statusError` that also implements `HTTPStatusCode`. Both are exposed via `utils.HTTPStatusCode`, so callers can tell a 404 (endpoint absent) from a 401 (credential error) on every path.

Add `GetWhoami` and its response types (`WhoamiResponse`, etc.) to `api/auth`. Only the application branch is modeled in detail; user principals are still resolved through `iam.GetCurrentUser`.

Replace login verification with a whoami-based path. `classifyWhoamiVerification` is fail-closed: nil means verified, a literal 404 is treated as an old server without whoami and falls back to legacy prefix verification (`verifyLoginLegacy`), and 401 plus every other error (parse failure, network) aborts login.

The `whoami` command now shows application identity (application name, service type, service account, scopes) for a service token. The whoami round-trip is gated by `config.IsServiceToken`, so user credentials skip the call and flow through the existing user path unchanged. The table output pads labels to the widest printed label so columns stay aligned on both the user and application paths.

## Safety

Login verification is fail-closed. It falls back to legacy verification only when the whoami call returns 404; every other error, including 401, aborts login rather than being absorbed by the fallback. An invalid credential therefore cannot slip through the fallback path.

## Testing

`go build ./...` passes. `go test` passes for the `api/auth`, `cmd`, `client`, and `utils` packages. The added tests cover whoami response parsing (application/user principal), status-code propagation (404/401/non-API error), the fail-closed branches of `classifyWhoamiVerification`, and the whoami command's application field mapping plus field omission for a user principal.

Verified live against the dev server (`https://alpacax.dev.alpacon.io`) with both a browser login (user principal) and a service token (application principal). The service-token login logged `Authenticated as application 'alpacon'`, confirming the whoami-based verification path. Phone number below is masked; the raw output showed the real value.

User principal (`alpacon whoami`):

```
User:        chaejisung
Email:       js.chae@alpacax.com
Phone:       +82 10-XXXX-XXXX
Workspace:   alpacax (https://alpacax.dev.alpacon.io)
Auth:        Browser login
Auth class:  browser_login
UID:         2006
Shell:       /bin/bash
Home:        /home/chaejisung
Role:        superuser
Groups:      Alpacon users (member)
WS required: yes—run 'alpacon work-session list' to see available sessions
```

User principal (`alpacon whoami --output json`):

```json
{
  "username": "chaejisung",
  "email": "js.chae@alpacax.com",
  "phone": "+82 10-XXXX-XXXX",
  "workspace_name": "alpacax",
  "workspace_url": "https://alpacax.dev.alpacon.io",
  "auth_method": "Browser login",
  "auth_classification": "browser_login",
  "uid": 2006,
  "shell": "/bin/bash",
  "home_directory": "/home/chaejisung",
  "role": "superuser",
  "groups": [
    { "name": "Alpacon users", "role": "member" }
  ],
  "work_session_required": true,
  "worksession_required_for_access": true,
  "active_work_session": null,
  "active_worksession": null
}
```

Application principal (`alpacon whoami`):

```
Application:  alpacon
Service type: integration
Scopes:       server:read, uploaded_file:create, uploaded_file:upload, uploaded_file:status, uploaded_file:bulk, uploaded_file:bulk_upload, uploaded_file:read
Workspace:    alpacax (https://alpacax.dev.alpacon.io)
Auth:         Service token
Auth class:   service_token
WS required:  no
```

Application principal (`alpacon whoami --output json`):

```json
{
  "workspace_name": "alpacax",
  "workspace_url": "https://alpacax.dev.alpacon.io",
  "auth_method": "Service token",
  "auth_classification": "service_token",
  "principal_type": "application",
  "application_name": "alpacon",
  "service_type": "integration",
  "scopes": [
    "server:read",
    "uploaded_file:create",
    "uploaded_file:upload",
    "uploaded_file:status",
    "uploaded_file:bulk",
    "uploaded_file:bulk_upload",
    "uploaded_file:read"
  ],
  "work_session_required": false,
  "worksession_required_for_access": false,
  "active_work_session": null,
  "active_worksession": null
}
```

## Checklist

- [x] Build passes
- [x] Tests pass
- [x] gofmt / go vet clean

Closes #231

### DIFF
--- a/api/auth/auth.go
+++ b/api/auth/auth.go
@@ -23,7 +23,22 @@ const (
 	tokenURL       = "/api/auth/tokens/"
 	tokenScopesURL = "/api/auth/tokens/scopes/"
 	statusURL      = "/api/status/"
+	whoamiURL      = "/api/auth/whoami/"
 )
+
+// GetWhoami fetches the caller's identity; a 404 means the endpoint is absent (old server).
+func GetWhoami(ac *client.AlpaconClient) (*WhoamiResponse, error) {
+	body, err := ac.SendGetRequest(whoamiURL)
+	if err != nil {
+		return nil, err
+	}
+
+	var resp WhoamiResponse
+	if err = json.Unmarshal(body, &resp); err != nil {
+		return nil, err
+	}
+	return &resp, nil
+}
 
 func LoginAndSaveCredentials(loginReq *LoginRequest, token string, insecure bool) error {
 	httpClient := &http.Client{
--- a/api/auth/auth.go
+++ b/api/auth/auth.go
@@ -37,6 +37,10 @@ func GetWhoami(ac *client.AlpaconClient) (*WhoamiResponse, error) {
 	if err = json.Unmarshal(body, &resp); err != nil {
 		return nil, err
 	}
+	// Fail closed: a 200 without a principal_type is a malformed identity response.
+	if resp.PrincipalType == "" {
+		return nil, errors.New("whoami response missing principal_type")
+	}
 	return &resp, nil
 }
 
--- a/api/auth/types.go
+++ b/api/auth/types.go
@@ -2,6 +2,9 @@ package auth
 
 import "time"
 
+// PrincipalTypeApplication is the whoami principal_type for service/application tokens.
+const PrincipalTypeApplication = "application"
+
 type LoginRequest struct {
 	WorkspaceURL  string `json:"workspace_url"`
 	Username      string `json:"username"`
@@ -59,4 +62,22 @@ type TokenScopeAttributes struct {
 	Name    string `json:"name" table:"Resource"`
 	Actions string `json:"actions" table:"Actions"`
 	ACL     string `json:"acl" table:"ACL"`
+}
+
+// WhoamiResponse is the GET /api/auth/whoami/ identity. Only the application
+// branch is consumed; user principals fall through to iam.GetCurrentUser.
+type WhoamiResponse struct {
+	PrincipalType string             `json:"principal_type"`
+	Auth          WhoamiAuth         `json:"auth"`
+	Application   *WhoamiApplication `json:"application,omitempty"`
+}
+
+type WhoamiAuth struct {
+	Scopes []string `json:"scopes"`
+}
+
+type WhoamiApplication struct {
+	Name        string `json:"name"`
+	ServiceType string `json:"service_type"`
+	Username    string `json:"username"` // service account name
 }
--- a/client/client.go
+++ b/client/client.go
@@ -30,6 +30,17 @@ type apiError struct {
 	statusCode int
 }
 
+// statusError carries an HTTP status on errors that aren't *apiError (e.g. an
+// HTML 404 page), so utils.HTTPStatusCode can still read it.
+type statusError struct {
+	err        error
+	statusCode int
+}
+
+func (e *statusError) Error() string      { return e.err.Error() }
+func (e *statusError) Unwrap() error       { return e.err }
+func (e *statusError) HTTPStatusCode() int { return e.statusCode }
+
 func NewAlpaconAPIClient() (*AlpaconClient, error) {
 	validConfig, err := config.LoadConfig()
 	if err != nil {
@@ -212,7 +223,7 @@ func readJSONResponse(resp *http.Response) ([]byte, error) {
 
 	// Empty content type is allowed for responses without content (e.g. PATCH).
 	if ct := resp.Header.Get("Content-Type"); ct != "" && !strings.Contains(ct, "application/json") {
-		return nil, fmt.Errorf("unexpected response from server (HTTP %d, Content-Type: %s)", resp.StatusCode, ct)
+		return nil, withStatus(fmt.Errorf("unexpected response from server (HTTP %d, Content-Type: %s)", resp.StatusCode, ct), resp.StatusCode)
 	}
 	return body, nil
 }
@@ -415,13 +426,18 @@ func newAPIError(message, code, source string) error {
 	return &apiError{message: message, code: code, source: source}
 }
 
-// withStatus tags an *apiError with its HTTP status so callers can tell 404 from 401.
+// withStatus tags err with its HTTP status so callers can tell 404 from 401,
+// wrapping non-*apiError errors so the status survives the error chain.
 func withStatus(err error, statusCode int) error {
+	if err == nil {
+		return nil
+	}
 	var ae *apiError
 	if errors.As(err, &ae) {
 		ae.statusCode = statusCode
+		return err
 	}
-	return err
+	return &statusError{err: err, statusCode: statusCode}
 }
 
 // parseAPIError extracts a human-readable error message from a JSON API error response.
--- a/client/client.go
+++ b/client/client.go
@@ -37,7 +37,7 @@ type statusError struct {
 	statusCode int
 }
 
-func (e *statusError) Error() string      { return e.err.Error() }
+func (e *statusError) Error() string       { return e.err.Error() }
 func (e *statusError) Unwrap() error       { return e.err }
 func (e *statusError) HTTPStatusCode() int { return e.statusCode }
 
--- a/client/client.go
+++ b/client/client.go
@@ -24,9 +24,10 @@ const (
 )
 
 type apiError struct {
-	message string
-	code    string
-	source  string
+	message    string
+	code       string
+	source     string
+	statusCode int
 }
 
 func NewAlpaconAPIClient() (*AlpaconClient, error) {
@@ -203,7 +204,7 @@ func (ac *AlpaconClient) createRequest(method, url string, body io.Reader) (*htt
 func readJSONResponse(resp *http.Response) ([]byte, error) {
 	body, readErr := io.ReadAll(resp.Body)
 	if err := checkAuthStatus(resp.StatusCode, body); err != nil {
-		return nil, err
+		return nil, withStatus(err, resp.StatusCode)
 	}
 	if readErr != nil {
 		return nil, readErr
@@ -229,7 +230,7 @@ func (ac *AlpaconClient) sendRequest(req *http.Request) ([]byte, error) {
 	}
 
 	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
-		return nil, parseAPIError(respBody)
+		return nil, withStatus(parseAPIError(respBody), resp.StatusCode)
 	}
 
 	return respBody, nil
@@ -310,7 +311,7 @@ func (ac *AlpaconClient) SendMultipartStreamRequest(url, contentType string, bod
 	}
 
 	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
-		return nil, parseAPIError(respBody)
+		return nil, withStatus(parseAPIError(respBody), resp.StatusCode)
 	}
 
 	return respBody, nil
@@ -406,18 +407,28 @@ func (e *apiError) ErrorSource() string {
 	return e.source
 }
 
+func (e *apiError) HTTPStatusCode() int {
+	return e.statusCode
+}
+
 func newAPIError(message, code, source string) error {
 	return &apiError{message: message, code: code, source: source}
+}
+
+// withStatus tags an *apiError with its HTTP status so callers can tell 404 from 401.
+func withStatus(err error, statusCode int) error {
+	var ae *apiError
+	if errors.As(err, &ae) {
+		ae.statusCode = statusCode
+	}
+	return err
 }
 
 // parseAPIError extracts a human-readable error message from a JSON API error response.
 // Handles common formats: {"detail": "..."}, {"field": ["error", ...]}, {"non_field_errors": ["..."]}
 func parseAPIError(body []byte) error {
-	message, code, source, ok := parseAPIErrorPayload(body)
-	if ok {
-		return newAPIError(message, code, source)
-	}
-	return errors.New(message)
+	message, code, source, _ := parseAPIErrorPayload(body)
+	return newAPIError(message, code, source)
 }
 
 func parseAPIErrorPayload(body []byte) (message string, code string, source string, ok bool) {
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -17,7 +17,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const defaultBaseDomain = "alpacon.io"
+const (
+	defaultBaseDomain = "alpacon.io"
+
+	// whoami verification outcomes (see classifyWhoamiVerification).
+	whoamiVerified = "verified"
+	whoamiFallback = "fallback"
+	whoamiFail     = "fail"
+)
 
 // knownCloudRegions are the Alpacon Cloud regions shown when --region is omitted.
 // Source: 10-alpacon-web constants.ts, 06-account settings.py. Update on release.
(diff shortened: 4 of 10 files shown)


api/auth/types.go
```diff
@@ -60,3 +63,21 @@ type TokenScopeAttributes struct {
...
+
+type WhoamiAuth struct {
+	Scopes []string `json:"scopes"`
+}
+
+type WhoamiApplication struct {
+	Name        string `json:"name"`
+	ServiceType string `json:"service_type"`
+	Username    string `json:"username"` // service account name
+}
```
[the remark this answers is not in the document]
jisung-02: Not applicable in Go. `json.Unmarshal` of a JSON `null` into a non-pointer `string` is a documented no-op: it leaves the field at its zero value (`""`) and returns no error—it does not fail. Verified:

```go
var a struct{ Username string `json:"username"` }
err := json.Unmarshal([]byte(`{"username":null}`), &a)
// err == nil, a.Username == ""
```

So `GetWhoami` does not break and login verification is not affected by a null `username`. No change needed; keeping `Username string`.

cmd/whoami.go
```diff
@@ -162,6 +180,18 @@ func getExpiresAt(cfg config.Config) string {
...
 
+// applyApplicationPrincipal fills out's application fields from an application-principal whoami response.
+func applyApplicationPrincipal(out whoamiOutput, who *auth.WhoamiResponse) whoamiOutput {
+	out.PrincipalType = auth.PrincipalTypeApplication
+	if who.Application != nil {
+		out.ApplicationName = who.Application.Name
+		out.ServiceType = who.Application.ServiceType
+		out.ServiceAccount = who.Application.Username
+	}
+	out.Scopes = who.Auth.Scopes
```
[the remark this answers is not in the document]
jisung-02: Follows from the `types.go:83` thread. `WhoamiApplication.Username` is a value type (`string`), not a pointer, so `who.Application.Username` can never be nil and the assignment cannot panic. When the API sends `username: null`, the field is `""` and `printWhoami` omits the empty `Service account` row. No nil-handling needed.

client/client.go
```diff
@@ -406,18 +407,28 @@ func (e *apiError) ErrorSource() string {
...
 }
 
+// withStatus tags an *apiError with its HTTP status so callers can tell 404 from 401.
+func withStatus(err error, statusCode int) error {
+	var ae *apiError
+	if errors.As(err, &ae) {
+		ae.statusCode = statusCode
+	}
+	return err
+}
```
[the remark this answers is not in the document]
jisung-02: Good catch—confirmed and fixed in 41726f2.

`checkAuthStatus` ignores 404 (only 401/403), so an HTML 404 fell through to the non-JSON Content-Type guard in `readJSONResponse`, which returned a plain `fmt.Errorf` that `sendRequest` passed through untagged. `classifyWhoamiVerification` then saw status 0 instead of 404 and aborted login instead of falling back to legacy verification.

Fix: `withStatus` now wraps non-`*apiError` errors in a status-carrying `statusError` (implements `HTTPStatusCode` + `Unwrap`), and the content-type guard tags the error with `resp.StatusCode`. Added `TestSendRequest_404HTMLBodyExposesStatusCode` covering the HTML 404 path. A body-read failure is left untagged on purpose—aborting (fail-closed) on a flaky connection is the desired behavior, not a legacy fallback.

cmd/whoami.go
```diff
@@ -299,6 +329,10 @@ func printWhoami(output whoamiOutput) {
 		value string
 	}{
 		{"User", output.Username},
+		{"Application", output.ApplicationName},
+		{"Service type", output.ServiceType},
+		{"Service account", output.ServiceAccount},
+		{"Scopes", strings.Join(output.Scopes, ", ")},
```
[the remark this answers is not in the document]
jisung-02: Fixed in 140e1ac. printWhoami now computes the label width from the widest label actually printed and uses %-*s, so columns stay aligned on both the user and application paths. Verified live against the dev server with a service token — "Service type: integration" and "Scopes:" now render with correct spacing.

cmd/whoami.go
```diff
@@ -103,6 +50,17 @@ commands, especially for AI agents and operators managing multiple workspaces.`,
...
+		// Only a service token can be an application principal, so skip the whoami
+		// round-trip for user credentials, which fall through to the user path below.
+		if config.IsServiceToken(cfg.Token) {
+			if who, whoErr := auth.GetWhoami(ac); whoErr == nil && who.PrincipalType == auth.PrincipalTypeApplication {
+				output = applyApplicationPrincipal(output, who)
+				warnIfExpiringSoon(cfg)
+				printWhoami(output)
+				return
+			}
+		}
```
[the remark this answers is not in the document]
jisung-02: Good catch—fixed in 8c4babc.

A service token has no user profile, so falling through to `iam.GetCurrentUser()` on a whoami failure was guaranteed to fail and swallowed the real error behind a misleading "Could not fetch user info" warning (the old-server 404 is exactly that case). The service-token branch now resolves identity via whoami and returns unconditionally: it applies the application principal on success, and otherwise warns with the actual whoami error (or an unexpected-principal message) and prints config-based output—never touching the user-only IAM endpoint.

api/auth/auth.go
```diff
@@ -23,8 +23,23 @@ const (
...
+	body, err := ac.SendGetRequest(whoamiURL)
+	if err != nil {
+		return nil, err
+	}
+
+	var resp WhoamiResponse
+	if err = json.Unmarshal(body, &resp); err != nil {
+		return nil, err
+	}
+	return &resp, nil
```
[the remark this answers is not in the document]
jisung-02: Good catch. GetWhoami now rejects a 200 with an empty principal_type (returns an error), so classifyWhoamiVerification maps it to whoamiFail and login aborts instead of treating a malformed payload as verified. Added a regression test (TestGetWhoamiEmptyPrincipalTypeFailsClosed). Fixed in 386bf72.

cmd/login.go
```diff
@@ -481,6 +490,33 @@ func formatHostURL(host string) string {
 	return fmt.Sprintf("%s://%s", scheme, strings.TrimSuffix(host, "/"))
 }
 
+// classifyWhoamiVerification maps a whoami error to verified / fallback (404) /
+// fail (401 or any other error—never swallowed as a fallback).
+func classifyWhoamiVerification(whoErr error) string {
```
geunwoonoh, line 495: Warning: fail-closed side effect. Previously a personal-token profile-preload failure was non-fatal (warn and continue). Now any non-404 whoami error — e.g. a 403 if the server ever WorkSession-gates whoami — hits `whoamiFail` and aborts login. This is the intended, safer behavior, but it couples login success to the assumption that whoami is never policy-gated for a valid credential. Low risk — worth pinning down as a server contract (whoami answers 200/404, never a gated 4xx, once authenticated).

cmd/login.go
```diff
@@ -17,7 +17,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const defaultBaseDomain = "alpacon.io"
+const (
+	defaultBaseDomain = "alpacon.io"
+
+	// whoami verification outcomes (see classifyWhoamiVerification).
+	whoamiVerified = "verified"
```
geunwoonoh, line 24: Suggestion: `whoamiVerified/Fallback/Fail` are raw `string` constants and `classifyWhoamiVerification` returns `string`. A dedicated type would be more idiomatic and prevent typo / mismatch comparisons at compile time:

```go
type whoamiVerdict int
const (
    whoamiVerified whoamiVerdict = iota
    whoamiFallback
    whoamiFail
)
```
Works fine as-is; optional.

cmd/whoami.go
```diff
@@ -313,10 +354,22 @@ func printWhoami(output whoamiOutput) {
 		{"WS required", formatWSRequired(output.WorksessionRequiredForAccess, output.ActiveWorksessionCanonical)},
 	}
 
+	// Pad to the widest printed label so columns align regardless of which rows appear.
+	pad := 0
+	for _, l := range lines {
+		if l.value == "" || l.value == "0" {
```
geunwoonoh, line 360: Suggestion (DRY): the `l.value == "" || l.value == "0"` skip condition is duplicated here (padding loop) and in the print loop below. Extracting a small closure keeps it in one place:

```go
visible := func(v string) bool { return v != "" && v != "0" }
```
Negligible perf either way (~17 rows).

cmd/whoami.go
```diff
@@ -103,6 +50,24 @@ commands, especially for AI agents and operators managing multiple workspaces.`,
 			return
 		}
 
+		// Only a service token can be an application principal. Service tokens have no
+		// user profile, so resolve identity via whoami and return without falling
+		// through to the user-only IAM endpoint, which would always fail for them.
+		if config.IsServiceToken(cfg.Token) {
```
geunwoonoh, line 56: Suggestion (test seam): the service-token branches here — `whoErr != nil` (warn) and the unexpected-principal-type default (warn) — aren't unit-tested; only `applyApplicationPrincipal` is. If the `Run` closure is hard to test directly, extracting the whoami → whoamiOutput decision into a small pure function would create a seam. Optional.

geunwoonoh (approved):
## PR Review: #245 — integrate /api/auth/whoami/ for login verification and identity display

**Reviewer verdict:** ✅ APPROVE — mergeable. All Warnings/Suggestions below are optional.

### Local verification
- `go build ./...` ✅
- `go vet ./...` ✅ (clean)
- `gofmt -l` on changed files ✅ (clean)
- `go test ./api/auth/... ./client/... ./cmd/... ./utils/...` ✅ (all pass)

### Summary
Wires the new `GET /api/auth/whoami/` endpoint into the CLI: (1) replaces prefix-inference login verification with a principal-agnostic whoami path, (2) shows application identity (name / service type / service account / scopes) for service tokens in `whoami`, and (3) propagates the HTTP status code through the error chain so callers can tell 404 (old server) from 401 (bad credential). The **fail-closed** design (only 404 falls back to legacy; every other error aborts login) is the core of the change.

Copilot's 6 findings were all resolved or rebutted with justification (null-username no-op, `statusError` wrapping, service-token fallthrough removal, `{}` fail-closed, dynamic padding, em-dash). My independent review confirms the points below.

### Critical issues (0)
None. No critical bug or security issue found.
- **Nil safety:** `GetWhoami` only returns `(nil, err)` or `(&resp, nil)`, and `who` is dereferenced only in the `whoErr == nil` branch → safe.
- **Security:** no hardcoded secrets; token / refresh token never exposed in `whoamiOutput` or logs; fail-closed verification prevents an invalid credential from slipping through the fallback.
- **Declaration order:** moving the types below `var whoamiCmd` actually **improves** conformance to the `var → type → func` convention.

### Warnings (1)
1. **cmd/login.go — fail-closed side effect.** A personal-token profile-preload failure used to be non-fatal; now any non-404 whoami error aborts login. Intended and safer, but it couples login to the assumption that whoami is never policy-gated (e.g. never returns 403 for a valid credential). Low risk — worth pinning down as a server contract.

### Suggestions (3)
1. **cmd/login.go — stringly-typed state.** `whoamiVerified/Fallback/Fail` are raw strings; a dedicated `type whoamiVerdict int` with iota constants would be more idiomatic and catch typo-comparison bugs at compile time. Optional.
2. **cmd/whoami.go — DRY.** The `l.value == "" || l.value == "0"` skip condition is duplicated across the padding loop and the print loop; a small `visible(v string) bool` closure would keep it in one place. No perf impact.
3. **cmd/whoami.go — test seam.** The command's service-token branches (`whoErr != nil` warning / unexpected-principal warning) aren't unit-tested; only `applyApplicationPrincipal` is. Optional.

*Out of scope (pre-existing):* `getRole` (cmd/whoami.go) and `getUserPrivileges` (client/client.go) are near-duplicate `(isStaff, isSuperuser) → role` mappings differing only in the default (`"user"` vs `"general"`).

### Performance
- Login: whoami-verified path = 1 network call (same as the previous `LoadCurrentUser`). Fallback (old server) is 2 calls but is the exception path.
- `whoami` for a service token: 1 whoami call that succeeds, vs. the previous `GetCurrentUser` call that 401'd — same cost, better UX.
- No regression, no unused / dead code (all new symbols are referenced).

### Highlights
- fail-closed design and its rationale are clearly documented in comments.
- `statusError` + `withStatus` cleanly propagate status even for non-`*apiError` (HTML 404), guarded by a regression test.
- Dynamic label-width (`%-*s`) fixes column alignment on both the user and application paths.
- Good test coverage; Copilot findings rebutted with accurate reasoning (notably the Go `json.Unmarshal(null → string)` no-op).

**Verdict:** ✅ APPROVE

**Checklist**
- [x] Follows stack conventions (declaration order, errcheck, gofmt, English comments, sentence case)
- [x] Documentation / PR description complete

cmd/login.go
jisung-02: Confirmed the server contract on alpacon-server origin/main (#2364). The whoami view uses only IsAuthenticatedOr401 — no scope or WorkSession gate. An authenticated credential returns 200; 400 only in the pathological case where the principal is neither user nor application; 401 means unauthenticated; 404 only on an old server without the endpoint. No path returns a gated 4xx like 403 for a valid credential. This is pinned by server tests (test_no_scope_required, test_unauthenticated_returns_401, test_unsupported_principal_returns_400), so a regression would fail server CI. Keeping the fail-closed behavior as-is.

cmd/whoami.go
jisung-02: Only two occurrences today, so by the rule of three I'll leave it inline for now and extract a visible() helper if a third use shows up. Thanks for flagging.

cmd/login.go
jisung-02: Good call, and agreed it'd be more idiomatic. Since the constants are only used locally by classifyWhoamiVerification and its single switch, there's no place where a stringly-typed comparison could drift, so the compile-time safety wouldn't buy much here yet. I'll keep it as-is for now (YAGNI) and reach for the whoamiVerdict type the moment these outcomes grow or get used elsewhere. Thanks!

cmd/whoami.go
jisung-02: Good instinct on the missing coverage. I looked at extracting the whoami->output decision into a pure function, but the two warning branches have no logic to test beyond the switch itself — the one branch with actual behavior (applyApplicationPrincipal) is already covered. Pulling out a single-caller function whose only purpose is to make the switch testable would add indirection and pin the human-facing warning copy in a unit test, so it reads as a net loss here. Keeping the inline switch and treating these two lines as covered by review rather than a unit test. Thanks for flagging.

PR #260, opened 2026-07-05
feat(cli): add work-session cancel

### LINKED ISSUE #256: feat(cli): add work-session cancel
## Background

The web already lets a requester cancel their own *pending* Work Session before approval, flipping it to `cancelled` (`CancelSessionModal.tsx`, POST `.../cancel/`). The CLI has no equivalent, so a requester who created a session from the CLI has no way to withdraw it there.

This is not the same as the existing `work-session revoke`, which force-terminates an *already approved or active* session and is superuser-only (`isRevoke = session.status === "approved"`). Cancel is a self-request action—the requester pulling back their own pending request—so it sits inside the range the CLI is allowed to carry under ADR 0015.

## Proposal

- [ ] Add `alpacon work-session cancel SESSION_ID`

## Source

alpacon-handbook `reference/cli-parity-sweep-2026-07.md`

jisung-02:
## Background

The web console already lets a requester cancel their own pending work session before approval, flipping it to `cancelled` (POST `.../cancel/`). The CLI had no equivalent, so a requester who created a session from the CLI had no way to withdraw it there.

This is not the same as `work-session revoke`, which force-terminates an already approved or active session and is superuser-only. Cancel is a self-request action, the requester pulling back their own pending request, so it sits inside the range the CLI is allowed to carry under ADR 0015.

## Changes

Add `alpacon work-session cancel SESSION_ID`. The command mirrors the existing `revoke`/`complete` leaf commands: it posts to the session's `cancel` action via the new `CancelWorkSession` API client, then prints a success line. Registration, the `opCancel` operation identifier for JSON envelopes, the no-subcommand help string, and the README were updated alongside it.

The CLI delegates all authorization and state checks to the server rather than pre-validating, matching the sibling commands.

## Permissions

The server gates the two termination actions differently, which is why cancel is allowed from the CLI while approve/reject are not:

| Action | Server permission | CLI |
| --- | --- | --- |
| `approve` / `reject` | superuser + approval-channel | blocked (ADR 0015) |
| `revoke` | superuser | allowed |
| `cancel` | session creator or superuser (no approval-channel) | allowed |

Cancel additionally accepts only `pending` sessions; a non-pending session is rejected server-side with `work_session_not_pending`.

## Testing

- `go build`, `go test -race ./...`, and `golangci-lint run` all pass.
- Added `TestCancelWorkSession` asserting the POST path suffix `.../{id}/cancel/`.
- Verified end-to-end against dev (`alpacax.dev.alpacon.io`): created a pending session on `web-editor`, cancelled it (`status: pending` to `cancelled`, `completed_at` set), then re-cancelled the same session and got `Failed to cancel work session: code: work_session_not_pending.` with exit code 1.

## Checklist

- [x] Build, tests, and lint pass
- [x] Unit test for the new API client
- [x] End-to-end verification against a live workspace
- [x] README updated

Closes #256

### DIFF
--- a/cmd/worksession/worksession.go
+++ b/cmd/worksession/worksession.go
@@ -47,7 +47,7 @@ Gated operations (require an active WorkSession under interactive auth):
 
 Bypass: Token auth (API token or Service token) skips the requirement.
 
-Lifecycle:  pending → approved → active → complete | expired | revoked
+Lifecycle:  pending → approved → active → complete | expired | revoked | cancelled
 
 Error codes returned when a session check fails:
   work_session_required           no session selected for this shell
--- a/cmd/worksession/worksession.go
+++ b/cmd/worksession/worksession.go
@@ -9,6 +9,7 @@ import (
 // Operation identifiers carried in JSON outputs: success "operation" and error context.operation.
 const (
 	opActivate  = "activate"
+	opCancel    = "cancel"
 	opComplete  = "complete"
 	opCreate    = "create"
 	opCurrent   = "current"
@@ -62,7 +63,7 @@ Run 'alpacon whoami' to check your WorkSession requirement and active session.`,
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session approve', 'alpacon work-session reject', 'alpacon work-session revoke', 'alpacon work-session timeline', or 'alpacon work-session recording'. Run 'alpacon work-session --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session approve', 'alpacon work-session reject', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Run 'alpacon work-session --help' for more information")
 	},
 }
 
@@ -81,4 +82,5 @@ func init() {
 	WorkSessionCmd.AddCommand(workSessionApproveCmd)
 	WorkSessionCmd.AddCommand(workSessionRejectCmd)
 	WorkSessionCmd.AddCommand(workSessionRevokeCmd)
+	WorkSessionCmd.AddCommand(workSessionCancelCmd)
 }
--- a/cmd/worksession/worksession.go
+++ b/cmd/worksession/worksession.go
@@ -47,7 +47,7 @@ Gated operations (require an active WorkSession under interactive auth):
 
 Bypass: Token auth (API token or Service token) skips the requirement.
 
-Lifecycle:  pending → approved → active → complete | expired | revoked | cancelled
+Lifecycle:  pending → approved → active → complete | expired | revoked  (pending only → rejected | cancelled)
 
 Error codes returned when a session check fails:
   work_session_required           no session selected for this shell
--- a/cmd/worksession/worksession.go
+++ b/cmd/worksession/worksession.go
@@ -51,7 +51,7 @@ Lifecycle:  pending → approved → active → complete | expired | revoked | c
 
 Error codes returned when a session check fails:
   work_session_required           no session selected for this shell
-  work_session_not_active         session not active (pending, approved, completed, or revoked)
+  work_session_not_active         session not active (pending, approved, completed, revoked, or cancelled)
   work_session_expired            session has expired
   work_session_scope_not_allowed  operation not in session scopes
   work_session_server_not_allowed target server not in session
--- a/cmd/worksession/worksession_cancel.go
+++ b/cmd/worksession/worksession_cancel.go
@@ -23,6 +23,11 @@ var workSessionCancelCmd = &cobra.Command{
 			utils.CliErrorEnvelopeWithExit(opCancel, err, "Failed to cancel work session: %s.", err)
 		}
 
-		utils.CliSuccess("Work session %s cancelled.", args[0])
+		output := newWorkSessionCancelOutput(args[0])
+		if utils.OutputFormat == utils.OutputFormatJSON {
+			printWorkSessionMutationJSON(output)
+			return
+		}
+		utils.CliSuccess("%s", output.Message)
 	},
 }
--- a/cmd/worksession/worksession_cancel.go
+++ b/cmd/worksession/worksession_cancel.go
@@ -1,0 +1,28 @@
+package worksession
+
+import (
+	wsapi "github.com/alpacax/alpacon-cli/api/worksession"
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+var workSessionCancelCmd = &cobra.Command{
+	Use:     "cancel SESSION_ID",
+	Short:   "Withdraw your own pending work session request",
+	Long:    "Withdraw a pending work session you requested before it is reviewed. Restricted to the session's requester (or a superuser); only pending sessions can be cancelled.",
+	Args:    cobra.ExactArgs(1),
+	Example: `  alpacon work-session cancel ses-abc123`,
+	Run: func(cmd *cobra.Command, args []string) {
+		ac, err := client.NewAlpaconAPIClient()
+		if err != nil {
+			utils.CliErrorEnvelopeWithExit(opCancel, err, "Connection to Alpacon API failed: %s. Consider re-logging.", err)
+		}
+
+		if err := wsapi.CancelWorkSession(ac, args[0]); err != nil {
+			utils.CliErrorEnvelopeWithExit(opCancel, err, "Failed to cancel work session: %s.", err)
+		}
+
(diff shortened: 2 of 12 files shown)


cmd/worksession/worksession.go
```diff
@@ -62,7 +63,7 @@ Run 'alpacon whoami' to check your WorkSession requirement and active session.`,
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session approve', 'alpacon work-session reject', 'alpacon work-session revoke', 'alpacon work-session timeline', or 'alpacon work-session recording'. Run 'alpacon work-session --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session approve', 'alpacon work-session reject', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Run 'alpacon work-session --help' for more information")
```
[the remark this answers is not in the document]
jisung-02: Fixed in 8af3f62 — added the pending→cancelled path to the lifecycle line.

While addressing this I also checked for the same class of omission elsewhere and fixed two more spots:
- fix(worksession) 2f334b8: pollForApproval treated rejected/expired/revoked/completed as terminal but fell through on cancelled, so a session cancelled during create --wait polled to timeout instead of exiting. Added a cancelled case mirroring revoked, with a test.
- docs(worksession) 41ff328: the --status filter help and the work_session_not_active description (help text + README) enumerated statuses but omitted cancelled.

cmd/worksession/worksession.go
```diff
@@ -46,11 +47,11 @@ Gated operations (require an active WorkSession under interactive auth):
 
 Bypass: Token auth (API token or Service token) skips the requirement.
 
-Lifecycle:  pending → approved → active → complete | expired | revoked
+Lifecycle:  pending → approved → active → complete | expired | revoked | cancelled
```
geunwoonoh: Suggestion: This line now reads as if `cancelled` is reachable from `active`, but the server only allows cancelling *pending* sessions (as the PR body itself notes). `rejected` being absent is pre-existing, but since this line is already being touched, consider splitting out the pending-only outcomes, e.g.:

```
Lifecycle:  pending → approved → active → complete | expired | revoked  (pending only → rejected | cancelled)
```

This help text is consumed by AI agents, so state-machine accuracy here has outsized value.

cmd/worksession/worksession_cancel.go
```diff
@@ -0,0 +1,28 @@
...
+		ac, err := client.NewAlpaconAPIClient()
+		if err != nil {
+			utils.CliErrorEnvelopeWithExit(opCancel, err, "Connection to Alpacon API failed: %s. Consider re-logging.", err)
+		}
+
+		if err := wsapi.CancelWorkSession(ac, args[0]); err != nil {
+			utils.CliErrorEnvelopeWithExit(opCancel, err, "Failed to cancel work session: %s.", err)
+		}
+
+		utils.CliSuccess("Work session %s cancelled.", args[0])
```
geunwoonoh: Suggestion: With `--output json`, a successful cancel prints nothing to stdout (`CliSuccess` writes to stderr), while `create`/`update`/`use`/`extend` emit a `printWorkSessionMutationJSON` success envelope. This matches the closest siblings (`revoke`/`complete`), so it's not a defect of this PR—but since cancel explicitly targets machine consumers too, consider a follow-up issue to give cancel/revoke/complete mutation envelopes as well.

cmd/worksession/worksession_create.go
```diff
@@ -35,6 +34,9 @@ const (
 
 var validScopePresets = []string{"command", "editor", "sudo", "tunnel", "webftp", "websh"}
 
+// pollInterval is a var (not const) so tests can shrink it.
+var pollInterval = 10 * time.Second
```
geunwoonoh: Suggestion: The const→var change isn't strictly needed for the new test's happy path—the server returns `cancelled` immediately, so `pollForApproval` returns from the switch on the first iteration and never reaches `time.Sleep`. It does have defensive value (a regression of the cancelled case fails in ~0s instead of hanging ~290s), so it's fine as-is; passing the interval as a parameter to `pollForApproval` would be the cleaner alternative. Optional.

geunwoonoh (dismissed):
## PR review: work-session cancel

### Summary

Adds `alpacon work-session cancel SESSION_ID`, letting a requester withdraw their own pending session before approval—bringing the CLI to parity with the web console's cancel. The command mirrors the existing `revoke`/`complete` leaf pattern and delegates authorization/state checks to the server. It also adds cancelled-state handling to `pollForApproval`.

### Critical issues (0)

None. Build/tests/lint pass, e2e verified, no security concerns.

### Warnings (1)

1. **`utils/worksession_error.go:133`**—the next-action hint for `work_session_not_active` was not updated for cancelled. This PR updates the same status enumeration in the README refusal table and the `work-session` help text, but misses the runtime hint string that is actually printed on refusal:

   ```go
   "alpacon work-session current  # pending/approved: wait until active; completed/revoked: create or reuse below",
   ```

   A pending session referenced via `--work-session <id>` that later gets cancelled hits exactly this path. Cancelled belongs in the same "create or reuse" branch as completed/revoked, so the hint should read `completed/revoked/cancelled: ...`. This string is consumed mechanically by AI agents, so consistency matters here in particular. (No inline comment—the file isn't part of this diff.)

### Suggestions (3)

Posted inline: lifecycle-line accuracy in the help text (`cmd/worksession/worksession.go:50`), a missing JSON success envelope for machine consumers (`cmd/worksession/worksession_cancel.go:26`, follow-up material), and a note on the `pollInterval` const→var change (`cmd/worksession/worksession_create.go:38`).

### Requested review dimensions

**Go conventions**—compliant. POSIX `Use` string, leaf-command `Run` + `CliErrorEnvelopeWithExit` pattern, and the `wsapi` import alias all match the sibling commands; `opCancel` is inserted in alphabetical order in the const block; error strings are lowercase without trailing punctuation (ST1005); declaration order (const → var → type → func) is respected; the errcheck patterns in tests are correct.

**Duplicate/unused code**—no dead code introduced: `opCancel` (two error envelopes) and `cancelledWorkSessionStatus` (`pollForApproval`) are both used. The `revoke`/`complete`/`cancel` Run bodies are copies differing only in identifiers and strings (~28 lines each); a shared constructor (e.g. `newSessionActionCmd(op, apiFn, ...)`) would cut ~60 lines, but CLAUDE.md documents the one-file-per-leaf pattern as the repo convention, so keeping it as-is is also defensible—optional.

**Performance**—no regressions, and one real improvement. The command itself is a single POST. The `pollForApproval` cancelled case is a genuine fix: previously, a session cancelled from the web console while `create --wait` was polling would spin for up to 5 minutes (30 × 10s) printing "Waiting for approval..." before timing out; now it fails fast with a clear error.

### Highlights

- Thorough sweep of status-enumeration sites beyond the minimal feature—`pollForApproval` fail-fast, README refusal table, `--status` flag help (one miss, see the warning)
- The `CancelWorkSession` doc comment cleanly explains the authorization semantics vs `RevokeWorkSession`—useful for readers who only see the API layer
- Tests follow each layer's existing pattern exactly (api: `newTestClient` + path-suffix assertion; cmd: direct client construction like the `update`/`create` tests)
- Excellent PR description: permissions table, ADR 0015 rationale vs revoke, and e2e verification with actual outputs

---

**Verdict**: APPROVED—nothing blocks the merge. It would be nice to fold in the warning (a one-line hint string) before or shortly after merging.

**Checklist**:
- [x] Code follows stack conventions (golang + repo CLAUDE.md rules)
- [x] Tests included and passing (one per layer + e2e verification)
- [x] No security vulnerabilities (authz delegated to the server; no credential exposure)
- [x] Documentation updated (README, help text—except the one hint string above)

cmd/worksession/worksession.go
jisung-02: Done in 58b6bab. Split cancelled/rejected out as pending-only outcomes, including rejected which was previously absent:

`Lifecycle:  pending → approved → active → complete | expired | revoked  (pending only → rejected | cancelled)`

cmd/worksession/worksession_cancel.go
jisung-02: Done in ef48fbb. cancel now emits a mutation envelope in `--output json` mode, mirroring the extend helper — added `newWorkSessionCancelOutput` with a deterministic `status: cancelled` and a unit + command-level JSON test. Kept the scope to cancel; revoke/complete stay envelope-less as they were, and I left a `Directive:` trailer noting the helper pattern should be reused if they get envelopes in a follow-up.

cmd/worksession/worksession_create.go
jisung-02: Done in 8de709a. Went with your cleaner alternative: `pollInterval` is back to a const and `pollForApproval` now takes an explicit `interval time.Duration`. The caller passes the const and the test passes 0 directly, so it no longer swaps and restores shared package state. No behavior change.

jisung-02:
Addressed the warning (the `work_session_not_active` runtime hint) in 9e4a065. `utils/worksession_error.go:133` now reads `completed/revoked/cancelled: create or reuse below`, so a pending session cancelled while referenced via `--work-session <id>` is routed to the same create-or-reuse branch, and the hint stays consistent with the README refusal table and help text.

The three inline suggestions were also folded in: lifecycle-line accuracy (58b6bab), `pollForApproval` interval as a parameter (8de709a), and a JSON success envelope for cancel (ef48fbb, scoped to cancel; revoke/complete left for a follow-up).

geunwoonoh (approved):

PR #261, opened 2026-07-07
feat(websh): add session-record search command

### LINKED ISSUE #258: feat(cli): add session-record command-text search
## Background

The web can search inside a session's records both by name and by command text (`useSearchRecordQuery`, `src/queries/Websh.ts:337`, hitting `websh/sessions/{id}/search/`; the UI is `WebshSearchInput.tsx` with modes `"name" | "command"`). The CLI has nothing close—`audit` only filters by `--tail`/`--user`/`--app`/`--model`, with no free-text or command search at all.

## Proposal

- [ ] Add a CLI command that exposes session-record (command-text) search

Audit-log free-text search is a secondary follow-up.

## Source

alpacon-handbook `reference/cli-parity-sweep-2026-07.md`

jisung-02:
## Background

The web can search a session's records by command text (`websh/sessions/{id}/search/`, `useSearchRecordQuery`), but the CLI had no equivalent. `audit` only offers the `--tail`/`--user`/`--app`/`--model` field filters, with no way to find command text inside a session. This is the gap flagged by the alpacon-handbook CLI parity sweep (Finding 9). The server API already exists, so this is a pure client-side change.

## What changed

Adds `alpacon websh records SESSION_ID [-q TEXT] [-n LIMIT]` (alias `rec`).

- Without `-q`/`--query` it calls `sessions/{id}/records/` (list all); with a query it calls `sessions/{id}/search/?q=` (fuzzy command-text search).
- It follows the `next` cursor up to `-n`/`--limit` (default 100). The server uses cursor-token pagination, so `api.FetchAllPages` (integer page-based) can't be reused; a dedicated `GetSessionRecords` and `recordCursorPage` type were added in `api/websh`.
- Output honors the global `--output` flag. `json` emits records verbatim (for machine/script consumers); the default table output is cleaned by `sanitizeRecord`.
- Refactor: the ANSI/control-char stripping helpers were promoted from `cmd/worksession` to `utils` (`StripANSIEscapes`, `StripControlChars`, `IsControlRune`), and worksession was converged onto the shared helpers. This removes duplication and preserves behavior.

## Access control

Both `records` and `search` endpoints are gated server-side by `IsPaidPlan`. The free plan returns 403, and the CLI surfaces the server error message as-is. No client-side pre-guard was added, since the server is the authority on authorization.

## Safety

Records are secret-masked terminal recordings. In shared or invited sessions a lower-privileged participant can influence their content, which another user later renders. Given that input path:

- Table output strips ANSI escapes (CSI/OSC/DCS/Fe) and sanitizes control characters across C0, DEL, and C1 (U+0080 to U+009F). C1 can act as a CSI/OSC introducer on 8-bit terminals. Control characters are replaced with spaces so tokens do not merge, then collapsed with `strings.Fields`.
- `json` output is verbatim by contract. `encoding/json` escapes ESC (0x1b), which blocks sequence reconstruction, and sanitizing it would corrupt data consumers rely on and diverge from other commands, so sanitization is applied to the table path only.
- The server timestamp (`added_at`) is rendered raw, consistent with how server timestamps are handled across the codebase.

## Tests

- `api/websh/websh_test.go`: cursor following, `--limit` capping, `page_size` value (`min(100, limit-len)`), and `search/` endpoint routing when `-q` is set.
- `cmd/websh/websh_test.go`: `sanitizeRecord` cases for ANSI, OSC, C1, and the double-space edge.
- `utils/utils_test.go`: `StripANSIEscapes` (CSI/OSC-BEL/OSC-ST/DCS branches) and `StripControlChars` (C0/DEL/C1).
- Local CI-equivalent checks pass: `go build -v .`, `go test -race ./...`, `golangci-lint run ./...` (0 issues), `govulncheck` (0 affecting), and go.mod/go.sum tidy.

Manual E2E verification against an Alpacon Cloud workspace (paid plan, Auth0 browser login). Created a websh session on `web-editor`, ran `ls -al` / `docker ps` / `echo`, then queried its records by session ID:

- `websh records SESSION_ID` returned the session record with the table path sanitized — ANSI escapes, SGR colors, and backspace/autocomplete replay stripped, then truncated at the column width.
- `--output json` returned the record verbatim, with ESC emitted as its JSON unicode escape by `encoding/json`, confirming the sequence-reconstruction block described in Safety.
- Search filtering confirmed by positive/negative contrast on the same session: `-q docker` and `-q reboot` returned the matching record, while `-q zzzznope` returned an empty result. This proves the query is routed to `sessions/{id}/search/` and filtered server-side, not ignored.
- `rec -n 20` exercised the alias and `--limit`.

## Checklist

- [x] `golangci-lint run ./...` reports 0 issues
- [x] Existing behavior preserved (worksession sanitizer convergence is behavior-neutral)
- [x] Manual E2E verified on `web-editor` (list, search, json, alias/limit)

Closes #258

### DIFF
--- a/api/websh/types.go
+++ b/api/websh/types.go
@@ -95,8 +95,7 @@ type SessionRecord struct {
 	Record  string `json:"record"   table:"Record"`
 }
 
-// recordCursorPage is the ESCursorPagination envelope for session records.
-// ListResponse[T] can't be reused: its Next is an int, this is a cursor token.
+// recordCursorPage decodes the cursor-pagination response; ListResponse[T] can't be reused (its Next is an int, not a cursor token).
 type recordCursorPage struct {
 	Next    string          `json:"next"`
 	Results []SessionRecord `json:"results"`
--- a/api/websh/types.go
+++ b/api/websh/types.go
@@ -89,3 +89,15 @@ type ConnectRequest struct {
 	IsMaster bool   `json:"is_master"`
 	ReadOnly bool   `json:"read_only"`
 }
+
+type SessionRecord struct {
+	AddedAt string `json:"added_at" table:"Added At"`
+	Record  string `json:"record"   table:"Record"`
+}
+
+// recordCursorPage is the ESCursorPagination envelope for session records.
+// ListResponse[T] can't be reused: its Next is an int, this is a cursor token.
+type recordCursorPage struct {
+	Next    string          `json:"next"`
+	Results []SessionRecord `json:"results"`
+}
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"strconv"
 	"syscall"
 	"time"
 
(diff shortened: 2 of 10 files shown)


eunyoung14 (approved):
LGTM — approving. Well-scoped feature that fills a real parity gap, with the security-sensitive path handled carefully.

Verified locally: `go build ./...`, `go test -race ./api/websh/... ./cmd/websh/... ./cmd/worksession/... ./utils/...` pass; no merge conflict (up to date with `main`).

Notable: I checked the cursor pagination against the actual server contract rather than assuming DRF defaults — `alpacon-server`'s custom `ESCursorPagination` returns `next` as a **bare base64 token** (not a full URL) and honors `page_size`, so `cursor = page.Next`, `page_size = min(100, limit-len)`, the limit truncation, and the empty-next/empty-results termination are all correct. The bare-token test mock is representative.

Security path is sound: table output strips ANSI + C0/DEL/**C1** (8-bit CSI/OSC introducer) via replace-with-space + `strings.Fields` collapse; JSON stays verbatim and `encoding/json` escapes ESC, blocking sequence reconstruction. Good threat model for masked recordings in shared/invited sessions.

One minor wording nit (non-blocking, no code change required):
- The checklist says "worksession sanitizer convergence is behavior-neutral", but the shared `StripControlChars` (`IsControlRune`) now also strips **C1 (U+0080–U+009F)**, which the old worksession `stripControlChars` (C0 + DEL only) did not. So worksession timeline/preview output additionally strips C1 now — a positive hardening, but a behavior change rather than strictly neutral. Worth tweaking the wording for accurate history.

Nice work.

jisung-02:
I'll take care of the wording fix in a separate minor PR. Thanks.

PR #265, opened 2026-07-15
feat(exec, worksession): add --wait-approval custom timeout flag

### LINKED ISSUE #254: enhance(approval): make approval wait timeouts configurable via --timeout
## Background

Both approval wait loops use hardcoded timeouts:

- `exec --wait` gives up after 3 minutes (`approvalWaitTimeout`, `cmd/exec/run.go:26`), polling every 5 seconds.
- `work-session create --wait` gives up after 5 minutes (`pollMaxAttempts=30` × `pollInterval=10s`, `cmd/worksession/worksession_create.go:21-22`).

A human reviewer routinely takes longer than 3–5 minutes to approve. When the window elapses, the CLI falls back to the same pending message as if `--wait` had never been passed, and the caller has to re-invoke the identical wait loop. There is no flag to change this; the only comparable knob in the codebase is the `ALPACON_EXEC_TIMEOUT` env var, which covers command execution, not approval waits.

## Proposal

- Add a `--timeout <duration>` flag (e.g. `30m`) to `exec` and `work-session create`, effective only with `--wait`; reject it without `--wait` as a usage error.
- Keep the current values as defaults so existing behavior is unchanged.
- On timeout, keep the existing fallback (pending message + exit 4).

## Scope

- `cmd/exec/exec.go`, `cmd/exec/parse.go`, `cmd/exec/run.go`
- `cmd/worksession/worksession_create.go`
- README `--wait` mentions and the exit-code 4 row

jisung-02:
## Background

Both `alpacon exec` and `alpacon work-session create` already accept a boolean `--wait` flag that blocks while a sudo command or work session awaits human approval in the Alpacon console, then exits with code 4 if approval does not arrive in time. The timeout was hard-coded and asymmetric: exec waited 3m, work-session create waited 5m (30 attempts of 10s). Human approval routinely takes longer than that, and there was no way to raise the ceiling — the caller just re-ran the same wait. Internal dogfooding on the CI plus human-in-the-loop flow surfaced the same gap and proposed the `--wait-approval <duration>` spelling.

## Changes

Adds `--wait-approval DURATION` to both commands, alongside the existing `--wait`. `--wait` blocks with the default timeout; `--wait-approval 30m` implies `--wait` and sets a custom timeout parsed with `time.ParseDuration`, which must be positive and has no upper cap. Passing both is allowed and `--wait-approval` wins. Neither flag means no waiting, unchanged from before.

Implementation notes. `cmd/exec` parses flags by hand (`DisableFlagParsing: true`), so the flag is handled in `cmd/exec/parse.go` via the existing `extractFlagValue` path and `RemoteExecArgs.WaitTimeout()` derives the effective timeout; `cmd/worksession` uses pflag, and `resolveWaitTimeout` in `cmd/worksession/worksession_create.go` derives it there. A shared `utils.ParsePositiveDuration` (`utils/utils.go`) is the single duration-flag validator, now used by `--wait-approval` on both commands and by the pre-existing `--expires-in`. work-session create's poll loop switched from attempt-count to deadline-based: the old count-based loop collapsed to a single 0s attempt when the timeout was shorter than one interval (for example `--wait-approval 15s`) and timed out immediately, now guarded by `TestPollForApproval_WaitsFullTimeout`. websh command mode constructs `RemoteExecArgs` without the wait fields, so `--wait-approval` has no effect there and the doc string says so.

## Behavior change

`exec --wait` default timeout changes from 3m to 5m, so it now matches `work-session create`. `--wait` has shipped since exec v1.7.2, so a CI job that assumed a 3m ceiling now receives the same exit 4 two minutes later. No functionality is lost, and `--wait-approval 3m` restores the old timeout explicitly. The original 3m was chosen to mirror the MFA step-up retry ceiling (`maxRetryDuration` in `utils/error_handler.go`), but that is a self-service wait measured in seconds whereas approval is a third-party wait measured in minutes, so the two were never meant to match.

## Fixes found in review

Two correctness bugs surfaced during review and are fixed here. In exec, the resumed approval stream re-armed `StreamApprovedCommand` with the full timeout after the poll loop had already consumed part of the window, so a run could wait up to twice the requested duration; it now anchors a deadline up front and passes the remaining time. In work-session create, `--wait-approval=` (explicit empty) was indistinguishable from an unset flag, so the requested wait was silently dropped; it now threads `cmd.Flags().Changed("wait-approval")` so an explicit empty or blank value is rejected the same way exec's parser rejects it.

## Testing

`go test -race ./...` passes across all packages. `gofmt -l`, `go vet ./...`, and `golangci-lint run ./...` are clean with 0 issues. New table-driven tests `TestResolveWaitTimeout` and `TestWaitTimeout` cover precedence, the unset-vs-explicit-empty distinction, and invalid, zero, and negative durations; new cases in `TestParseRemoteExecArgs` cover the exec parser for the spaced and `=` forms, a missing value, and the `--detach` conflict. The 5m defaults are pinned as literals in `TestWaitTimeout` and `TestResolveWaitTimeout`, so a change to either default breaks a test.

## Follow-up

`parseExpiryFlag`'s tests (`TestParseExpiryFlag_*`) predate the table-driven style used by the new tests and could be converted for consistency. Left as a separate task to keep this PR scoped.

## Checklist

- [ ] Review the exec 3m to 5m default change with anyone relying on the old ceiling
- [x] Tests, gofmt, vet, and golangci-lint pass locally
- [x] `--wait-approval` documented in `README.md` and both commands' help

Closes #254


utils/utils.go
```diff
@@ -594,3 +594,15 @@ func SplitAndTrim(s, sep string) []string {
 	}
 	return result
 }
+
+// ParsePositiveDuration parses a duration flag value, rejecting non-positive ones.
+func ParsePositiveDuration(flagName, raw string) (time.Duration, error) {
+	d, err := time.ParseDuration(raw)
+	if err != nil {
+		return 0, fmt.Errorf("invalid %s value %q: %w", flagName, raw, err)
+	}
```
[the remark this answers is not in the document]
jisung-02: Fixed in 3fee3d7. ParsePositiveDuration now trims surrounding whitespace, so exec and work-session create normalize duration flag values identically.

cmd/worksession/worksession_create.go
```diff
@@ -376,11 +394,14 @@ func buildSudoPolicies(specs []string, reason string) []wsapi.SudoPolicyInline {
...
+// timeout elapses. untilActive=false returns on approved or active; untilActive=true
+// returns only on active (continues polling on approved until the server
+// auto-activates). Deadline-based rather than attempt-count-based so a timeout
+// under one interval (e.g. --wait-approval 15s) still waits the full duration.
+func pollForApproval(ac *client.AlpaconClient, id string, untilActive bool, interval, timeout time.Duration) (*wsapi.WorkSession, error) {
+	deadline := time.Now().Add(timeout)
+	for {
 		s, err := wsapi.GetWorkSession(ac, id)
 		if err != nil {
 			return nil, fmt.Errorf("polling failed: %w", err)
```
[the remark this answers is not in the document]
jisung-02: The poll at the deadline is intentional, not an off-by-one. The loop polls at t=0, interval, 2*interval, ... up to and including t=deadline, and that final poll is the last chance to detect an approval that landed during the closing sub-interval window.

- Returning before that poll would shrink the effective approval-detection window by up to one interval, which works against the flag's whole purpose (detecting approval within the requested duration).
- The only overshoot is the round-trip of that single final GetWorkSession, which is negligible next to a minutes-scale approval wait, and the loop never oversleeps past the deadline (time.Sleep(min(remaining, interval)) caps the last sleep at exactly remaining).

cmd/worksession/worksession_create_test.go
```diff
@@ -21,11 +21,31 @@ func TestPollForApprovalCancelled(t *testing.T) {
...
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"id":"ses-p","status":"pending"}`))
+	}))
+	defer ts.Close()
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	timeout := 30 * time.Millisecond
+	start := time.Now()
+	_, err := pollForApproval(ac, "ses-p", false, 10*time.Millisecond, timeout)
+	elapsed := time.Since(start)
```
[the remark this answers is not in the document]
jisung-02: This test only asserts a lower bound (elapsed >= timeout), so a slow or loaded runner cannot make it flaky.

The current value is appropriate because:
- A slower runner only increases elapsed (extra round-trip + scheduling jitter push it further above 30ms), which still satisfies elapsed >= timeout — the failure direction Copilot describes cannot occur here.
- The lower bound is guaranteed by construction: the loop returns "timed out" only once time.Until(deadline) <= 0, i.e. elapsed >= timeout, and time.Sleep never sleeps less than requested.
- Bumping the timeout would only slow the suite without adding coverage.

cmd/worksession/worksession_create.go
```diff
@@ -219,12 +227,12 @@ so it is recorded and scoped accordingly.`,
 				return
 			}
 		case useDecisionErrorNeedsWait:
-			if !waitApproval {
+			if !shouldWait {
 				utils.CliUsageErrorEnvelopeWithExit(opCreate, "--use requires the session to be active. Pass --wait to wait for approval, or run 'alpacon work-session use %s' after approval.", session.ID)
```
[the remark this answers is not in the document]
jisung-02: Fixed in f5e8c90. The --use usage error now reads "Pass --wait (or --wait-approval <duration> for a longer timeout) to wait for approval, ...".

cmd/exec/parse_test.go
```diff
@@ -434,6 +435,65 @@ func TestParseRemoteExecArgs(t *testing.T) {
...
+				WaitApproval: 10 * time.Minute,
+				Server:       "server",
+				Command:      "ls",
+			},
+		},
+		{
+			name:     "wait-approval missing value",
+			args:     []string{"--wait-approval"},
+			expected: RemoteExecArgs{Err: "flag needs an argument: --wait-approval"},
+		},
```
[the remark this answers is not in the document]
jisung-02: Fixed in f5e8c90. Added a "wait-approval explicit empty value" case to TestParseRemoteExecArgs asserting `--wait-approval=` is rejected, matching work-session create validation.

cmd/exec/run.go
```diff
@@ -169,31 +171,32 @@ func isApprovalDenial(err error) bool {
...
 		return err
 	}
 
 	spinner := utils.NewSpinner("Waiting for approval in the Alpacon console...")
 	spinner.Start()
 
-	timer := time.NewTimer(approvalWaitTimeout)
+	deadline := time.Now().Add(waitTimeout)
+	timer := time.NewTimer(waitTimeout)
 	defer timer.Stop()
```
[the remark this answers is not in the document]
jisung-02: Fixed in f5e8c90. On timeout the loop now emits a warning before returning the denial: "Approval wait timed out after <d>; the command is still pending. Pass --wait-approval <duration> to wait longer." So a user who already waited sees what happened and how to extend the window, rather than only the generic "use --wait" line.

README.md
```diff
@@ -291,7 +291,7 @@ What each refusal code means and what to do next:
...
+| `4`  | Pending human approval—the action is awaiting an out-of-band approve/reject in the Alpacon console (web/Slack), not refused. For `exec`, re-run the command after approval (or pass `--wait` on the original command to block; `--wait-approval <duration>` raises the wait timeout, default 5m); `websh` command mode has no `--wait`, so re-run via `alpacon exec --wait` to block; for `work-session create` the session already exists—after approval attach it with `alpacon work-session use <id>` (or pass `--wait` on the original create to block; `--wait-approval <duration>` raises the wait timeout, default 5m). Under `--output json`, a `{"status":"pending_approval", ...}` object is emitted. Returned by `exec` (and `websh` when running a command) on a `SUDO_APPROVAL_REQUIRED` sudo denial and by `work-session create` when the session lands pending |
```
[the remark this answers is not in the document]
jisung-02: The placeholder is already inside a backtick code span (`--wait-approval <duration>`), and GitHub-flavored Markdown does not parse HTML inside code spans, so `<duration>` renders literally.

The current form is appropriate because:
- The same table cell already uses `alpacon work-session use <id>` in a code span, and it renders fine — the `<placeholder>` pattern is the README prose convention here.
- Switching this one to DURATION would make it inconsistent with the adjacent `<id>`. The uppercase DURATION style is used in the CLI help output (`--wait-approval DURATION`), which this doc prose intentionally does not mirror.

cmd/exec/exec.go
```diff
@@ -54,7 +54,9 @@ Flags:
                                 Use 'alpacon exec logs JOB_ID' to retrieve the result.
   --wait                        When a sudo command needs human approval, block and
                                 re-attempt until a reviewer approves it in the Alpacon
-                                console (web), or the wait times out.
+                                console (web), or the wait times out (default 5m).
+  --wait-approval DURATION      Like --wait with a custom wait timeout (e.g. 30m;
+                                default 5m). Implies --wait.
```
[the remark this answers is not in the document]
jisung-02: Fixed in 8acbbeb. The exit-code 4 paragraph now reads "...or pass --wait to block—--wait-approval DURATION to block with a longer timeout)".

cmd/exec/run.go
```diff
@@ -16,14 +16,15 @@ import (
...
-	approvalWaitTimeout      = 3 * time.Minute
+	// approvalWaitPollInterval and defaultApprovalWaitTimeout bound the --wait poll
+	// for a SUDO_APPROVAL_REQUIRED denial. The interval is slower than the MFA
+	// step-up poll (api/mfa/mfa.go) because each tick re-submits and re-runs the
+	// remote command, and a human approving out of band in the console works on a
+	// seconds-to-minutes timescale, not sub-second. The timeout matches
+	// work-session create's so both approval waits behave alike; it's a default,
+	// not a ceiling—--wait-approval can exceed it uncapped.
+	approvalWaitPollInterval   = 5 * time.Second
+	defaultApprovalWaitTimeout = 5 * time.Minute
```
geunwoonoh: Warning: exec `--wait` default timeout changes 3m → 5m here. `--wait` has shipped since exec v1.7.2, so a CI job assuming a 3m ceiling now receives exit 4 two minutes later. Deliberate and documented (`--wait-approval 3m` restores the old value), but the PR checklist item "Review the exec 3m to 5m default change" is still unchecked — worth confirming with anyone relying on the old ceiling before merge.

cmd/exec/run.go
```diff
@@ -212,7 +217,13 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
 			// mid-wait; honor --wait by resuming the held job instead of exiting.
 			if errors.As(err, &pendingErr) {
 				spinner.Stop()
-				return event.StreamApprovedCommand(ac, pendingErr.CommandID, out, approvalWaitTimeout)
+				// Resume inside the original window: this loop already consumed part
+				// of it, so re-arming with the full timeout would double the wait.
+				remaining := time.Until(deadline)
```
geunwoonoh, line 217: Warning: this is the core correctness fix (anchor the deadline up front, resume with the remaining time so the wait can't stretch toward 2×), but no unit test covers it directly. The work-session side has `TestPollForApproval_WaitsFullTimeout`; the exec path relies on manual reasoning. Mocking the stream/poll makes it hard to test, but flagging it as a regression-safety gap.

cmd/exec/run.go
```diff
@@ -169,40 +171,43 @@ func isApprovalDenial(err error) bool {
...
 	for {
 		select {
 		case <-timer.C:
 			spinner.Stop()
-			// Return the last pending denial so the caller emits the standard
-			// pending-approval signal and exit code.
+			// The caller's generic pending-approval message says "use --wait", which
+			// misleads here since the user already waited; name the timeout and the
+			// flag that extends it before returning the denial.
+			utils.CliWarning("Approval wait timed out after %s; the command is still pending. Pass --wait-approval <duration> to wait longer.", waitTimeout)
```
geunwoonoh: Suggestion: this `CliWarning` names `--wait-approval`, and then the caller's standard pending-approval message (run.go:263) names it again — the user effectively sees the same guidance twice. Intentional reinforcement, but slightly redundant.

cmd/worksession/worksession_create.go
```diff
@@ -18,10 +18,10 @@ import (
 )
 
 const (
-	pollMaxAttempts   = 30
-	pollInterval      = 10 * time.Second
-	waitMsgApproval   = "Waiting for approval..."
-	waitMsgActivation = "Waiting for activation..."
+	pollInterval               = 10 * time.Second
+	defaultApprovalWaitTimeout = 5 * time.Minute // preserves the old 30 × 10s window
```
geunwoonoh: Suggestion: `defaultApprovalWaitTimeout = 5 * time.Minute` is duplicated here and in `cmd/exec/run.go:27`. The PR states the two approval waits are meant to behave alike, but the invariant isn't structurally enforced. A single `utils.DefaultApprovalWaitTimeout` would enforce it. Low priority — the two constants carry different rationales and both are pinned as literals in tests, so drift already breaks a test.

cmd/worksession/worksession_create.go
```diff
@@ -308,6 +313,19 @@ func parseExpiryFlag(expiresIn, expiresAt string) (string, error) {
 	return expiresAt, nil
 }
 
+// resolveWaitTimeout uses waitApprovalSet to separate an unset --wait-approval
+// from an explicitly empty value, which must be rejected rather than silently ignored.
+func resolveWaitTimeout(wait bool, waitApprovalRaw string, waitApprovalSet bool) (time.Duration, error) {
```
geunwoonoh: Suggestion (very minor): the `wait bool` parameter shadows the package-level `var wait`. Behavior is correct and the explicit-parameter design is good for testability — renaming the parameter (e.g. `waitFlag`) would just remove the shadow.

cmd/worksession/worksession_create.go
```diff
@@ -403,16 +424,18 @@ func pollForApproval(ac *client.AlpaconClient, id string, untilActive bool, inte
...
 		waitMsg := waitMsgApproval
 		if s.Status == approvedWorkSessionStatus {
 			waitMsg = waitMsgActivation
 		}
-		utils.CliInfo("%s (attempt %d/%d)", waitMsg, attempt, pollMaxAttempts)
-		if attempt < pollMaxAttempts {
-			time.Sleep(interval)
-		}
+		utils.CliInfo("%s (%s elapsed of %s)", waitMsg, (timeout - remaining).Round(time.Second), timeout)
+		time.Sleep(min(remaining, interval))
```
geunwoonoh: Suggestion (very minor): `time.Sleep(min(remaining, interval))` sleeps 0 when `interval <= 0`, which would busy-loop the poll. Not a real problem — production always passes `pollInterval = 10s` and `interval=0` is only reachable from a test that returns before this line. A small guard would make it more robust.

geunwoonoh (dismissed):
## Review summary

**Stack**: golang
**Changes**: +283 -64 across 11 files

I ran the full local verification before reviewing:

- `go build` ✅  ·  `go vet ./cmd/exec/... ./cmd/worksession/... ./utils/...` ✅  ·  `gofmt -l` clean ✅
- `go test -race ./cmd/exec/... ./cmd/worksession/... ./utils/...` → all three packages `ok` ✅
- No dangling references to the removed `pollMaxAttempts` or the renamed `waitApproval` (bool → string); every caller of the changed signatures (`RunExecWithApprovalWait`, `pollForApproval`, `WaitTimeout()`) is consistent ✅

### Overview

Adds `--wait-approval DURATION` alongside the existing boolean `--wait` on both `exec` and `work-session create`. `--wait-approval` implies `--wait` and sets a custom timeout; when both are passed `--wait-approval` wins. Alongside it: (1) three scattered duration parse/validate blocks are unified into `utils.ParsePositiveDuration`, (2) work-session polling moves from attempt-count to deadline-based (fixing a sub-interval timeout bug), (3) the exec resume path's double-wait bug is fixed, and (4) the exec `--wait` default timeout is unified 3m → 5m.

Overall this is a very polished PR. **No critical issues.**

---

### Warnings (2)

1. **exec `--wait` default 3m → 5m — potentially breaking.** `--wait` has shipped since exec v1.7.2; a CI job assuming a 3m ceiling now receives exit 4 two minutes later. It is deliberate and documented, and `--wait-approval 3m` restores the old value — but the PR checklist item *"Review the exec 3m to 5m default change"* is still unchecked. Not a code defect; flagging it as a **pre-merge confirmation** item.
2. **The core correctness fix (passing `remaining` on resume) has no unit test.** The double-wait fix in `RunExecWithApprovalWait` is logically correct, but no test covers the deadline path directly — the work-session side has `TestPollForApproval_WaitsFullTimeout`, whereas the exec path relies on manual reasoning. Mocking the stream/poll makes it hard to test, but it remains a regression-safety gap.

### Suggestions (4)

1. **`defaultApprovalWaitTimeout` is duplicated across two packages.** The PR states the two waits are meant to match, but the invariant isn't structurally enforced. Extracting a single `utils.DefaultApprovalWaitTimeout` would enforce it. Low priority — the two constants carry different rationales and both are pinned as literals in tests, so drift breaks a test.
2. **`resolveWaitTimeout(wait bool, ...)` shadows the package-level `var wait`.** Behavior is correct; the explicit-parameter design is good for testability, so just renaming the parameter (e.g. `waitFlag`) removes the shadow. Very minor nit.
3. **Duplicate messaging on exec timeout.** The `CliWarning` names `--wait-approval`, then the caller's standard pending-approval message names it again — the user effectively sees the same guidance twice. Intentional reinforcement, but slightly redundant.
4. **Busy-loop when `interval <= 0`.** `time.Sleep(min(remaining, interval))` sleeps 0 when `interval` is 0, hammering the poll. Not a real problem — production always passes `pollInterval = 10s` and `interval=0` is only reachable from a test that returns immediately. A guard would make it more robust.

### Highlights

- **Performance / correctness:** the deadline-based poll fixes a real latent bug (attempt-count collapsed to a single 0s attempt for sub-interval timeouts like `--wait-approval 15s`), guarded by `TestPollForApproval_WaitsFullTimeout`. `min(remaining, interval)` also keeps the final wait from overshooting the deadline.
- **DRY:** three inline parse+positive-check blocks (`--wait-approval` ×2, the pre-existing `--expires-in`) collapse into `utils.ParsePositiveDuration`, with `parseExpiryFlag`'s behavior preserved (existing tests pass).
- **exec double-wait fix:** anchoring the deadline up front and passing the remaining time bounds a wait that could otherwise stretch to nearly 2× the requested window.
- **unset vs explicit-empty:** `cmd.Flags().Changed("wait-approval")` rejects `--wait-approval=` rather than silently dropping it, matching the exec parser. Careful detail.
- **Test coverage:** precedence, unset/empty, invalid/zero/negative, and the `--detach` conflict are all covered with table-driven tests, and the comments explain the *why* well.

---

**Verdict: APPROVE** (with minor, non-blocking comments).

Checklist:
- [x] Follows Go conventions (const→var→type→func ordering, errcheck patterns, lowercase error strings)
- [x] Tests included and passing (verified locally with `-race`)
- [x] No security vulnerabilities (no hardcoded secrets/injection; the uncapped duration is intentional)
- [x] Documentation updated (README + both commands' help)

cmd/exec/run.go
jisung-02: Checked the three IaC repos that could drive hosts via `alpacon exec` (iac-onprem, iac-aws, alpacon-gitops) on origin/main — none pass `--wait` or `--wait-approval`. iac-onprem's `clean-runner-cache.yml` is the only workflow that uses `alpacon exec`, and every invocation is `exec … -- <cmd>` with no wait flag; the `--wait` hits in alpacon-gitops are all `kubectl delete --wait=`, unrelated to the CLI. So no CI relying on the old 3m ceiling exists in these repos. Keeping the default at 5m — `--wait-approval 3m` restores the old value for anyone who needs it.

cmd/exec/run.go
jisung-02: Added direct coverage in 5c96b23. `TestRunExecWithApprovalWait_ResumePassesRemainingNotFull` drives the resume path through new test seams (`runPresenceStepUp` / `streamApprovedCommand`) and asserts the resumed stream receives less than the full timeout — so a regression back to re-arming with the full window fails the test. `TestRunExecWithApprovalWait_TimesOutAfterWindow` covers the deadline path itself (waits the full anchored window, returns the last pending denial when approval never lands).

cmd/exec/run.go
jisung-02: Fixed in 928273b. The timeout `CliWarning` now reports only the timeout ("Approval wait timed out after <d>; the command is still pending."); the caller's pending-approval message keeps the `--wait-approval` guidance, so the flag is named once instead of twice.

cmd/worksession/worksession_create.go
jisung-02: Fixed in 928273b. Extracted `utils.DefaultApprovalWaitTimeout` and removed the duplicated constant from both `cmd/exec` and `cmd/worksession`, so the two approval waits now share one source of truth.

cmd/worksession/worksession_create.go
jisung-02: Fixed in 928273b. Renamed the parameter to `waitFlag`, removing the shadow of the package-level `var wait`.

cmd/worksession/worksession_create.go
jisung-02: Fixed in 928273b. `pollForApproval` now falls back to the remaining time when `interval` is non-positive (remaining > 0 is guaranteed by the deadline check just above), so a zero interval can no longer turn the poll into a busy-loop.

geunwoonoh (approved):

PR #272, opened 2026-07-20
fix(event): back off and give up chunk gap-fill re-fetch when a missing seq never fills

### LINKED ISSUE #269: fix(event): back off and give up chunk gap-fill re-fetch when a missing seq never fills
## Summary

When streaming a command's output, if a gap appears in the chunk `seq` sequence (a missing seq that never becomes available), the gap-fill logic re-issues the **same** `GET /api/events/commands/{id}/chunks/?...&seq__gte=N` request repeatedly with **no backoff and no give-up**, until the client's exec timeout (default 30m). A single command can generate thousands of identical requests.

## Observed behavior

- A single command was seen re-fetching the same `seq__gte=<N>` ~6,500 times over ~30 minutes (~3–4 req/s). The re-request rate is bounded only by network round-trip time, not by any interval.
- For automation clients that drive the CLI, these stuck streams dominate backend request volume during load spikes.

## Root cause (code)

- `api/event/event.go` → `applyChunk`: on a WS chunk with `seq > lastSeq+1`, it calls `GetCommandChunks(lastSeq+1)` to fill the gap. If seq `lastSeq+1` is never returned by the server, `lastSeq` never advances, so **every subsequent WS chunk re-issues the identical gap-fill request**. There is no backoff, no cap, and no eventual give-up.

## Impact

- Sustained unnecessary load (~3–4 req/s per stuck stream) for up to the exec timeout (default 30m).
- Multiple concurrent stuck streams from one client can produce hundreds of req/min against the backend.

## Proposed fix

Keep re-fetching a missing seq, but stop hammering when it isn't arriving:

- Apply **exponential backoff** while `lastSeq` does not advance between fetches (e.g. 0.3s → 1s → 2s → 5s cap).
- **Reset** to fast fetching as soon as `lastSeq` advances (a real gap-fill should stay responsive).
- **Give up** after a bounded number of no-progress attempts (or a max wait), surfacing a warning and ending the stream instead of looping to the exec timeout.

The loop already tracks `lastSeq`, so the backoff state can key directly off "no seq progress".

## Notes

- The upstream trigger — *why* a chunk seq goes missing / is never persisted server-side — is a separate concern to be tracked in the server/agent repos. This issue only covers making the client degrade gracefully instead of hammering.
- Behavior is unchanged since v1.8.1; no related safety net has been added (the recent adaptive-backoff work was for FTP transfer polling, not command output chunks).

jisung-02:
## Background

When streaming command output, if a gap appears in the chunk `seq` sequence (a seq that never arrives), the gap-fill logic in `applyChunk` re-issues the same `GET /api/events/commands/{id}/chunks/?...&seq__gte=N` request with no backoff and no give-up. Because `lastSeq` cannot advance, every subsequent WS chunk re-triggers the identical gap-fill request, and the re-request interval is bounded only by network round-trip time, continuing until the exec timeout (default 30m). A single command was observed re-requesting the same `seq__gte=<N>` roughly 6,500 times over about 30 minutes. For automation clients driving the CLI, these stuck streams dominate backend request volume during load spikes.

Why a chunk seq goes missing in the first place (server/agent-side persistence) is a separate concern; this change only covers making the client degrade gracefully instead of hammering indefinitely.

## What changed

`api/event/event.go` was modified so gap-fill keeps trying but stops hammering the server when a seq is not arriving.

- Exponential backoff: while `lastSeq` does not advance, the re-request interval grows from 0.3s, doubling up to a 5s cap (`gapFillInterval`). Requests now fire only after the backoff window elapses, instead of on every WS chunk.
- Reset on progress: once `lastSeq` advances, `noProgress` resets to 0 so a legitimate gap-fill stays responsive.
- Give up: when no-progress attempts reach `gapFillMaxNoProgress` (11), the seq is given up and skipped so streaming resumes (`giveUpGap`). The given-up seq is recorded in `gapFillState.skipped`.
- Recover at command end: on command completion, `recoverSkipped` makes one final fetch for the skipped seqs. If they were persisted in the meantime, it prints them out of order and reports it; if they are still absent, it warns explicitly about the loss.
- Terminal drain: `drainRemainingChunks` flushes the remaining chunks and reports any leftover gap in a single warning.

State is held in the `gapFillState` struct (`lastAttempt`, `noProgress`, `skipped`), and the backoff decision keys directly off the "no seq progress" signal the loop already tracked.

## Safety

Gap enumeration is bounded (`maxGapWidth`, 100,000) so a buggy or hostile server cannot stall the CLI or exhaust memory via a server-controlled seq delta. A command's chunk count is capped well below this server-side, so exceeding it can only come from an abnormal server. The bound is applied both per individual gap and cumulatively (across `skipped`, and across `missing` in the drain), so neither one huge gap nor the sum of many small gaps can grow the slices without limit. Gaps beyond the bound are not enumerated but reported with a summarized warning.

## Tests

Gap-fill unit tests were added to `api/event/event_test.go` (backoff interval, skip/recovery paths, drain warnings, cumulative/per-gap bound edges). `go test -race ./...` passes.

Before/after behavior was compared against the `web-editor` live server using debug instrumentation that force-drops a specific seq from both the WS stream and the REST endpoint. The instrumentation was not committed.

Normal run (no loss): main and this branch produce byte-identical stdout (5000/5000 lines, exit 0), confirming the normal streaming path is untouched.

Issue-shaped reproduction (a long-lived stream emitting one line per second, with an early seq permanently missing), comparing instrumented `GetCommandChunks` calls:

- main: re-issued the same `seq__gte=3` 57 times over the roughly 62-second life of the stream (about 1 req/s) and exited with no warning. The request rate is bounded only by the stream's lifetime, which scales directly to the 30-minute, thousands-of-requests case the issue observed.
- this branch: the re-request interval widened via backoff and then gave up after 11 no-progress attempts (re-requests capped at 12), skipping mid-stream and resuming the rest of the output normally.

Recover-at-command-end path (the skipped seq becomes persisted server-side after the give-up):

```
Warning: chunk seq(s) [3] not arrived after 11 attempts; skipping for now (will retry at command end)
Warning: late chunk seq(s) [3] recovered at command end (printed out of order)
```

The skipped chunk's content is printed out of order at the tail and the command exits 0. If the seq is still absent at command end, it instead reports the loss with `chunk seq(s) [3] never arrived; output may be incomplete`. Either way, the user is always notified.

## Checklist

- [x] `gofmt`, `go vet`, `golangci-lint run ./api/event/...` report no issues
- [x] Normal streaming output confirmed identical to pre-fix against the live server
- [x] Debug instrumentation not included in any commit

Closes #269


api/event/event.go
```diff
@@ -280,15 +395,21 @@ func streamSubscribed(ac *client.AlpaconClient, session *EventSessionResponse, l
...
 	if chunk.Seq > lastSeq+1 {
+		// Throttle: while lastSeq is stuck, re-fetch only after the backoff window.
+		if g.noProgress > 0 && gapFillNow().Sub(g.lastAttempt) < gapFillInterval(g.noProgress) {
+			return lastSeq
+		}
+		before := lastSeq
 		missing, err := GetCommandChunks(ac, cmdID, lastSeq+1)
+		g.lastAttempt = gapFillNow()
 		if err != nil {
 			utils.CliWarning("failed to fetch missing chunks (seq %d..%d): %v; output may be incomplete",
```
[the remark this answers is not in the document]
jisung-02: Fixed. Moved the give-up/skip decision into the successful-fetch branch. On a fetch error `noProgress` still increments so the backoff window keeps growing (no hammering), but `giveUpGap` is only reached when a successful fetch confirms the seq is absent.

So a transient network/API error no longer emits the misleading "not arrived" warning or prematurely skips output — a persistent failure now backs off and defers to the terminal drain instead. Added `TestApplyChunk_FetchErrorBacksOffButNeverGivesUp` to cover this.

api/event/event_test.go
```diff
@@ -1084,3 +1085,324 @@ func TestRunCommandStreaming_FallbackToResultWhenNothingStreamed(t *testing.T) {
...
+	assert.Contains(t, stderr, "chunk seq(s) [5] never arrived")
+}
+
+// A hostile/buggy server can set a chunk seq arbitrarily high; giveUpGap must
+// not enumerate the whole hole (which would spin the CLI and grow g.skipped
+// without bound), yet still deliver the chunks it did fetch and resume past it.
+func TestGiveUpGap_BoundsHugeServerSeq(t *testing.T) {
+	g := &gapFillState{}
+	out := &bytes.Buffer{}
+	nextSeq := 1 << 40 // would hang if enumerated seq-by-seq
```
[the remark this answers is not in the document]
jisung-02: Fixed. Changed `1 << 40` to `1 << 30`, which still far exceeds maxGapWidth (100_000) so the huge-gap path is exercised, but fits a 32-bit int so the `386` GoReleaser build compiles. Verified with `GOARCH=386 go vet ./api/event/`.

api/event/event_test.go
```diff
@@ -1084,3 +1085,324 @@ func TestRunCommandStreaming_FallbackToResultWhenNothingStreamed(t *testing.T) {
...
+	})
+
+	assert.Equal(t, 12, last)
+	assert.Equal(t, "c3\nc6\nc9\nc12\n", out.String(), "all delivered chunks still printed")
+}
+
+// The terminal drain must likewise bound a server-supplied huge seq instead of
+// enumerating every missing seq into the warning slice.
+func TestDrainRemainingChunks_BoundsHugeServerSeq(t *testing.T) {
+	huge := 1 << 40
```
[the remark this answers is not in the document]
jisung-02: Fixed. Changed `1 << 40` to `1 << 30`, which still far exceeds maxGapWidth (100_000) so the huge-gap path is exercised, but fits a 32-bit int so the `386` GoReleaser build compiles. Verified with `GOARCH=386 go vet ./api/event/`.

api/event/event.go
```diff
@@ -20,6 +20,118 @@ const (
...
+		}
+		lastSeq = s
+	}
+	if recorded == len(lost) {
+		utils.CliWarning("chunk seq(s) %v not arrived after %d attempts; skipping for now (will retry at command end)",
+			lost, gapFillMaxNoProgress)
+	} else {
+		utils.CliWarning("chunk seq(s) %v not arrived after %d attempts; skipping (skip budget exhausted, no retry); output may be incomplete",
+			lost, gapFillMaxNoProgress)
+	}
```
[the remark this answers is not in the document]
jisung-02: Fixed. Split the warning three ways: fully recorded ("will retry at command end"), partially recorded ("recorded N of M for retry at command end, skip budget exhausted"), and none recorded ("skip budget exhausted, no retry"). The partial case now states exactly how many of the lost seqs get the command-end retry.

api/event/event_test.go
```diff
@@ -1084,3 +1085,324 @@ func TestRunCommandStreaming_FallbackToResultWhenNothingStreamed(t *testing.T) {
...
+			next := last + 3 + 1 // gap of 3 seqs, each under maxGapWidth=4
+			last = g.giveUpGap(last, next, nil, out)
+		}
+	})
+
+	assert.Equal(t, maxGapWidth, len(g.skipped), "cumulative skipped capped at maxGapWidth")
+	// The first gap is fully recorded (retryable); once the cap fills, later gaps
+	// must not claim a retry they won't get.
+	assert.Contains(t, stderr, "will retry at command end")
+	assert.Contains(t, stderr, "skip budget exhausted, no retry")
```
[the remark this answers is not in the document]
jisung-02: Fixed. The test now asserts all three wordings emitted across the four gaps: the fully-recorded "will retry at command end" (gap 1), the new partial "recorded 1 of 3 for retry at command end" (gap 2, where the budget fills mid-gap), and "skip budget exhausted, no retry" (gaps 3-4).

api/event/event.go
```diff
@@ -311,18 +450,48 @@ func applyChunk(ac *client.AlpaconClient, cmdID string, lastSeq int, chunk Chunk
...
+		// hostile server can't grow missing without limit; excess is summarized.
+		if gap := c.Seq - lastSeq - 1; gap > 0 {
+			if len(missing)+gap > maxGapWidth {
+				truncated = true
+			} else {
+				for s := lastSeq + 1; s < c.Seq; s++ {
+					missing = append(missing, s)
+				}
+			}
 		}
```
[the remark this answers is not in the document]
jisung-02: Fixed. Changed the guard to compare against the remaining budget by subtraction: `gap > maxGapWidth-len(missing)`. The guard itself keeps `len(missing) <= maxGapWidth`, so the RHS is always >= 0 and no operand can wrap regardless of the server-controlled seq.

For the record, the old sum could not actually overflow through this path — `final` is sorted ascending and `c.Seq <= lastSeq` dedupes, so every value in `missing` is a distinct seq below the current `lastSeq`, giving `len(missing)+gap <= c.Seq <= MaxInt`. But that safety rested on a subtle invariant a future refactor could break; the subtraction form is safe by construction. Behavior unchanged, existing bound tests (`TestDrainRemainingChunks_Bounds*`) still pass.

api/event/event.go
```diff
@@ -280,16 +400,27 @@ func streamSubscribed(ac *client.AlpaconClient, session *EventSessionResponse, l
...
 	if chunk.Seq <= lastSeq {
 		return lastSeq
 	}
 	if chunk.Seq > lastSeq+1 {
+		// Throttle: while lastSeq is stuck, re-fetch only after the backoff window.
+		if g.noProgress > 0 && gapFillNow().Sub(g.lastAttempt) < gapFillInterval(g.noProgress) {
+			return lastSeq
+		}
+		before := lastSeq
 		missing, err := GetCommandChunks(ac, cmdID, lastSeq+1)
```
geunwoonoh, line 416: Warning: gap-fill re-fetch pulls the entire output tail instead of only the gap window (performance).

`GetCommandChunks(ac, cmdID, lastSeq+1)` resolves to `FetchAllPages` with `seq__gte=lastSeq+1`, i.e. it downloads *every page* from the hole to the current end of output. But this call only needs `[lastSeq+1, chunk.Seq-1]` — the contiguous-advance loop just below already `break`s at `c.Seq > chunk.Seq`, so anything past the live chunk is fetched and discarded.

While the hole stays open on a command that keeps producing output, the tail grows, so even though backoff caps the *count* at ~11, each re-fetch is O(current total tail). If the endpoint supports `seq__lte` (as it does `seq__gte`), bounding the fetch to `seq__lte=chunk.Seq-1` would make each re-fetch constant-sized.

(This fetch-to-end predates the PR, but this is the natural place to bound it since the change already adds `maxGapWidth`-style bounds nearby.)

api/event/event.go
```diff
@@ -20,6 +20,123 @@ const (
...
+}
+
+// recoverSkipped makes a final fetch for seqs skipped mid-stream (giveUpGap),
+// printing late content out of order and reporting still-absent seqs as lost.
+// g.skipped is ascending, so the fetch starts at its first seq.
+func (g *gapFillState) recoverSkipped(ac *client.AlpaconClient, cmdID string, out io.Writer) {
+	if len(g.skipped) == 0 {
+		return
+	}
+	final, err := GetCommandChunks(ac, cmdID, g.skipped[0])
```
geunwoonoh, line 106: Suggestion: `recoverSkipped` has the same fetch-to-end pattern (performance).

`GetCommandChunks(ac, cmdID, g.skipped[0])` fetches from the earliest skipped seq all the way to the end of output, then keeps only the few seqs in `skipped`. For a long-running command where an early seq was skipped, this re-downloads the whole tail into `bySeq` at command end. Since `skipped` is ascending, bounding with `seq__lte=skipped[len-1]` would avoid the large fetch. Low severity — it runs once, only on the abnormal give-up path.

api/event/event.go
```diff
@@ -20,6 +20,123 @@ const (
...
+			if len(g.skipped) < maxGapWidth {
+				g.skipped = append(g.skipped, s)
+				recorded++
+			}
+		}
+		lastSeq = s
+	}
+	switch {
+	case recorded == len(lost):
+		utils.CliWarning("chunk seq(s) %v not arrived after %d attempts; skipping for now (will retry at command end)",
```
geunwoonoh: Suggestion: large sub-`maxGapWidth` gaps print the full seq slice via `%v`.

Here (and in the drain path around `drainRemainingChunks`) `lost`/`missing` are printed with `%v`. A gap just under `maxGapWidth` (100k) could dump ~100k ints as a single stderr warning line. The `> maxGapWidth` branch already uses a compact `%d..%d` range format, so this sub-threshold path is the gap in that defense. Consider collapsing to a range/summary once the count exceeds a small threshold, for consistency with the oversized-gap branch.

api/event/event_test.go
```diff
@@ -1084,3 +1085,325 @@ func TestRunCommandStreaming_FallbackToResultWhenNothingStreamed(t *testing.T) {
...
+	return buf.String()
+}
+
+// holeServer serves /chunks/?seq__gte=N returning every seq >= N except those
+// in `hole`. fetches counts the chunk requests. Used to simulate a seq that is
+// never persisted server-side.
+func holeServer(t *testing.T, maxSeq int, hole map[int]bool, fetches *int) *client.AlpaconClient {
+	t.Helper()
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		*fetches++
```
geunwoonoh: Suggestion: test counter uses a plain `int` while this file elsewhere uses `atomic.Int32`.

`holeServer` increments `*fetches` from the handler goroutine and reads it from the test goroutine. `-race` passes (requests are sequential in these tests), but `TestGetEventList` / `TestPollCommandExecution` already use `atomic.Int32` for the same purpose — matching that would be more consistent and robust. Minor: `TestRecoverSkippedChunks_RecoversLatePersistedSeq` also asserts only the output, not the `recovered`/`never arrived` warnings on the mixed recover/lost path — asserting those would tighten coverage.

geunwoonoh (commented):
## PR Review: #272 — fix(event): back off and give up chunk gap-fill re-fetch when a missing seq never fills

**Stack**: Go (golang)
**Changes**: +502 -8 across 2 files (`api/event/event.go`, `api/event/event_test.go`)

---

### Summary

When a permanent hole appears in the chunk `seq` sequence during WS streaming, `applyChunk` previously re-issued the same `seq__gte=N` REST request with no backoff and no give-up on every subsequent WS chunk (observed ~6,500 times over 30 min). This PR fixes it with (1) exponential backoff (0.3s→5s cap), (2) reset-on-progress, (3) give-up + skip after 11 no-progress attempts (`giveUpGap`), (4) a single end-of-command recovery pass (`recoverSkipped`), and (5) a terminal-drain warning (`drainRemainingChunks`), plus per-gap and cumulative `maxGapWidth` bounds to defend against a buggy/hostile server. The design, test coverage, and comments are all high quality.

I re-verified the risky scenarios locally and found no correctness bugs: throttled WS chunks are safely dropped because REST is the source of truth; the live chunk is still delivered after give-up; recovered skipped seqs are not double-printed; the budget comparison is overflow-safe; and a transient fetch error never triggers a give-up. `go build`, `go vet`, `gofmt -l`, and `go test -race ./api/event/...` all pass locally.

### Critical Issues (0)

None — no merge blockers.

### Warnings (1)

1. **[event.go:416]** Gap-fill re-fetch pulls the entire output tail instead of only the gap window (performance). See inline comment.

### Suggestions (3)

1. **[event.go:106]** `recoverSkipped` has the same fetch-to-end pattern and could bound the upper seq. See inline comment.
2. **[event.go:85]** Large sub-`maxGapWidth` gaps print the full seq slice via `%v`. See inline comment.
3. **[event_test.go:1112]** Test counter uses a plain `int` while the file elsewhere uses `atomic.Int32`. See inline comment.

### Highlights

- **Excellent adversarial test coverage**: hostile huge seq (`1<<30`, 32-bit-safe), per-gap and cumulative `maxGapWidth` bounds, and a deterministic clock seam (`gapFillNow`) driving backoff timing.
- **Careful correctness**: gives up only after a *successful* fetch confirms absence (transient errors back off but never drop output); the contiguous-advance loop stops at the first hole so a chunk racing ahead of persistence is not skipped as a duplicate.
- **Overflow-safe budget check**: compares `maxGapWidth - len(missing)` by subtraction rather than adding a server-controlled gap.
- **Go conventions clean**: `const→var→type→func` ordering, errcheck (`_, _ = fmt.Fprint`), gofmt/vet clean; comments focus on the non-obvious "why".

---

**Verdict**: COMMENTED — no merge blockers. Only the one Warning (bounding the gap-fill fetch) is worth considering; the rest are optional.

**Checklist**
- [x] Follows stack conventions (declaration order / errcheck / gofmt / vet)
- [x] Tests included and passing (`go test -race` re-confirmed locally, ~53s ok)
- [x] No security vulnerabilities (no hardcoded secrets/injection; actually hardens against a hostile server)
- [x] No duplicated/unused code (all new symbols are referenced)

choongbum (dismissed):

api/event/event.go
jisung-02: Verified against alpacon-server: the chunks endpoint (`events/api/views.py`, `chunks` action) only parses `seq__gte` manually—there is no FilterSet, so a `seq__lte` param would be silently ignored and the full tail would still be returned. Bounding the fetch window therefore requires a server-side change first. I'll add `seq__lte` support to alpacon-server and then bound this fetch (and `recoverSkipped`) in a follow-up PR.

api/event/event.go
jisung-02: Same as the gap-fill re-fetch above: the server does not support `seq__lte` yet, so this will be bounded together with that fix in a follow-up PR once alpacon-server accepts the param.

api/event/event.go
jisung-02: Fixed in f3eb912: seq slices in warnings now go through `formatSeqs`, which lists up to 20 seqs verbatim and collapses larger lists to a `lo..hi (N seqs)` summary. Applied to the giveUpGap, recoverSkipped, and drain paths.

api/event/event_test.go
jisung-02: Fixed in 56493f8: `holeServer`'s counter is now `atomic.Int32`, and `TestRecoverSkippedChunks_RecoversLatePersistedSeq` asserts the recovered/never-arrived warnings on the mixed recover/lost path.

geunwoonoh (approved):

choongbum (approved):

Issue #275, opened 2026-07-20
webftp-log --server/--user filters return nothing for hyphenated names; resolve names to ids instead


jisung-02:
## Background

`alpacon webftp-log` filters are forwarded to the server as display-name strings (`api/webftp/webftp.go:16-29`):

```go
if serverName != "" { params["server_name"] = serverName }
if userName != ""   { params["user_name"]   = userName }
if action != ""     { params["action"]      = action }
```

On the server, `/api/history/webftp-logs/` maps `server_name` / `user_name` to analyzed `Text` fields and filters them with an Elasticsearch `term` query. `term` does not analyze the query string, so any value the analyzer tokenizes (hyphens, spaces, uppercase) never matches. A server named `web-editor` is indexed as `["web", "editor"]`, so `?server_name=web-editor` always returns zero hits even when logs exist.

## Symptom

```
alpacon webftp-log --server web-editor   # returns nothing, even though web-editor has upload logs
alpacon webftp-log --user "Some Name"    # same, whenever the name is tokenized
```

Names that happen to be a single token (e.g. `staging`) work by accident, which hides the bug on small setups.

## Root cause

The CLI filters on the denormalized, analyzed display-name fields (`server_name`, `user_name`) instead of the stable `Keyword` identifier fields (`server`, `user`) that the endpoint also accepts. `server` / `user` are indexed as `Keyword` (not analyzed), so exact-match filtering on them works on all historical data with no server change.

The web frontend already does this: its audit/files history view resolves the picked server/user to their UUID and sends `?server=<uuid>` / `?user=<uuid>`, never `server_name` / `user_name`. The CLI should match that behavior.

## Suggested fix

In `runWebFTP` (`cmd/webftp/webftp.go`), before calling `GetWebFTPLogList`, resolve names to ids and send the `Keyword` params:

- `--server <name>` -> `server.GetServerIDByName(ac, name)` -> `params["server"] = <uuid>` (helper already exists at `api/server/server.go:82`, used by `cp`/`exec`)
- `--user <name>` -> `iam.GetUserIDByName(ac, name)` -> `params["user"] = <uuid>` (helper already exists at `api/iam/iam.go:210`)
- `--action` stays as-is (`action` is a `Keyword`, already works)

[...]

PR #277, opened 2026-07-20
fix: decode ES cursor pagination in webftp-log, audit, and log commands

### LINKED ISSUE #274: webftp-log (and audit/log) crash with unmarshal error past the first page due to ES cursor pagination
## Background

`alpacon webftp-log` (alias `webftp-logs`) reads `/api/history/webftp-logs/`. On the server this endpoint uses a **different pagination contract** from the regular DB endpoints.

- Regular DB endpoints: `MyPageNumberPagination` — `next`/`previous`/`current`/`last` are integer page numbers.
- Elasticsearch-backed endpoints (webftp-logs, server logs, audit activity, websh records): `ESCursorPagination` — `next`/`previous` are base64 **cursor strings**, and `current`/`last` are absent from the response.

The shared `api.ListResponse[T]` type (`api/types.go`) only models the former, declaring `Next int` / `Previous int`. When `webftp.GetWebFTPLogList` (`api/webftp/webftp.go:36`) unmarshals a cursor response into this type, the cursor string cannot fit into an `int` and the call fails.

## Symptom

Whenever the result set is larger than the server's first page (default `page_size` 15, `max_page_size` 100) so that `next` carries a cursor string, the command always exits with:

```
Error: Failed to get WebFTP logs: json: cannot unmarshal string into Go struct field ListResponse[github.com/alpacax/alpacon-cli/api/webftp.WebFTPLogEntry].next of type int.
```

It happens to work when the response fits in a single page (`next` is `null`), e.g. small `--tail` values on a low-traffic workspace. That is why it works on small workspaces and reproduces on workspaces with accumulated logs.

## Reproduce

On a workspace with more than 100 transfer logs:

```
alpacon webftp-log --tail 200
alpacon webftp-log            # also reproduces with no filter
```

## Root cause

- `ListResponse[T]` in `api/types.go` is declared with `Next int` / `Previous int` / `Current int` / `Last int` — the page-number contract only.
- `api/webftp/webftp.go:36` parses a cursor response into this type.

The `ESCursorPagination` cursor design is intentional on the server (keyset `search_after` pagination that avoids the ES `max_result_window` ceiling). So reverting the server to page numbers is not the fix — the **CLI must understand the cursor contract**.

## Suggested direction

1. Add a response type dedicated to the ES cursor endpoints, e.g.:
   ```go
   type CursorListResponse[T any] struct {
       Count    int    `json:"count"`
       Next     string `json:"next"`
       Previous string `json:"previous"`
       Results  []T    `json:"results"`
   }
   ```
   (`current`/`last` are not part of this contract, so they are omitted.)
2. Change `GetWebFTPLogList` to unmarshal into this type.
3. When `--tail N` exceeds the server `max_page_size` (100), follow the `next` cursor and accumulate pages until `N` entries are collected (currently a single request caps the result at 100).

There is already prior art for this in the repo: `recordCursorPage` in `api/websh/types.go:97` decodes the same cursor contract for websh records, with the explicit comment that `ListResponse[T]` can't be reused because its `Next` is an int, not a cursor token. This fix should generalize that approach.

## Scope

Other commands hitting the same cursor contract have the identical bug and should be fixed together:

- `alpacon audit` — `/api/audit/activity/` (`api/audit/audit.go:41`), server `ActivityViewSet` uses `ESCursorPagination`.
- `alpacon log` — `/api/history/logs/` (`api/log/log.go:34`), server `LogViewSet` uses `ESCursorPagination`.

Not affected: `alpacon websh` record listing already uses a cursor-specific type; `alpacon event` (`/api/events/commands/`) uses page-number pagination.

jisung-02:
## Background

`alpacon webftp-log`, `alpacon audit`, and `alpacon log` read Elasticsearch-backed endpoints that the server paginates with `ESCursorPagination`: `next`/`previous` carry base64 cursor strings (or null) and `current`/`last` are absent. The shared `api.ListResponse[T]` in `api/types.go` models only the page-number contract (`Next int`), so whenever a result set exceeded the first page the commands failed with `json: cannot unmarshal string into Go struct field ... next of type int`. A single request also capped output at the server's `max_page_size` (100), so `--tail` values above 100 were never honored. The cursor design is intentional on the server (keyset `search_after` pagination), so the fix teaches the CLI the cursor contract.

## Changes

- Add `api.CursorListResponse[T]` (`api/types.go`) for the cursor contract and `api.FetchCursorPages[T]` (`api/pagination.go`), which requests `page_size = min(100, remaining)`, follows the `next` cursor until `tail` entries are accumulated or pages run out, and truncates any overshoot. The loop generalizes the previously working inline implementation in `websh.GetSessionRecords`.
- Guard the helper against a non-positive `tail` (returns an empty result without sending any request) and copy the caller's params map on entry so the pagination loop never leaves `page_size`/`cursor` residue behind. `FetchAllPages` gets the same map-copy fix.
- Switch `GetWebFTPLogList` (`api/webftp/webftp.go`), `GetAuditLogList` (`api/audit/audit.go`), and `GetSystemLogList` (`api/log/log.go`) to the helper. Signatures keep their positional shape, so the `cmd/` layer is unchanged.
- Refactor `websh.GetSessionRecords` onto the same helper and delete the now-redundant `recordCursorPage` type.
- Reject a non-positive `--tail` at the `cmd/` layer with a usage error (`utils.RequirePositiveInt`), matching `websh records`' `--limit` handling. The existing `--limit` and `--valid-days` guards are routed through the same helper.

Behavior change: `--tail 0` (or a negative value) now exits with a usage error instead of printing one server-default page (15 entries). All three commands default `--tail` to 25, so this only affects an explicit non-positive `--tail`.

## Tests

- New `api/pagination_test.go` covers the helper: single page (null `next`), cursor following with limit truncation, `page_size` derivation, non-positive limit (0 and -1, the latter reproducing the slice-bounds panic), a pathological page (`next` token with no results) that must not loop, a stale caller-supplied `cursor` param that must not leak into the first request, that the caller's params map is not mutated, and the error paths (a mid-pagination HTTP error and a malformed JSON body both propagate the error and discard partial results).
- Regression tests in `api/webftp/webftp_test.go` and `api/audit/audit_test.go` reproduce the original crash shape (base64 `next`) and verify multi-page accumulation.
- Call-site filter params are pinned to the wire format: audit resolves the username to a user ID and sends `user`/`app`/`model`, webftp sends `server_name`/`user_name`/`action`, and log sends the resolved server ID alongside `page_size`.
- `api/log/log_test.go` now encodes the cursor contract and asserts no follow-up request once `tail` is satisfied.
- Existing `TestGetSessionRecords_*` tests are retained as the refactor safety net; the limit-wiring assertion was folded into `TestGetSessionRecords_FollowsCursor`.
- `go test -race ./...` passes and `golangci-lint run ./...` reports 0 issues. Not verified against a live workspace with over 100 accumulated logs; coverage relies on the httptest-based reproduction of the server contract.

Closes #274

### DIFF
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -57,6 +57,9 @@ func FetchCursorPages[T any](ac *client.AlpaconClient, endpoint string, params m
 		params["page_size"] = strconv.Itoa(min(maxPageSize, limit-len(result)))
 		if cursor != "" {
 			params["cursor"] = cursor
+		} else {
+			// Drop any caller-supplied cursor so the first request starts from the first page.
+			delete(params, "cursor")
 		}
 
 		responseBody, err := ac.SendGetRequest(utils.BuildURL(endpoint, "", params))
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -41,3 +41,43 @@ func FetchAllPages[T any](ac *client.AlpaconClient, endpoint string, params map[
 
 	return result, nil
 }
+
+// FetchCursorPages follows the Elasticsearch cursor contract, accumulating up to limit items.
+func FetchCursorPages[T any](ac *client.AlpaconClient, endpoint string, params map[string]string, limit int) ([]T, error) {
+	if params == nil {
+		params = make(map[string]string)
+	}
+
+	// The server caps page_size at 100 (ESCursorPagination.max_page_size).
+	const maxPageSize = 100
+
+	var result []T
+	cursor := ""
+	for len(result) < limit {
+		params["page_size"] = strconv.Itoa(min(maxPageSize, limit-len(result)))
+		if cursor != "" {
+			params["cursor"] = cursor
+		}
+
+		responseBody, err := ac.SendGetRequest(utils.BuildURL(endpoint, "", params))
+		if err != nil {
+			return nil, err
+		}
+
+		var page CursorListResponse[T]
+		if err = json.Unmarshal(responseBody, &page); err != nil {
+			return nil, err
+		}
+
+		result = append(result, page.Results...)
+		if page.Next == "" || len(page.Results) == 0 {
+			break
+		}
+		cursor = page.Next
+	}
+
+	if len(result) > limit {
+		result = result[:limit]
+	}
+	return result, nil
+}
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -44,6 +44,9 @@ func FetchAllPages[T any](ac *client.AlpaconClient, endpoint string, params map[
 
 // FetchCursorPages follows the Elasticsearch cursor contract, accumulating up to limit items.
 func FetchCursorPages[T any](ac *client.AlpaconClient, endpoint string, params map[string]string, limit int) ([]T, error) {
+	if limit <= 0 {
+		return nil, nil
+	}
 	if params == nil {
 		params = make(map[string]string)
 	}
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -3,20 +3,26 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"maps"
 	"strconv"
 
 	"github.com/alpacax/alpacon-cli/client"
 	"github.com/alpacax/alpacon-cli/utils"
 )
 
+// copyParams returns a shallow copy so the pagination loop never mutates the caller's map.
+func copyParams(params map[string]string) map[string]string {
+	out := make(map[string]string, len(params)+2)
+	maps.Copy(out, params)
+	return out
+}
+
 func FetchAllPages[T any](ac *client.AlpaconClient, endpoint string, params map[string]string) ([]T, error) {
 	var result []T
 	page := 1
 	const pageSize = 100
 
-	if params == nil {
-		params = make(map[string]string)
-	}
+	params = copyParams(params)
 	params["page"] = strconv.Itoa(page)
 	params["page_size"] = strconv.Itoa(pageSize)
 
@@ -48,12 +54,13 @@ func FetchCursorPages[T any](ac *client.AlpaconClient, endpoint string, params m
 	if limit <= 0 {
 		return nil, nil
 	}
-	if params == nil {
-		params = make(map[string]string)
-	}
 
 	// The server caps page_size at 100 (ESCursorPagination.max_page_size).
 	const maxPageSize = 100
+
+	params = copyParams(params)
+	// Drop any caller-supplied cursor so the first request starts from the first page.
+	delete(params, "cursor")
 
 	result := make([]T, 0, min(limit, maxPageSize))
 	cursor := ""
@@ -61,9 +68,6 @@ func FetchCursorPages[T any](ac *client.AlpaconClient, endpoint string, params m
 		params["page_size"] = strconv.Itoa(min(maxPageSize, limit-len(result)))
 		if cursor != "" {
 			params["cursor"] = cursor
-		} else {
-			// Drop any caller-supplied cursor so the first request starts from the first page.
-			delete(params, "cursor")
 		}
 
 		responseBody, err := ac.SendGetRequest(utils.BuildURL(endpoint, "", params))
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"strconv"
 
 	"github.com/alpacax/alpacon-cli/client"
@@ -67,12 +68,12 @@ func FetchCursorPages[T any](ac *client.AlpaconClient, endpoint string, params m
 
 		responseBody, err := ac.SendGetRequest(utils.BuildURL(endpoint, "", params))
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("fetching cursor page from %s: %w", endpoint, err)
 		}
 
 		var page CursorListResponse[T]
 		if err = json.Unmarshal(responseBody, &page); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("decoding cursor page from %s: %w", endpoint, err)
 		}
 
 		result = append(result, page.Results...)
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -54,7 +54,7 @@ func FetchCursorPages[T any](ac *client.AlpaconClient, endpoint string, params m
 	// The server caps page_size at 100 (ESCursorPagination.max_page_size).
 	const maxPageSize = 100
 
-	var result []T
+	result := make([]T, 0, min(limit, maxPageSize))
 	cursor := ""
 	for len(result) < limit {
 		params["page_size"] = strconv.Itoa(min(maxPageSize, limit-len(result)))
--- a/api/types.go
+++ b/api/types.go
@@ -8,3 +8,9 @@ type ListResponse[T any] struct {
 	Last     int `json:"last"`
 	Results  []T `json:"results"`
 }
+
+// CursorListResponse decodes the Elasticsearch cursor-pagination contract (base64 next token, not ListResponse's int page).
+type CursorListResponse[T any] struct {
+	Next    string `json:"next"`
+	Results []T    `json:"results"`
+}
--- a/api/audit/audit.go
+++ b/api/audit/audit.go
@@ -1,9 +1,6 @@
 package audit
 
 import (
-	"encoding/json"
-	"fmt"
-
 	"github.com/alpacax/alpacon-cli/api"
 	"github.com/alpacax/alpacon-cli/api/iam"
 	"github.com/alpacax/alpacon-cli/client"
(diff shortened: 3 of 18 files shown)


api/pagination.go
```diff
@@ -41,3 +41,43 @@ func FetchAllPages[T any](ac *client.AlpaconClient, endpoint string, params map[
...
+
+	var result []T
+	cursor := ""
+	for len(result) < limit {
+		params["page_size"] = strconv.Itoa(min(maxPageSize, limit-len(result)))
+		if cursor != "" {
+			params["cursor"] = cursor
+		}
+
+		responseBody, err := ac.SendGetRequest(utils.BuildURL(endpoint, "", params))
```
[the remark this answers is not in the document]
jisung-02: Fixed in 440c79d. The loop now deletes the cursor key when unset, so a caller-supplied (or stale reused) cursor can never leak into the first request. Added TestFetchCursorPages_IgnoresStaleCursorParam covering this. Kept the in-place params mutation itself, matching the existing FetchAllPages convention; all current call sites build a fresh local map.

api/types.go
```diff
@@ -8,3 +8,9 @@ type ListResponse[T any] struct {
 	Last     int `json:"last"`
 	Results  []T `json:"results"`
 }
+
+// CursorListResponse decodes the Elasticsearch cursor-pagination contract (base64 next token, not ListResponse's int page).
+type CursorListResponse[T any] struct {
+	Next    string `json:"next"`
+	Results []T    `json:"results"`
+}
```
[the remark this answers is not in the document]
jisung-02: The current implementation is correct as-is. Per the encoding/json documentation, unmarshaling a JSON null into a Go type other than a pointer, interface, map, or slice has no effect on the value and produces no error - so `"next": null` simply leaves Next as its zero value "", which is exactly the loop's termination condition in FetchCursorPages.

This is verified by TestFetchCursorPages_SinglePageNullNext (api/pagination_test.go), which feeds the raw wire form `{"next":null,...}` through json.Unmarshal and passes. A *string or custom UnmarshalJSON would add complexity without changing behavior.

api/pagination.go
```diff
@@ -41,3 +41,49 @@ func FetchAllPages[T any](ac *client.AlpaconClient, endpoint string, params map[
...
+		params = make(map[string]string)
+	}
+
+	// The server caps page_size at 100 (ESCursorPagination.max_page_size).
+	const maxPageSize = 100
+
+	var result []T
+	cursor := ""
+	for len(result) < limit {
+		params["page_size"] = strconv.Itoa(min(maxPageSize, limit-len(result)))
```
geunwoonoh, line 68: Warning: `FetchCursorPages` mutates the caller's `params` map — it writes `page_size` here and adds/deletes `cursor` below, leaving residue after the function returns. This is a latent bug for any future caller that reuses the map or uses it for another purpose. That `TestFetchCursorPages_IgnoresStaleCursorParam` has to pin the defensive `delete` is itself the signal.

Copying the map on entry removes the footgun and lets the `delete` branch disappear:

```go
merged := make(map[string]string, len(params)+2)
for k, v := range params {
    merged[k] = v
}
delete(merged, "cursor") // then use only `merged` in the loop
```

Note: the existing `FetchAllPages` shares this same mutation pattern, so a full fix touches both — reasonable to defer outside this PR.

api/pagination.go
```diff
@@ -41,3 +41,49 @@ func FetchAllPages[T any](ac *client.AlpaconClient, endpoint string, params map[
...
+		return nil, nil
+	}
+	if params == nil {
+		params = make(map[string]string)
+	}
+
+	// The server caps page_size at 100 (ESCursorPagination.max_page_size).
+	const maxPageSize = 100
+
+	var result []T
```
geunwoonoh: Suggestion: preallocating avoids repeated growth in the common single-page case (`--tail 25`):

```suggestion
	result := make([]T, 0, min(limit, maxPageSize))
```

Optional — negligible over a CLI lifetime.

api/pagination.go
```diff
@@ -41,3 +41,49 @@ func FetchAllPages[T any](ac *client.AlpaconClient, endpoint string, params map[
...
+		if cursor != "" {
+			params["cursor"] = cursor
+		} else {
+			// Drop any caller-supplied cursor so the first request starts from the first page.
+			delete(params, "cursor")
+		}
+
+		responseBody, err := ac.SendGetRequest(utils.BuildURL(endpoint, "", params))
+		if err != nil {
+			return nil, err
```
geunwoonoh: Suggestion: the golang stack rules favor wrapping with context, e.g. `fmt.Errorf("fetching page for %s: %w", endpoint, err)`. That said, `FetchAllPages` returns bare errors too and the cmd layer adds context via `CliErrorWithExit`, so keeping it as-is is defensible for consistency with the existing repo pattern.

geunwoonoh (dismissed):
## Review Summary

**Stack**: golang · **Changes**: +377 −141 in 12 files
**Verification (run locally)**: `go vet ./api/...` clean, `go test -race ./api/...` passes across all packages.

### Summary

This PR fixes the crash where ES cursor pagination (`next` is a base64 token) was unmarshaled into an int page number. It adds a shared `api.CursorListResponse[T]` + `api.FetchCursorPages[T]` helper, migrates the `webftp-log` / `audit` / `log` commands onto it, and folds the previously-working inline cursor loop in `websh.GetSessionRecords` into the same helper — a net 35-line deletion.

### Critical Issues (0)

None. Cursor tokens are encoded through `BuildURL` via `url.Values.Encode()`, so base64 special chars (`=`, `+`) are safe; the loop terminates deterministically on `next == "" || len(Results) == 0`; and the non-positive-limit guard prevents the slice-bounds panic.

### Warnings (2)

**1. `FetchCursorPages` mutates the caller's `params` map** (`api/pagination.go`)
The helper writes `page_size` and adds/deletes `cursor` on the caller-supplied map, leaving residue after it returns. It's a latent bug for any future caller that reuses the map. The fact that `TestFetchCursorPages_IgnoresStaleCursorParam` has to pin the defensive `delete` is itself the signal. Copying the map on entry would also let the `delete` branch disappear. Note the existing `FetchAllPages` shares this pattern, so fixing it means fixing both — reasonable to defer outside this PR. (Inline comment added.)

**2. Inconsistent non-positive `--tail` handling** (`cmd/log/log.go:27`, `cmd/audit/audit.go:40`, `cmd/webftp/webftp.go:40` — not in this diff)
`websh records` errors out on `--limit 0` via `CliErrorWithExit`, but these three commands silently print an empty table on `--tail 0`. The behavior change is documented in the PR body, but from the exit-code-stability angle (CLAUDE.md), failing as a usage error like `websh` would be more consistent for scripts and AI agents.

### Suggestions (3)

**1. Rename `pageSize` → `tail` in the cmd layer** (`cmd/audit/audit.go`, `cmd/webftp/webftp.go`, `cmd/log/log.go` — not in this diff)
The API signatures were renamed `pageSize` → `tail`, but the cmd-layer locals are still `pageSize`. The semantics are now "last N accumulated", so the old name misleads. Pure rename, no risk.

**2. Preallocate `result`** (`api/pagination.go:57`) — optional. (Inline comment added.)

**3. Error wrapping** (`api/pagination.go:69-70`) — golang stack rules favor `fmt.Errorf(...: %w)`, but `FetchAllPages` behaves identically and the cmd layer adds context via `CliErrorWithExit`, so keeping it as-is is defensible. (Inline comment added.)

### Review of requested angles

**Duplicate / unused code**: Clean. The `recordCursorPage` type and the websh inline cursor loop are gone, and the freed `json` / `fmt` / `strconv` imports are all removed (`go vet` clean). I checked all 8 remaining `ListResponse[T]` sites (webhook, note, auth, server, iam, packages, event) — all are page-number DB endpoints, so no un-migrated cursor endpoint remains. `FetchAllPages` and `FetchCursorPages` share a similar loop skeleton, but their wire contracts differ (int page vs cursor token); merging them would cost more in abstraction than the duplication saves. Keeping them separate is the right call. Minor overlap: `TestGetSessionRecords_FollowsCursor`'s page_size assertion overlaps `TestFetchCursorPages_PageSize`, but the PR body states this is an intentional refactor safety net.

**Go conventions**: Followed. `const → var → type → func` ordering, modern idioms (builtin `min`, `for i := range 25`), doc comments starting with the identifier name, errcheck `_ =` pattern, English comments — all consistent with CLAUDE.md and the stack rules.

**Performance**: No regression, net improvement. The default `--tail 25` still issues a single request (`page_size=25`) — pinned by `TestGetSystemLogList_NoExtraPagination` with an atomic counter — and multiple requests only occur for `tail > 100`, which was previously impossible. Cursor chaining is inherently sequential so there's no parallelization to be had; the `result[:limit]` reslice retains the backing array but that's irrelevant over a CLI process lifetime.

### Highlights

- The helper defends against the negative-`tail` slice-bounds panic, the pathological "next token but empty results" infinite loop, and caller-supplied `cursor` leakage.
- Tests reproduce the original crash shape (base64 `next`) as regressions, and pin error paths (mid-pagination 500, malformed JSON) and wire format (filter params reaching the query string).
- The PR body honestly documents the behavior change and the verification limit (not validated against a live workspace).

---

**Verdict**: APPROVE — the 2 warnings either match existing repo conventions or are intentionally documented, so neither blocks merge.

**Checklist**
- [x] Tests included and passing (`go test -race ./api/...` verified locally)
- [x] Documentation updated (behavior change documented in PR body)

jisung-02:
Sequencing update: this PR will be merged first, ahead of #276 (fixes #275).

Both touch `api/webftp/webftp.go` and add `api/webftp/webftp_test.go`, so #276 will need to rebase onto this once it lands. On its side: adopt this PR's `FetchCursorPages` switch in `GetWebFTPLogList`, layer the `--server`/`--user` name-to-ID resolution on top, and merge the two `webftp_test.go` files (cursor/error-path cases here + name-resolution cases from #276).

geunwoonoh (approved):

PR #280, opened 2026-07-23
perf(event): bound gap-fill and recoverSkipped fetches with seq__lte

### LINKED ISSUE #273: perf(event): bound gap-fill and recoverSkipped fetches with seq__lte
## Background

Follow-up to the #272 review. The gap-fill re-fetch in `applyChunk` (`api/event/event.go:416`) and `recoverSkipped` (`api/event/event.go:106`) call `GetCommandChunks(ac, cmdID, fromSeq)`, which downloads everything from the hole to the current end of output and then discards all seqs past what is actually needed. While a hole stays open on a command that keeps producing output, each re-fetch is O(current total tail).

The server does not support `seq__lte` yet (the `chunks` action parses only `seq__gte` manually and silently ignores unknown params), so this could not be fixed in #272 and is split out here.

## Changes (after the server issue lands)

- Add an upper-bound seq parameter to `GetCommandChunks` (`seq__lte`); no bound keeps the current behavior
- Bound the gap-fill re-fetch with `seq__lte=liveSeq-1` and `recoverSkipped` with `seq__lte=skipped[len-1]`
- Compatibility with older servers that ignore the param: client-side filtering already exists, so behavior is identical—only the performance benefit is lost

## Tests

- Add `seq__lte` filtering to the test servers, then assert request params and response ranges

Blocked by: alpacax/alpacon-server#2571
Related: #272 (review Warning + Suggestion 1)

jisung-02:
## Background

While streaming command output, the re-fetches that fill seq holes only set a lower bound (`seq__gte`), so they downloaded everything from the hole to the current end of output and discarded the unneeded tail client-side. As long as a hole stayed open and the command kept producing output, each retry cost O(current tail). The server now supports a `seq__lte` upper-bound filter (alpacax/alpacon-server#2571), so the CLI can narrow these re-fetches to the window it actually needs.

## Changes

Add a `toSeq` upper-bound argument to `getCommandChunks`. The function stays single; unbounded is expressed with a negative sentinel `noSeqBound` (-1). Because seq is 0-indexed, `0` is a legitimate upper bound, so the sentinel must be negative rather than `0`. `seq__lte` is sent only when `toSeq >= 0`; when unbounded the parameter is omitted entirely (the server's strict integer filter rejects an empty `seq__lte`, so an empty value is never sent).

Call sites (5):

| Location | Range | Kind |
|---|---|---|
| `applyChunk` gap-fill | `[lastSeq+1, chunk.Seq-1]` | bounded |
| `recoverSkipped` | `[skipped[0], skipped[last]]` | bounded |
| `GetCommandOutput` | `noSeqBound` | unbounded |
| `streamSubscribed` initial fetch | `noSeqBound` | unbounded |
| `drainRemainingChunks` | `noSeqBound` | unbounded |

The two bounded upper values are already known to the code (`chunk.Seq-1`, and the last element of the ascending `g.skipped`), so no new state is introduced.

## Backward compatibility

If an older server ignores `seq__lte` and returns the full tail, behavior is unchanged. `applyChunk`'s consume loop stops at `c.Seq > chunk.Seq` (client-side clip), so it never consumes chunks past the bound, and `recoverSkipped` only looks up the exact seqs it cares about via the `bySeq` map, so any extra chunks are ignored. This fallback is pinned by a regression test for each call site (`applyChunk` and `recoverSkipped`).

This assumes a server that either supports `seq__lte` (alpacax/alpacon-server#2571 and later) or ignores unknown query parameters — the DRF default. The only unsupported case is a server that rejects unknown parameters with a 400: gap-fill re-fetches would fail repeatedly and fall through to `giveUpGap`, dropping output. No such server version is known. Note also that `recoverSkipped`'s `[skipped[0], skipped[last]]` window is not tight — it can still be wide when skips are far apart — but a single bounded request is strictly better than the previous unbounded fetch.

## Testing

- New tests: `seq__lte` is sent when bounded (including the `0` boundary) and omitted when unbounded; the `applyChunk` and `recoverSkipped` call sites derive the expected bound; output stays identical at both call sites when an old server ignores `seq__lte`.
- Test servers (`holeServer`/`chunkServer`) now honor `seq__lte` via a shared `parseSeqLte` helper.
- `go build ./...` OK, `go vet ./api/event/` OK, `golangci-lint run ./api/event/...` 0 issues, `go test -race ./api/event/` PASS.
- Verified end-to-end against a live server (`web-editor`, ubuntu 24.04) with the freshly built binary: a 300-line burst streamed back contiguous `1..300` with no gaps or duplicates, confirming the bounded and unbounded fetch paths preserve output integrity.

## Checklist

- [x] Build, vet, lint, and race tests pass
- [x] Backward-compatibility regression test added
- [x] Verified on a live server
- [ ] Reviewer approval

Closes #273

### DIFF
--- a/api/event/chunks.go
+++ b/api/event/chunks.go
@@ -10,7 +10,7 @@ import (
 	"github.com/alpacax/alpacon-cli/client"
 )
 
-// noSeqBound is the toSeq sentinel that leaves GetCommandChunks unbounded (seq
+// noSeqBound is the toSeq sentinel that leaves getCommandChunks unbounded (seq
 // is 0-indexed, so 0 is a valid upper bound and cannot mean "no bound").
 const noSeqBound = -1
 
@@ -20,11 +20,11 @@ type Chunk struct {
 	Content string `json:"content"`
 }
 
-// GetCommandChunks fetches chunks for cmdID with seq in [fromSeq, toSeq],
+// getCommandChunks fetches chunks for cmdID with seq in [fromSeq, toSeq],
 // sorted by seq ascending. A negative toSeq (noSeqBound) omits the upper
 // bound. The streaming consumers rely on the order, so we sort defensively
 // in case the server does not honor the ordering param.
-func GetCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq, toSeq int) ([]Chunk, error) {
+func getCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq, toSeq int) ([]Chunk, error) {
 	endpoint := "/api/events/commands/" + url.PathEscape(cmdID) + "/chunks/"
 	params := map[string]string{
 		"seq__gte": strconv.Itoa(fromSeq),
@@ -47,7 +47,7 @@ func GetCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq, toSeq int
 // order. Empty when no chunks were produced. Used by non-streaming paths (exec
 // logs, polling fallback) where Result is empty under the chunk contract.
 func GetCommandOutput(ac *client.AlpaconClient, cmdID string) (string, error) {
-	chunks, err := GetCommandChunks(ac, cmdID, 0, noSeqBound)
+	chunks, err := getCommandChunks(ac, cmdID, 0, noSeqBound)
 	if err != nil {
 		return "", err
 	}
--- a/api/event/chunks.go
+++ b/api/event/chunks.go
@@ -30,9 +30,9 @@ func getCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq, toSeq int
 		"seq__gte": strconv.Itoa(fromSeq),
 		"ordering": "seq",
 	}
+	// Send seq__lte only when bounded; omit rather than send empty, since the
+	// server's strict integer filter rejects an empty seq__lte.
 	if toSeq >= 0 {
-		// Omit rather than send empty when unbounded: the server's strict
-		// integer filter rejects an empty seq__lte.
 		params["seq__lte"] = strconv.Itoa(toSeq)
 	}
 	chunks, err := api.FetchAllPages[Chunk](ac, endpoint, params)
--- a/api/event/chunks.go
+++ b/api/event/chunks.go
@@ -10,20 +10,30 @@ import (
 	"github.com/alpacax/alpacon-cli/client"
 )
 
+// noSeqBound is the toSeq sentinel that leaves GetCommandChunks unbounded (seq
+// is 0-indexed, so 0 is a valid upper bound and cannot mean "no bound").
+const noSeqBound = -1
+
 // Chunk represents a single stdout/stderr chunk produced during command execution.
 type Chunk struct {
 	Seq     int    `json:"seq"`
 	Content string `json:"content"`
 }
 
-// GetCommandChunks fetches chunks for cmdID with seq >= fromSeq, sorted by seq
-// ascending. The streaming consumers rely on that order, so we sort defensively
+// GetCommandChunks fetches chunks for cmdID with seq in [fromSeq, toSeq],
+// sorted by seq ascending. A negative toSeq (noSeqBound) omits the upper
+// bound. The streaming consumers rely on the order, so we sort defensively
 // in case the server does not honor the ordering param.
-func GetCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq int) ([]Chunk, error) {
+func GetCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq, toSeq int) ([]Chunk, error) {
 	endpoint := "/api/events/commands/" + url.PathEscape(cmdID) + "/chunks/"
 	params := map[string]string{
 		"seq__gte": strconv.Itoa(fromSeq),
 		"ordering": "seq",
+	}
+	if toSeq >= 0 {
+		// Omit rather than send empty when unbounded: the server's strict
+		// integer filter rejects an empty seq__lte.
+		params["seq__lte"] = strconv.Itoa(toSeq)
 	}
 	chunks, err := api.FetchAllPages[Chunk](ac, endpoint, params)
 	if err != nil {
@@ -37,7 +47,7 @@ func GetCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq int) ([]Ch
 // order. Empty when no chunks were produced. Used by non-streaming paths (exec
 // logs, polling fallback) where Result is empty under the chunk contract.
 func GetCommandOutput(ac *client.AlpaconClient, cmdID string) (string, error) {
-	chunks, err := GetCommandChunks(ac, cmdID, 0)
+	chunks, err := GetCommandChunks(ac, cmdID, 0, noSeqBound)
 	if err != nil {
 		return "", err
 	}
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -103,7 +103,7 @@ func (g *gapFillState) recoverSkipped(ac *client.AlpaconClient, cmdID string, ou
 	if len(g.skipped) == 0 {
 		return
 	}
-	final, err := GetCommandChunks(ac, cmdID, g.skipped[0], g.skipped[len(g.skipped)-1])
+	final, err := getCommandChunks(ac, cmdID, g.skipped[0], g.skipped[len(g.skipped)-1])
 	if err != nil {
 		utils.CliWarning("failed to fetch skipped chunks (seq %s): %v; output may be incomplete", formatSeqs(g.skipped), err)
 		return
@@ -347,7 +347,7 @@ func streamSubscribed(ac *client.AlpaconClient, session *EventSessionResponse, l
 	// as a duplicate. Chunks past the gap are picked up by applyChunk or the
 	// terminal drain once the gap is filled.
 	lastSeq := -1
-	if existing, err := GetCommandChunks(ac, cmdID, 0, noSeqBound); err == nil {
+	if existing, err := getCommandChunks(ac, cmdID, 0, noSeqBound); err == nil {
 		for _, c := range existing {
 			if c.Seq != lastSeq+1 {
 				break
@@ -413,7 +413,7 @@ func applyChunk(ac *client.AlpaconClient, cmdID string, lastSeq int, chunk Chunk
 			return lastSeq
 		}
 		before := lastSeq
-		missing, err := GetCommandChunks(ac, cmdID, lastSeq+1, chunk.Seq-1)
+		missing, err := getCommandChunks(ac, cmdID, lastSeq+1, chunk.Seq-1)
 		g.lastAttempt = gapFillNow()
 		if err != nil {
 			// Fetch failed: back off (noProgress++ keeps the throttle window
@@ -471,7 +471,7 @@ func chunkContent(chunks []Chunk) map[int]string {
 }
 
 func drainRemainingChunks(ac *client.AlpaconClient, cmdID string, lastSeq int, out io.Writer) int {
-	final, err := GetCommandChunks(ac, cmdID, lastSeq+1, noSeqBound)
+	final, err := getCommandChunks(ac, cmdID, lastSeq+1, noSeqBound)
 	if err != nil {
 		utils.CliWarning("failed to fetch trailing chunks (from seq %d): %v; output may be incomplete",
 			lastSeq+1, err)
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -98,12 +98,12 @@ func (g *gapFillState) giveUpGap(lastSeq, nextSeq int, missing []Chunk, out io.W
 
 // recoverSkipped makes a final fetch for seqs skipped mid-stream (giveUpGap),
 // printing late content out of order and reporting still-absent seqs as lost.
-// g.skipped is ascending, so the fetch starts at its first seq.
+// g.skipped is ascending, so the fetch is bounded to [first seq, last seq].
 func (g *gapFillState) recoverSkipped(ac *client.AlpaconClient, cmdID string, out io.Writer) {
 	if len(g.skipped) == 0 {
 		return
 	}
-	final, err := GetCommandChunks(ac, cmdID, g.skipped[0])
+	final, err := GetCommandChunks(ac, cmdID, g.skipped[0], g.skipped[len(g.skipped)-1])
 	if err != nil {
 		utils.CliWarning("failed to fetch skipped chunks (seq %s): %v; output may be incomplete", formatSeqs(g.skipped), err)
 		return
@@ -347,7 +347,7 @@ func streamSubscribed(ac *client.AlpaconClient, session *EventSessionResponse, l
 	// as a duplicate. Chunks past the gap are picked up by applyChunk or the
 	// terminal drain once the gap is filled.
 	lastSeq := -1
-	if existing, err := GetCommandChunks(ac, cmdID, 0); err == nil {
+	if existing, err := GetCommandChunks(ac, cmdID, 0, noSeqBound); err == nil {
 		for _, c := range existing {
 			if c.Seq != lastSeq+1 {
 				break
@@ -413,7 +413,7 @@ func applyChunk(ac *client.AlpaconClient, cmdID string, lastSeq int, chunk Chunk
 			return lastSeq
 		}
 		before := lastSeq
-		missing, err := GetCommandChunks(ac, cmdID, lastSeq+1)
+		missing, err := GetCommandChunks(ac, cmdID, lastSeq+1, chunk.Seq-1)
 		g.lastAttempt = gapFillNow()
 		if err != nil {
 			// Fetch failed: back off (noProgress++ keeps the throttle window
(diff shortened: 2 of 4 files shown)


api/event/event.go
```diff
@@ -426,6 +426,8 @@ func applyChunk(ac *client.AlpaconClient, cmdID string, lastSeq int, chunk Chunk
 		} else {
 			// Advance only over contiguous seqs, stopping at the first hole, so a
 			// gap-fill racing ahead of persistence can't skip a not-yet-stored seq.
+			// The c.Seq > chunk.Seq clip is a safety net for old servers that
+			// ignore seq__lte and return the tail past the requested bound.
```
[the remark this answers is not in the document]
jisung-02: Good catch. Reworded in 66a521c: the comment now states the `c.Seq > chunk.Seq` clip caps consumption at the live chunk's seq (chunk.Seq) — one past the requested `seq__lte` of chunk.Seq-1 — rather than at the requested bound. This matches the framing already used in the applyChunk old-server test comments. Verified no other occurrence of the phrasing in the branch scope.

api/event/chunks.go
```diff
@@ -10,21 +10,31 @@ import (
...
 }
 
-// GetCommandChunks fetches chunks for cmdID with seq >= fromSeq, sorted by seq
-// ascending. The streaming consumers rely on that order, so we sort defensively
+// GetCommandChunks fetches chunks for cmdID with seq in [fromSeq, toSeq],
+// sorted by seq ascending. A negative toSeq (noSeqBound) omits the upper
+// bound. The streaming consumers rely on the order, so we sort defensively
 // in case the server does not honor the ordering param.
-func GetCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq int) ([]Chunk, error) {
+func GetCommandChunks(ac *client.AlpaconClient, cmdID string, fromSeq, toSeq int) ([]Chunk, error) {
```
geunwoonoh: Warning: `GetCommandChunks` is exported but has no callers outside `api/event`. A repo-wide grep shows only 4 in-package call sites; the only externally used function here is `GetCommandOutput` (`cmd/exec/logs.go:48`).

This PR adds the **unexported** constant `noSeqBound` to that exported signature's contract, so an external package has no way to express "unbounded" without hardcoding `-1`. Unexporting to `getCommandChunks` is the cheaper fix (4 call sites plus tests) and removes the asymmetry entirely.

api/event/chunks.go
```diff
@@ -10,21 +10,31 @@ import (
...
 	endpoint := "/api/events/commands/" + url.PathEscape(cmdID) + "/chunks/"
 	params := map[string]string{
 		"seq__gte": strconv.Itoa(fromSeq),
 		"ordering": "seq",
 	}
+	if toSeq >= 0 {
+		// Omit rather than send empty when unbounded: the server's strict
+		// integer filter rejects an empty seq__lte.
+		params["seq__lte"] = strconv.Itoa(toSeq)
+	}
```
geunwoonoh, line 37: Suggestion: the "Omit rather than send empty" rationale explains the *absent* branch but sits inside the `if toSeq >= 0` branch. Moving it above the `if` makes it describe the whole condition.

```suggestion
	// Send seq__lte only when bounded; omit rather than send empty, since the
	// server's strict integer filter rejects an empty seq__lte.
	if toSeq >= 0 {
		params["seq__lte"] = strconv.Itoa(toSeq)
	}
```

geunwoonoh (commented):
## PR Review: #280 - perf(event): bound gap-fill and recoverSkipped fetches with seq__lte

**Stack**: golang
**Changes**: +161 -20 in 4 files

Local verification: `go build ./...` OK, `go vet ./api/event/` OK, `go test -race ./api/event/` PASS (53.4s). `golangci-lint` is not installed on my machine, so it was not run.

---

### Summary

Gap-fill re-fetches during streaming only had a lower bound (`seq__gte`), so they downloaded everything from the hole to the current end of output and discarded the tail client-side. This PR narrows them to the window actually needed using the server's new `seq__lte` filter. `GetCommandChunks` gains a `toSeq` argument, and "unbounded" is expressed with a negative sentinel (`noSeqBound = -1`).

The core logic is correct. Things I verified directly:

- The gap-fill entry condition in `applyChunk` is `chunk.Seq > lastSeq+1`, which guarantees `chunk.Seq-1 >= lastSeq+1` — the bounds can never invert.
- I checked that `g.skipped` is genuinely ascending. `giveUpGap` appends in ascending order within a call, and `lastSeq` is monotonic across calls, so `g.skipped[len-1]` is indeed the maximum. Neither the budget-exhaustion branch (`len(g.skipped) < maxGapWidth`) nor the wide-gap early return breaks that ordering.
- `api.FetchAllPages` re-sends the params map on every page (`api/pagination.go:25-46`), so `seq__lte` survives pagination. Since `page_size` is pinned at 100, collapsing a long multi-page tail fetch into a single page is the real win here. The performance claim holds.
- `recoverSkipped` runs *after* `drainRemainingChunks`, and its range is entirely below `lastSeq` at that point, so the two fetches are disjoint.

**Critical issues: 0** — no behavioral regressions and no security concerns.

---

### Warnings (2)

**1. `api/event/chunks.go:27` — exported function paired with an unexported sentinel**

`GetCommandChunks` is exported but has **no callers outside the package**. A repo-wide grep shows only 4 in-package call sites; the only externally used function is `GetCommandOutput` (`cmd/exec/logs.go:48`).

This PR adds the **unexported** constant `noSeqBound` to that exported signature's contract. As a result an external package has no way to express "unbounded" without hardcoding `-1`. Unexporting the function to `getCommandChunks` is the cheaper fix (4 call sites plus tests) and removes the asymmetry entirely.

**2. `api/event/event_test.go:1106-1107` — the comment states the opposite of server behavior**

```go
// parseSeqLte reads the optional seq__lte upper bound; an absent or
// unparseable value means "no upper bound" (mirroring the server).
```

Both the PR description and the comment at `chunks.go:34-35` state that the server's strict integer filter **rejects** an empty `seq__lte`. So for an unparseable value the real server returns 400 while this helper silently ignores it — that is the opposite of mirroring. It is functionally harmless because the CLI never sends an invalid value, but the comment asserts a fidelity that does not hold, and a future reader writing a test against it would diverge from the real server. Fix the comment (preferred), or make the helper 400.

---

### Suggestions (7)

**3. `api/event/chunks.go:33-37` — comment placement**

The "Omit rather than send empty when unbounded" rationale explains the *absent* branch but sits inside the `if toSeq >= 0` branch. Moving it above the `if` makes it describe the whole condition.

**4. `api/event/event_test.go:1473` — stderr noise**

`TestRecoverSkippedChunks_SendsSeqLteBound` is the only one of these tests not wrapped in `captureStderr`, so the warning leaks into test output. Confirmed by running it:

```
=== RUN   TestRecoverSkippedChunks_SendsSeqLteBound
Warning: chunk seq(s) [1 4] never arrived; output may be incomplete
--- PASS
```

Every other `TestRecoverSkippedChunks_*` in this file (lines 1287, 1303) uses `captureStderr`.

**5. `api/event/event_test.go:1446-1491` — duplication and a loose assertion**

The two bound-assertion tests duplicate an identical `seq__lte`-capturing httptest handler; a small `lteCapturingServer(t, results)` helper would collapse it.

More importantly, `TestApplyChunk_SendsSeqLteBound` asserts only `seq__lte` and never looks at `seq__gte`. Since the handler returns chunks 1 and 2 regardless of the query, **the test would still pass with a wrong `fromSeq`**. Assert both bounds.

Also, `atomic.Pointer[string]` plus `require.NotNil` plus a dereference is four lines for capturing a single request, while the neighboring `chunks_test.go` does the same job with a plain `var capturedQuery string` and passes under `-race`. Worth picking one style. And `lastLte` implies repeated requests when there is exactly one — `capturedLte` reads truer.

**6. `api/event/event_test.go:1362` — out-of-scope change**

`for i := 0; i < 4; i++` → `for range 4` is a valid modernization but unrelated to a perf PR. Unless this is part of a repo-wide sweep, a separate commit would keep the diff clean.

**7. Coverage asymmetry**

The old-server fallback is pinned by a regression test for `applyChunk` but not for `recoverSkipped`. The PR body only argues in prose that `recoverSkipped` is safe because it does `bySeq` map lookups. A five-line test with a server that ignores `seq__lte` would make the coverage symmetric.

**8. `api/event/chunks.go:42` — `sort.Slice` (out of scope, performance)**

Not a line this PR touches, but noting it since performance was in scope. `sort.Slice` uses reflect-based swaps; with go.mod at 1.25.12, `slices.SortFunc(chunks, func(a, b Chunk) int { return cmp.Compare(a.Seq, b.Seq) })` is the idiomatic and faster form. Separate PR material.

**9. Low-risk compatibility note**

If any deployed server version rejects unknown query parameters with a 400 instead of ignoring them, gap-fill would repeatedly hit the `failed to fetch missing chunks` path and eventually fall through to `giveUpGap` — real output loss. DRF ignores unknown params by default, so this is very unlikely. A one-line note about the minimum server version in the PR body is probably enough; no code change needed.

Separately, `recoverSkipped`'s range `[skipped[0], skipped[last]]` can still be wide when skips are far apart (say seq 5 and seq 50000). It is still strictly better than the previous unbounded fetch, and a single request looks like the right trade against N per-run requests. No action — just recording that the bound is not tight.

---

### Highlights

- **The choice of `-1` as the sentinel is correctly reasoned.** Because seq is 0-indexed, `0` is a legitimate upper bound and cannot mean "no bound". That is pinned in three places: the code comment, the PR description, and the `"zero is a valid bound"` sub-test.
- **The backward-compatibility argument is actually pinned down, not hand-waved.** The `c.Seq > chunk.Seq` clip is deliberately kept, documented as to why, and guarded by `TestApplyChunk_OldServerIgnoringSeqLte_OutputIdentical`. It becomes unreachable against a compliant server, and keeping it is the right call.
- **No new state is introduced.** Both upper bounds derive from values already in hand (`chunk.Seq-1`, `g.skipped[last]`).
- **The test servers (`holeServer`/`chunkServer`) were updated to honor the new filter**, so the existing gap-fill suite actually exercises the bounded path instead of passing by accident on the old unbounded behavior.

---

**Verdict**: **COMMENTED** — nothing blocks the merge. Clearing the two warnings (the exported/unexported asymmetry and the inaccurate comment) would be enough to ship as is.

**Checklist**:
- [x] Code follows stack conventions (except warning #1)
- [x] Tests included and passing (verified with `go test -race`)
- [x] No security vulnerabilities (query built only via `strconv.Itoa` — no injection surface)
- [x] Documentation updated if needed (except the one comment in warning #2)

api/event/event_test.go
```diff
@@ -1103,16 +1103,31 @@ func captureStderr(t *testing.T, fn func()) string {
 	return buf.String()
 }
 
-// holeServer serves /chunks/?seq__gte=N returning every seq >= N except those
-// in `hole`. fetches counts the chunk requests. Used to simulate a seq that is
-// never persisted server-side.
+// parseSeqLte reads the optional seq__lte upper bound; an absent or
+// unparseable value means "no upper bound" (mirroring the server).
```
geunwoonoh: Warning: this comment states the opposite of the server's actual behavior. Both the PR description and the comment at `chunks.go:34-35` say the server's strict integer filter **rejects** an empty `seq__lte`. So for an unparseable value the real server returns 400 while this helper silently ignores it — that is not mirroring.

Functionally harmless (the CLI never sends an invalid value), but the comment asserts a fidelity that does not hold, and a future reader writing a test against it would diverge from the real server.

```suggestion
// parseSeqLte reads the optional seq__lte upper bound. An absent value means
// "no upper bound". An unparseable value is ignored here too, unlike the real
// server, whose strict integer filter rejects it — the CLI never sends one.
```

api/event/event_test.go
```diff
@@ -1423,3 +1440,78 @@ func TestDrainRemainingChunks_NoWarnWhenContiguous(t *testing.T) {
 	assert.Equal(t, "c3\nc4\nc5\nc6\n", out.String())
 	assert.NotContains(t, stderr, "never arrived")
 }
+
+// TestApplyChunk_SendsSeqLteBound verifies the gap-fill re-fetch is bounded by
+// the live chunk that exposed the hole (seq__lte == chunk.Seq-1).
+func TestApplyChunk_SendsSeqLteBound(t *testing.T) {
```
geunwoonoh: Suggestion: three small things on this test.

1. **Loose assertion.** It asserts only `seq__lte` and never looks at `seq__gte`. Since the handler returns chunks 1 and 2 regardless of the query, this test would still pass with a wrong `fromSeq`. Assert both bounds.
2. **Duplication.** This handler is identical to the one in `TestRecoverSkippedChunks_SendsSeqLteBound`; a small `lteCapturingServer(t, results)` helper would collapse both.
3. **Capture style.** `atomic.Pointer[string]` + `require.NotNil` + a dereference is four lines for capturing a single request, while the neighboring `chunks_test.go` does the same job with a plain `var capturedQuery string` and passes under `-race`. Also `lastLte` implies repeated requests when there is exactly one — `capturedLte` reads truer.

api/event/event_test.go
```diff
@@ -1423,3 +1440,78 @@ func TestDrainRemainingChunks_NoWarnWhenContiguous(t *testing.T) {
...
+	_ = applyChunk(ac, "cmd", 0, ChunkEvent{Seq: 3, Content: "c3\n"}, out, g)
+
+	got := lastLte.Load()
+	require.NotNil(t, got)
+	assert.Equal(t, "2", *got)
+}
+
+// TestRecoverSkippedChunks_SendsSeqLteBound verifies the final recovery fetch is
+// bounded by the last skipped seq (g.skipped is ascending).
+func TestRecoverSkippedChunks_SendsSeqLteBound(t *testing.T) {
```
geunwoonoh, line 1480: Suggestion: this is the only one of these tests not wrapped in `captureStderr`, so the warning leaks into test output. Confirmed by running it:

```
=== RUN   TestRecoverSkippedChunks_SendsSeqLteBound
Warning: chunk seq(s) [1 4] never arrived; output may be incomplete
--- PASS
```

Every other `TestRecoverSkippedChunks_*` in this file (lines 1287, 1303) uses `captureStderr`.

api/event/event_test.go
```diff
@@ -1423,3 +1440,78 @@ func TestDrainRemainingChunks_NoWarnWhenContiguous(t *testing.T) {
...
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+
+	g := &gapFillState{}
+	out := &bytes.Buffer{}
+	// Gap [1,2] exposed by live chunk seq=3; server also returns 4,5 (ignored).
+	lastSeq := applyChunk(ac, "cmd", 0, ChunkEvent{Seq: 3, Content: "c3\n"}, out, g)
+
+	assert.Equal(t, 3, lastSeq, "advances only through the live chunk, not the extra tail")
+	assert.Equal(t, "c1\nc2\nc3\n", out.String(), "extra tail beyond the bound is not emitted")
+}
```
geunwoonoh, line 1518: Suggestion: coverage asymmetry. The old-server fallback is pinned by a regression test for `applyChunk` (right above) but not for `recoverSkipped`. The PR body only argues in prose that `recoverSkipped` is safe because it does `bySeq` map lookups. A five-line test with a server that ignores `seq__lte` and returns extra chunks would make the coverage symmetric.

api/event/event_test.go
```diff
@@ -1342,7 +1359,7 @@ func TestGiveUpGap_BoundsCumulativeSkipped(t *testing.T) {
 	out := &bytes.Buffer{}
 	last := 0
 	stderr := captureStderr(t, func() {
-		for i := 0; i < 4; i++ {
+		for range 4 {
```
geunwoonoh: Suggestion: `for i := 0; i < 4; i++` → `for range 4` is a valid modernization but unrelated to a perf PR. Unless this is part of a repo-wide sweep, a separate commit would keep this diff focused.

api/event/chunks.go
jisung-02: Done in 741ae7e — unexported to `getCommandChunks`. A repo-wide grep confirms no callers outside `api/event` (4 in-package sites plus tests; `GetCommandOutput` stays the only exported entry), and `go build ./...` passes, so the asymmetry with the unexported `noSeqBound` sentinel is gone.

api/event/chunks.go
jisung-02: Done in 68409f5 — moved the "omit rather than send empty" rationale above the `if toSeq >= 0` so it describes the whole send-only-when-bounded condition.

api/event/event_test.go
jisung-02: Done in 94d75dc — reworded so the comment states the helper ignores an unparseable value **unlike** the real server (whose strict integer filter rejects it with 400), and that the CLI never sends one. It no longer claims to mirror the server.

api/event/event_test.go
jisung-02: Done in 797b00a — all three: (1) the `applyChunk` test now asserts both `seq__gte` and `seq__lte`, so a wrong `fromSeq` would fail; (2) extracted the shared `lteCapturingServer(t, results)` helper used by both bound tests; (3) dropped `atomic.Pointer[string]`/`require.NotNil` for plain vars read after the fetch via a `bounds()` accessor, and removed the `lastLte` name.

api/event/event_test.go
jisung-02: Done in 797b00a — wrapped the `recoverSkipped` call in `captureStderr` so the "never arrived" warning no longer leaks into test output.

api/event/event_test.go
jisung-02: Done in 797b00a — added `TestRecoverSkippedChunks_OldServerIgnoringSeqLte_OutputIdentical`: a server that ignores `seq__lte` and returns `1..6` while `g.skipped` is `{1,4}`, asserting only `c1`/`c4` are emitted. The old-server fallback is now pinned for `recoverSkipped` just like `applyChunk`.

api/event/event_test.go
jisung-02: Reverted in 797b00a — restored `for i := 0; i < 4; i++` to keep this perf PR's diff focused. The `for range 4` modernization can land in a separate repo-wide sweep.

geunwoonoh (approved):

Issue #271, opened 2026-07-20
feat(event): add generic event watch/wait commands for push-based agent notification


junho226:
## Summary

Add generic `alpacon event watch` (stream events to stdout) and `alpacon event wait` (block until a matching event, then exit) commands so external tools and AI agents can receive server-side events — e.g. work session approval — as push instead of polling.

## Problem Statement

When a work session requires admin approval, the requester has no push channel to learn the decision:

- `alpacon event` (`cmd/event/event.go`) is a one-shot REST list (`--tail`), not a stream.
- `alpacon exec --wait` polls the job over REST on a ticker (`cmd/exec/run.go`).
- The only WS event consumer is the sudo MFA listener (`api/event/sudolistener.go`), hardwired to `event_type: "sudo"` inside websh (`cmd/websh/websh.go`).

Meanwhile the server already publishes approval transitions to the event channel: approving a work session emits `work_session` events (`approved`, `activated`, ...) via `publish_work_session_status_change` → `publish_event_to_subscribers`, and subscribing is permitted to anyone passing `user_can_view_work_session` (session assignee included). The infrastructure exists end to end; the CLI just never exposes it generically.

Primary use case — agent integration: a coding agent (Claude Code, Codex, Gemini CLI, plain scripts) starts a work session and needs to be woken when approval lands. Agents with background-task support attach `event watch` to their process monitor (one stdout line = one notification); any other agent runs `event wait` as a blocking foreground command.

## Proposed Solution

Reuse existing building blocks (`CreateEventSession` / subscription API in `api/event/subscribe.go`, reconnect/backoff pattern from `SudoListener`):

1. Generalize the subscription helper to accept any `event_type` (the request struct already carries the field; only the helper hardcodes `sudo`).
2. Extract a generic WS listener from `sudolistener.go` that prints events as NDJSON lines to stdout instead of driving the MFA flow.
3. New commands:
   - `alpacon event watch --type work_session --target <id>` — stream matching events until interrupted.
   - `alpacon event wait --type work_session --target <id> [--status approved] [--timeout 300]` — subscribe first, then one REST catch-up check (avoids the subscribe/publish race), print the first matching event and exit; distinct exit codes for timeout vs. rejection (align with #253).
4. Keep output minimal and structured (event type, target, status, timestamp) — lines land in AI-agent contexts, so no free-form text passthrough.

No server changes required.

## Priority

medium

## Related

#253, #254 (closed), #158

[...]

PR #287, opened 2026-07-28
refactor(event): collapse duplicate WebSocket listener skeletons

jisung-02:
## Background

`SudoListener` (`api/event/sudolistener.go`) and `CommandOutputListener` (`api/event/commandoutput_listener.go`) each carried their own copy of nearly the same code. `listenLoop`, `connectAndListen`, `WaitConnected`, `Stop`, the lifecycle fields, and even the 1s/30s backoff constants existed in both files. The only thing genuinely different between them was frame handling.

Adding the generic `event watch` listener that #271 calls for would have made that loop a third copy. This collapses the skeleton first. It is PR1 of the #271 plan and does not close the issue on its own.

## What changed

A shared `wsListener` in `api/event/wslistener.go`, embedded by both listeners. The skeleton owns dialing, exponential-backoff reconnect, and shutdown; each listener keeps only its own frame handling.

| File | Lines |
|---|---|
| `api/event/sudolistener.go` | 313 to 189 |
| `api/event/commandoutput_listener.go` | 200 to 90 |
| `api/event/wslistener.go` (new) | 160 |

Two signature divergences reconciled:

`SudoListener.connectAndListen` returned `(bool, error)` while the command-output one returned `bool`. The sudo error was nil only when `done` had been closed, which `listenLoop` rechecks independently, and it never reached a user or a log. The shared version returns `bool`.

The dial handshake timeouts, 10s for sudo and 5s for command output, are preserved per-listener through the `handshakeTimeout` field. `commandOutputConnectTimeout` keeps its name because `api/event/event.go:301` and `:328` also use it as the `WaitConnected` budget.

The `stopped` channel now exists on both listeners. It is the goroutine-termination signal; without it `TestSudoListener_StopClosesConnection` degrades to polling for `conn == nil`, which cannot distinguish termination from the gap between reconnect attempts.

## Dependencies

No change to the package graph. The `api/event` to `client` edge already existed via `event.go`, `subscribe.go`, `chunks.go` and others, and `client` imports no `api/*` package, so there is no cycle. `newWSListener` takes a `*client.AlpaconClient` rather than an `http.Header` so the nil-client branch lives in one place instead of being duplicated across both constructors.

## Safety

The goal was zero behavior change, and the existing listener tests pass with their assertions untouched. Only the construction sites changed: moving fields into the embedded struct broke the struct literals, so those were mechanically replaced with `NewSudoListener` / `NewCommandOutputListener` calls. Access to `sl.mu`, `sl.conn`, and `sl.stopped` still works through field promotion.

Two failure modes that the refactor made shared, and therefore worth failing loudly on, are now guarded:

`Start` panics with `event: wsListener.handleFrame must be assigned before Start` when a constructor forgets the assignment. Previously the nil call sat in `connectAndListen`, so it only fired after a successful dial, on the read goroutine, far from the cause. `Start` has no error return and all three call sites (`cmd/websh/websh.go:350`, `api/event/event.go:300` and `:327`) call it bare, so a panic on the caller's stack is the available shape. This is a programmer error on an unexported type, not a user-facing failure.

`handleSudoMFA` returns early when `ac` is nil. `NewSudoListener` documents nil `ac` as a test-only mode, but nothing enforced it, and an `mfa_request` frame would have nil-dereferenced inside the spawned goroutine. The guard sits at the top of `handleSudoMFA` rather than at the spawn site so it is synchronously testable and covers every deref below it. The nil-`ac` test mode is load-bearing rather than incidental: `&client.AlpaconClient{}` makes `SetWebsocketHeader()` emit an empty `Origin`, which gorilla's `checkSameOrigin` rejects, while a nil client yields a header with no `Origin` at all.

`wsListener.mu` guards `conn` and nothing else. `CommandOutputListener` carries its own `cmdMu` for `commandID` rather than reusing the skeleton lock, so the invariant stays readable from one file.

`newWSListener` returns `*wsListener` and both listeners embed the pointer. Returning the struct by value copied a type holding a `sync.Mutex` and two `sync.Once`. It was safe in practice because the value is pristine at that point, but `go vet` could not confirm it — copylocks skips call expressions, since `lockPathRhs` returns nil for an `*ast.CallExpr`. Returning a pointer removes the value-copy shape from the type instead of relying on a reader to know it is fresh.

## Tests

Tests that never dial — `WaitConnected` success, timeout and shutdown, plus a `StopIsIdempotent` that existed identically in both listener files — moved to `api/event/wslistener_test.go`. They hand-close `connected` and `done`, so they asserted nothing about either concrete listener. Tests that do dial stayed where they were: they double as proof that each embedder is wired correctly.

Three skeleton blocks were unreachable from the existing tests and are now covered. Each new test was mutation-checked: inverting the logic it guards makes it fail.

| Uncovered block | Test |
|---|---|
| backoff cap | `TestWSListener_NextReconnectDelay` |
| dial-failure return | `TestWSListener_ConnectAndListen_ReturnsFalseOnFailedHandshake` |
| `listenLoop` entry check for a closed `done` | `TestWSListener_ListenLoop_DoesNotDialWhenAlreadyStopped` |

The cap needed the doubling extracted to `nextReconnectDelay`, because reaching it through `listenLoop` costs 1+2+4+8+16 seconds of real waiting. The extraction also pins the operation order, which the loop never asserted — clamping before doubling would silently stretch the wait to twice the ceiling, and the mutation run confirms the test catches exactly that.

Coverage of `api/event/wslistener.go`: `listenLoop` 81.8 to 100 percent, `connectAndListen` 90.5 to 95.2 percent, everything else at 100 percent. One block stays uncovered, the read loop's `case <-w.done` at `api/event/wslistener.go:148`. `Stop` closes `conn`, so `ReadMessage` always wakes with an error first and that arm is not deterministically reachable.

```
go build ./... && go vet ./...
go test -race ./...
golangci-lint run ./...
gofmt -l ./api/ ./cmd/
```

All packages pass, lint reports 0 issues, gofmt reports no differences.

## Checklist

- [x] Existing listener tests pass with assertions unchanged
- [x] No import cycle introduced
- [x] New tests mutation-checked against the logic they guard
- [ ] Reviewer confirms behavior parity by comparing pre- and post-refactor listener code

Refs #271

### DIFF
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -48,7 +48,7 @@ type sudoMFAEvent struct {
 // http.Client is concurrency-safe. Token refresh and grant verification are
 // serialized by mfaMu so only one MFA flow runs at a time.
 type SudoListener struct {
-	wsListener
+	*wsListener
 	ac         *client.AlpaconClient
 	serverName string
 	mfaMu      sync.Mutex // serializes handleSudoMFA so only one MFA flow runs at a time
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -55,7 +55,7 @@ type SudoListener struct {
 }
 
 // NewSudoListener creates a SudoListener but does not connect yet. ac may be
-// nil only in tests that never reach the MFA flow, which dereferences it.
+// nil in tests; MFA request frames are then dropped instead of dereferencing it.
 func NewSudoListener(ac *client.AlpaconClient, wsURL, serverName string) *SudoListener {
 	sl := &SudoListener{
 		wsListener: newWSListener(ac, wsURL, sudoHandshakeTimeout),
@@ -84,6 +84,11 @@ func (sl *SudoListener) handleMessage(message []byte) {
 }
 
 func (sl *SudoListener) handleSudoMFA(event sudoMFAEvent) {
+	// Every step below dereferences ac, which is nil only in tests.
+	if sl.ac == nil {
+		return
+	}
+
 	// Check if shutdown was requested before acquiring the lock or doing side effects
 	select {
 	case <-sl.done:
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -3,7 +3,6 @@ package event
 import (
 	"encoding/json"
 	"fmt"
-	"net/http"
 	"net/url"
 	"os"
 	"sync"
@@ -12,7 +11,6 @@ import (
 	"github.com/alpacax/alpacon-cli/api/mfa"
 	"github.com/alpacax/alpacon-cli/client"
 	"github.com/alpacax/alpacon-cli/utils"
-	"github.com/gorilla/websocket"
 )
 
 const (
@@ -28,11 +26,7 @@ const (
 	// extra buffer over that so a slow browser MFA does not race the expiry.
 	mfaPollingTimeout = 60 * time.Second
 
-	// reconnectBaseDelay is the initial delay for WebSocket reconnection.
-	reconnectBaseDelay = 1 * time.Second
-
-	// reconnectMaxDelay is the maximum delay between reconnection attempts.
-	reconnectMaxDelay = 30 * time.Second
+	sudoHandshakeTimeout = 10 * time.Second
 )
 
 // sudoMFAEvent represents the MFA request payload from the event WebSocket.
@@ -54,145 +48,22 @@ type sudoMFAEvent struct {
 // http.Client is concurrency-safe. Token refresh and grant verification are
 // serialized by mfaMu so only one MFA flow runs at a time.
 type SudoListener struct {
-	ac          *client.AlpaconClient
-	serverName  string
-	wsURL       string
-	wsHeader    http.Header
-	done        chan struct{}
-	stopped     chan struct{} // closed when listenLoop exits
-	connected   chan struct{} // closed after first successful WebSocket connection
-	connectOnce sync.Once
-	closeOnce   sync.Once
-	mu          sync.Mutex
-	conn        *websocket.Conn
-	mfaMu       sync.Mutex // serializes handleSudoMFA so only one MFA flow runs at a time
+	wsListener
+	ac         *client.AlpaconClient
+	serverName string
+	mfaMu      sync.Mutex // serializes handleSudoMFA so only one MFA flow runs at a time
 }
 
-// NewSudoListener creates a SudoListener but does not connect yet.
+// NewSudoListener creates a SudoListener but does not connect yet. ac may be
+// nil only in tests that never reach the MFA flow, which dereferences it.
 func NewSudoListener(ac *client.AlpaconClient, wsURL, serverName string) *SudoListener {
-	return &SudoListener{
+	sl := &SudoListener{
+		wsListener: newWSListener(ac, wsURL, sudoHandshakeTimeout),
 		ac:         ac,
 		serverName: serverName,
-		wsURL:      wsURL,
-		wsHeader:   ac.SetWebsocketHeader(),
-		done:       make(chan struct{}),
-		stopped:    make(chan struct{}),
-		connected:  make(chan struct{}),
 	}
-}
-
-// Start begins listening for sudo MFA events in a background goroutine.
-// It automatically reconnects on disconnection. Call Stop() to shut down.
-func (sl *SudoListener) Start() {
-	go func() {
-		defer close(sl.stopped)
-		sl.listenLoop()
-	}()
-}
-
-// WaitConnected blocks until the WebSocket connection is established or the
-// timeout expires. Returns true if connected, false on timeout or shutdown.
-func (sl *SudoListener) WaitConnected(timeout time.Duration) bool {
-	select {
-	case <-sl.connected:
-		return true
-	case <-sl.done:
-		return false
-	case <-time.After(timeout):
-		return false
-	}
-}
-
-// Stop signals the listener to shut down and closes the WebSocket connection
-// to unblock any pending ReadMessage call.
-func (sl *SudoListener) Stop() {
-	sl.closeOnce.Do(func() {
-		close(sl.done)
-		sl.mu.Lock()
-		if sl.conn != nil {
-			_ = sl.conn.Close()
-		}
-		sl.mu.Unlock()
-	})
-}
-
-func (sl *SudoListener) listenLoop() {
-	delay := reconnectBaseDelay
-
-	for {
-		select {
-		case <-sl.done:
-			return
-		default:
-		}
-
-		connected, err := sl.connectAndListen()
-		if err == nil {
-			return
-		}
-
-		// Reset backoff if we had a successful connection that later dropped
-		if connected {
-			delay = reconnectBaseDelay
-		}
-
-		select {
-		case <-sl.done:
-			return
-		case <-time.After(delay):
-			delay *= 2
-			if delay > reconnectMaxDelay {
-				delay = reconnectMaxDelay
-			}
-		}
-	}
-}
-
-// connectAndListen dials the event WebSocket and reads messages until
-// the connection drops or Stop() is called. Returns (true, err) if the
-// connection was established, (false, err) if the dial itself failed.
-func (sl *SudoListener) connectAndListen() (connected bool, err error) {
-	dialer := websocket.Dialer{
-		HandshakeTimeout: 10 * time.Second,
-	}
-	conn, _, dialErr := dialer.Dial(sl.wsURL, sl.wsHeader)
-	if dialErr != nil {
-		return false, fmt.Errorf("event websocket connection failed: %w", dialErr)
-	}
-
-	sl.mu.Lock()
-	sl.conn = conn
-	sl.mu.Unlock()
-
-	// Signal that we have successfully connected (first time only)
-	sl.connectOnce.Do(func() { close(sl.connected) })
-
-	defer func() {
-		sl.mu.Lock()
-		sl.conn = nil
-		sl.mu.Unlock()
-		_ = conn.Close()
-	}()
-
-	for {
-		select {
-		case <-sl.done:
-			return true, nil
-		default:
-		}
-
-		_, message, readErr := conn.ReadMessage()
-		if readErr != nil {
-			select {
-			case <-sl.done:
-				return true, nil
-			default:
-			}
-			return true, fmt.Errorf("event websocket read error: %w", readErr)
-		}
-
-		sl.handleMessage(message)
-	}
+	sl.handleFrame = sl.handleMessage
+	return sl
 }
 
 func (sl *SudoListener) handleMessage(message []byte) {
--- a/api/event/sudolistener_test.go
+++ b/api/event/sudolistener_test.go
@@ -69,6 +69,17 @@ func TestSudoListener_HandleMessage_IgnoresNonMFA(t *testing.T) {
 	}
 }
 
+func TestSudoListener_HandleSudoMFA_DropsRequestWhenClientIsNil(t *testing.T) {
+	sl := NewSudoListener(nil, "", "")
+
+	var event sudoMFAEvent
+	event.Payload.Type = "auth"
+	event.Payload.Query = "mfa_request"
+	event.Payload.SudoGrantID = "test-grant-id"
+
+	assert.NotPanics(t, func() { sl.handleSudoMFA(event) })
+}
+
 func TestSudoListener_StopIsIdempotent(t *testing.T) {
 	sl := NewSudoListener(nil, "", "")
 
--- a/api/event/sudolistener_test.go
+++ b/api/event/sudolistener_test.go
@@ -145,9 +145,10 @@ func TestSudoListener_ConnectAndListen_ExitsOnDisconnect(t *testing.T) {
 
 	sl := NewSudoListener(nil, wsURL, "")
 
-	// Run connectAndListen and signal when the read loop processes a message
+	// Local signal, so the test does not close the skeleton's stopped channel
+	readDone := make(chan struct{})
 	go func() {
-		defer close(sl.stopped)
+		defer close(readDone)
 		_ = sl.connectAndListen()
 	}()
 
@@ -161,7 +162,7 @@ func TestSudoListener_ConnectAndListen_ExitsOnDisconnect(t *testing.T) {
 	close(clientRead) // server closes → client read loop exits
 
 	select {
-	case <-sl.stopped:
+	case <-readDone:
 		// connectAndListen returned cleanly after server disconnect
 	case <-time.After(2 * time.Second):
 		t.Fatal("timeout waiting for connectAndListen to return")
--- a/api/event/sudolistener_test.go
+++ b/api/event/sudolistener_test.go
@@ -60,11 +60,7 @@ func TestSudoListener_HandleMessage_IgnoresNonMFA(t *testing.T) {
 		},
 	}
 
-	sl := &SudoListener{
-		done:      make(chan struct{}),
-		stopped:   make(chan struct{}),
-		connected: make(chan struct{}),
-	}
+	sl := NewSudoListener(nil, "", "")
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -74,11 +70,7 @@ func TestSudoListener_HandleMessage_IgnoresNonMFA(t *testing.T) {
 }
 
 func TestSudoListener_StopIsIdempotent(t *testing.T) {
-	sl := &SudoListener{
-		done:      make(chan struct{}),
-		stopped:   make(chan struct{}),
-		connected: make(chan struct{}),
-	}
+	sl := NewSudoListener(nil, "", "")
 
 	sl.Stop()
 	sl.Stop()
@@ -102,13 +94,7 @@ func TestSudoListener_StopClosesConnection(t *testing.T) {
 	defer server.Close()
 
 	wsURL := "ws" + strings.TrimPrefix(server.URL, "http")
-	sl := &SudoListener{
-		wsURL:     wsURL,
-		wsHeader:  http.Header{},
-		done:      make(chan struct{}),
-		stopped:   make(chan struct{}),
-		connected: make(chan struct{}),
-	}
+	sl := NewSudoListener(nil, wsURL, "")
 	sl.Start()
 
 	// Wait for connection to establish by polling sl.conn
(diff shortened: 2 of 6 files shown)


api/event/wslistener.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// Start reads events in a background goroutine, reconnecting automatically
+// until Stop is called.
+func (w *wsListener) Start() {
+	go func() {
+		defer close(w.stopped)
+		w.listenLoop()
+	}()
+}
```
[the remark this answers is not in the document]
jisung-02: Fixed in 3753333. `Start` now panics with `event: wsListener.handleFrame must be assigned before Start` before spawning the goroutine.

Agreed on the diagnosis, and the delayed failure is the worse half of it: the nil call sits in `connectAndListen`, so it only fires after a successful dial, on the read goroutine, far from the constructor that forgot the assignment. Checking at `Start` puts it on the caller's stack, synchronously.

Panic rather than an error return because `Start` has no error return and three call sites (`cmd/websh/websh.go:350`, `api/event/event.go:300` and `:327`) call it bare. This is a programmer error on an unexported type, not a user-facing failure, so failing loudly is the right shape.

Branch-scope sweep: `handleFrame` is the only func field introduced here, and both constructors assign it, so there was no second occurrence to fix. Covered by `TestWSListener_StartPanicsWithoutHandleFrame`.

api/event/sudolistener.go
```diff
@@ -54,145 +48,22 @@ type sudoMFAEvent struct {
...
 
-// NewSudoListener creates a SudoListener but does not connect yet.
+// NewSudoListener creates a SudoListener but does not connect yet. ac may be
+// nil only in tests that never reach the MFA flow, which dereferences it.
 func NewSudoListener(ac *client.AlpaconClient, wsURL, serverName string) *SudoListener {
-	return &SudoListener{
+	sl := &SudoListener{
+		wsListener: newWSListener(ac, wsURL, sudoHandshakeTimeout),
 		ac:         ac,
 		serverName: serverName,
```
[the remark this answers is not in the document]
jisung-02: Fixed in 3753333. `handleSudoMFA` now returns early when `ac` is nil.

You are right that the doc comment stated the contract without enforcing it. The guard went at the top of `handleSudoMFA` rather than in `handleMessage`, because the deref happens on the goroutine `handleMessage` spawns—guarding at the spawn site would leave the check untestable synchronously, since a regression would panic on the goroutine and take the test binary down instead of failing an assertion. At the top of `handleSudoMFA` it also covers a future direct caller, and one early return covers every deref below it (`RefreshToken`, `GetMFALinkForSudo`, `CheckMFACompletion` in `pollMFACompletion`, `SendPostRequest` in `verifySudoGrant`).

Branch-scope sweep for the same class: `CommandOutputListener` stores no client and only hands `ac` to `newWSListener`, which already nil-checks it, so `SudoListener` was the only occurrence.

The nil-`ac` test mode is load-bearing rather than incidental: `&client.AlpaconClient{}` would make `SetWebsocketHeader()` emit an empty `Origin`, which gorilla's `checkSameOrigin` rejects, while a nil client yields a header with no `Origin` at all. Covered by `TestSudoListener_HandleSudoMFA_DropsRequestWhenClientIsNil`, and the constructor comment now describes the enforced behavior.

api/event/wslistener.go
```diff
@@ -0,0 +1,148 @@
...
+func (w *wsListener) WaitConnected(timeout time.Duration) bool {
+	select {
+	case <-w.connected:
+		return true
+	case <-w.done:
+		return false
+	case <-time.After(timeout):
+		return false
+	}
+}
```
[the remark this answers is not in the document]
jisung-02: Not changing this one—the lingering-timer premise no longer holds on this module.

Go 1.23 changed timer reclamation: an unfired `Timer` becomes eligible for garbage collection as soon as the program stops referring to it, without `Stop` being called. The behavior is gated on the main module's `go` directive, and `go.mod` declares `go 1.25.12`, so the new semantics apply. The timer `time.After` allocates when `WaitConnected` returns early via `connected` or `done` is therefore collected, not held until it fires.

Call frequency also does not push it the other way. `WaitConnected` runs at most twice per command execution (`api/event/event.go:301` and `:328`) and once per websh session, never in a loop.

Branch-scope sweep, since the same pattern would need the same treatment: `time.After` also appears in `listenLoop` at `api/event/wslistener.go:103` (reconnect backoff) and predates this branch in `pollMFACompletion` at `api/event/sudolistener.go:151`. Both are covered by the same Go 1.23 reclamation, so all three stay as they are rather than one being rewritten for consistency. Where a timer genuinely outlives a single select—`pollCommandExecution` at `api/event/event.go:248`, which reuses it across loop iterations—the codebase already uses `time.NewTimer` with `defer Stop()`.

api/event/wslistener.go
```diff
@@ -0,0 +1,154 @@
...
+
+// connectAndListen dials the event WebSocket and reads until it drops or Stop
+// is called; returns whether it connected, so the caller can reset backoff.
+func (w *wsListener) connectAndListen() (connected bool) {
+	dialer := websocket.Dialer{HandshakeTimeout: w.handshakeTimeout}
+	conn, _, dialErr := dialer.Dial(w.wsURL, w.wsHeader)
+	if dialErr != nil {
+		return false
+	}
```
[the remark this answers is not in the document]
jisung-02: Not changing this one—gorilla/websocket already handles both halves of the concern on the failure path.

Checked against the pinned version, `gorilla/websocket v1.5.3`, in `client.go`:

- The underlying connection is closed by the library, not the caller. `DialContext` installs `defer func() { if netConn != nil { netConn.Close() } }()` at line 331 and only clears `netConn` at line 425, after a successful 101. Every error return between those points closes the socket.
- The `*http.Response` handed back on a non-101 is not backed by that connection. Immediately before returning `ErrBadHandshake`, the library slurps up to 1 KiB and swaps the body for an in-memory buffer: `resp.Body = ioutil.NopCloser(bytes.NewReader(buf[:n]))` at line 403. Closing that is a no-op, so repeated reconnect attempts have nothing to accumulate.

So the discarded response holds no file descriptor and no unread socket, and `listenLoop`'s reconnect loop does not leak across attempts.

Branch-scope sweep, since a real leak here would apply everywhere the same call shape appears: the repository has three other `Dial` sites—`api/websh/websh.go:207` and `:257`, and `pkg/tunnel/runtime/runtime.go:98`—all of which discard the response identically. They are consistent with each other and correct for the same reason, so none of them needed a change either.

api/event/wslistener.go
```diff
@@ -0,0 +1,160 @@
...
+	stopped     chan struct{} // closed when listenLoop exits
+	connected   chan struct{} // closed after first successful WebSocket connection
+	connectOnce sync.Once
+	closeOnce   sync.Once
+	mu          sync.Mutex // guards conn; embedders may reuse it for their own state
+	conn        *websocket.Conn
+}
+
+// newWSListener builds the shared skeleton. ac may be nil (empty header, for tests).
+func newWSListener(ac *client.AlpaconClient, wsURL string, handshakeTimeout time.Duration) wsListener {
```
junho226: Suggestion: this returns a `wsListener` by value, and the type holds a `sync.Mutex` plus two `sync.Once`.

It is safe as written — the returned value is pristine, no goroutine has touched the mutex yet — and `go vet` will never flag it, but not because it is fine: copylocks deliberately gives up on call expressions, since `lockPathRhs` returns nil for a `*ast.CallExpr` on the assumption a call may return a zero value. So the safety here rests on a reader knowing the value is fresh, not on tooling.

The cost is that copy-by-value becomes the shape of the type. A later `w := *someListener` or a value receiver would read as consistent with this constructor while being a real bug. Returning a pointer removes the question:

```suggestion
func newWSListener(ac *client.AlpaconClient, wsURL string, handshakeTimeout time.Duration) *wsListener {
```

Embed `*wsListener` in both listeners; promotion for `sl.mu` / `sl.conn` / `sl.done` / `sl.stopped` is unchanged, so the call sites and tests stay as they are.

api/event/wslistener.go
```diff
@@ -0,0 +1,160 @@
...
+	wsHeader         http.Header
+	handshakeTimeout time.Duration
+	handleFrame      func(message []byte)
+
+	done        chan struct{}
+	stopped     chan struct{} // closed when listenLoop exits
+	connected   chan struct{} // closed after first successful WebSocket connection
+	connectOnce sync.Once
+	closeOnce   sync.Once
+	mu          sync.Mutex // guards conn; embedders may reuse it for their own state
```
junho226: Suggestion: `embedders may reuse it for their own state` is honest, and `CommandOutputListener` does exactly that for `commandID` — but it means the lock's invariant now spans two files, which is the reasoning burden the PR description flags rather than removes.

A dedicated mutex on `CommandOutputListener` for `commandID` costs three lines and lets this field document itself as *guards conn*, full stop. Worth it since the whole point of the refactor is that a third listener will read this skeleton without having read the embedders.

No deadlock risk either way today — `connectAndListen` releases `mu` before calling `handleFrame` — so this is about the comment being able to state one invariant instead of two.

api/event/sudolistener_test.go
```diff
@@ -154,18 +143,12 @@ func TestSudoListener_ConnectAndListen_ExitsOnDisconnect(t *testing.T) {
...
-		wsHeader:  http.Header{},
-		done:      make(chan struct{}),
-		stopped:   make(chan struct{}),
-		connected: make(chan struct{}),
-	}
+	sl := NewSudoListener(nil, wsURL, "")
 
 	// Run connectAndListen and signal when the read loop processes a message
 	go func() {
 		defer close(sl.stopped)
```
junho226: Suggestion: `close(w.stopped)` is now the shared `Start`'s deferred responsibility, so this test is reaching into a channel it no longer owns.

No double-close today, because this test drives `connectAndListen` directly rather than going through `Start`. But if anyone later adds `sl.Start()` here — a natural thing to try when extending the test — it panics on close of a closed channel, and the cause is two files away.

A local signal keeps the test's own lifecycle separate from the skeleton's:

```go
readDone := make(chan struct{})
go func() {
	defer close(readDone)
	_ = sl.connectAndListen()
}()
```

and then select on `readDone` below instead of `sl.stopped`.

api/event/wslistener.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		select {
+		case <-w.done:
+			return
+		case <-time.After(delay):
+			delay = nextReconnectDelay(delay)
+		}
+	}
+}
```
junho226, line 113: Highlight: extracting this was worth it for more than the 1+2+4+8+16 seconds of real waiting it avoids — it pins the operation order, which the inline version in `listenLoop` never asserted.

Doubling before clamping is the correct semantics: the cap bounds the delay actually waited, not the value that gets doubled. Clamping first would let the wait stretch to twice the ceiling, and `"overshoot clamps to cap"` in the table test is exactly the case that catches it. Confirmed against `origin/main` that both original loops did `delay *= 2` then clamp, so this is parity and not a retuning.

junho226 (dismissed):
## Review summary

Collapsing the skeleton before adding a third listener is the right order of operations, and the refactor lands where a pure refactor should: net −50 lines, zero behavior change, existing assertions untouched.

### Verified independently

I reproduced the claimed checks in a worktree and compared the control flow against `origin/main` rather than taking parity on faith.

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l ./api/ ./cmd/` | no differences |
| `go test -race ./...` | exit 0 |
| `go test -race -count=1 ./api/event/...` | **ok 54.4s, race-clean** |

Parity points I checked against `origin/main`:

- **Backoff constants** — `main` really did carry `1s`/`30s` in both files (`reconnectBaseDelay`/`reconnectMaxDelay` and `commandOutputReconnectBase`/`commandOutputReconnectMax`), so unifying them is not a silent retuning.
- **`listenLoop` control flow** — sudo's `if err == nil { return }` fired only on the two `return true, nil` sites, both guarded by a closed `done`. The new loop reaches the same return through `select { case <-w.done: return ... }`, since only that arm is ready. Equivalent.
- **Backoff reset semantics** — dial failure returns `false` (no reset), read error returns `true` (reset). Both preserved on each path.
- **Dropped error wrapping** — the wrapped dial/read errors were only ever compared against `nil`; nothing logged or surfaced them. No observability loss, as the description claims.

### Critical Issues (0)

None.

### Warnings (0)

None.

### Suggestions (3)

**1. `api/event/wslistener.go:35` — `newWSListener` returns a lock-bearing struct by value**

Safe today because the returned value is pristine, and `go vet`'s copylocks misses it by design (`lockPathRhs` returns nil for a `*ast.CallExpr`, on the assumption a call may return a zero value). But it makes copy-by-value the shape of a type holding a `sync.Mutex` and two `sync.Once`, so a later real copy reads as consistent. Returning `*wsListener` and embedding the pointer keeps every promotion working.

**2. `api/event/commandoutput_listener.go:71` — reusing the shared `mu` for `commandID` couples two files**

Not a regression (the original single-file listener used one mutex for both), and the field comment is honest about it. But the invariant now spans the skeleton and the embedder. A dedicated mutex on `CommandOutputListener` costs three lines and lets `wsListener.mu` document itself as *guards conn*, full stop.

**3. `api/event/sudolistener_test.go:150` — the test closes `stopped`, a channel `Start` now owns**

No double-close today because this test drives `connectAndListen` directly. But `close(w.stopped)` is now `Start`'s deferred responsibility, so anyone adding `sl.Start()` here gets a close-of-closed-channel panic.

### Highlights

- **The `(bool, error)` → `bool` reconciliation is correctly argued**, and I confirmed the equivalence against `main` rather than trusting the claim.
- **`nextReconnectDelay` extraction pins the operation order.** Doubling before clamping is the correct semantics — the cap bounds the delay actually waited, not the value that gets doubled — the loop never asserted it, and the table test now does, including `"overshoot clamps to cap"`, exactly the mutation that would silently double the ceiling.
- **The `Start` panic is placed where the cause is.** The pre-existing nil `handleFrame` call sat in `connectAndListen`, so it could only fire on the read goroutine after a successful dial. With a third listener coming in #271, failing at assignment time on the caller's stack is the right trade for an unexported type.
- **The nil-`ac` test mode rationale is load-bearing, not incidental.** `&client.AlpaconClient{}` emits an empty `Origin` that gorilla's `checkSameOrigin` rejects, while nil yields no `Origin` header at all. Documenting *why* prevents someone "fixing" it into a broken state.
- **Honest coverage accounting**, including the one arm (`wslistener.go:148`) that stays uncovered, with a correct explanation of why it is not deterministically reachable: `Stop` closes `conn`, so `ReadMessage` always wakes with an error first.
- Moving only the tests that never dial, and leaving the dialing ones in place as per-embedder wiring proof, is the right split.

---

**Verdict**: APPROVED

- [x] Follows stack conventions (`gofmt`, `go vet`, unexported shared type, doc comments on exported API)
- [x] Tests included and passing — `go test -race ./...` exit 0, `api/event` race-clean
- [x] No import cycle introduced (`client` imports no `api/*` package)
- [x] Behavior parity confirmed by diffing against `origin/main` — the reviewer checklist item is satisfied

api/event/wslistener.go
jisung-02: Applied in `c2cddcf`. `newWSListener` now returns `*wsListener` and both listeners embed the pointer.

The copylocks detail is the part that settled it — I had been reading `go vet` silence as evidence the value return was fine, and `lockPathRhs` returning nil for `*ast.CallExpr` means it was never checked at all. Agreed that leaving the value shape available is the real cost, not the current call sites.

Field promotion held as you said: `sl.mu`, `sl.conn`, `sl.done`, `sl.stopped` all resolve unchanged, and no call site or test needed an edit.

api/event/wslistener.go
jisung-02: Applied in `49312ba`. `CommandOutputListener` now has its own `cmdMu` for `commandID`, and `wsListener.mu` is documented as `guards conn` with nothing appended.

You are right that the shared-lock comment described the burden rather than removing it. The three lines are cheap next to a third listener having to read both embedders to learn what `mu` covers.

`handleMessage` and `setCommandID` were the only two `mu` users on the command-output side, so the split is complete — no remaining embedder reuses the skeleton lock.

api/event/sudolistener_test.go
jisung-02: Applied in `40734c5`, using the local `readDone` channel exactly as suggested.

This was a leftover from before `stopped` moved into the skeleton — the test predates `Start` owning the close, and I carried the line across without re-checking who owns the channel now.

I swept the branch for the same pattern: `sudolistener_test.go` was the only test closing `stopped` itself. Every other reference (`TestSudoListener_StopClosesConnection:114`, `TestSudoListener_ReconnectsAfterDisconnect:240`) only receives from it, which is the correct direction.

api/event/wslistener.go
jisung-02: Thanks for checking the order against `origin/main` independently — that was the one thing the extraction could have silently changed, and parity is exactly what it needed to be.

The table case you name is also the one I mutation-checked: swapping the clamp ahead of the doubling turns `"overshoot clamps to cap"` from `16s -> 30s` into `16s -> 32s`, and the test fails. Same for `"cap stays at cap"`, which goes to `1m0s`.

junho226 (approved):
## Re-review (delta only)

My earlier review on `4543892` was dismissed by the three follow-up commits, so this pass covers only `4543892..40734c5` — the three suggestions I raised, and nothing already settled above.

| Commit | Suggestion it answers | Verdict |
| --- | --- | --- |
| `c2cddcf` | `newWSListener` returned a lock-bearing struct by value | Resolved |
| `49312ba` | `commandID` reused the skeleton's `mu` | Resolved |
| `40734c5` | the test closed `stopped`, a channel `Start` owns | Resolved |

### What I checked on the delta

**`c2cddcf` — pointer return and embed.** `newWSListener` now returns `*wsListener` and both listeners embed the pointer, so copy-by-value is no longer the shape of a type holding a `sync.Mutex` and two `sync.Once`. The embedded field keeps the name `wsListener`, so every promoted method and field access is unchanged, and `sl.handleFrame = sl.handleMessage` still writes through to the same struct.

The one thing a pointer embed can introduce is a nil-deref on a zero value, so I checked every construction site: `sudolistener.go:60` and `commandoutput_listener.go:50` are the only ones, both through the constructors, and `grep` finds no bare `SudoListener{}` / `CommandOutputListener{}` anywhere. An external package can still write `event.SudoListener{}`, but the embedded field is unexported so it cannot be populated — and the value-embed version was equally unusable there (nil channels, `close(nil)` in `Start`). Not a regression.

**`49312ba` — dedicated `cmdMu`.** `commandID` is now guarded by its own mutex and `wsListener.mu`'s comment drops to `guards conn`. I confirmed the two are independent: `handleMessage` and `setCommandID` are the only readers/writers of `commandID` and neither touches `conn`, so there is no cross-invariant that needed one lock. The remaining `sl.mu` uses in `sudolistener_test.go` (105, 120, 157) all guard `sl.conn` — the field the comment now claims — so they read correctly against the narrowed contract.

**`40734c5` — test-owned signal.** The goroutine now closes a local `readDone` instead of `sl.stopped`, which removes the latent double-close for anyone who later adds `sl.Start()` to this test. `sl.stopped` is left untouched and nothing else waits on it here.

### Verification

| Check | Result |
| --- | --- |
| `gofmt -l ./api/ ./cmd/` | no differences |
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -race -count=1 ./api/event/...` | **ok 54.9s, race-clean** |
| `go test -race -count=1 ./...` | all packages ok |

No new findings. Everything in the previous review's Highlights still holds — the delta touched none of it.

---

**Verdict**: APPROVED

PR #289, opened 2026-07-28
refactor(event): accept any event type in the subscription helper

jisung-02:
## Background

`api/event/subscribe.go` carried `SubscribeSudoEvent` and `SubscribeCommandOutput`, whose bodies were identical apart from the `EventType` literal and the error string. `EventSubscriptionRequest` already held an `event_type` field, so the only thing pinning the type was the helper itself.

`event watch --type work_session`, which #271 calls for, is a third event type. Under the current shape every new type means another copy of the same function. This is PR2 of the #271 plan and does not close the issue on its own.

## What changed

Both helpers are replaced by `SubscribeEvent(ac, channelID, eventType, targetID)`. The plan left the option of reducing them to wrappers, but a wrapper only moves the hardcoded type from the helper to the wrapper, and `api/` has no external consumers, so the call sites were updated instead. Precedent is #237.

There are two call sites.

| Location | Change |
|---|---|
| `cmd/websh/websh.go:357` | `event.SubscribeEvent(ac, eventSession.ChannelID, event.EventTypeSudo, sessionID)` |
| `api/event/event.go:339` | `SubscribeEvent(ac, session.ChannelID, EventTypeCommandOutput, cmdID)` |

`EventTypeSudo` and `EventTypeCommandOutput` are new constants in `subscribe.go`. Without them the call sites pass bare strings, which relocates the hardcoding rather than removing it. They are exported because `cmd/websh` references them. Only the two types the CLI consumes are declared; the server owns the full set, and `SubscribeEvent` forwards types that have no constant here.

A second commit replaces the remaining `"command_output"` literal in the frame filter at `api/event/commandoutput_listener.go:70` with the same constant. It is not part of the subscription helper, so it sits outside the literal scope of PR2, but it was the last hardcoded copy of that string. One line, no behavior change, and it was separated into its own commit so it could be dropped if a reviewer preferred. Review has since kept it.

A named `type EventType string` was suggested in review and deliberately not taken here: it would have to propagate to `EventSubscriptionRequest.EventType` and to everything that builds or reads that struct, which is more than a PR scoped to the subscription helper should carry. The JSON wire shape is not part of the cost — a named type with underlying kind `string` marshals identically to `string`, so `POST /api/events/subscriptions/` would be unaffected. Worth revisiting when `work_session` lands in PR4/PR5 and there are three call-site types instead of two.

## Error message

The wrap moves from per-type literals to `failed to subscribe to %s events: %w`, reproducing the previous output of both functions with `eventType` as a format argument.

The shape matters. `isNotFoundError` at `cmd/websh/websh.go:371` only checks whether the message ends with a literal `": not found"` or `": not found."`, and it is what silently skips the 404 from older servers that lack the subscription API. If `%w` stops being last, that path starts emitting a warning instead.

## Tests

`TestSubscribeSudoEvent` and `TestSubscribeCommandOutput_SendsExpectedPayload` were near-identical and are now one table-driven `TestSubscribeEvent_SendsExpectedPayload`. The cases are sudo, command output, and `work_session`, which has no constant in the CLI. That last case covers what this PR actually sells: an arbitrary `event_type` reaching the payload untouched.

`TestSubscribeSudoEvent_ServerError` is renamed `TestSubscribeEvent_ServerError` and keeps the 500 case.

A third commit adds `TestSubscribeEvent_NotFoundKeepsServerMessageLast`, which pins the wrap order described above. The test server returns a real 404 with `{"detail": "Not found."}` and the assertion is `strings.HasSuffix(strings.ToLower(err.Error()), ": not found.")` — the exact predicate `isNotFoundError` evaluates. It replaces a prefix assertion that only described that property in a comment without testing it. `isNotFoundError` is unexported in `cmd/websh`, so the test mirrors the suffix match rather than calling it.

The `TestCreateEventSession` tests are unchanged.

```
go build ./... && go vet ./...
go test -race ./...
golangci-lint run ./...
gofmt -l .
```

All packages pass, lint reports 0 issues, gofmt reports no differences. A repo-wide search confirms neither deleted symbol is referenced in code, comments, or docs.

## Checklist

- [x] No remaining references to the two deleted functions
- [x] Error wrap compatibility with `isNotFoundError` pinned by a test
- [x] Reviewer decided the `commandoutput_listener.go` one-liner stays in this PR

Refs #271

### DIFF
--- a/api/event/subscribe.go
+++ b/api/event/subscribe.go
@@ -10,6 +10,10 @@ import (
 const (
 	eventSessionsURL      = "/api/events/sessions/"
 	eventSubscriptionsURL = "/api/events/subscriptions/"
+
+	// Event types the CLI subscribes to; the server defines the full set.
+	EventTypeSudo          = "sudo"
+	EventTypeCommandOutput = "command_output"
 )
 
 // EventSessionResponse is returned when creating an event session.
@@ -42,34 +46,18 @@ func CreateEventSession(ac *client.AlpaconClient) (*EventSessionResponse, error)
 	return &resp, nil
 }
 
-// SubscribeSudoEvent subscribes the given channel to sudo events for a websh session.
-func SubscribeSudoEvent(ac *client.AlpaconClient, channelID, sessionID string) error {
+// SubscribeEvent subscribes the given channel to eventType events, scoped to
+// targetID (a websh session for sudo, a command for command_output).
+func SubscribeEvent(ac *client.AlpaconClient, channelID, eventType, targetID string) error {
 	req := &EventSubscriptionRequest{
 		Channel:   channelID,
-		EventType: "sudo",
-		TargetID:  sessionID,
+		EventType: eventType,
+		TargetID:  targetID,
 	}
 
 	_, err := ac.SendPostRequest(eventSubscriptionsURL, req)
 	if err != nil {
-		return fmt.Errorf("failed to subscribe to sudo events: %w", err)
-	}
-
-	return nil
-}
-
-// SubscribeCommandOutput subscribes the channel to command_output events for the
-// given command (target_id scopes delivery to that command's chunks).
-func SubscribeCommandOutput(ac *client.AlpaconClient, channelID, commandID string) error {
-	req := &EventSubscriptionRequest{
-		Channel:   channelID,
-		EventType: "command_output",
-		TargetID:  commandID,
-	}
-
-	_, err := ac.SendPostRequest(eventSubscriptionsURL, req)
-	if err != nil {
-		return fmt.Errorf("failed to subscribe to command_output events: %w", err)
+		return fmt.Errorf("failed to subscribe to %s events: %w", eventType, err)
 	}
 
 	return nil
--- a/api/event/subscribe_test.go
+++ b/api/event/subscribe_test.go
@@ -40,31 +40,47 @@ func TestCreateEventSession(t *testing.T) {
 	assert.Equal(t, expected.ChannelID, resp.ChannelID)
 }
 
-func TestSubscribeSudoEvent(t *testing.T) {
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		assert.Equal(t, http.MethodPost, r.Method)
-		assert.Contains(t, r.URL.Path, "events/subscriptions")
-
-		var req EventSubscriptionRequest
-		err := json.NewDecoder(r.Body).Decode(&req)
-		assert.NoError(t, err)
-		assert.Equal(t, "channel-456", req.Channel)
-		assert.Equal(t, "sudo", req.EventType)
-		assert.Equal(t, "session-123", req.TargetID)
-
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusCreated)
-		_, _ = w.Write([]byte(`{"id":"sub-789"}`))
-	}))
-	defer ts.Close()
-
-	ac := &client.AlpaconClient{
-		HTTPClient: ts.Client(),
-		BaseURL:    ts.URL,
+func TestSubscribeEvent_SendsExpectedPayload(t *testing.T) {
+	tests := []struct {
+		name      string
+		eventType string
+		targetID  string
+	}{
+		{name: "sudo", eventType: EventTypeSudo, targetID: "session-123"},
+		{name: "command output", eventType: EventTypeCommandOutput, targetID: "command-uuid"},
+		{name: "type the CLI does not know yet", eventType: "work_session", targetID: "ws-uuid"},
 	}
 
-	err := SubscribeSudoEvent(ac, "channel-456", "session-123")
-	assert.NoError(t, err)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				assert.Equal(t, http.MethodPost, r.Method)
+				assert.Contains(t, r.URL.Path, "events/subscriptions")
+
+				body, err := io.ReadAll(r.Body)
+				require.NoError(t, err)
+
+				var req EventSubscriptionRequest
+				require.NoError(t, json.Unmarshal(body, &req))
+				assert.Equal(t, "channel-456", req.Channel)
+				assert.Equal(t, tt.eventType, req.EventType)
+				assert.Equal(t, tt.targetID, req.TargetID)
+
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusCreated)
+				_, _ = w.Write([]byte(`{"id":"sub-789"}`))
+			}))
+			defer ts.Close()
+
+			ac := &client.AlpaconClient{
+				HTTPClient: ts.Client(),
+				BaseURL:    ts.URL,
+			}
+
+			err := SubscribeEvent(ac, "channel-456", tt.eventType, tt.targetID)
+			assert.NoError(t, err)
+		})
+	}
 }
 
 func TestCreateEventSession_ServerError(t *testing.T) {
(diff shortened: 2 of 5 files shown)


api/event/subscribe.go
```diff
@@ -10,6 +10,10 @@ import (
 const (
 	eventSessionsURL      = "/api/events/sessions/"
 	eventSubscriptionsURL = "/api/events/subscriptions/"
+
+	// Event types the CLI subscribes to; the server defines the full set.
+	EventTypeSudo          = "sudo"
+	EventTypeCommandOutput = "command_output"
```
junho226, line 16: Suggestion (optional): a named type would make the generalization typo-proof at zero runtime cost, since every call site already passes a constant.

```suggestion
	// Event types the CLI subscribes to; the server defines the full set.
	EventTypeSudo          EventType = "sudo"
	EventTypeCommandOutput EventType = "command_output"
```

This needs `type EventType string` and a matching change to `EventSubscriptionRequest.EventType`, so it does ripple a little. Fine to leave as-is if you would rather keep the request struct untouched — the constants already remove the literal-typo risk at the call sites that matter.

api/event/subscribe_test.go
```diff
@@ -94,35 +110,8 @@ func TestSubscribeSudoEvent_ServerError(t *testing.T) {
...
-	defer ts.Close()
-
-	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
-
-	err := SubscribeCommandOutput(ac, channelID, cmdID)
-	assert.NoError(t, err)
+	err := SubscribeEvent(ac, "channel-456", EventTypeSudo, "session-123")
+	require.Error(t, err)
+	// websh's isNotFoundError matches a literal ": not found" suffix, so the wrap must keep the server's message last.
+	assert.Contains(t, err.Error(), "failed to subscribe to sudo events: ")
```
junho226, line 116: Suggestion (nitpick): the comment above states the server's message must come *last*, but `assert.Contains` only proves the prefix is present — it would still pass if something were appended after the wrapped error.

`%w` at the tail of the format string already guarantees the property structurally, so this is not a real gap. But if you want the test to actually pin what the comment claims, asserting the suffix instead would:

```go
assert.True(t, strings.HasPrefix(err.Error(), "failed to subscribe to sudo events: "))
```

or have the test server write a `not found` body and assert `strings.HasSuffix(err.Error(), ": not found")` — that would guard the actual `isNotFoundError` contract end to end.

junho226 (dismissed):
## Review: #289 — refactor(event): accept any event type in the subscription helper

### Summary

`SubscribeSudoEvent` / `SubscribeCommandOutput` collapse into one `SubscribeEvent(ac, channelID, eventType, targetID)`, with `EventTypeSudo` / `EventTypeCommandOutput` as exported constants. Net −23 lines, no behavior change.

### Verification

- `gofmt -l` clean, `go vet` clean, `go test ./api/event/... ./cmd/websh/...` → ok
- All call sites migrated; no leftover `"command_output"` / `"sudo"` event-type literals in production code (the remaining `"sudo"` hits are work-session *scope* names and approval `request_type` — unrelated)
- **Error-message parity confirmed**: `"failed to subscribe to %s events: %w"` renders byte-identical to both previous messages, so `cmd/websh/websh.go:371` `isNotFoundError`'s `": not found"` suffix match still holds. Documenting that cross-package coupling in the test is a genuinely good catch — it is exactly the kind of thing a consolidation like this quietly breaks.

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (2, both optional)

See inline comments. Neither blocks merge.

### Highlights

- Test coverage is strictly better than what it replaced: the table adds an unknown-type case (`work_session`) that pins the "accepts any type" contract, and the error test now pins the wrap format instead of only asserting non-nil.
- `commandoutput_listener.go:70` picking up the constant closes the last literal — the refactor is complete, not half-done.

### Checklist

- [x] No documentation update needed (internal helper)

**Verdict: APPROVE**

api/event/subscribe_test.go
jisung-02: Applied in 11be92c, taking the second option.

The first suggestion (`strings.HasPrefix` on the 500 case) does not actually close the gap you identified — a prefix assertion proves no more about trailing content than `assert.Contains` did. So instead I added `TestSubscribeEvent_NotFoundKeepsServerMessageLast`: the test server returns a real 404 with `{"detail": "Not found."}`, and the assertion is `strings.HasSuffix(strings.ToLower(err.Error()), ": not found.")` — the exact predicate `isNotFoundError` evaluates. Removed the misleading comment from the 500 case since the property is now pinned by a test rather than described in prose.

`isNotFoundError` is unexported in `cmd/websh`, so the test mirrors its suffix match rather than calling it; the constraint is recorded in the commit trailer.

api/event/subscribe.go
jisung-02: Leaving as-is for this PR, but correcting the reason I gave — you are right that the wire contract is not the cost. A named type with underlying kind `string` marshals identically to `string`, so `EventSubscriptionRequest.EventType` would keep the exact same JSON shape on `POST /api/events/subscriptions/`. My earlier reading of that was wrong.

What actually argues against it here is the ripple: the type has to propagate to the struct and to everything that builds or reads it, which is more than a PR scoped to the subscription helper should carry. The exported constants already remove the typo risk at every call site, which is the part that mattered.

Worth revisiting when `work_session` lands (PR4/PR5 of #271) and there are three call-site types instead of two. The description now records the ripple rather than wire compatibility.

jisung-02:
Pushed 11be92c and updated the description to match.

Applied your second suggestion. `TestSubscribeEvent_NotFoundKeepsServerMessageLast` has the test server return a real 404 with `{"detail": "Not found."}` and asserts `strings.HasSuffix(strings.ToLower(err.Error()), ": not found.")` — the exact predicate `isNotFoundError` evaluates, rather than a prefix assertion with a comment claiming more than it proved. `isNotFoundError` is unexported in `cmd/websh`, so the test mirrors the suffix match instead of calling it.

Did not take the `type EventType string` suggestion in this PR; reasoning is in the inline reply and now in the description.

Description changes: the Error message section no longer claims `TestSubscribeEvent_ServerError` pins the wrap, the Tests section covers the new test, the `commandoutput_listener.go` checklist item is resolved as kept, and the named-type decision is recorded.

Re-verified on 11be92c: `go build ./...`, `go vet ./...`, `gofmt -l .` clean, `golangci-lint run ./...` 0 issues, `go test -race ./...` all packages pass.

junho226 (approved):
## Re-review of the delta `f4b11c53..11be92c`

My earlier APPROVE was dismissed by the push, so this only covers what changed since. The delta is one file, `api/event/subscribe_test.go`:

| Change | Assessment |
|---|---|
| Misleading comment dropped from `TestSubscribeEvent_ServerError` | Correct — the property moved to a dedicated test, so the prose no longer needs to claim it |
| `TestSubscribeEvent_NotFoundKeepsServerMessageLast` added | Closes the gap from thread `3663524071` |

## Verification

| Check | Result |
|---|---|
| `go build ./...`, `go vet ./...`, `gofmt -l .` | clean |
| `go test ./api/event/... ./cmd/websh/...` | ok |
| Deleted symbols (`SubscribeSudoEvent`, `SubscribeCommandOutput`) still referenced anywhere | 0 hits in code, comments, docs |
| Remaining `"sudo"` literals | work-session *scope* names and approval `request_type` only — unrelated, re-confirmed |

**The decisive check.** I injected the exact regression the property guards against — kept `%w` in place and appended text *after* it (`"...events: %w (retry with --verbose)"`):

- `TestSubscribeEvent_ServerError` (the old `assert.Contains`) → **PASS** ← the gap I originally flagged
- `TestSubscribeEvent_NotFoundKeepsServerMessageLast` → **FAIL**

So the new test catches a wrap-order change that would silently break `isNotFoundError`, and my originally-suggested `strings.HasPrefix` variant would **not** have caught it. You were right to reject it and take the second option instead.

## Threads

**`type EventType string`** (thread `3663524055`) — deferral accepted; three call-site types in PR4/PR5 is the right moment to revisit.

One correction for the record, since the reasoning will be carried forward: a named type whose underlying kind is `string` marshals identically to `string`, so `EventSubscriptionRequest.EventType` would keep the exact same JSON wire shape. The wire contract is not the cost here. The real cost is the ripple across the struct and its users plus the widening of this PR's scope — which is a sufficient reason on its own. Worth having the record say that rather than wire compatibility.

**Suffix assertion** (thread `3663524071`) — resolved.

## Nitpick (informational, no action needed)

The assertion omits `strings.TrimSpace`, which `isNotFoundError` (`cmd/websh/websh.go:375`) applies before matching, so it is not literally the same predicate. It is *stricter* rather than looser, though: a passing assertion implies the predicate is true, so it can only ever false-alarm, never false-pass. Sound as a guard — leaving it as-is is fine.

## Critical issues (0) · Warnings (0)

---

**Verdict: APPROVE**

PR #290, opened 2026-07-29
refactor(event): relocate command history to exec ls

jisung-02:
## Background

`alpacon event` lists command executions from `/api/events/commands/`. The subscription channel #271 wants is a different API — `/api/events/sessions/` plus `/api/events/subscriptions/`. Under one name the parent is command-scoped while the children cover every event type.

So `event` is promoted to mean the event channel, and the history table moves next to `alpacon exec logs JOB_ID`. PR3 of the #271 plan and the only break in it. PR1 (#287) and PR2 (#289) are merged. Does not close the issue.

## What changed

| Commit | Change |
|---|---|
| `1c1b82d` | `type EventType string` in `api/event` |
| `25fde70` | New `alpacon exec ls` |
| `75fc725` | `alpacon event` warns, then delegates to `exec.RunList` |
| `dcac960` | Event-type constants back into one `const` block (review) |
| `09477cd` | Shorten the deprecated `Short` so root help stops wrapping (review) |
| `e7942ce` | Share the list flag set between the two commands (review) |
| `983faad` | Drop unreachable returns after `CliErrorWithExit` (review) |

`exec ls` keeps alias `list` and the same `--tail`/`--server`/`--user`. The API layer is untouched — `GetEventList` stays in `api/event`, which `cmd/exec` already imports.

`event` still works, warns on stderr, and delegates instead of holding a second copy. Delegation over duplication is empirical: the two copies drifted three times during review — error wording, flag help, `Example` syntax. `cmd/token/acl_list.go:12` already uses this shape.

Nothing about the flag set is duplicated either. Both commands call `exec.AddListFlags(cmd)` to declare the three flags and `exec.RunListFromFlags(cmd)` to read them, so the flag names appear only in `cmd/exec/list.go` and `cmd/event/event.go` contains none of them. Two exported symbols rather than one because `cmd/event` needs its own `FlagSet`, so registration and read cannot collapse into a single call.

## Reasoning

Cobra's `Deprecated` field is unused: a non-empty value makes `IsAvailableCommand()` false (v1.10.2, `command.go:1607-1610`), dropping `event` from root help and with it any way to find the subcommands landing here next. Its built-in notice also prints via `c.Printf` to stdout, which would corrupt `--output json`. The `cmd/token/acl_*` precedent covers leaf commands being fully replaced.

Correction to PR2: #289 deferred the named type partly on the grounds that it would change the JSON wire shape. That was wrong — an underlying kind of `string` marshals identically. The real cost was struct propagation and scope growth, reason enough on its own. `EventType` applies to `SubscribeEvent`'s parameter, `EventSubscriptionRequest.EventType`, and the WS envelope field; the constants stay untyped so every const sits in one block. No validation, because `event watch --type` must accept server types the CLI has no constant for — the `"work_session"` case in `subscribe_test.go` pins that.

## Behavior changes

`alpacon event --tail 0` is now rejected. Previously `page_size` was omitted (`api/event/event.go:158`) and the server default of 15 applied. `--tail` already defaults to 25, so 0 was only a second path to a smaller value and `--tail 15` is identical. Matches `exec ls` and the same flag on `audit`, `log`, `webftp`.

The fetch-failure message on `event` changes from "Failed to get events" to "Failed to retrieve the commands". Exit code unchanged at 1.

## Known limitations

- A server named `ls` or `list` is unreachable through `exec ls` — it routes to the subcommand and prints the table with no error. Pre-existing for `logs`; the `--` escape is now documented in `cmd/exec/exec.go`'s `Long`.
- During the transition `event` still prints command history rather than event-channel output. Deliberate.
- The deprecation notice names no removal version. None has been decided, and the removal is effectively gated on PR4/PR5 landing rather than on a time-based window, since that is when a bare `alpacon event` printing command history becomes actively wrong.

## Tests

`cmd/exec/list_test.go` pins that a child routes and parses its own flags while the parent `ExecCmd` sets `DisableFlagParsing`. Per cobra the child is unaffected, but nothing pinned it and `exec` had no flag-bearing subcommand to show it. All four cases reset flags from their own `DefValue` to remove order dependence.

`cmd/event/event.go` has no unit test: `utils.CliWarning` writes straight to `os.Stderr`, so capturing it would mean refactoring production code, and `cmd/log`, `cmd/audit`, `cmd/webftp` have none either. Verified on the binary instead.

```
go build -o alpacon . && go vet ./... && gofmt -l .
golangci-lint run ./...
go test -race ./...
go test -race -shuffle=on -count=2 ./cmd/exec/
```

All packages pass (37 ok), lint reports 0 issues, gofmt reports no differences. Each commit was checked out into a temporary worktree and builds and vets independently.

Checked on the binary: `event` still in root help; the notice goes to stderr only, leaving `--output json` stdout intact; `--tail 0` and `--tail -1` exit 1 before any network call; `exec -- ls uptime` treats `ls` as the server name. The two flag blocks are now diffed rather than eyeballed — `diff` of the `Flags:` section of `event --help` against `exec ls --help` reports only cobra's auto `--help` line.

## Checklist

- [x] Each of the seven commits builds independently
- [x] `event` confirmed still present in root help
- [x] stderr/stdout separation confirmed under `--output json`
- [x] Wire format unchanged
- [x] `--tail 0` rejection confirmed acceptable — it matches `audit`, `log`, `webftp`, and `--tail 15` reproduces the old behavior exactly
- [x] Flag declarations collapsed; the two help outputs are byte-identical apart from cobra's `--help` line

Refs #271

### DIFF
--- a/api/event/subscribe.go
+++ b/api/event/subscribe.go
@@ -10,10 +10,15 @@ import (
 const (
 	eventSessionsURL      = "/api/events/sessions/"
 	eventSubscriptionsURL = "/api/events/subscriptions/"
+)
 
-	// Event types the CLI subscribes to; the server defines the full set.
-	EventTypeSudo          = "sudo"
-	EventTypeCommandOutput = "command_output"
+// EventType is a server event channel type, marshalled as a plain string. The server
+// defines the full set; untyped literals convert, so types the CLI lacks a constant for stay usable.
+type EventType string
+
+const (
+	EventTypeSudo          EventType = "sudo"
+	EventTypeCommandOutput EventType = "command_output"
 )
 
 // EventSessionResponse is returned when creating an event session.
@@ -25,9 +30,9 @@ type EventSessionResponse struct {
 
 // EventSubscriptionRequest is sent to subscribe to an event type.
 type EventSubscriptionRequest struct {
-	Channel   string `json:"channel"`
-	EventType string `json:"event_type"`
-	TargetID  string `json:"target_id"`
+	Channel   string    `json:"channel"`
+	EventType EventType `json:"event_type"`
+	TargetID  string    `json:"target_id"`
 }
 
 // CreateEventSession creates a new event session and returns the WebSocket URL
@@ -48,7 +53,7 @@ func CreateEventSession(ac *client.AlpaconClient) (*EventSessionResponse, error)
 
 // SubscribeEvent subscribes the given channel to eventType events, scoped to
 // targetID (a websh session for sudo, a command for command_output).
-func SubscribeEvent(ac *client.AlpaconClient, channelID, eventType, targetID string) error {
+func SubscribeEvent(ac *client.AlpaconClient, channelID string, eventType EventType, targetID string) error {
 	req := &EventSubscriptionRequest{
 		Channel:   channelID,
 		EventType: eventType,
--- a/api/event/subscribe.go
+++ b/api/event/subscribe.go
@@ -10,16 +10,16 @@ import (
 const (
 	eventSessionsURL      = "/api/events/sessions/"
 	eventSubscriptionsURL = "/api/events/subscriptions/"
+
+	// Left untyped so every const stays in one block; they remain assignable and
+	// comparable to EventType.
+	EventTypeSudo          = "sudo"
+	EventTypeCommandOutput = "command_output"
 )
 
 // EventType is a server event channel type, marshalled as a plain string. The server
 // defines the full set; untyped literals convert, so types the CLI lacks a constant for stay usable.
 type EventType string
-
-const (
-	EventTypeSudo          EventType = "sudo"
-	EventTypeCommandOutput EventType = "command_output"
-)
 
 // EventSessionResponse is returned when creating an event session.
 type EventSessionResponse struct {
--- a/cmd/event/event.go
+++ b/cmd/event/event.go
@@ -9,7 +9,7 @@ import (
 var EventCmd = &cobra.Command{
 	Use:     "event",
 	Aliases: []string{"events"},
-	Short:   "List recent remote command executions (deprecated: use 'exec ls')",
+	Short:   "List recent remote command executions (deprecated)",
 	Long: `List recent remote command executions, most recent last.
 
 This command has moved to 'alpacon exec ls' and will be removed in a future
--- a/cmd/event/event.go
+++ b/cmd/event/event.go
@@ -1,8 +1,7 @@
 package event
 
 import (
-	"github.com/alpacax/alpacon-cli/api/event"
-	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/cmd/exec"
 	"github.com/alpacax/alpacon-cli/utils"
 	"github.com/spf13/cobra"
 )
@@ -10,47 +9,32 @@ import (
 var EventCmd = &cobra.Command{
 	Use:     "event",
 	Aliases: []string{"events"},
-	Short:   "Retrieve and display recent Alpacon events",
-	Long: `
-	Retrieve and display a list of recent events from the Alpacon, with options to filter by server, user, and the number of events. 
-	Use the '--tail' flag to limit the output to the last N event entries. 
-	Specify a server with '--server' or filter events by user with '--user' to narrow down the results.
-	`,
-	Example: `
-	alpacon event
-	alpacon events
-	alpacon event -tail 10 -s myserver -u admin
-	alpacon event --tail=10 --server=myserver --user=admin
-	`,
+	Short:   "List recent remote command executions (deprecated: use 'exec ls')",
+	Long: `List recent remote command executions, most recent last.
+
+This command has moved to 'alpacon exec ls' and will be removed in a future
+release. It still works today and takes the same flags.
+
+The 'event' name is being repurposed for the Alpacon event channel.`,
+	Example: `  alpacon event
+  alpacon event --tail 10 --server my-server --user admin`,
+	// Deprecated is deliberately unused: it makes IsAvailableCommand() false, dropping
+	// 'event' from root help and with it the event-channel subcommands landing here.
 	Run: runEvent,
 }
 
 func init() {
-	var pageSize int
-	var serverName string
-	var userName string
-
-	EventCmd.Flags().IntVarP(&pageSize, "tail", "t", 25, "Number of event entries to show from the end")
-	EventCmd.Flags().StringVarP(&serverName, "server", "s", "", "Specify server for events")
-	EventCmd.Flags().StringVarP(&userName, "user", "u", "", "Specify request user for events")
+	EventCmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
+	EventCmd.Flags().StringP("server", "s", "", "Filter by server name")
+	EventCmd.Flags().StringP("user", "u", "", "Filter by requesting user")
 }
 
-func runEvent(cmd *cobra.Command, args []string) {
+func runEvent(cmd *cobra.Command, _ []string) {
+	utils.CliWarning("'alpacon event' has moved to 'alpacon exec ls' and will be removed in a future release.")
+
 	pageSize, _ := cmd.Flags().GetInt("tail")
 	serverName, _ := cmd.Flags().GetString("server")
 	userName, _ := cmd.Flags().GetString("user")
 
-	alpaconClient, err := client.NewAlpaconAPIClient()
-	if err != nil {
-		utils.CliErrorWithExit("Connection to Alpacon API failed: %s. Consider re-logging.", err)
-		return
-	}
-
-	eventList, err := event.GetEventList(alpaconClient, pageSize, serverName, userName)
-	if err != nil {
-		utils.CliErrorWithExit("Failed to get events: %s.", err)
-		return
-	}
-
-	utils.PrintTable(eventList)
+	exec.RunList(pageSize, serverName, userName)
 }
--- a/cmd/event/event.go
+++ b/cmd/event/event.go
@@ -24,17 +24,11 @@ The 'event' name is being repurposed for the Alpacon event channel.`,
 }
 
 func init() {
-	EventCmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
-	EventCmd.Flags().StringP("server", "s", "", "Filter by server name")
-	EventCmd.Flags().StringP("user", "u", "", "Filter by requesting user")
+	exec.AddListFlags(EventCmd)
 }
 
 func runEvent(cmd *cobra.Command, _ []string) {
 	utils.CliWarning("'alpacon event' has moved to 'alpacon exec ls' and will be removed in a future release.")
 
-	pageSize, _ := cmd.Flags().GetInt("tail")
-	serverName, _ := cmd.Flags().GetString("server")
-	userName, _ := cmd.Flags().GetString("user")
-
-	exec.RunList(pageSize, serverName, userName)
+	exec.RunListFromFlags(cmd)
 }
--- a/cmd/exec/list.go
+++ b/cmd/exec/list.go
@@ -19,25 +19,34 @@ and --user to filter by the requesting user.`,
   alpacon exec ls --tail 10
   alpacon exec ls --tail 10 --server my-server --user admin`,
 	Run: func(cmd *cobra.Command, _ []string) {
-		pageSize, _ := cmd.Flags().GetInt("tail")
-		serverName, _ := cmd.Flags().GetString("server")
-		userName, _ := cmd.Flags().GetString("user")
-
-		RunList(pageSize, serverName, userName)
+		RunListFromFlags(cmd)
 	},
 }
 
 func init() {
-	listCmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
-	listCmd.Flags().StringP("server", "s", "", "Filter by server name")
-	listCmd.Flags().StringP("user", "u", "", "Filter by requesting user")
+	AddListFlags(listCmd)
 
 	ExecCmd.AddCommand(listCmd)
 }
 
-// RunList is exported for the deprecated 'alpacon event', which delegates here so the
-// two cannot drift. Their flag sets are still declared separately and must stay in sync.
-func RunList(pageSize int, serverName, userName string) {
+// AddListFlags and RunListFromFlags are exported for the deprecated 'alpacon event', which
+// delegates here. Both live in this file so the flag names exist in exactly one place.
+func AddListFlags(cmd *cobra.Command) {
+	cmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
+	cmd.Flags().StringP("server", "s", "", "Filter by server name")
+	cmd.Flags().StringP("user", "u", "", "Filter by requesting user")
+}
+
+// RunListFromFlags requires the flag set registered by AddListFlags.
+func RunListFromFlags(cmd *cobra.Command) {
+	pageSize, _ := cmd.Flags().GetInt("tail")
+	serverName, _ := cmd.Flags().GetString("server")
+	userName, _ := cmd.Flags().GetString("user")
+
+	runList(pageSize, serverName, userName)
+}
+
+func runList(pageSize int, serverName, userName string) {
 	utils.RequirePositiveInt("tail", pageSize)
 
 	alpaconClient, err := client.NewAlpaconAPIClient()
--- a/cmd/exec/list.go
+++ b/cmd/exec/list.go
@@ -1,0 +1,56 @@
+package exec
+
+import (
+	"github.com/alpacax/alpacon-cli/api/event"
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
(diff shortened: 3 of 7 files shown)


api/event/subscribe.go
```diff
@@ -10,10 +10,15 @@ import (
...
+// EventType is a server event channel type, marshalled as a plain string. The server
+// defines the full set; untyped literals convert, so types the CLI lacks a constant for stay usable.
+type EventType string
 
-	// Event types the CLI subscribes to; the server defines the full set.
-	EventTypeSudo          = "sudo"
-	EventTypeCommandOutput = "command_output"
+const (
+	EventTypeSudo          EventType = "sudo"
+	EventTypeCommandOutput EventType = "command_output"
```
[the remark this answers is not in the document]
jisung-02: Applied in dcac960. The event-type constants are untyped again, so all constants stay in one `const` block above `type EventType string`, matching the const -> var -> type -> func rule.

The type safety this PR was after is unaffected: `SubscribeEvent`'s parameter, `EventSubscriptionRequest.EventType`, and the WS envelope field are all still `EventType`, so a plain `string` variable still cannot reach them. The only thing given up is assigning a constant to a `string` variable, which nothing does.

One note on the cited precedent: `api/event/event.go:19-39` groups const -> var -> var but has no `type` declaration, so it does not cover the enum case directly. The counter-precedent is `cmd/worksession/worksession_create.go:20-31`, which declares `type useDecision int` followed by its own `const` block — the shape stdlib uses for enums too (`time.Month`). Going with the written rule here since the rest of `api/event` groups strictly by kind.

cmd/exec/list.go
```diff
@@ -0,0 +1,56 @@
...
+		pageSize, _ := cmd.Flags().GetInt("tail")
+		serverName, _ := cmd.Flags().GetString("server")
+		userName, _ := cmd.Flags().GetString("user")
+
+		RunList(pageSize, serverName, userName)
+	},
+}
+
+func init() {
+	listCmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
```
junho226: Warning: the flag sets are still duplicated, which is the very drift this PR set out to remove.

These three flags are declared independently here and in `cmd/event/event.go:27-29`, guarded only by the comment below saying they "must stay in sync". The PR's rationale for delegating rather than duplicating `RunList` is empirical — the two copies drifted three times during review. The flag declarations are the remaining copy, and they carry the same failure mode: a default, shorthand, or help string changed on one side diverges silently, and a flag *renamed* on one side turns `GetString` into a silently empty filter.

This closes with about six lines — export a registrar next to `RunList`:

```suggestion
	AddListFlags(listCmd)
```

with

```go
// AddListFlags registers the shared flag set so 'exec ls' and the deprecated
// 'event' cannot drift.
func AddListFlags(cmd *cobra.Command) {
	cmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
	cmd.Flags().StringP("server", "s", "", "Filter by server name")
	cmd.Flags().StringP("user", "u", "", "Filter by requesting user")
}
```

and `exec.AddListFlags(EventCmd)` on the `cmd/event` side. `cmd/event` already imports `cmd/exec`, so this needs no new dependency.

If you'd rather keep the surface at one exported symbol, an `exec.RunListFromFlags(cmd *cobra.Command)` that both `Run` bodies call would collapse the duplicated `GetInt`/`GetString` triple as well and let `RunList` go unexported. Either is fine; the registrar alone is the smaller change.

cmd/exec/list.go
```diff
@@ -0,0 +1,56 @@
...
+
+// RunList is exported for the deprecated 'alpacon event', which delegates here so the
+// two cannot drift. Their flag sets are still declared separately and must stay in sync.
+func RunList(pageSize int, serverName, userName string) {
+	utils.RequirePositiveInt("tail", pageSize)
+
+	alpaconClient, err := client.NewAlpaconAPIClient()
+	if err != nil {
+		utils.CliErrorWithExit("Connection to Alpacon API failed: %s. Consider re-logging.", err)
+		return
```
junho226: Suggestion: dead `return` after `CliErrorWithExit`.

`utils.CliErrorWithExit` ends in `os.Exit(1)` (`utils/log.go:28`), so this `return` and the one at line 52 are unreachable. They came along with the code moved out of `cmd/event`, so this is pre-existing rather than new — but `list.go` is a new file and it's the cheap moment to drop them.

```suggestion
	}
```

cmd/event/event.go
```diff
@@ -1,56 +1,40 @@
...
-	EventCmd.Flags().StringVarP(&serverName, "server", "s", "", "Specify server for events")
-	EventCmd.Flags().StringVarP(&userName, "user", "u", "", "Specify request user for events")
+	EventCmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
+	EventCmd.Flags().StringP("server", "s", "", "Filter by server name")
+	EventCmd.Flags().StringP("user", "u", "", "Filter by requesting user")
 }
 
-func runEvent(cmd *cobra.Command, args []string) {
+func runEvent(cmd *cobra.Command, _ []string) {
+	utils.CliWarning("'alpacon event' has moved to 'alpacon exec ls' and will be removed in a future release.")
```
junho226, line 31: Suggestion: the deprecation notice has no removal milestone.

"will be removed in a future release" gives a scripted caller nothing to plan against. If a target version is known, naming it here and in the `Long` above makes the transition window explicit.

cmd/event/event.go
```diff
@@ -1,56 +1,40 @@
...
 
 func init() {
-	var pageSize int
-	var serverName string
-	var userName string
-
-	EventCmd.Flags().IntVarP(&pageSize, "tail", "t", 25, "Number of event entries to show from the end")
-	EventCmd.Flags().StringVarP(&serverName, "server", "s", "", "Specify server for events")
-	EventCmd.Flags().StringVarP(&userName, "user", "u", "", "Specify request user for events")
+	EventCmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
```
junho226: Warning (same issue as `cmd/exec/list.go:31`): this is the second copy of the flag set.

See the comment there for the proposed `exec.AddListFlags(cmd)` registrar. Applied here:

```suggestion
	exec.AddListFlags(EventCmd)
```

Note the failure mode is not symmetric with the `RunList` delegation you already did: a rename here makes `GetString` below return an empty string and the filter silently stops working, while a changed default or help string just diverges from `exec ls` unnoticed. Sharing the declaration removes both.

junho226 (dismissed):
## Summary

PR3 of the #271 plan: `alpacon event`'s command-history table moves to a new `alpacon exec ls`, and `event` becomes a thin deprecated shim that warns on stderr and delegates to `exec.RunList`. `EventType` is introduced in `api/event` as a named string type for the subscription/WS surface.

The design reasoning is unusually well documented, and I verified the load-bearing claims independently rather than taking them on trust:

- **The `Deprecated` field argument is correct.** cobra v1.10.2 `command.go:IsAvailableCommand()` returns false when `len(c.Deprecated) != 0`, so setting it really would drop `event` from root help — which defeats the purpose of keeping the name reachable for the event-channel subcommands landing next. The `cmd/token/acl_list.go` precedent uses `Deprecated` **with** `Hidden: true`, i.e. a leaf being fully retired; that is a different case, exactly as the description says.
- **No import cycle.** `cmd/event` → `cmd/exec` → `cmd/worksession`; `cmd/exec` does not reach back into `cmd/event`, and only `cmd/root.go` imports `cmd/event`. `cmd/exec` already importing `cmd/worksession` sets the precedent for a cmd→cmd edge.
- **`--tail 0` rejection is consistent, not a one-off.** `RequirePositiveInt("tail", ...)` is already what `cmd/audit`, `cmd/log`, and `cmd/webftp` do. Since `--tail` defaults to 25 and `0` merely fell through to the server's default of 15, `--tail 15` reproduces the old behavior exactly.
- `go build ./...`, `go vet ./...`, `gofmt -l .`, and `go test ./cmd/exec/ ./api/event/` are all clean on `09477cd`.

I'd consider the last checklist item ("Confirm `--tail 0` rejection is acceptable as a breaking change") answered: yes — it aligns this command with its three siblings and has a trivially documented equivalent.

Nothing here blocks the merge. One maintainability point and two small suggestions below.

---

## Warnings (1)

### 1. The flag sets are still duplicated, which is the very drift this PR set out to remove

`cmd/exec/list.go:31-33` and `cmd/event/event.go:27-29` declare the same three flags independently, guarded only by a comment saying they "must stay in sync". The PR's own rationale for delegating rather than duplicating `RunList` is empirical — the two copies drifted three times during review. The flag declarations are the remaining copy, and they carry the same failure mode: a default, shorthand, or help-string changed on one side diverges silently, and a flag *renamed* on one side turns `GetString` into a silently empty filter.

This closes with about six lines. Export a registrar next to `RunList`:

```go
// AddListFlags registers the shared flag set so 'exec ls' and the deprecated
// 'event' cannot drift.
func AddListFlags(cmd *cobra.Command) {
	cmd.Flags().IntP("tail", "t", 25, "Number of command entries to show from the end")
	cmd.Flags().StringP("server", "s", "", "Filter by server name")
	cmd.Flags().StringP("user", "u", "", "Filter by requesting user")
}
```

then `AddListFlags(listCmd)` in `cmd/exec/list.go`'s `init()` and `exec.AddListFlags(EventCmd)` in `cmd/event/event.go`'s. `cmd/event` already imports `cmd/exec`, so this needs no new dependency.

If you'd rather keep the surface at one exported symbol, an `exec.RunListFromFlags(cmd *cobra.Command)` that both `Run` bodies call would collapse the duplicated `GetInt`/`GetString` triple as well and let `RunList` go unexported. Either is fine; the registrar alone is the smaller change.

---

## Suggestions (2)

### 2. Dead `return` after `CliErrorWithExit`

`cmd/exec/list.go:46,52` — `utils.CliErrorWithExit` ends in `os.Exit(1)` (`utils/log.go:28`), so both `return`s are unreachable. They came along with the code moved out of `cmd/event`, so this is pre-existing rather than new, but `list.go` is a new file and it's the cheap moment to drop them.

### 3. The deprecation notice has no removal milestone

"will be removed in a future release" gives a scripted caller nothing to plan against. If a target version is known, naming it in both the `Long` and the runtime warning makes the transition window explicit.

---

## Highlights

- The old `Example` block advertised `alpacon event -tail 10 -s myserver -u admin`. Under pflag, `-tail` parses as the `-t` shorthand with value `ail`, so that line never worked — it's quietly fixed by the rewrite.
- `TestExecLsRoutesAndParsesFlags` pins something genuinely non-obvious: `ExecCmd` sets `DisableFlagParsing: true` (`cmd/exec/exec.go:98`), and until now `exec` had no flag-bearing subcommand to demonstrate that a child still parses its own flags. Resetting each flag from its own `DefValue` rather than a hardcoded literal keeps the "defaults" case honest about what is actually registered *and* removes order dependence — that's why `-shuffle=on -count=2` passes.
- `subscribe_test.go`'s third case (`eventType: "work_session"`) is the right way to pin the design decision: it proves an untyped literal for a server event type the CLI has no constant for still compiles, which is exactly the property that makes validation-free `EventType` safe.
- Documenting the `--` escape for a server literally named `ls`/`logs` in `exec.go`'s `Long` addresses a pre-existing sharp edge that this PR merely widens by one name.
- The correction to PR2's stated reasoning (a named type with underlying kind `string` marshals identically, so the wire-shape concern was wrong) is the kind of thing that's easy to leave unsaid.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (build/vet/gofmt/test verified locally on `09477cd`)
- [x] Breaking change (`--tail 0`) documented and consistent with sibling commands
- [ ] Optional: collapse the duplicated flag declarations before merge

cmd/exec/list.go
jisung-02: Agreed, and taken as the fuller of the two options you offered — `e7942ce`.

You pointed out the failure modes are asymmetric: a rename breaks the *read* side (`GetString` silently returns `""`), while a changed default or help string diverges on the *declaration* side. A registrar alone only closes the second one, so both are shared now:

- `exec.AddListFlags(cmd)` — the three declarations
- `exec.RunListFromFlags(cmd)` — the `GetInt`/`GetString` triple
- `RunList` is unexported as `runList`

Net effect is that `cmd/event/event.go` no longer contains a single flag name, which was not true with the registrar alone. Exported surface stays at two symbols because `cmd/event` needs a `FlagSet` of its own, so registration and read cannot collapse into one call.

Verified the two help outputs are now byte-identical apart from cobra's auto `--help` line:

```
$ diff <(alpacon event --help | sed -n "/^Flags:/,/^$/p") <(alpacon exec ls --help | sed -n "/^Flags:/,/^$/p")
2c2
<   -h, --help            help for event
---
>   -h, --help            help for ls
```

The `RunList` doc comment that claimed the flag sets "must stay in sync" is gone with the thing it warned about, and the matching bullet under Known limitations is dropped from the description.

cmd/event/event.go
jisung-02: Done in `e7942ce` — `exec.AddListFlags(EventCmd)` here, and `exec.RunListFromFlags(cmd)` in `runEvent`, so the read side is shared too. Full reasoning in the `cmd/exec/list.go:31` thread.

cmd/exec/list.go
jisung-02: Dropped both in `983faad`, and the majority-form argument turned out stronger than "it is a new file":

```
$ grep -rl "CliErrorWithExit" --include="*.go" . | wc -l
135
$ grep -rn -A1 "CliErrorWithExit" --include="*.go" . | grep -c "return$"
14
```

Only three files ever wrote the trailing return — `cmd/ftp/cp.go` (3), `cmd/exec/logs.go` (3), `cmd/exec/exec.go` (6), plus the 2 here. So it is not house style; 132 files already omit it, and removing these moves `list.go` onto the common form rather than away from it.

Per this repo's review convention I checked the rest rather than only the flagged line. The remaining 12 are unrelated to this PR — `logs.go` and `exec.go` are the same package, so it is a one-line-each cleanup, but the diff would touch code this PR has no other reason to enter. Happy to do it here if you prefer, otherwise I will file it separately.

cmd/event/event.go
jisung-02: No target version has been decided, so I have not put one in — naming a release I cannot commit to would be worse than the vague wording.

Two things worth noting for whoever picks it:

- The repo has no precedent to copy. The three `cmd/token/acl_*` commands use cobra's `Deprecated` with the form `use 'alpacon token acl command ls' instead` — no version, no window. Latest tag is `v1.9.1`.
- Unlike those, `event` is not merely renamed: the name is being reused for the event channel in PR4/PR5. Once `event watch`/`event wait` exist, a bare `alpacon event` printing command history is actively confusing, so the removal is effectively gated on those PRs landing rather than on a time-based window.

Leaving this thread open for you to name a version. If you give me one I will put it in both the `Long` and the runtime warning in a follow-up commit.

junho226 (approved):
## Re-review scope

My APPROVE on `09477cd` was auto-dismissed by the two pushes that followed, so this covers only `09477cd..983faad` — the fixes for the two items I raised. Everything verified in the previous review still holds and is not repeated here.

---

## Previous findings

### Warning 1 — duplicated flag sets → resolved in `e7942ce`

Taken as the fuller of the two options, which is the right call: a registrar alone would have closed the declaration-side drift but left the `GetInt`/`GetString` triple, and that is the side where a rename fails silently. Verified on the head:

- `grep -rn "RunList\|AddListFlags\|runList" --include="*.go" .` returns 10 hits, all in `cmd/exec/list.go` and `cmd/event/event.go`. `RunList` is gone entirely; `runList` is unexported.
- `cmd/event/event.go` contains no flag name at all — `init()` is one line, `runEvent` is two.
- Help parity on the built binary: a `diff` of the `Flags:` section of `event --help` against `exec ls --help` reports only cobra's auto line (`help for event` vs `help for ls`). That parity is now structural rather than something a reviewer has to keep eyeballing.

Two exported symbols instead of one is correctly justified — `cmd/event` needs its own `FlagSet`, so registration and read cannot collapse into a single call.

### Suggestion 2 — dead `return` after `CliErrorWithExit` → resolved in `983faad`

Confirmed `utils/log.go:28` ends in `os.Exit(1)`, so both were unreachable. The majority-form data you gathered is a stronger argument than the one I gave, and leaving the other 12 out of scope is right — they sit in files this PR has no other reason to enter.

### Suggestion 3 — removal milestone → closing it, no version needed

Your reasoning answers it. The removal is gated on PR4/PR5 landing rather than on a time window, since that is the point at which a bare `alpacon event` printing command history becomes actively wrong; a date-based deprecation would be inventing a commitment the schedule does not have. The `cmd/token/acl_*` precedent naming no version points the same way. Please resolve the thread.

---

## Verification on `983faad`

```
go build ./...                            OK
go vet ./...                              OK
gofmt -l .                                clean
go test -race ./cmd/exec/ ./api/event/    ok (5.5s / 55.6s)
```

Binary checks: `event` is still present in root help; the warning goes to stderr only with stdout clean; `--tail 0` exits 1 before any network call on both `event` and `exec ls`.

---

## Verdict: APPROVED

No new findings in the delta.

**Checklist**

- [x] Tests included and passing (verified locally on `983faad`)
- [x] Breaking change (`--tail 0`) documented and consistent with sibling commands
- [x] Flag declarations collapsed — the remaining item from my previous review

Issue #253, opened 2026-07-03
fix(approval): unify pending/rejected exit-code contract for machine consumers


jisung-02:
## Background

The pending-approval state has a dedicated exit code 4 (`ExitCodePendingApproval`) and a JSON envelope (`"status": "pending_approval"`, `next_actions`) so that machine consumers can branch without parsing text. However, this contract only applies to the happy path of `exec`. On other paths, consumers that rely on exit codes or JSON alone—AI agents, CI scripts—misread the state. Lightweight-model agents in particular tend to branch on exit codes without interpreting stderr prose, so the inconsistencies below turn into real misbehavior.

## Problems

1. A reviewer rejection (`rejected`) exits with the same code 1 as a generic failure. It is distinguishable only by the message `"command was rejected by a reviewer"` (`api/event/event.go:358`, `cmd/exec/logs.go:90-94`), so an agent can misread it as a transient error, retry the rejected command repeatedly, and keep generating approval requests. This is asymmetric with pending approval, which does get its own exit code 4.
2. `alpacon websh SERVER "command"` (non-interactive execution) goes through `HandleCommandResult` only and is not wired to `HandlePendingApproval` (`cmd/websh/websh.go:227-230`). In the same pending-approval situation it prints the bare line `"command is awaiting human approval"` and exits 1, so a consumer expecting exit 4 (as on `exec`) branches into the failure path.
3. `alpacon exec logs JOB_ID` returns exit 0 for the pending-approval state (`cmd/exec/logs.go:82-88`). The "still in progress" intent is reasonable, but a consumer that only checks the exit code reads it as "succeeded with empty output" and marks the command as done. With `--output json`, the pending state is not emitted in a structured form either.

## Proposal

- Give `rejected` a dedicated exit code, or at minimum an `error_code` in the JSON envelope (e.g., `command_rejected`), so machines get a "do not retry" signal. The code itself now exists—see Status below. The original suggestion of `5` is no longer available: `ExitCodeServerBusy` took it, and the settled-without-a-grant code is `6`.
- Route the non-interactive command branch of `websh` through the same `RunExecWithApprovalWait` + `HandlePendingApproval` path as `exec`, so it shares the exit-4 contract.
- For the pending-approval case in `exec logs`, emit a `"status": "pending_approval"` envelope when `--output json` is set. Whether the plain-text path should keep exit 0 needs discussion (changing it is breaking).

## Status

[...]

PR #291, opened 2026-07-30
feat(event): add event watch to stream the event channel

jisung-02:
## Background

Half of the push-based notification #271 asks for, built on PR1 #287, PR2 #289, and PR3 #290.

The open question this PR had to settle first — event session expiry — changed its scope. An event channel token cannot be reused: the event server's check requires `opened_at IS NULL`, and since alpacon-server #2760 the token also carries a 60s TTL stamped at creation. The existing `wsListener` re-dialed that dead URL silently on a 30s backoff forever. Since a watch is meant to stay up for hours, a reconnect there means creating a new session and re-subscribing, so the command and that reprovisioning logic land together.

## What changed

| Commit | Change |
| --- | --- |
| `f277a01` | `provision` / `onConnected` / `onDialFailed` hooks and `readLimit` / `reconnectBaseDelay` fields on `wsListener` |
| `b3428e6` | `omitempty` on `EventSubscriptionRequest.TargetID` |
| `3619368` | `Watcher` — provisions a session per attempt, re-subscribes once connected |
| `2a7c9e8` | `renderEvent` — one line per frame |
| `2d05ddd` | `alpacon event watch`, and the deprecated `event` notice now points at it |
| `a5236e5` | review follow-up: usage line and the `--output json` help wording |
| `5a7087d` | review follow-up: `defer signal.Stop` on the watch signal channel |
| `30e1560` | review follow-up: a transient session-create failure retries instead of ending the command |
| `a64419e` | review follow-up: name the provision step in the connect-timeout comment |
| `671d4bf` | review follow-up: 408 and 429 leave the fatal session-create band |
| `f06bcad` | review follow-up: the test session-create failure body names its own status |

`--type` is required and is not validated locally: the server owns the type set, and PR3 pinned that a type the CLI has no constant for must still work. With `--output json` each line is the server frame compacted to a single line (NDJSON) — insignificant whitespace is dropped, but every field and the server's key order survive, including fields the CLI has no constant for; the table default prints four fixed fields — receive time, event type, sub type, and the target. stdout carries event lines only, everything else goes to stderr. `utils.PrintTable` is deliberately unused: it pages through `WriteToPager` and assumes a whole dataset.

`watch` does not interpret event types and stays up until SIGINT or SIGTERM. Terminal-state detection and the exit-code contract belong to PR5 (`wait`) and #253; this PR uses only 0 and 1. Per-decision rationale lives in each commit's Lore trailers.

## Safety

Server-supplied text reaches a TTY and an agent's log verbatim. Both frame fields and server error messages are stripped of ANSI escapes, control bytes, and Unicode format characters, and line breaks and tabs become spaces. A bidi override could otherwise render `denied` as `approved`, and a deleted newline would glue two server sentences into one word.

The WebSocket URL from the session response is validated with `url.Parse` before the dialer sees it. Letting the dialer fail on it yields a `*url.Error` that quotes the whole URL, and that URL carries a token that has not been consumed yet. Frames are capped at 1 MiB.

The `--output json` path is deliberately not sanitized, so C1 and DEL bytes pass through. That mode exists for lossless delivery to a machine consumer.

## Behavior changes

`connected` now closes after `onConnected` returns nil rather than right after the dial, so `WaitConnected` means connected and subscribed. The two existing listeners leave `onConnected` nil and are unaffected.

A failure before any successful subscribe is fatal: the watcher stops and exits 1 with the server's own message. Session creation is the exception, split by status — the request body is empty, so it cannot carry a caller mistake, and only a 4xx other than 408 and 429 proves retrying pointless. A 5xx, a 408, a 429, or a network error retries inside the connect window like a dial failure does. Failures after a first success are ordinary reconnect material, with the first failure of each outage announced once.

## Tests

```
gofmt -l .                                                     no output
go vet ./...                                                   pass
golangci-lint run ./...                                        0 issues
go test -race ./...                                            all packages pass
go test -race -shuffle=on -count=2 ./api/event/ ./cmd/event/    pass
```

Each of the eleven commits builds and vets independently in a throwaway worktree.

A review round found that `watch` called `signal.Notify` without a matching `signal.Stop`, unlike `cmd/tunnel/run.go:93` and `cmd/tunnel/tunnel.go:141`. Fixed in `5a7087d`. `api/websh/websh.go:220` has the same gap plus a `sigChan` goroutine that outlives the call, but it is not touched here — websh holds the terminal in raw mode, so changing its signal path needs its own verification.

Review also flagged that a pre-subscribe session-create failure was fatal while a dial failure in the same window was retried. Fixed in `30e1560`. `TestWatcher_SessionCreateFailureStopsAndSurfaces` asserted the old behavior on a 500 and would have kept passing for the wrong reason, so it was re-pointed at a 401 and joined by a retryable-status case. One consequence is left as is: a persistent 5xx now reports the connect timeout rather than the server's message, matching how a persistent pre-subscribe dial failure already behaves.

A follow-up round flagged that the resulting fatal band swallowed 408 and 429, the two 4xx that ask to be retried. Fixed in `671d4bf`, with the retryable case now table-driven over 502, 408, and 429 — the two new rows fail without the fix, so they are not passing for the wrong reason. `api/ftp/ftp.go:550` returns on any 4xx in the same way inside `cp`'s download retry loop. It is pre-existing and untouched here, and it needs more than the same two constants: `maxAttempts` reaches `fetchFromURLToFile` as 100 from `api/ftp/ftp.go:726` with a flat `time.Sleep(time.Second)`, so making 429 retryable there without changing the schedule would send 100 requests one second apart at a server that just asked for fewer. The event path is bounded instead by a doubling backoff under the 15s connect ceiling. That fix wants its own PR, with the backoff first.

Verified end to end against a local native stack — alpacon-server main `11e63ad9c` on :8000, event-server on :8082, backhaul-server, proxy-server, alpamon, Postgres 16, Redis:

- `notification` in table mode: `10:47:37	notification	probe	-`
- `work_session` in NDJSON: completing a session delivered `sub_type=completed` with `previous_status=active` and the full status payload
- `command_output` in NDJSON: `seq=0`, `content="hello-from-exec\n"`
- NDJSON is exactly one line per frame and preserves fields the CLI does not know
- Ctrl+C exits 0. A nonexistent target, an unknown `--type`, and a missing required target exit 1 immediately with `work_session_not_found`, `invalid_choice`, and `invalid_input` respectively, with no retry
- The channel row showed `token_created_at` plus a 60s TTL with `opened_at` set, confirming the token cannot be reused
- Killing and restarting the event server produced exactly one outage warning and one recovery warning, a new session row per attempt, and delivery resumed. The dial error carried only host:port, so no token leaked

Four types were not observed: `servers_commit` only fires on the first commit of a server that is not yet commissioned (`servers/api/views.py:299`), `sudo` runs inside websh's interactive MFA flow, and the four `cloud_*` types need a cloud account and a celery worker — the area #271 explicitly defers. `command_fin` and `command_reject` have no publisher in server code at all and exist only in the enum. Subscribing succeeds in every one of those cases.

## Known limitations

Once a work session reaches a terminal state the server deletes the subscription (`work_sessions/services.py:765`), so `watch` goes quiet rather than exiting. Events published while disconnected are lost, and if stdout stalls for over 60s the pong lapses and the connection is re-established.

`SudoListener` and `CommandOutputListener` stay on the static-URL constructor, so after a drop they re-dial a URL whose token can no longer be used and retry every 30s until `Stop`. Both were traced rather than assumed:

- `command_output` is covered. `streamSubscribed` runs a REST poller alongside the WebSocket, drains the remaining chunks over REST when the command finishes, and calls `Stop` there, so no output is lost and the retry loop is bounded by the command's lifetime. The only cost is that output during an outage arrives in a burst at the end.
- `sudo` has no equivalent fallback. The MFA request reaches the CLI only as a pushed event, and websh keeps the listener for the whole terminal session, so a drop mid-session means later `sudo` prompts never appear.

Neither is addressed here. Moving them onto provisioners changes websh and exec at runtime and needs its own verification, and the `sudo` case has not been reported in practice — so this is deliberately deferred and will be fixed if it comes up. Note commit `f277a01`'s `Directive` trailer calls it a follow-up, which predates this decision.

## Follow-ups

The server never deletes `EventSession` or `EventUserChannel` rows — there is no reaper, and DELETE only sets `closed_at` — so a long watch accumulates rows. That cannot be fixed in the CLI. Extending `utils.StripControlChars` to cover Cf runes would give websh and worksession rendering the same protection. Host validation on the server-supplied URL is pre-existing behavior, but this change exercises it in a permanent reconnect loop, which raises its priority.

Refs #271

## Checklist

- [x] `go test -race ./...` passes, golangci-lint reports 0 issues
- [x] Each of the eleven commits builds independently
- [x] Three event types received against a live local stack, with reconnect and exit codes measured
- [x] stdout/stderr separation and flag isolation confirmed on the binary

### DIFF
--- a/api/event/watcher.go
+++ b/api/event/watcher.go
@@ -3,11 +3,13 @@ package event
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"sync"
 	"time"
 
 	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
 )
 
 const (
@@ -54,8 +56,8 @@ func NewWatcher(ac *client.AlpaconClient, eventType EventType, targetID string) 
 	return w
 }
 
-// Only after a first success: before that, a dial is worth retrying inside
-// WaitConnected's window rather than ending the command.
+// Only after a first success: before that, a retryable attempt is worth another try
+// inside WaitConnected's window rather than ending the command.
 func (w *Watcher) announceOutage(cause error) {
 	w.stateMu.Lock()
 	subscribed := w.subscribed
@@ -83,10 +85,21 @@ func (w *Watcher) Err() error {
 	return w.err
 }
 
+// The session request carries neither --type nor --target, so only a 4xx proves retrying
+// pointless. Anything else gets the retry window a dial failure already gets.
+func (w *Watcher) sessionCreateFailed(cause error) error {
+	if status := utils.HTTPStatusCode(cause); status >= http.StatusBadRequest && status < http.StatusInternalServerError {
+		return w.fail(cause)
+	}
+
+	w.announceOutage(cause)
+	return cause
+}
+
 func (w *Watcher) provisionSession() (string, error) {
 	session, err := CreateEventSession(w.ac)
 	if err != nil {
-		return "", w.fail(err)
+		return "", w.sessionCreateFailed(err)
 	}
 
 	// Rejected here because the dialer's own *url.Error quotes the whole URL, and this
@@ -137,10 +150,9 @@ func (w *Watcher) subscribe() error {
 	return nil
 }
 
-// A failure before any successful subscribe is fatal, so a bad target or an
-// unreachable server surfaces the server's own message rather than a generic connect
-// timeout. Session creation counts: before a first success nothing proves the request
-// valid. Later failures are ordinary reconnect material.
+// A failure before any successful subscribe is fatal, so a bad target or an expired
+// login surfaces the server's own message rather than a generic connect timeout.
+// Later failures are ordinary reconnect material.
 func (w *Watcher) fail(cause error) error {
 	w.stateMu.Lock()
 	fatal := !w.subscribed
--- a/api/event/watcher.go
+++ b/api/event/watcher.go
@@ -1,0 +1,182 @@
+package event
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/alpacax/alpacon-cli/client"
+)
+
+const (
+	watchHandshakeTimeout = 10 * time.Second
+	watchFrameBuffer      = 256
+
+	// Bounded so a compromised server cannot grow the CLI's memory without limit.
+	watchFrameLimit = 1024 * 1024
+)
+
+// Watcher streams raw event frames for one event type. Event channel tokens are
+// single-use, so every dial attempt provisions its own session and re-subscribes once
+// connected. When WaitConnected returns false, Err holds the reason if it was fatal.
+type Watcher struct {
+	*wsListener
+	ac              *client.AlpaconClient
+	eventType       EventType
+	targetID        string
+	frames          chan []byte
+	reconnected     chan struct{}
+	reconnectFailed chan error
+
+	stateMu    sync.Mutex // guards channelID, subscribed, warned, err
+	channelID  string
+	subscribed bool
+	warned     bool
+	err        error
+}
+
+func NewWatcher(ac *client.AlpaconClient, eventType EventType, targetID string) *Watcher {
+	w := &Watcher{
+		ac:              ac,
+		eventType:       eventType,
+		targetID:        targetID,
+		frames:          make(chan []byte, watchFrameBuffer),
+		reconnected:     make(chan struct{}, 1),
+		reconnectFailed: make(chan error, 1),
+	}
+	w.wsListener = newProvisionedWSListener(ac, w.provisionSession, watchHandshakeTimeout)
+	w.readLimit = watchFrameLimit
+	w.handleFrame = w.handleMessage
+	w.onConnected = w.subscribe
+	w.onDialFailed = w.announceOutage
+	return w
+}
+
+// Only after a first success: before that, a dial is worth retrying inside
+// WaitConnected's window rather than ending the command.
+func (w *Watcher) announceOutage(cause error) {
+	w.stateMu.Lock()
+	subscribed := w.subscribed
+	w.stateMu.Unlock()
+
+	if subscribed {
+		_ = w.fail(cause)
+	}
+}
+
+func (w *Watcher) Frames() <-chan []byte { return w.frames }
+
+// Reconnected fires on each re-subscribe. Events published while disconnected are
+// lost for good: the event channel has no history to replay.
+func (w *Watcher) Reconnected() <-chan struct{} { return w.reconnected }
+
+// ReconnectFailed reports the first failure of each outage, so a stream whose
+// reconnects keep failing is distinguishable from a quiet one.
+func (w *Watcher) ReconnectFailed() <-chan error { return w.reconnectFailed }
+
+// Err returns the error that stopped the Watcher before it ever subscribed.
+func (w *Watcher) Err() error {
+	w.stateMu.Lock()
+	defer w.stateMu.Unlock()
+	return w.err
+}
+
+func (w *Watcher) provisionSession() (string, error) {
+	session, err := CreateEventSession(w.ac)
+	if err != nil {
+		return "", w.fail(err)
+	}
+
+	// Rejected here because the dialer's own *url.Error quotes the whole URL, and this
+	// URL carries a live channel token. Only the inner reason is safe to keep.
+	if _, err := url.Parse(session.WebsocketURL); err != nil {
+		var urlErr *url.Error
+		if errors.As(err, &urlErr) {
+			err = urlErr.Err
+		}
+		return "", w.fail(fmt.Errorf("server returned a malformed event channel URL: %w", err))
+	}
+
+	w.stateMu.Lock()
+	w.channelID = session.ChannelID
+	w.stateMu.Unlock()
+
+	return session.WebsocketURL, nil
+}
+
+func (w *Watcher) subscribe() error {
+	w.stateMu.Lock()
+	channelID := w.channelID
+	first := !w.subscribed
+	w.stateMu.Unlock()
+
+	if err := SubscribeEvent(w.ac, channelID, w.eventType, w.targetID); err != nil {
+		return w.fail(err)
+	}
+
+	w.stateMu.Lock()
+	w.subscribed = true
+	w.warned = false
+	// An unread notice describes an outage that is over: leaving it queued would
+	// misreport a healthy stream and crowd out the next outage's notice.
+	select {
+	case <-w.reconnectFailed:
+	default:
+	}
+	w.stateMu.Unlock()
+
+	if !first {
+		select {
+		case w.reconnected <- struct{}{}:
+		default:
+		}
+	}
+
+	return nil
+}
+
+// A failure before any successful subscribe is fatal, so a bad target or an
+// unreachable server surfaces the server's own message rather than a generic connect
+// timeout. Session creation counts: before a first success nothing proves the request
+// valid. Later failures are ordinary reconnect material.
+func (w *Watcher) fail(cause error) error {
+	w.stateMu.Lock()
+	fatal := !w.subscribed
+	if fatal {
+		w.err = cause
+	}
+	announce := !fatal && !w.warned
+	w.stateMu.Unlock()
+
+	if fatal {
+		w.Stop()
+		return cause
+	}
+
+	if announce {
+		// Never park the dial loop on an undrained notice; warned is set only on a
+		// delivered one, so a dropped notice is re-announced next attempt.
+		select {
+		case w.reconnectFailed <- cause:
+			w.stateMu.Lock()
+			w.warned = true
+			w.stateMu.Unlock()
+		default:
+		}
+	}
+
+	return cause
+}
+
+func (w *Watcher) handleMessage(raw []byte) {
+	// Copied because raw is handed to another goroutine through frames.
+	frame := make([]byte, len(raw))
+	copy(frame, raw)
+
+	select {
+	case w.frames <- frame:
+	case <-w.done:
+	}
+}
--- a/api/event/watcher.go
+++ b/api/event/watcher.go
@@ -86,9 +86,13 @@ func (w *Watcher) Err() error {
 }
 
 // The session request carries neither --type nor --target, so only a 4xx proves retrying
-// pointless. Anything else gets the retry window a dial failure already gets.
+// pointless—except 408 and 429, which ask for exactly that. Anything else gets the retry
+// window a dial failure already gets.
 func (w *Watcher) sessionCreateFailed(cause error) error {
-	if status := utils.HTTPStatusCode(cause); status >= http.StatusBadRequest && status < http.StatusInternalServerError {
+	status := utils.HTTPStatusCode(cause)
+	retryLater := status == http.StatusRequestTimeout || status == http.StatusTooManyRequests
+
+	if !retryLater && status >= http.StatusBadRequest && status < http.StatusInternalServerError {
 		return w.fail(cause)
 	}
 
--- a/api/event/watcher_test.go
+++ b/api/event/watcher_test.go
@@ -1,0 +1,332 @@
+package event
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// newWatcherTestServer serves the session, subscription, and WebSocket endpoints on one
+// httptest server so a Watcher can run end to end. shouldUpgrade and subscribeStatus each
+// receive their own 1-based attempt count, in the order the runtime reaches them.
+func newWatcherTestServer(t *testing.T, shouldUpgrade func(attempt int32) bool, subscribeStatus func(attempt int32) int, wsHandler func(conn *websocket.Conn, n int32)) (*httptest.Server, *atomic.Int32, *atomic.Int32) {
+	t.Helper()
+
+	var sessions, conns, subscribes atomic.Int32
+	upgrader := websocket.Upgrader{}
+	mux := http.NewServeMux()
+
+	ts := httptest.NewServer(mux)
+
+	mux.HandleFunc(eventSessionsURL, func(w http.ResponseWriter, r *http.Request) {
+		n := sessions.Add(1)
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(EventSessionResponse{
+			ID:           "session",
+			WebsocketURL: "ws" + strings.TrimPrefix(ts.URL, "http") + "/ws/",
+			ChannelID:    fmt.Sprintf("channel-%d", n),
+		})
+	})
+
+	mux.HandleFunc(eventSubscriptionsURL, func(w http.ResponseWriter, r *http.Request) {
+		status := subscribeStatus(subscribes.Add(1))
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(status)
+		if status >= 400 {
+			_, _ = w.Write([]byte(`{"detail":"Invalid input."}`))
+			return
+		}
+		_, _ = w.Write([]byte(`{"id":"sub"}`))
+	})
+
+	mux.HandleFunc("/ws/", func(w http.ResponseWriter, r *http.Request) {
+		n := conns.Add(1)
+		if !shouldUpgrade(n) {
+			return
+		}
+		conn, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			return
+		}
+		wsHandler(conn, n)
+	})
+
+	t.Cleanup(ts.Close)
+	return ts, &sessions, &subscribes
+}
+
+func alwaysCreated(int32) int { return http.StatusCreated }
+
+func alwaysUpgrade(int32) bool { return true }
+
+func alwaysRejected(int32) int { return http.StatusBadRequest }
+
+func TestWatcher_ForwardsFramesVerbatim(t *testing.T) {
+	frame := `{"event_type":"work_session","payload":{"category":"status","sub_type":"approved","unknown":1}}`
+
+	ts, _, _ := newWatcherTestServer(t, alwaysUpgrade, alwaysCreated, func(conn *websocket.Conn, _ int32) {
+		_ = conn.WriteMessage(websocket.TextMessage, []byte(frame))
+		for {
+			if _, _, err := conn.ReadMessage(); err != nil {
+				return
+			}
+		}
+	})
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	w := NewWatcher(ac, "work_session", "ws-uuid")
+	w.Start()
+	defer w.Stop()
+
+	require.True(t, w.WaitConnected(3*time.Second))
+	require.NoError(t, w.Err())
+
+	select {
+	case got := <-w.Frames():
+		assert.JSONEq(t, frame, string(got))
+	case <-time.After(3 * time.Second):
+		t.Fatal("no frame received")
+	}
+}
+
+func TestWatcher_CreatesANewSessionOnReconnect(t *testing.T) {
+	ts, sessions, subscribes := newWatcherTestServer(t, alwaysUpgrade, alwaysCreated, func(conn *websocket.Conn, n int32) {
+		if n == 1 {
+			// The token is single-use, so recovering means a whole new session.
+			_ = conn.Close()
+			return
+		}
+		for {
+			if _, _, err := conn.ReadMessage(); err != nil {
+				return
+			}
+		}
+	})
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	w := NewWatcher(ac, "work_session", "ws-uuid")
+	w.reconnectBaseDelay = testReconnectBaseDelay
+	w.Start()
+	defer w.Stop()
+
+	require.True(t, w.WaitConnected(3*time.Second))
+
+	select {
+	case <-w.Reconnected():
+	case <-time.After(5 * time.Second):
+		t.Fatalf("expected a reconnect, sessions=%d subscribes=%d", sessions.Load(), subscribes.Load())
+	}
+
+	assert.GreaterOrEqual(t, sessions.Load(), int32(2), "each attempt must create its own event session")
+}
+
+func TestWatcher_FirstSubscribeFailureStopsAndSurfaces(t *testing.T) {
+	ts, _, _ := newWatcherTestServer(t, alwaysUpgrade, alwaysRejected, func(conn *websocket.Conn, _ int32) {
+		for {
+			if _, _, err := conn.ReadMessage(); err != nil {
+				return
+			}
+		}
+	})
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	w := NewWatcher(ac, "work_session", "not-a-uuid")
+	w.Start()
+	defer w.Stop()
+
+	assert.False(t, w.WaitConnected(3*time.Second), "a rejected subscription must not report connected")
+
+	err := w.Err()
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "failed to subscribe to work_session events")
+}
+
+func TestWatcher_FailureAfterFirstSuccessIsNonFatal(t *testing.T) {
+	firstOnly := func(attempt int32) int {
+		if attempt == 1 {
+			return http.StatusCreated
+		}
+		return http.StatusBadRequest
+	}
+
+	ts, _, subscribes := newWatcherTestServer(t, alwaysUpgrade, firstOnly, func(conn *websocket.Conn, n int32) {
+		if n == 1 {
+			// Dropped only after its subscribe succeeded, so the reconnect is the
+			// attempt that gets rejected.
+			_ = conn.Close()
+			return
+		}
+		for {
+			if _, _, err := conn.ReadMessage(); err != nil {
+				return
+			}
+		}
+	})
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	w := NewWatcher(ac, "work_session", "ws-uuid")
+	w.reconnectBaseDelay = testReconnectBaseDelay
+	w.Start()
+	defer w.Stop()
+
+	require.True(t, w.WaitConnected(3*time.Second))
+	require.NoError(t, w.Err())
+
+	require.Eventually(t, func() bool {
+		return subscribes.Load() >= 2
+	}, 5*time.Second, 10*time.Millisecond, "expected a rejected re-subscribe")
+
+	assert.NoError(t, w.Err())
+	select {
+	case <-w.done:
+		t.Fatal("watcher stopped after a post-success subscribe failure")
+	default:
+	}
+
+	select {
+	case failErr := <-w.ReconnectFailed():
+		assert.ErrorContains(t, failErr, "failed to subscribe to work_session events")
+	case <-time.After(5 * time.Second):
+		t.Fatal("expected a ReconnectFailed notice after a post-success failure")
+	}
+
+	require.Eventually(t, func() bool {
+		return subscribes.Load() >= 3
+	}, 5*time.Second, 10*time.Millisecond, "expected the listener to keep retrying")
+
+	assert.NoError(t, w.Err())
+	select {
+	case <-w.done:
+		t.Fatal("watcher stopped after a post-success subscribe failure")
+	default:
+	}
+}
+
+func TestWatcher_DialFailureAfterFirstSuccessIsAnnounced(t *testing.T) {
+	// A refused handshake is a real dial error, which reaches neither provision nor
+	// subscribe.
+	firstConnOnly := func(attempt int32) bool { return attempt == 1 }
+
+	ts, _, _ := newWatcherTestServer(t, firstConnOnly, alwaysCreated, func(conn *websocket.Conn, _ int32) {
+		_ = conn.Close()
+	})
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	w := NewWatcher(ac, "work_session", "ws-uuid")
+	w.reconnectBaseDelay = testReconnectBaseDelay
+	w.Start()
+	defer w.Stop()
+
+	require.True(t, w.WaitConnected(3*time.Second))
+
+	select {
+	case failErr := <-w.ReconnectFailed():
+		assert.Error(t, failErr)
+	case <-time.After(5 * time.Second):
+		t.Fatal("a dial failure after a first success must be announced")
+	}
+
+	assert.NoError(t, w.Err(), "a dial failure after a first success must not be fatal")
+}
+
+func TestWatcher_StaleFailureNoticeIsDroppedOnRecovery(t *testing.T) {
+	secondOnly := func(attempt int32) int {
+		if attempt == 2 {
+			return http.StatusBadRequest
+		}
+		return http.StatusCreated
+	}
+
+	ts, _, subscribes := newWatcherTestServer(t, alwaysUpgrade, secondOnly, func(conn *websocket.Conn, n int32) {
+		if n == 1 {
+			_ = conn.Close()
+			return
+		}
+		for {
+			if _, _, err := conn.ReadMessage(); err != nil {
+				return
+			}
+		}
+	})
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	w := NewWatcher(ac, "work_session", "ws-uuid")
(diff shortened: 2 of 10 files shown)


cmd/event/watch.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+// Covers one dial (watchHandshakeTimeout in api/event) plus the subscribe that follows.
+const watchConnectTimeout = 15 * time.Second
+
+var watchCmd = &cobra.Command{
+	Use:   "watch --type TYPE [--target ID] [flags]",
+	Short: "Stream events from the Alpacon event channel",
```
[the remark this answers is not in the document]
jisung-02: Fixed in a5236e5 — `Use` is now `watch --type TYPE [--target TARGET_ID]`.

One correction on the reasoning: dropping `[flags]` does not change `--help`. Cobra 1.10.2 `UseLine()` appends `" [flags]"` whenever `HasAvailableFlags()` and the `Use` string does not already contain the token, so the rendered usage is identical either way — verified on the built binary, which still prints `alpacon event watch --type TYPE [--target TARGET_ID] [flags]`. The explicit token was redundant in source, not visible in output.

The consistency point stands, though. Only 6 of ~160 `Use` strings spell `[flags]` out, and the four non-trivial ones (`websh`, `tunnel`, `exec`, `websh records`) place it *before* the positionals precisely because Cobra's append would otherwise land after `COMMAND...`. `watch` is `cobra.NoArgs`, so it has no such reason.

On the placeholder: `TARGET_ID` rather than something type-specific. `--target` names a work session, a command, or a server depending on `--type`, so `SESSION_ID` would be wrong for most values. `TARGET_ID` matches the flag and follows the `SESSION_ID` / `CERT_ID` / `JOB_ID` convention used elsewhere.

cmd/event/watch.go
```diff
@@ -0,0 +1,92 @@
...
+	Long: `Stream events from the Alpacon event channel until interrupted.
+
+One line is written to stdout per event; everything else goes to stderr, so the
+stream stays machine-readable when piped. With --output json each line is the
+server frame verbatim (NDJSON). The default table format prints four fixed
+fields: receive time, event type, sub type, and the target given by --target.
```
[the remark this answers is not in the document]
jisung-02: Agreed and fixed in a5236e5. `renderEvent` runs `json.Compact`, so the emitted line is not byte-identical to the frame, and "verbatim" invited exactly the wrong reading for a machine consumer. The help text now reads:

> With --output json each line is the server frame compacted to a single line (NDJSON)—whitespace is dropped, but every field and the server's key order survive, including fields this CLI does not know.

That states the property the mode actually guarantees: `json.Compact` is a byte filter rather than an unmarshal/marshal round trip, so unknown fields and key order are preserved even though insignificant whitespace is not.

The same loose wording appeared twice in the PR description, so both occurrences were corrected as well. `cmd/event/render_test.go` already asserted the accurate property ("exact byte-for-byte compaction of raw") and needed no change.

api/event/watcher.go
```diff
@@ -0,0 +1,182 @@
...
+func (w *Watcher) ReconnectFailed() <-chan error { return w.reconnectFailed }
+
+// Err returns the error that stopped the Watcher before it ever subscribed.
+func (w *Watcher) Err() error {
+	w.stateMu.Lock()
+	defer w.stateMu.Unlock()
+	return w.err
+}
+
+func (w *Watcher) provisionSession() (string, error) {
```
junho226, line 103: Suggestion: consider making a pre-subscribe `CreateEventSession` failure retryable for 5xx/network, and fatal only for 4xx.

`provisionSession` routes a session-create failure through `fail()`, which is fatal before the first subscribe — while a *dial* failure in the same window is retried (`announceOutage` no-ops when `!subscribed`).

The stated rationale is that "before a first success nothing proves the request valid", but that argument is about the request's content, and `CreateEventSession` carries none of it: no `--type`, no `--target`. The messages that make fail-fast worth having — `work_session_not_found`, `invalid_choice`, `invalid_input` — all come from `SubscribeEvent`.

What the current split actually buys is fast-fail on a 401, which is real. What it costs is that a single transient 502 on `POST /api/events/sessions/` ends the command, where a dial failure at the same instant would have been absorbed by the 15s window. Splitting on the status code keeps the 401 behavior and drops the fragility.

cmd/event/watch.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+
+	eventapi "github.com/alpacax/alpacon-cli/api/event"
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+// Covers one dial (watchHandshakeTimeout in api/event) plus the subscribe that follows.
+const watchConnectTimeout = 15 * time.Second
```
junho226, line 16: Suggestion: name the provision step in this comment.

The comment says the timeout covers "one dial (`watchHandshakeTimeout` in api/event) plus the subscribe that follows", but the window also has to fit `CreateEventSession`'s HTTP round trip ahead of the dial — and on a retried attempt, a backoff step too. 15s over a 10s handshake leaves 5s for two HTTP calls, so the budget is tighter than the comment implies.

```suggestion
// Covers one provision + dial (watchHandshakeTimeout in api/event) plus the subscribe.
```

junho226 (dismissed):
## PR review: #291 — feat(event): add event watch to stream the event channel

**Stack**: golang · **Changes**: +1038 −24 in 10 files

---

### Summary

Adds `alpacon event watch`, and with it the reprovisioning the event channel actually requires: a channel token is single-use and now carries a 60 s TTL, so every dial attempt provisions its own session and re-subscribes. `wsListener` grows `provision` / `onConnected` / `onDialFailed` hooks, `connected` now means *connected and subscribed*, and the deprecated `event` command becomes a pointer to `alpacon exec ls`.

I verified the load-bearing claims rather than taking them from the description, and all of them held (below). Two suggestions, neither blocking.

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (2)

1. **[api/event/watcher.go:86-100]** Consider making a pre-subscribe `CreateEventSession` failure retryable for 5xx/network, fatal only for 4xx.

   `provisionSession` routes a session-create failure through `fail()`, which is fatal before the first subscribe — while a *dial* failure in the same window is retried (`announceOutage` no-ops when `!subscribed`). The stated rationale is that "before a first success nothing proves the request valid", but that argument is about the request's content, and `CreateEventSession` carries none of it: no `--type`, no `--target`. The messages that make fail-fast worth having — `work_session_not_found`, `invalid_choice`, `invalid_input` — all come from `SubscribeEvent`.

   What the current split actually buys is fast-fail on a 401, which is real. What it costs is that a single transient 502 on `POST /api/events/sessions/` ends the command, where a dial failure at the same instant would have been absorbed by the 15 s window. Splitting on the status code keeps the 401 behavior and drops the fragility.

2. **[cmd/event/watch.go:15-16]** The comment says the timeout covers "one dial (`watchHandshakeTimeout`) plus the subscribe that follows", but the window also has to fit `CreateEventSession`'s HTTP round trip ahead of the dial — and on a retried attempt, a backoff step too. 15 s over a 10 s handshake leaves 5 s for two HTTP calls. Worth naming the provision step in the comment so the budget is visible next to the number.

### Highlights

**Verified against the code, not the description**

- **Deprecated-command flag parity holds.** `exec.AddListFlags` registers `--tail`/`-t` (default 25), `--server`/`-s`, `--user`/`-u` — identical names, shorthands, and default to the flags `event` previously declared inline (`cmd/exec/list.go:34-38`). "Takes the same flags" is exact, and `GetEventList` is not orphaned: `cmd/exec/list.go:57` still calls it.
- **stdout/stderr separation holds.** Every `utils.Cli*` helper writes to `os.Stderr` (`utils/log.go:24-62`), so stdout genuinely carries event lines only.
- **The `connected` semantics change is contained.** `connectOnce.Do` now fires after `onConnected`, and both existing listeners reach `wsListener` through `newWSListener`, which leaves `onConnected` nil — so `WaitConnected` is unchanged for them.
- `go vet ./...` clean; `go test -race ./api/event/ ./cmd/event/` passes. The new tests cost nothing measurable — the slow cases in `api/event` are the pre-existing `RunCommandStreaming_*` ones (2-12 s each), not the `Watcher`/`wsListener` additions.

**Design**

- The URL pre-parse is justified precisely, not defensively. gorilla's `Dialer.Dial` calls `url.Parse` itself and returns the resulting `*url.Error`, whose `Error()` quotes the whole URL — which at that moment carries an unconsumed channel token. Catching it first and keeping only `urlErr.Err` is the right narrow fix, and the comment says exactly why.
- Worth recording for the host-validation follow-up in the description: `SetWebsocketHeader` sets only `Origin` and `User-Agent` (`client/client.go:179-185`), no bearer token. So a spoofed `websocket_url` would expose the single-use 60 s channel token, not the API credential. That bounds the follow-up's severity usefully.
- The sanitization split is right in both directions: stripping `unicode.Cf` to defeat a bidi override that could render `denied` as `approved`, mapping `\n`/`\r`/`\t` to spaces *before* trimming so a multi-line server error cannot glue into one word — and deliberately leaving `--output json` unfiltered, because that mode's contract is lossless delivery to a machine.
- `renderEvent` writing nothing at all on failure is the correct choice for a stream something else is parsing, and checking `eventType` *after* stripping catches both an escape-only value and a bare JSON `null` (which would otherwise print as a row of dashes). The comment calls out both.
- The reconnect-notice state machine is careful in a way that is easy to get wrong: `warned` is set only on a *delivered* notice, so a dropped one is re-announced next attempt; a stale notice is drained on re-subscribe so it cannot misreport a stream that has already recovered; and the frame copy in `handleMessage` is explained by the fact that `raw` crosses a goroutine boundary.
- Test coverage matches the risk surface — provision-per-attempt, `onConnected` ordering vs. `connected`, first-failure-fatal, later-failure-non-fatal, stale-notice drop, malformed URL, session-create failure. `TestRenderEvent_JSONIsOneLinePerFrameAndPreservesUnknownFields` pins the property the help text now claims, which is the right pairing after the `verbatim` → `compacted NDJSON` correction.
- Scope discipline is good: `SudoListener` / `CommandOutputListener` were traced rather than assumed, the `command_output` case shown to be bounded by the REST poller, and the `sudo` gap named honestly instead of being swept into this PR.

---

**Verdict**: APPROVED

**Checklist**
- [x] Tests included and passing (`go vet` clean, `-race` green, verified locally)
- [x] No security vulnerabilities (sanitization and token-leak paths reviewed)
- [x] Documentation updated (help text, deprecation notice, PR description all consistent with behavior)

api/event/watcher.go
jisung-02: Agreed and fixed in 30e1560. The argument holds: the session request body is `struct{}{}` (`api/event/subscribe.go:42`), so it carries nothing the caller could have gotten wrong, and every message that makes fail-fast worth having comes from `SubscribeEvent`.

`provisionSession` now routes a create failure through `sessionCreateFailed`, which splits on the status:

```go
func (w *Watcher) sessionCreateFailed(cause error) error {
	if status := utils.HTTPStatusCode(cause); status >= http.StatusBadRequest && status < http.StatusInternalServerError {
		return w.fail(cause)
	}

	w.announceOutage(cause)
	return cause
}
```

A 4xx keeps the old path, so the fast 401 survives. Everything else — 5xx, a network error, an unreadable body — goes to `announceOutage`, which is the same helper a dial failure already uses: silent before the first subscribe so the retry window absorbs it, announced on `ReconnectFailed` after. `utils.HTTPStatusCode` walks the wrapped chain, and `withStatus` tags the error at `client/client.go:244`, so the status survives `CreateEventSession`'s `%w`.

Also corrected `fail`'s doc comment, which asserted the opposite ("Session creation counts: before a first success nothing proves the request valid").

One consequence worth naming rather than hiding: a *persistent* 5xx now ends in `Timed out connecting to the event channel after 15s` instead of the server's 502 message, because a retryable pre-subscribe failure is not recorded in `Err()`. That is exactly how a persistent pre-subscribe dial failure already behaves, so the two paths stay symmetric — which was the point of the change. Surfacing the last transient cause on timeout would improve both, but it needs `announceOutage` to change too, so I left it out of this PR rather than fixing one path and not the other.

Tests: `TestWatcher_SessionCreateFailureStopsAndSurfaces` used a 500 to assert fatality, so it would have passed for the wrong reason after this change. It is now `TestWatcher_SessionCreateRejectionStopsAndSurfaces` on a 401, plus a new `TestWatcher_SessionCreateServerErrorIsRetried` that returns 502 on the first attempt and asserts the Watcher connects anyway with a second session created — the transient case from your comment. `newWatcherTestServer` gained a `sessionStatus` hook to make that expressible.

`go vet` clean, `go test -race ./...` green, `golangci-lint` 0 issues.

cmd/event/watch.go
jisung-02: Applied your suggestion verbatim in a64419e:

```go
// Covers one provision + dial (watchHandshakeTimeout in api/event) plus the subscribe.
```

The comment was understating the budget in two ways you named: `CreateEventSession`'s round trip sits ahead of the dial, and a retried attempt pays a backoff step on top. 15s over a 10s handshake really does leave 5s for two HTTP calls, and that is worth being visible next to the constant — more so now that 30e1560 makes a 5xx on the provision step retryable inside the same window.

api/event/watcher.go
```diff
@@ -0,0 +1,194 @@
...
+func (w *Watcher) Err() error {
+	w.stateMu.Lock()
+	defer w.stateMu.Unlock()
+	return w.err
+}
+
+// The session request carries neither --type nor --target, so only a 4xx proves retrying
+// pointless. Anything else gets the retry window a dial failure already gets.
+func (w *Watcher) sessionCreateFailed(cause error) error {
+	if status := utils.HTTPStatusCode(cause); status >= http.StatusBadRequest && status < http.StatusInternalServerError {
```
junho226: Suggestion (non-blocking): exclude 429 and 408 from the fatal band.

The comment says "only a 4xx proves retrying pointless", but 429 (and 408) are the 4xx that prove the opposite — they *are* retry-later signals, and under this band they end the command.

Reachability, checked rather than assumed: the server's only throttle today is `ServiceTokenRateThrottle`, whose `get_cache_key` returns `None` for JWT / APIToken / Session auth, so `POST /api/events/sessions/` cannot 429 a CLI credential as things stand. Two ways it becomes live: a proxy or WAF in front of the API, which emits 429 on its own, and alpacon-server#1319 (API throttling), which is open. A long-running `event watch` is exactly the client shape that draws a rate limit.

Fine as a follow-up, or to leave until #1319 lands — the band is not wrong today, just narrower than the comment claims.

junho226 (dismissed):
## Re-review: delta since 5a7087d

My earlier approval was dismissed by the new push, so this is a delta pass over `5a7087d..a64419e` (+65 −19 in 3 files), not a re-read of the whole PR.

### Both suggestions resolved

**1. Retryable pre-subscribe session-create failure (30e1560)** — resolved, and I verified the mechanism rather than taking it from the reply.

- `utils.HTTPStatusCode` walks the unwrap chain and returns 0 for an untagged error (`client/client_test.go:123-124`), so a network error joins the retry path rather than the fatal one.
- `withStatus` tags at `client/client.go:218`, `:226`, `:244`, so both the `checkAuthStatus` 401 and a 502 that arrives as HTML carry a readable status.
- The retry is real and single: `connectAndListen` returns false on a provision error *without* calling `onDialFailed` (`api/event/wslistener.go:146-149`), so there is no double-announce, and `listenLoop` retries after backoff without resetting it.
- It is bounded: `WaitConnected` false exits at `cmd/event/watch.go:66-73`, so a persistent 5xx cannot loop forever.

The test pair is sound in a way worth naming. `TestWatcher_SessionCreateServerErrorIsRetried` would pass even if the status never got tagged, since 0 also retries — but `TestWatcher_SessionCreateRejectionStopsAndSurfaces` on 401 fails in that case. The two together pin both the tagging and the retry; either alone would not.

The consequence you named — a persistent 5xx ending in the generic timeout instead of the server's message — is the right call to defer. Fixing one path and not the other would cost more than the symmetry you kept.

**2. Provision step named in the timeout comment (a64419e)** — applied.

### Suggestion (1, non-blocking)

**[api/event/watcher.go:88-91]** The comment says "only a 4xx proves retrying pointless", but 429 (and 408) are the 4xx that prove the opposite — they *are* retry-later signals, and under this band they end the command.

Reachability, checked rather than assumed: the server's only throttle today is `ServiceTokenRateThrottle`, whose `get_cache_key` returns `None` for JWT / APIToken / Session auth, so `POST /api/events/sessions/` cannot 429 a CLI credential as things stand. Two ways it becomes live: a proxy or WAF in front of the API, which emits 429 on its own, and alpacon-server#1319 (API throttling), which is open. A long-running `event watch` is exactly the client shape that draws a rate limit.

Cheap version, keeping everything else: exclude 429/408 from the fatal band. Fine as a follow-up, or to leave until #1319 lands — the band is not wrong today, just narrower than the comment claims.

### Verified

- `go vet ./...` clean; `go test -race -run 'TestWatcher|TestRenderEvent|TestWSListener' ./api/event/ ./cmd/event/` green.
- The test helper's new hook order (`sessionStatus, shouldUpgrade, subscribeStatus, wsHandler`) matches the runtime order its doc comment claims.

---

**Verdict**: APPROVED

api/event/watcher.go
jisung-02: Fixed in 671d4bf rather than deferred — the fix is two constants and the reachability you traced is not gated on #1319.

```go
status := utils.HTTPStatusCode(cause)
retryLater := status == http.StatusRequestTimeout || status == http.StatusTooManyRequests

if !retryLater && status >= http.StatusBadRequest && status < http.StatusInternalServerError {
	return w.fail(cause)
}
```

The comment was the part that was actually wrong, so it now says so: "only a 4xx proves retrying pointless—except 408 and 429, which ask for exactly that."

On reachability, I take your read: `ServiceTokenRateThrottle.get_cache_key` returning `None` for JWT, APIToken, and Session auth means the server cannot throttle a CLI credential on this endpoint today, but a proxy or WAF emits 429 without any server change, so this was never only about #1319 landing.

Tests: `TestWatcher_SessionCreateServerErrorIsRetried` is now table-driven as `TestWatcher_SessionCreateRetryableStatusIsRetried` over 502, 408, and 429. I checked the two new rows fail without the fix — with `watcher.go` stashed, `request_timeout` and `too_many_requests` fail on `WaitConnected` while `server_error` still passes, so they are not green for the wrong reason. `TestWatcher_SessionCreateRejectionStopsAndSurfaces` still pins the 401 as fatal.

Two things I did not do:

`Retry-After` is ignored. The listener owns its own backoff schedule, and threading a header out of `CreateEventSession` and into the dial loop's delay is a bigger change than this one.

`api/ftp/ftp.go:550` has the same shape inside `cp`'s download retry loop — `if resp.StatusCode >= 400 && resp.StatusCode < 500 { return ... }`, commented "Client errors (4xx) will never succeed on retry" — so a throttled download aborts instead of retrying. Pre-existing and in a different command, so I left it and named it in the PR body instead. Happy to fix it here if you would rather it not sit.

`go vet` clean, `go test -race ./...` green, `golangci-lint` 0 issues. 671d4bf also builds and vets on its own.

api/event/watcher_test.go
```diff
@@ -0,0 +1,380 @@
...
+	mux := http.NewServeMux()
+
+	ts := httptest.NewServer(mux)
+
+	mux.HandleFunc(eventSessionsURL, func(w http.ResponseWriter, r *http.Request) {
+		n := sessions.Add(1)
+		w.Header().Set("Content-Type", "application/json")
+		if status := sessionStatus(n); status >= 400 {
+			w.WriteHeader(status)
+			_, _ = w.Write([]byte(`{"detail":"Bad gateway."}`))
```
junho226: Nit (cosmetic): this detail now describes one of the three rows.

`TestWatcher_SessionCreateRetryableStatusIsRetried` runs 502, 408 and 429 through this handler, and all three get `"Bad gateway."`. Nothing asserts on the body — `TestWatcher_SessionCreateRejectionStopsAndSurfaces` matches on `failed to create event session` — so this is presentation only, and it only shows up when a failure gets printed while debugging a row.

```suggestion
			_, _ = w.Write([]byte(fmt.Sprintf(`{"detail":"Session create failed with %d."}`, status)))
```

junho226 (dismissed):
## Re-review: delta since a64419e

Delta pass over `a64419e..671d4bf` (one commit, +41 −23 in 2 files), not a re-read of the whole PR — my earlier approval was dismissed by the push.

### The remaining suggestion is resolved (671d4bf)

The band now reads `!retryLater && 4xx`, so 408 and 429 get the same retry window a dial failure already gets, and the comment states the exception instead of overstating the rule.

I re-ran your "not green for the wrong reason" check rather than taking it from the reply. With the `retryLater` line removed from `sessionCreateFailed`, `request_timeout` and `too_many_requests` fail on `WaitConnected` while `server_error` still passes, and `TestWatcher_SessionCreateRejectionStopsAndSurfaces` stays green. The two new rows pin the exclusion, the 401 row pins the tagging, and neither set alone would do both.

Worth recording, because it is the case that actually matters: the reachable 429 is the proxy/WAF one, and that shape carries no JSON body — so it never reaches `parseAPIError`. It still arrives tagged. `readJSONResponse` rejects the content type at `client/client.go:226` and wraps with `withStatus(..., 429)`, so `HTTPStatusCode` finds 429 on the unwrap chain even though the error text is "unexpected response from server (HTTP 429, Content-Type: text/html)". The fix covers the path it was written for, not only the JSON path the test exercises.

Post-subscribe behavior is unchanged and consistent: 408/429 reach `announceOutage`, which is notice-only while subscribed, and the dial loop retries on its existing schedule. The fatal-before-first-subscribe contract did not move.

### On `api/ftp/ftp.go:550` — follow-up, and not only for scope

You asked whether to fix it here. My answer is no, and the reason is stronger than "different command".

It is not the same two-constant change there. `fetchFromURLToFile` retries with a flat `time.Sleep(time.Second)` (`api/ftp/ftp.go:559`) and `maxAttempts` reaches it as **100** from `api/ftp/ftp.go:726`, passed through `saveDownloadedURL`. Adding 429 to that function's retryable set without touching the schedule turns a throttle into up to 100 requests one second apart against the thing that just asked the client to slow down — a worse outcome than today's abort, not a better one.

Compare the path you just fixed: base 1 s doubling (`wsReconnectBaseDelay`, `nextReconnectDelay`) under a 15 s `watchConnectTimeout` ceiling, so a persistent 429 costs about five session-create attempts and then exits. That bound is what makes the event-side change safe on its own. The ftp side has no equivalent bound, so the equivalent fix needs a backoff — and at that point `Retry-After` stops being a separate nicety and becomes the natural input to it.

So it is a real defect of the same class, and it wants its own PR: a schedule change plus a table row on `TestFetchFromURL_ClientErrorNoRetry` (`api/ftp/ftp_test.go:779`), which already pins 403 → exactly one request and is the right place for 429 → retried. That is also the call I made on alpamon#392's write-only `RequestID` — correct change, separate PR — so I would rather be consistent about it than take this one inline.

Deferring `Retry-After` on the event path is right for the same reason it is load-bearing on the ftp path: here the 15 s window already bounds the damage, and the listener owning its own schedule is the seam you would have to break to thread a header through.

### Nit (1, cosmetic)

**[api/event/watcher_test.go:36]** The shared session handler answers every `>= 400` with `{"detail":"Bad gateway."}`, which now describes one of the three table rows. Nothing asserts on it, so this is presentation only.

### Verified

- `go vet ./...` clean; `go test -race -run TestWatcher ./api/event/` green.
- Fatal band re-derived by hand: 408/429 → retry, other 4xx → `fail`, 5xx and untagged (0) → retry. Matches the comment as now written.

---

**Verdict**: APPROVED

api/event/watcher_test.go
jisung-02: Applied in f06bcad, with one deviation from the suggestion as written:

```go
_, _ = fmt.Fprintf(w, `{"detail":"Session create failed with %d."}`, status)
```

`w.Write([]byte(fmt.Sprintf(...)))` verbatim fails the lint gate — staticcheck `QF1012`, "Use fmt.Fprintf(...) instead of Write([]byte(fmt.Sprintf(...)))". Same change, past the linter. Rendered bodies confirmed as `Session create failed with 502.` / `408.` / `429.`

I checked the two sibling handlers for the same drift rather than only the flagged line. Both are accurate as written: the subscribe handler's `{"detail":"Invalid input."}` is only ever reached with a 400 (`alwaysRejected`, `firstOnly`, `secondOnly` all return `StatusBadRequest`), and `TestWatcher_SessionCreateRejectionStopsAndSurfaces` has its own server that pairs 401 with `{"detail":"Invalid token."}`. Neither needed a change.

On `api/ftp/ftp.go:550` — taking your no, and the reason is better than my scope argument. `maxAttempts` reaching `fetchFromURLToFile` as 100 from `api/ftp/ftp.go:726` with a flat `time.Sleep(time.Second)` means adding 429 to the retryable set without touching the schedule turns one throttle into 100 requests a second apart at the thing that just asked for less traffic. That is worse than today's abort, and it is the part I had not worked out when I offered to fix it inline. The bound you name on this side — base 1s doubling under the 15s `watchConnectTimeout` ceiling, so a persistent 429 costs about five attempts then exits — is what makes the event-side change safe alone, and the ftp side has no equivalent. So: separate PR, backoff first, `Retry-After` as its input, and the 429 row on `TestFetchFromURL_ClientErrorNoRetry` next to the existing 403 one. Tracking it separately rather than here.

Also useful, and something my own reply had not established: the tagging on the bodyless proxy/WAF 429 through `readJSONResponse`'s content-type rejection at `client/client.go:226`. That is the reachable shape, and it was covered by accident rather than by the test I wrote — worth having on the record.

`gofmt` clean, `go vet ./...` clean, `go test -race ./...` green, `golangci-lint` 0 issues. `f06bcad` builds and vets on its own.

cmd/event/watch.go
```diff
@@ -0,0 +1,96 @@
...
+	watchCmd.Flags().String("target", "", "Target resource ID; omit where the server allows a subscription without one")
+
+	EventCmd.AddCommand(watchCmd)
+}
+
+func runWatch(cmd *cobra.Command, _ []string) {
+	eventType, _ := cmd.Flags().GetString("type")
+	target, _ := cmd.Flags().GetString("target")
+
+	if eventType == "" {
```
junho226, line 53: Suggestion: the manual presence check deviates from the repo's `MarkFlagRequired` convention (`cmd/token/acl_file_add.go:33`, `cmd/websh/websh_join.go:40`). `CliErrorWithExit` also appends the "report an issue" footer, which miscasts a usage mistake as an internal error — cobra's required-flag error prints usage instead.

```go
_ = watchCmd.MarkFlagRequired("type")
```

api/event/watcher.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
+)
+
+const (
+	watchHandshakeTimeout = 10 * time.Second
+	watchFrameBuffer      = 256
+
+	// Bounded so a compromised server cannot grow the CLI's memory without limit.
+	watchFrameLimit = 1024 * 1024
```
junho226, line 20: Suggestion: the 1 MiB `watchFrameLimit` has no test. An oversized frame makes gorilla close the connection, which should surface as an ordinary announced reconnect rather than a fatal stop — a test pinning that would protect the invariant when the read loop is next touched.

api/event/watcher.go
```diff
@@ -0,0 +1,198 @@
...
+func (w *Watcher) Err() error {
+	w.stateMu.Lock()
+	defer w.stateMu.Unlock()
+	return w.err
+}
+
+// The session request carries neither --type nor --target, so only a 4xx proves retrying
+// pointless—except 408 and 429, which ask for exactly that. Anything else gets the retry
+// window a dial failure already gets.
+func (w *Watcher) sessionCreateFailed(cause error) error {
```
junho226, line 91: Suggestion (non-blocking design question): after a first success, a session-create 401/403 (revoked credential) is announced once and then retried silently forever — a permanently dead watch that looks identical to a quiet one once the single warning scrolls by. Deliberate per the "ordinary reconnect material" rule, and a transient 401 during token refresh argues against fatal; but auth-class statuses post-success may deserve either a fatal exit or a periodic re-announce. Fine to defer to PR5's exit-code work.

junho226 (approved):
## Summary

PR4 of the #271 push-notification series. Adds `alpacon event watch`, streaming the Alpacon event channel until interrupted. Because event channel tokens are single-use (60s TTL since alpacon-server #2760), the PR extends `wsListener` with `provision`/`onConnected`/`onDialFailed` hooks so every dial attempt provisions a fresh session and re-subscribes — the two legacy listeners keep nil hooks and are behaviorally untouched. Rendering is NDJSON (byte-exact `json.Compact`, unknown fields and key order preserved) or a four-field table, with ANSI/control/bidi stripping on everything that reaches a TTY.

Verified locally: `gofmt -l` clean, `go vet ./...` clean, `go test -race -count=1 ./api/event/ ./cmd/event/` both pass.

### Critical issues (0)

None.

### Warnings (0)

None. The concurrency design was checked specifically: all hook callbacks (`provision`, `onConnected`, `onDialFailed`) run on the single listener goroutine, `Stop()` does not wait on `stopped` so `fail()`'s fatal path cannot self-deadlock, and the non-blocking sends on `reconnected`/`reconnectFailed` with the `warned` re-arm logic are sound.

### Suggestions (3)

Posted inline — `MarkFlagRequired` convention, a missing `watchFrameLimit` test, and a non-blocking design question on post-success auth-class session-create failures.

### Highlights

- The fatal/retryable session-create split is table-driven with rows (408, 429) that fail without the fix — not passing for the wrong reason.
- Token-leak awareness throughout: pre-dial `url.Parse` validation keeps the channel token out of `*url.Error`, with a regression test asserting the token never appears.
- The byte-for-byte NDJSON compaction assertion would catch an unmarshal-into-map regression that alphabetizes keys — a failure mode the order-insensitive assertions above it would miss.
- `connected` closing only after `onConnected` returns makes `WaitConnected` mean "connected *and* subscribed", and the test pins that ordering.
- stdout/stderr separation is real: every `Cli*` helper writes to stderr, so stdout stays NDJSON-parseable.

---

**Verdict**: APPROVED

- [x] Code follows stack conventions (gofmt/vet/golangci clean; one minor `MarkFlagRequired` deviation noted inline)
- [x] Tests included and passing (`go test -race` verified locally)
- [x] No security vulnerabilities (token-leak prevention, frame cap, TTY sanitization all present)
- [x] Documentation updated (command help + deprecation notice repointed)

cmd/event/watch.go
jisung-02: You are right about the convention — `MarkFlagRequired` is what the rest of the repo does, seven call sites including the two you named. I would rather not argue that away.

One detail worth putting on the record before we pick a fix, because it changes what the change actually buys. The footer survives either way: `cmd/root.go:73` routes every error `Execute()` returns through `utils.CliErrorWithExit`, so cobra's required-flag error picks up the same "report an issue" line. Measured on an existing command rather than reasoned about:

```
$ alpacon token acl file add sometoken
...usage...
Error: While executing the command: required flag(s) "action", "path" not set
For issues, check the latest version or report on https://github.com/alpacax/alpacon-cli/issues
```

So the gain is the usage block and the convention alignment — both real — but not the miscast-as-internal-error part, and the exit code is 1 on both paths regardless. If we want the footer gone for usage mistakes it is a `FlagErrorFunc` or a `SilenceErrors` change at the root, which is a wider change than this line and touches every command.

Given that, I would rather not push it here and dismiss your approval for it. Carrying it as a follow-up alongside the root-level footer question, where the two can be decided together.

api/event/watcher.go
jisung-02: Agreed that the invariant is worth pinning, and the framing is the useful part — the thing to assert is that an oversized frame reads as an ordinary announced reconnect, not a fatal stop.

One note on where the risk actually sits, which is why I would rather take it as a follow-up than squeeze it in here. The half that could regress is `connectAndListen` returning true on a read error, which is pre-existing `wsListener` behavior and unchanged by this PR; what is new is only the `SetReadLimit` call that turns an oversized frame into a read error. So the test would be guarding a seam this PR did not move, and it needs a WebSocket server writing a >1 MiB frame — enough setup that I would rather give it its own review than append it to an approved branch.

Happy to write it as a follow-up, next to the read-loop tests rather than the watcher ones, since that is where the behavior lives.

api/event/watcher.go
jisung-02: Taking your own suggestion here and deferring it to PR5, for the reason you gave plus one more.

The symptom you describe is real and I do not want to write it off: a revoked credential after a first success is announced once and then retried silently, and once that single warning scrolls past, a permanently dead watch is indistinguishable from a quiet one. That is the honest cost of the "ordinary reconnect material" rule.

The reason to leave it is that both plausible fixes are exit-code decisions, not watcher decisions. Exiting fatally on a post-success 401 needs a code that says "credential revoked" rather than the generic 1 this PR uses, and a periodic re-announce needs a rule for how a long-lived stream reports degraded-but-alive — which is the same question `wait` has to answer for terminal states. PR5 and #253 own that contract, so deciding it here would likely get revisited there.

Your counter-argument against fatal is also the one I keep landing on: a transient 401 during a token refresh is exactly the shape that would kill a healthy watch, and telling it apart from a revocation is not something the status code alone does.

Carrying it into PR5's scope rather than resolving it on this branch.

PR #292, opened 2026-07-31
feat(event)!: add event wait to block until one event decides

### LINKED ISSUE #271: feat(event): add generic event watch/wait commands for push-based agent notification

jisung-02:
## Background

`alpacon event watch` (#291) streams events but never ends on its own, so nothing could block on an outcome. Waiting on an approval was only possible as a side effect of `work-session create --wait`, which polls REST and only works for a session it just created — there was no way to wait on a session that already exists, and no machine-readable result beyond prose on stderr.

This is the last PR for #271. It adds `alpacon event wait`: subscribe, block until one event decides the wait, print it, exit with a code that carries the outcome.

## Changes

`alpacon event wait --type TYPE [--target ID] [--until SUBTYPE,...] [--timeout DURATION]`

The wait loop in `api/event/waiter.go` knows nothing about event types. Which sub types end a wait, and how to read the current state over REST, are injected through `WaitOptions`. The per-type knowledge lives in `cmd/event/wait_adapter.go`, which currently has one entry (`work_session`). Any other type works through `--until`.

Because the event channel has no history, an outcome published between the subscribe and the first frame would otherwise be lost. After the subscription stands, `work_session` with a `--target` reads the session once over REST and synthesizes a frame marked `"source": "catch_up"` so a consumer can tell it from a server frame. The read repeats after every reconnect. It runs off the wait loop: the REST call has no deadline of its own, so inline it would outlast `--timeout` and ignore `Stop`.

The event channel and the REST projection disagree in one place — the channel publishes `activated` where the resource reports `active` — so `workSessionStatusSubType` normalizes it. The `work_session` sub types were confirmed against alpacon-server `2cafc0f67` by reading every `publish_work_session_status_change` call site.

`--type` is never validated locally; the server is the authority on which types exist.

`--until` values are trimmed and empty entries dropped before matching. Cobra's `StringSlice` splits on commas but does not trim, so `--until "approved, activated"` would otherwise carry a leading space on the second entry and never equal a sub type — leaving the flag visibly set while the wait had no condition that could end it. Passing `--until` with nothing left after trimming is a usage error, not a silent fall back to the built-in condition.

## Exit codes

New code `6`, and `work-session create --wait` is aligned onto the same contract.

| Code | `event wait` | `work-session create --wait` |
|------|--------------|------------------------------|
| `0` | a success sub type arrived | approved (unchanged) |
| `1` | connect failed, subscription rejected, usage error | polling failed (unchanged) |
| `4` | timed out or interrupted | timed out, outcome still open |
| `6` | a failure sub type arrived | rejected, expired, revoked, cancelled, completed |

Usage errors sit in `1`, not `2`. The README table documents `2` for usage errors, but nothing implements it: `utils/error.go` has no such constant, `os.Exit(2)` appears nowhere in production code, and `utils.CliUsageErrorEnvelopeWithExit` exits `1` at its ~35 call sites. `cmd/event/watch.go` already validates `--type` the same way and exits `1`. Returning `2` here would make `event wait` the only surface in the CLI that does. The documentation gap is repo-wide and predates this PR — filed as #302. Under `--output json` the `usage_error` envelope already distinguishes these from network and server failures.

Under `--output json`, `4` emits the same `{"status":"pending_approval", ...}` envelope every other surface returning `4` emits.

`6` differs by command. `event wait` prints the deciding event and nothing else — the frame's `payload.sub_type` is the structured refusal, and adding an envelope would put a second JSON object on stdout for one invocation, breaking the one-line contract it shares with `event watch`. `work-session create --wait` has no event to print, so it emits the standard error envelope with `exit_code: 6`.

## Breaking changes

Both affect `work-session create --wait` only:

- rejected / expired / revoked / cancelled / completed: `1` → `6`
- timeout with the session still pending: `1` → `4`

Neither changes a success path, and both replace the general-error code with one that says what actually happened.

## Testing

`go test -race ./...` passes. `api/event/waiter_test.go` drives a real WebSocket over `httptest` to cover catch-up ordering (it must not run before the subscription stands), catch-up deciding without an event, frame classification, malformed frames not ending a wait, `Stop`, reconnect re-running catch-up, and a rejected subscription.

Verified live against a workspace: approval delivered over the WebSocket (exit `0`, no `catch_up` marker), catch-up match (exit `0`), timeout (exit `4`), interrupt (exit `4`), and a rejected subscription (exit `1`). `work-session create --wait` was checked separately for the `exit_code: 6` envelope.

Not verified live: a `work_session` failure transition delivered over the WebSocket. The test workspace auto-activates every session, so catch-up always decides first. That branch is covered by the unit test, which writes a real frame over the socket.

## Checklist

- [x] `go build ./...`, `go vet ./...`, `gofmt` clean
- [x] README exit-code table updated, and the abbreviated lists in `CLAUDE.md` and `.github/copilot-instructions.md` completed — both declared a stable convention but stopped at `3`
- [x] Title carries the `!` breaking marker, so `--generate-notes` surfaces it in the release draft
- [ ] Whoever publishes the release expands the draft with the two specifics under Breaking changes — `--generate-notes` copies titles only

Closes #271


utils/error.go
```diff
@@ -44,6 +44,13 @@ const (
 	// a hard failure (1): scripts and AI agents branch on it to retry when idle
 	// (or re-run with --force) rather than give up.
 	ExitCodeServerBusy = 5
+
+	// ExitCodeNotApproved is the process exit code for an awaited approval that
+	// ended without being granted—rejected, expired, revoked, or cancelled. It is
+	// the counterpart of ExitCodePendingApproval (4): 4 means the outcome is still
+	// open, 6 means it is settled and negative. Scripts and AI agents branch on it
+	// to stop retrying rather than keep re-requesting approval.
+	ExitCodeNotApproved = 6
```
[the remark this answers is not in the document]
jisung-02: Fixed in 484da1b.

`completed` is in `pollForApproval`'s terminal set (`cmd/worksession/worksession_create.go`), in the `event wait` adapter's failure list (`cmd/event/wait_adapter.go`), and in the README exit-code table. The doc comment was the only place the contract disagreed with itself.

cmd/event/wait.go
```diff
@@ -0,0 +1,153 @@
...
+		// Disarmed here, not only in the deferred call above: waiter.Stop can hang closing
+		// the WebSocket, and every exit path below it uses os.Exit, which skips defers—so
+		// this goroutine is the only place a second Ctrl+C is guaranteed to restore the
+		// default terminate disposition before the (possibly stuck) Stop call.
+		signal.Stop(sigChan)
+		waiter.Stop()
+	}()
+
+	go reportOutages(waiter)
+
```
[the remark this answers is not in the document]
jisung-02: Leaving this as is. Both goroutines are bounded by the process, not by `runWait`.

`runWait` is a leaf command's `Run`. Every path out of it ends the process: the `--type`/`--timeout` validation, the client and options failures, the `Wait` error, and both `PrintPendingApproval` branches all call `os.Exit` directly, and the only plain return is the last statement of the function, after which cobra's `Execute` returns and `main` exits. Nothing calls `runWait` in-process and no test drives it.

`reportOutages` on the line above is in the same position and is stronger evidence than the signal goroutine: it is an unbounded `for { select }` with no exit path at all. A `done` channel would have to be threaded through both, and the only thing it would change is which microsecond before process exit the goroutines stop.

If `runWait` ever becomes testable in-process, that requires factoring the `os.Exit` calls out first — the `done` channel belongs with that change, not ahead of it.

cmd/event/wait.go
```diff
@@ -0,0 +1,153 @@
...
+	// A render failure must not change a settled outcome's exit code—renderEvent writes
+	// nothing to stdout on failure, so the failure itself only ever reaches stderr.
+	if renderErr := renderEvent(os.Stdout, frame, utils.OutputFormat, target, time.Now()); renderErr != nil {
+		utils.CliWarning("%s.", renderErr)
+	}
+
+	if outcome == eventapi.OutcomeFailed {
+		os.Exit(utils.ExitCodeNotApproved)
+	}
```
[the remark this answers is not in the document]
jisung-02: The code is intentional, but you caught a real error in the PR description — I have corrected it.

`event wait` shares `watch`'s output contract: exactly one line on stdout, the deciding event, through the same `renderEvent`. Emitting an error envelope on the failure path would put a second JSON object on stdout for a single invocation and break that shape. The refusal is already structured: the frame carries `payload.sub_type` (`rejected`, `expired`, `revoked`, `cancelled`, `completed`), which is strictly more than `exit_code: 6` would tell a consumer.

`work-session create --wait` is different because it has no event to print — the envelope is its only structured output. Same exit code, different payload shape, because they are different kinds of command.

What was wrong was the PR body, which said "6 emits the standard error envelope" without scoping that to `create --wait`. Corrected.

jisung-02:
Replying to the three suppressed comments from the latest review. All three rest on the same premise—that the CLI returns `2` for usage errors—so I checked that first, and it does not hold today.

**`cmd/event/wait.go:82` (and `:89`)** — local flag validation exits 1

Leaving these as they are. There is no usage-error exit code in the implementation: `utils/error.go` defines `ExitCodeWorkSessionDenied` (3), `ExitCodePendingApproval` (4), `ExitCodeServerBusy` (5), and `ExitCodeNotApproved` (6), and no constant for `2`. `os.Exit(2)` does not appear anywhere in production code. The repository's own usage-error helper, `utils.CliUsageErrorEnvelopeWithExit`, exits 1 and says so in its doc comment (`// ... error_code is fixed to UsageErrorCode. Exits(1).`) across roughly 35 production call sites.

The closest precedent is `cmd/event/watch.go:54`, which validates `--type` the same way and exits 1. It is already on `main`. Changing `event wait` alone would make it the only command in the CLI returning `2`, which is a worse position for a script author than the current uniform 1.

**`cmd/event/wait.go:92`** — `resolveWaitOptions` errors

Same reasoning. The characterization is accurate: `resolveWaitOptions` returns exactly one error, `--until is required for %s: this CLI has no built-in end condition for it`, which is indeed a flag-contract failure rather than a network or server failure. But routing it to `2` while every sibling surface returns 1 would not make it more distinguishable in practice.

Worth noting that under `--output json` this is less opaque than it looks: the `usage_error` envelope already carries `error_code`, so a JSON consumer can separate the cases. It is only the bare exit code that is ambiguous.

**`cmd/event/wait.go:27`** — help text says usage errors exit 1

Keeping the help text. It matches what the command actually does, so changing it would introduce the inconsistency rather than remove it.

**On the underlying point**

The contradiction itself is real, and it is a fair catch—the README table has documented `2` since `fa29159` while nothing implements it. That is a repository-wide gap rather than something this PR introduced, and resolving it means moving ~35 call sites from 1 to 2 at once, which is breaking for anyone treating 1 as "any failure". Filed separately as #302 so it can be decided on its own terms rather than settled inside one command.

cmd/event/wait.go
```diff
@@ -0,0 +1,162 @@
...
+	target, _ := cmd.Flags().GetString("target")
+	untilRaw, _ := cmd.Flags().GetStringSlice("until")
+	timeout, _ := cmd.Flags().GetDuration("timeout")
+
+	// The slice reaches here as typed: '--until "approved, activated"' leaves a leading
+	// space on the second entry, which would never match a sub type.
+	until := utils.CompactStrings(untilRaw)
+
+	if eventType == "" {
+		utils.CliErrorWithExit("--type is required.")
```
junho226: Warning: this and the other five error paths in `runWait` (`:78`, `:81`, `:86`, `:91`, `:118`) call `utils.CliErrorWithExit`, which has no JSON branch at all — see `utils/log.go:23-29`. It prints colored prose to stderr and appends the `reportCLIError()` "report an issue on GitHub" footer.

The PR description says:

> Under `--output json` the `usage_error` envelope already distinguishes these from network and server failures.

That envelope comes from `utils.CliUsageErrorEnvelopeWithExit` (`utils/error_envelope.go:61`), which `cmd/event/` never calls — I checked all 33 production call sites of it elsewhere in the CLI. So under `--output json`, exit `1` is the one `event wait` outcome with no machine-readable output, while `0`, `4`, and `6` all emit structured data — on a command whose premise is "a script never has to parse prose". A missing `--type` also gets a "report a bug" footer for what is usually a typo.

`cmd/event/watch.go` behaves the same way, so this is self-consistent with the sibling command and I'm not blocking on it. But the stated reason for leaving it is wrong, and it was the reason given for dismissing a Copilot comment on these lines. Either fix it here (the swap keeps exit `1`) or correct the PR body and fold it into #302.

The suggestion below inlines the operation string; the package convention elsewhere is a bare verb constant (`opCreate = "create"` in `cmd/worksession/worksession.go`), so an `opWait = "wait"` in `cmd/event/event.go` would match better if you take this.

```suggestion
		utils.CliUsageErrorEnvelopeWithExit("wait", "--type is required.")
```

For the record, the other half of the reply to Copilot checks out: `os.Exit(2)` appears nowhere in production code, so returning `2` here would indeed have made `event wait` the only surface that does.

junho226 (commented):
## Summary

Adds `alpacon event wait`, a blocking consumer on top of the `event watch` streaming primitive: subscribe, block until one event decides the outcome, print the deciding frame, exit with a code that carries the result. Introduces exit code `6` (`ExitCodeNotApproved`) and retrofits `work-session create --wait` onto the same contract (breaking: `1` → `6` for settled-negative, `1` → `4` for timeout-still-pending).

The design is well-factored. `api/event/waiter.go` holds a wait loop that knows nothing about event types; all per-type knowledge (which sub types end a wait, how to read current state over REST) is injected via `WaitOptions` and lives in `cmd/event/wait_adapter.go`. The catch-up read — which closes the subscribe→first-publish window that the history-less event channel would otherwise lose — is correctly sequenced *after* the subscription stands, runs off the wait loop so a hung REST call can't outlast `--timeout` or ignore `Stop`, and re-runs on every reconnect.

**Verification run locally on the PR head:**

```
go build ./...   → clean
go vet ./...     → clean
gofmt -l .       → clean
go test -race ./api/event/... ./cmd/event/... ./cmd/worksession/... ./utils/...
  ok  api/event         58.041s
  ok  cmd/event          2.621s
  ok  cmd/worksession    3.892s
  ok  utils              2.525s
```

## Critical issues (0)

None.

## Warnings (1)

**1. `cmd/event/wait.go:73,78,81,86,91,118` — no JSON envelope on any exit-1 path, contradicting the PR description**

The PR body justifies keeping usage errors at exit `1` with:

> Under `--output json` the `usage_error` envelope already distinguishes these from network and server failures.

That is not what the code does. Every error path in `runWait` calls `utils.CliErrorWithExit`, which has no JSON branch at all (`utils/log.go:23-29`) — it prints colored text to stderr and appends the `reportCLIError()` "report an issue on GitHub" footer. The `usage_error` envelope comes from `utils.CliUsageErrorEnvelopeWithExit` (`utils/error_envelope.go:61`), which `cmd/event/` never calls. I verified 33 production call sites of that helper elsewhere in the CLI, so it is the established local-validation pattern.

Concrete effect: `alpacon event wait --output json` with a missing `--type`, a failed connect, or a rejected subscription produces unparseable prose plus a "report a bug" footer for what is often a user typo. Exit `1` therefore stays the one `event wait` outcome with no machine-readable output, while `0`, `4`, and `6` all emit structured data — on a command whose whole premise is "a script never has to parse prose."

`cmd/event/watch.go` behaves the same way, so this is at least self-consistent with the sibling command, and I'm not treating it as a merge blocker. But the stated reason for leaving it is factually wrong, and that reason was used to dismiss a Copilot comment on the same lines. Two options: fix it here (swap the three local-validation calls to `CliUsageErrorEnvelopeWithExit`, which keeps exit `1`), or correct the PR body and fold it into #302 alongside the documented-but-unimplemented `2`.

I did independently confirm the other half of the reply to Copilot: `os.Exit(2)` appears nowhere in production code, so returning `2` here would indeed have made `event wait` the only surface that does.

## Suggestions (2)

**1. `api/event/waiter_test.go:45-49` — dead binding.** `subs` is bound and immediately discarded with `_ = subs`.

**2. `cmd/event/wait.go:91,113` — user-supplied `--type` reaches the terminal unstripped.** Line 118 strips server text with the comment *"a bidi override could render 'denied' as 'approved'"*. The same terminal receives `eventType` verbatim at line 113 and, via `resolveWaitOptions`'s error, at line 91. The stated threat model is server-written text and a user typing their own flag is self-inflicted — but a CI script or AI agent may build `--type` from an untrusted source. Optional.

## Highlights

- **The `api`/`cmd` seam is the right cut**, and `TestResolveWaitOptions_OptionsAreEventAPITyped` explicitly guards it.
- **`startCatchUp` returning `nil`** so a nil channel blocks forever in the select is exactly the idiomatic Go move, and the comment says why.
- **Malformed frames return `""` from `subTypeOf`, which `classify` treats as undecided** — a garbage frame can't end a wait. Covered by a test that writes real bytes over the socket.
- **Typed sentinel errors (`terminalWaitError` / `pendingWaitError`)** rather than string matching on poll results, with the rationale — an agent reading only `$?` would retry a rejected request forever — stated in the type comment and pinned by `TestPollForApproval_APIFailureIsNotTerminal`.
- **Tests drive a real WebSocket over `httptest`** and cover catch-up ordering, reconnect re-running catch-up, `Stop`, and a rejected subscription — not just the happy path.
- **The `active` → `activated` REST/channel vocabulary mismatch is isolated in one named function** with both constants named, and the sub-type sets are traced to a specific alpacon-server commit.
- Comments consistently explain *why*, not *what*.

---

**Verdict**: COMMENTED

- [x] Code follows stack conventions (gofmt/vet clean, errors wrapped, no goroutine leaks beyond process lifetime, cancellation via `done`)
- [x] Tests included and passing (`-race` clean, table-driven, error paths covered)
- [x] No security vulnerabilities (no hardcoded secrets; the pre-existing token-bearing URL redaction in `provisionSession` is preserved)
- [x] Documentation updated (README exit-code table, CLAUDE.md, copilot-instructions.md)
- [x] Breaking changes marked with `!` and enumerated in the body

cmd/event/wait.go
```diff
@@ -0,0 +1,162 @@
...
+		// the WebSocket, and every exit path below it uses os.Exit, which skips defers—so
+		// this goroutine is the only place a second Ctrl+C is guaranteed to restore the
+		// default terminate disposition before the (possibly stuck) Stop call.
+		signal.Stop(sigChan)
+		waiter.Stop()
+	}()
+
+	go reportOutages(waiter)
+
+	utils.CliInfo("Waiting for a %s event (timeout %s). Press Ctrl+C to stop.", eventType, timeout)
```
junho226: Suggestion (optional): `eventType` reaches the terminal verbatim here, and again through `resolveWaitOptions`'s error at `:91`. Line 118 strips server text precisely because "a bidi override could render 'denied' as 'approved'" — the same terminal, the same risk class.

The stated threat model is server-written text, and a user typing their own flag is self-inflicted, so this is genuinely optional. It matters only where `--type` is built from an untrusted source by a CI script or an AI agent.

```suggestion
	utils.CliInfo("Waiting for a %s event (timeout %s). Press Ctrl+C to stop.", strip(eventType), timeout)
```

api/event/waiter_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+func TestWaiter_CatchUpRunsOnlyAfterTheSubscriptionStands(t *testing.T) {
+	var subscribes atomic.Int32
+	var catchUpSawSubscribes atomic.Int32
+
+	ts, _, subs := newWatcherTestServer(t, alwaysCreated, alwaysUpgrade, func(n int32) int {
+		subscribes.Store(n)
+		return http.StatusCreated
+	}, holdOpen)
+	_ = subs
```
junho226: Suggestion: `subs` is bound only to be discarded on the next line. Taking the third return as `_` drops both (deleting just the `_ = subs` line would leave `subs` declared and unused, which does not compile).

```suggestion
	ts, _, _ := newWatcherTestServer(t, alwaysCreated, alwaysUpgrade, func(n int32) int {
		subscribes.Store(n)
		return http.StatusCreated
	}, holdOpen)
```

cmd/event/wait.go
jisung-02: Fixed in f2ef08b — took the first option, since the swap keeps exit `1`.

You were right that the PR body's claim was the wrong part. The four local-validation paths now call `CliUsageErrorEnvelopeWithExit`: `:73` (`--type is required`), `:78` (`--until needs at least one sub type`), `:81` (`--timeout must be positive`), and `:91` (the `resolveWaitOptions` error). Added `opWait = "wait"` in `cmd/event/event.go` following the bare-verb convention — `cmd/username/username.go:10` is the single-constant precedent.

`:86` (connect failure) and `:118` (the `Wait` error) stay on `CliErrorWithExit`. Neither is a usage error, and a failed connect is exactly what the report-an-issue footer is for.

Verified against a built binary rather than by reading:

```
$ alpacon event wait --output json
{
  "ok": false,
  "exit_code": 1,
  "error_code": "usage_error",
  "message": "--type is required.",
  "context": {
    "operation": "wait"
  }
}
$ echo $?
1
```

All four paths checked, including `:91` against a live workspace (`--until is required for notification: ...`). Table mode is unchanged, and the exit code stays `1` on every one, so nothing a script branches on moved.

The PR body sentence is now true as written, so I left it. #302 still covers the documented-but-unimplemented `2`.

cmd/event/wait.go
jisung-02: Taken, in 1d06bab. The CI-script and agent case is the one that decided it: `--type` typed by hand is self-inflicted, but this CLI is also driven by pipelines and coding agents, and a value built from an untrusted source sits at the same trust level as server text.

Extended past the line you marked, since the same value reaches the terminal in four more places: the timeout and interrupt pending messages at `:128`/`:133`, the `resolveWaitOptions` error at `:91` (it embeds `eventType` in its message, so that one is `strip(err.Error())`, matching `:118`), and `cmd/event/watch.go:75`, which has the identical `CliInfo`. `:113`/`:128`/`:133` share one `displayType := strip(eventType)`.

What is deliberately not stripped is the value passed to `NewWaiter`. Stripping there would have the CLI subscribe to something other than what was asked for, and which types exist is the server's call — the command validates no type locally. Display only.

Verified by bytes, not by eye — `--type` given `notif` + U+202E + `ication`:

```
in:   6e6f 7469 66 e280ae 6963 6174 696f 6e
out:  ... 6e6f 7469 6669 6361 7469 6f6e        "notification"
```

The `e2 80 ae` is gone from the output. Checked the error path in both table and JSON mode the same way.

api/event/waiter_test.go
jisung-02: Applied in e713d45, with the third return taken as `_` as suggested — dropping only the `_ = subs` line leaves `subs` declared and unused, which does not compile.

`go test -race ./api/event/...` still passes.

cmd/event/wait.go
```diff
@@ -0,0 +1,166 @@
...
+	if cmd.Flags().Changed("until") && len(until) == 0 {
+		utils.CliUsageErrorEnvelopeWithExit(opWait, "--until needs at least one sub type.")
+	}
+	if timeout <= 0 {
+		utils.CliUsageErrorEnvelopeWithExit(opWait, "--timeout must be positive.")
+	}
+
+	alpaconClient, err := client.NewAlpaconAPIClient()
+	if err != nil {
+		utils.CliErrorWithExit("Connection to Alpacon API failed: %s. Consider re-logging.", err)
```
junho226: Warning: with `f2ef08b` in, exit `1` no longer has one shape under `--output json`. The `Long` help declares three causes for it — connection failed, subscription rejected, usage error — and only the third now emits a structured envelope. This path and `:122` still print prose plus the report-an-issue footer.

Your reply said neither of these is a usage error, and that is right — `CliUsageErrorEnvelopeWithExit` would have been the wrong helper here. The one that fits is `CliErrorEnvelopeWithExit`, the non-usage variant, which is what all 13 call sites in `cmd/worksession/` use for this exact connect-failure message.

It also answers the objection you raised: `CliErrorEnvelopeWithExit` falls through to `CliErrorWithExit` in table mode (`utils/error_envelope.go:41`), so the footer you wanted to keep for a failed connect stays exactly as it is. Only the JSON branch is added.

```suggestion
		utils.CliErrorEnvelopeWithExit(opWait, err, "Connection to Alpacon API failed: %s. Consider re-logging.", err)
```

cmd/event/wait.go
```diff
@@ -0,0 +1,166 @@
...
+	// Stripped for the same reason server text is: --type lands in a terminal, and a CI
+	// script or an agent may build it from a source the caller does not control.
+	displayType := strip(eventType)
+
+	utils.CliInfo("Waiting for a %s event (timeout %s). Press Ctrl+C to stop.", displayType, timeout)
+
+	frame, outcome, err := waiter.Wait()
+	if err != nil {
+		// Stripped because the server wrote it and it lands in a terminal.
+		utils.CliErrorWithExit("%s.", strip(err.Error()))
```
junho226: Warning: this is the `:86` issue with an extra loss. The error here is server-written — a rejected subscription — so `ParseErrorResponse` can pull an `error_code` out of it. `CliErrorWithExit` discards that, which means under `--output json` a script sees exit `1` and prose for the one exit-1 cause the server actually explained.

`CliErrorEnvelopeWithExit` takes `err` precisely to surface that code, and keeps the current table-mode output unchanged.

```suggestion
		utils.CliErrorEnvelopeWithExit(opWait, err, "%s.", strip(err.Error()))
```

junho226 (commented):
## Re-review — all three previous comments addressed

Verified locally at `1d06bab`: `go build ./...`, `go vet`, `gofmt -l` clean, and
`go test -race ./api/event/... ./cmd/event/... ./utils/...` passing.

| Previous comment | Resolution | Check |
|---|---|---|
| Exit-1 paths in `runWait` had no machine-readable output under `--output json` | `f2ef08b` — the four local-validation paths (`:73` `:78` `:81` `:91`) now call `CliUsageErrorEnvelopeWithExit`, with a new `opWait` constant | `opWait = "wait"` matches the bare-verb convention (`opActivate`, `opCreate`, `opGet`). `resolveWaitOptions` makes no network call, so classifying `:91` as `usage_error` is correct — a mixed local/remote helper would have made that a misclassification |
| `--type` reached the terminal verbatim | `1d06bab` — `displayType := strip(eventType)`, extended to the timeout/interrupt messages and `watch.go:75` | Leaving the value passed to `NewWaiter` unstripped is the right boundary: stripping there would subscribe to something other than what was asked for |
| Dead binding in `waiter_test.go` | `e713d45` — third return taken as `_` | Compiles and passes |

### Warnings (1)

**`cmd/event/wait.go:86` and `:122` — exit `1` now has two shapes under `--output json`.**
See the inline comments. Short version: `CliUsageErrorEnvelopeWithExit` was the wrong
helper for these two, as you said, but `CliErrorEnvelopeWithExit` is the right one, and it
falls back to `CliErrorWithExit` in table mode — so the report-an-issue footer you wanted
to keep survives the switch untouched.

### Suggestions (2)

**`cmd/event/watch.go:54` — two sibling commands in one package now report the same
`--type is required.` two different ways.** `wait.go:73` emits an envelope, `watch.go:54`
emits prose. Same flag, same message, same package. This PR already touches `watch.go`, so
an `opWatch` constant is cheap. Leaving `watch` alone is also defensible as pre-existing
behavior — just say so if that is the intent. (Outside the diff, so no inline comment.)

**`README.md:284` — the envelope-scope sentence is now stale.** It reads:

> `work-session` subcommand failures (`create`, `use`, `extend`, ...) also emit a JSON
> error envelope under `--output json`

`event wait` emits one too as of `f2ef08b`. Same class of contract drift as the
`ExitCodeNotApproved` doc comment fixed earlier in this PR.

### Highlights

- Checking that `resolveWaitOptions` is local-only before labelling its failure
  `usage_error`. A network call hidden in there would have made the code wrong.
- The display-vs-protocol split on `strip()`: stripped everywhere it reaches a terminal,
  untouched where it identifies a subscription.
- `render.go:47` already strips the server frame's `event_type` and rejects an
  escape-only value before it can print as a row of dashes, so the stdout path has no gap.
- Not quietly redefining exit `1` vs `2` inside a feature PR, and filing #302 instead.
- Verifying the fixes against a built binary rather than by reading the diff.

---

**Verdict**: COMMENTED — nothing blocking. The warning is worth an answer before this
ships as a public CLI contract; the two suggestions are optional.

cmd/event/wait.go
jisung-02: Applied in 57d5407. You're right that `CliErrorEnvelopeWithExit` resolves both of my earlier objections—it isn't the usage variant, and table mode falls through to `CliErrorWithExit`, so the report-an-issue footer survives unchanged.

Verified with a built binary against an empty HOME: table output is byte-identical to before, JSON emits the envelope with `error_code` omitted (local error, nothing to parse), and the exit code stays `1` in both modes.

One small correction to the figures: the connect-failure message has 12 call sites in `cmd/worksession/`, not 13—all of them on this helper, so the consistency point holds.

cmd/event/wait.go
jisung-02: Applied in 57d5407, as suggested. I confirmed the chain you described: `SubscribeEvent` wraps the client error with `%w`, `client.apiError` implements `ErrorCode()`, and `ParseErrorResponse` walks the unwrap chain—so the code was indeed being discarded here.

Verified live with a rejected subscription (`--type bogus_type`):

```json
{
  "ok": false,
  "exit_code": 1,
  "error_code": "invalid_choice",
  "message": "failed to subscribe to bogus_type events: code: invalid_choice.",
  "context": { "operation": "wait" }
}
```

When `Wait` fails with the local connect-timeout error instead, there is no code to extract and the envelope simply omits `error_code`, so that path stays safe too. Table mode and exit `1` are unchanged on both.

jisung-02:
On the two suggestions from the review body (no inline threads, so answering here):

`watch.go` — taken, in 3ffee59, and extended past the flagged line: after 57d5407, converting only `:54` would have left watch's connect-failure and rejected-subscription paths drifting from wait's in exactly the way this round fixed. All four watch error paths now use the envelope helpers with a new `opWatch` constant. The connect-timeout path has no error to parse, so it passes `nil` and the envelope omits `error_code`. Verified with a built binary: usage and connect-failure paths against an empty HOME, the rejected subscription live (`error_code: invalid_choice`, `operation: watch`); table output and exit `1` unchanged everywhere.

`README.md:284` — taken, in 38651f1. The sentence now names `event wait` / `event watch` alongside the work-session subcommands; both are listed since watch also emits the envelope as of 3ffee59. Grepped for the same sentence elsewhere in the repo—no other occurrence.

junho226 (approved):
## Re-review — resolved, and wider than asked

All three points from the last round landed, and the `watch` one went further than
the single line I flagged.

| Last round | Resolution |
|---|---|
| `wait.go:86`, `:122` — exit `1` had two shapes under `--output json` | `57d5407` — both on `CliErrorEnvelopeWithExit(opWait, err, ...)` |
| `watch.go:54` — a sibling command reported the same `--type is required.` differently | `3ffee59` — added `opWatch` and converted **all four** of `watch`'s exit-1 paths, not just the one named |
| `README.md:284` — the envelope-scope sentence was stale | `38651f1` — extended to `event wait` / `event watch` |

Net effect: `cmd/event/` has **zero** remaining `CliErrorWithExit`. Exit `1` now has one
shape under `--output json` in both commands.

### Verified

- `go build ./...`, `go vet`, `gofmt -l`, and `go test -race ./api/event/... ./cmd/event/... ./utils/...` all clean at `38651f17`
- CI green: Build and Test, golangci-lint, govulncheck, CodeQL, Analyze
- `watch.go:72` passing `nil` for `err` matches existing practice — `worksession_recording.go:44`,
  `worksession_create.go:207` and `:294`, and `username_get.go:33` do the same where there is
  no error to parse a code from
- The streaming loop's `CliWarning` calls are untouched, which is right: they are
  non-fatal diagnostics, not exits, so they are outside the envelope contract

### One correction to my own comment

Your count is the correct one: the connect-failure message has **12** call sites in
`cmd/worksession/`, not 13 — `activate`, `cancel`, `complete`, `create`, `describe`,
`extend`, `list`, `recording`, `revoke`, `timeline`, `update`, `use`. The consistency
argument stands, but the number I quoted was wrong.

Also worth noting you measured the thing I only asserted: I claimed `ParseErrorResponse`
could pull a code out of the rejected-subscription error, and you produced it live —
`error_code: "invalid_choice"` from `--type bogus_type`. That is the right way to close a
review point.

---

**Verdict**: APPROVED — nothing outstanding.

PR #297, opened 2026-07-31
fix(ftp): retry a throttled download and back the retries off

jisung-02:
## Background

`fetchFromURLToFile` (`api/ftp/ftp.go`) fetches the presigned download URL for `alpacon cp` and the bulk archive path. It rejected the whole 4xx range as unretryable:

```go
// Client errors (4xx) will never succeed on retry
if resp.StatusCode >= 400 && resp.StatusCode < 500 {
    return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
}
```

408 and 429 are the two that ask for exactly the retry this refuses. A throttled download fails on the first 429 rather than waiting it out. This is the same band `671d4bf` settled for the event session create; the download path was not updated with it.

The retry delay was also a flat `time.Sleep(time.Second)`, which is not a backoff — it hits a struggling server at a fixed rate regardless of how long it has been failing.

## Changes

408 and 429 now take the retry path. Every other 4xx still returns immediately.

The delay is exponential: 250ms, 500ms, then capped at one second. Both call sites pass `maxAttempts=100`, so the cap has to stay at one second or the worst-case total wait doubles — this keeps it where it was while letting a brief blip recover in a quarter of the time.

The shape already existed in this file as `nextPollInterval`. Extracted it as `backoffDelay(attempt, initial, limit)` and made `nextPollInterval` call it, rather than writing a second copy that differed only in two constants. `nextPollInterval` and `alignedPollDelay` behavior is unchanged, and their existing tests still pass.

## Not addressed

`Retry-After` is ignored. Honoring it means threading a header through a function that currently only sees the status, and no server path emits it today. Worth doing if a proxy or WAF starts setting it.

## Testing

`go build ./...`, `go vet ./...`, `gofmt` clean. `go test -race ./api/ftp/...` passes.

New tests against `httptest`:

- 408-then-200 and 429-then-200 both succeed on the second attempt
- 404 fails after exactly one request
- `backoffDelay` ramp and saturation

## Checklist

- [x] New behavior covered by tests
- [x] Existing poll-interval tests unaffected

### DIFF
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -49,7 +49,7 @@ const (
 // backoffDelay returns initial doubled once per 0-based attempt, capped at limit.
 func backoffDelay(attempt int, initial, limit time.Duration) time.Duration {
 	d := initial
-	for i := 0; i < attempt; i++ {
+	for range attempt {
 		d *= pollBackoffFactor
 		if d >= limit {
 			return limit
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -31,25 +31,37 @@ const (
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
 	pollBackoffFactor   = 2
-	basePollTimeout     = 30 * time.Second
-	perFilePollTimeout  = 10 * time.Second
-	perMBPollTimeout    = 5 * time.Second
+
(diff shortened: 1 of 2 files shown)


jisung-02:
Superseded by #298 — branch renamed to `fix/ftp-retry-backoff`. Copilot's review here is carried over into #298.

Issue #279, opened 2026-07-22
test: modernize assertion syntax and remove mixed testify usage


jisung-02:
## Background

In PR #278, `cmd/server/server_action_test.go:35` asserts a substring is absent with:

```go
assert.False(t, strings.Contains(msg, tt.wantAbsent), "message should not contain %q", tt.wantAbsent)
```

Reviewer feedback noted the idiomatic testify form is `assert.NotContains`, which drops the manual failure message and produces a clearer diff on failure. This was flagged as non-blocking and left as-is in #278 to keep that PR scoped.

## Scope

- Replace `assert.False(t, strings.Contains(...))` with `assert.NotContains(t, ...)` (and the `assert.True(strings.Contains(...))` counterpart with `assert.Contains`) across the test suite.
- Remove any now-unused `strings` imports left behind by the change.
- Sweep the suite for other mixed/roundabout testify usages and normalize to the dedicated helpers.

## Notes

Purely a test-quality/readability change—no production code or coverage impact.

PR #298, opened 2026-08-03
fix(ftp): retry a throttled download and back the retries off

jisung-02:
## Background

`fetchFromURLToFile` (`api/ftp/ftp.go`) fetches the presigned download URL for `alpacon cp` and the bulk archive path. It rejected the whole 4xx range as unretryable:

```go
// Client errors (4xx) will never succeed on retry
if resp.StatusCode >= 400 && resp.StatusCode < 500 {
    return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
}
```

408 and 429 are the two that ask for exactly the retry this refuses. A throttled download fails on the first 429 rather than waiting it out. This is the same band `671d4bf` settled for the event session create; the download path was not updated with it.

The retry delay was also a flat `time.Sleep(time.Second)`, which is not a backoff — it hits a struggling server at a fixed rate regardless of how long it has been failing.

## Changes

| Commit | Change |
| --- | --- |
| `02bd577` | 408 and 429 take the retry path; the flat one-second sleep becomes an exponential backoff |
| `2cdf1af` | modernize the new loop and its test counters |
| `47bbbb4` | `pollBackoffFactor` becomes `backoffFactor`, now that both callers share it |
| `78ec29d` | drain the error body before closing it, so a retry reuses the connection |
| `7a5b0f4` | scope the retry-budget comment to the caller it actually holds for |
| `bc582e9` | the retry check spans the band by `http.Status*` constant, not `400` and `500` |
| `220f116` | finish modernizing the two files this branch touched |
| `ba6f42a` | `backoffDelay` clamps a seed that already exceeds its limit |
| `c2a9788` | the attempt ceiling both call sites pass becomes `downloadMaxAttempts` |
| `0bd8d1c` | the error-body drain shrinks from 64 KiB to 8 KiB |
| `14aa221` | the fatal-4xx check both retry loops copied moves to `utils.IsFatalClientError` |
| `6e42210` | a throttled download stops spending the full retry budget |
| `e6e5af2` | the exhaustion error reports the attempts made, not the budget |
| `a517331`, `8970b87`, `d4750af`, `e1d1f88`, `ccfc8d4`, `ddc52dd` | tests: fold the fatal-4xx cases into one table, name the backoff subtests, cover the 5xx retry and both attempt budgets, and take the budget test off the production backoff |
| `c4a91ed` | keep the retry-delay comments free of the constants below them |

408 and 429 now take the retry path. Every other 4xx still returns immediately.

The delay is exponential: 250ms, 500ms, then capped at one second. A brief blip recovers in a quarter of the time the flat sleep took, and `downloadMaxAttempts = 100` still spans roughly the same ~99s as before for the classes that keep the full budget.

The shape already existed in this file as `nextPollInterval`. Extracted it as `backoffDelay(attempt, initial, limit)` and made `nextPollInterval` call it, rather than writing a second copy that differed only in two constants. `nextPollInterval` and `alignedPollDelay` behavior is unchanged, and their existing tests still pass.

A 429 does not draw that full budget. `throttledMaxAttempts = 10` caps the retry-later class (408 and 429) at ten requests over about 7.75s, instead of a hundred over about 98s. A server that answered "too many requests" is not persuaded by more of them, and without `Retry-After` the right window is a guess — so the cap errs small. The budget is `min(maxAttempts, throttledMaxAttempts)` evaluated per response, which means it can collapse below the attempts already spent: a run that draws 5xx for fifty attempts and then a 429 stops on that 429 and reports fifty-one, not ten.

A non-200 response is drained before it is closed. An unread body keeps `net/http` from reusing the connection, so under throttling every retry would open a new one — the opposite of what backing off is for. The drain is bounded at 8 KiB (`maxDrainedErrorBody`). The bound is a time defense, not a memory one: `client/client.go:50` builds the shared client with no `Timeout`, so an error body that stalls mid-read would block the retry loop indefinitely, and this bound is the only thing that stops it.

`utils.IsFatalClientError` and `utils.IsRetryLaterStatus` now hold the band. `api/ftp/ftp.go` and `api/event/watcher.go` were each spelling out `status >= 400 && status < 500` with their own exemptions, which is the part that drifts. `watcher.go` drops its `net/http` import entirely as a result, and `IsFatalClientError(0)` keeps its old behavior for an error carrying no status.

`backoffDelay` clamps its seed with `min(initial, limit)`. The cap previously applied only inside the doubling loop, so attempt 0 returned `initial` verbatim and the helper did not keep the guarantee its doc states. Neither caller passes such a pair today, so no runtime behavior changes; clamping the seed rather than the result preserves the early return that stops the doubling from overflowing at a large attempt count.

`initialDownloadRetryDelay` and `maxDownloadRetryDelay` are `var`, not `const`, so the attempt-budget test can swap in millisecond values and restore them. That test otherwise spends its entire runtime in `time.Sleep`: 16.80s of the package's 24.4s, none of it in an assertion. `api/event/event.go:30` already handles the same problem the same way.

The modernize commits are confined to the two files this branch already touches: counted `for` loops become `for range`, and the test counters that still used `atomic.AddInt32` / `atomic.LoadInt32` move to `atomic.Int32`, which every other counter in that file already used. The rest of the package is left for #279.

## Not addressed

`Retry-After` is ignored. Honoring it means threading a header through a function that currently only sees the status, and no server path emits it today. `Watcher.sessionCreateFailed` in `api/event/watcher.go` has the same gap, so closing it is a decision about both call sites rather than a local fix. `6e42210` carries a `Directive` trailer recording that `throttledMaxAttempts` is the value a parsed `Retry-After` should feed.

`http.Client.Timeout` is unset on the shared client, which is the real fix for a stalled error body. It also bounds the success-path read in `SaveStreamAtomic`, so its value is a decision about every API call the CLI makes — not something a download-retry fix should settle. The 8 KiB drain bound is the local containment until then.

`503 SlowDown` keeps the full `downloadMaxAttempts` budget. It is S3's own throttle signal, so the narrowing in `6e42210` lands on the proxy/WAF 429 while the storage backend's throttle still draws about a hundred requests. That is pre-existing behavior this branch did not change, and the fix is the same deadline conversion the `Directive` trailer points at: `100` was originally a 100-second budget expressed as a count (`ac13e7e`, next to a since-deleted "may timeout if it exceeds 100 seconds" warning), and this branch is what broke the one-attempt-equals-one-second identity that made the count mean a duration.

## Review

Copilot raised two findings; both were valid and are fixed. The undrained error body on the non-200 path was flagged as connection churn under throttling, now that the loop can run many times — fixed in `78ec29d`. `backoffDelay` was flagged for documenting a cap it did not apply at attempt 0 — fixed in `ba6f42a`.

@geunwoonoh raised six in round 1, all addressed:

| Finding | Resolution |
| --- | --- |
| the error-body drain blocks with no time bound | `0bd8d1c` — 64 KiB to 8 KiB, and the comment names time rather than memory as the exposure |
| a 429 draws the full 100-attempt budget | `6e42210` — `throttledMaxAttempts = 10` |
| the fatal-4xx band is duplicated across two retry loops | `14aa221` — `utils.IsFatalClientError` |
| `100` is a magic literal | `c2a9788` — `downloadMaxAttempts` |
| a new fatal-4xx test duplicates an existing one | `a517331` — folded in as a row |
| the backoff table needs `t.Run`, and 5xx and exhaustion are uncovered | `8970b87`, `d4750af`, `e1d1f88` |

`6e42210` introduced a defect of its own: the exhaustion error printed the budget, which shrinks to `throttledMaxAttempts` when the last response is a 429, so a run that spent fifty attempts on 5xx reported ten. Fixed in `e6e5af2`.

Round 2 approved with three non-blocking suggestions; the two that touch code are done. `e6e5af2` had no test that could distinguish it — both existing rows used a 429-only server, where the printed budget and the attempt count are numerically identical. `ccfc8d4` adds a row that serves 502 ten times and then a 429; reverting `count+1` to `budget` fails it, and swapping `count >= budget-1` for `==` fails it with a nil error after the loop runs to exhaustion and hands a closed body to `SaveStreamAtomic`. `ddc52dd` then takes that row and the existing one off the production backoff.

## Testing

```
gofmt -l .                       no output
go vet ./...                     pass
go build ./...                   pass
golangci-lint run ./...          0 issues
go test -race -count=1 ./...     38 packages ok
```

`api/ftp` runs in 7.8s. It was 24.4s before `ddc52dd`, against 7.46s before this branch.

New tests against `httptest`:

- 408, 429, and 502 each retried once, then 200
- 404 and 403 each fail after exactly one request, leaving no file behind
- the caller budget, the throttle budget, and a throttle budget that collapses below the attempts already spent
- an error body that stalls after `maxDrainedErrorBody` bytes returns rather than blocking
- `backoffDelay` ramp and saturation, plus a seed already over the limit
- `utils.IsFatalClientError` and `utils.IsRetryLaterStatus` across 404, 403, 408, 429, 500, 200, and a missing status

## Checklist

- [x] New behavior covered by tests
- [x] Existing poll-interval tests unaffected

### DIFF
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -33,11 +33,7 @@ const (
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
 
-	// A brief blip recovers sooner than the flat one-second retry did, while
-	// downloadMaxAttempts still spans roughly the same ~99s.
-	initialDownloadRetryDelay = 250 * time.Millisecond
-	maxDownloadRetryDelay     = time.Second
-	downloadMaxAttempts       = 100
+	downloadMaxAttempts = 100
 
 	// A server that answered "too many requests" is not persuaded by more of them,
 	// and without Retry-After the window is a guess—so this errs small.
@@ -53,6 +49,14 @@ const (
 
 	bulkUploadConcurrency = 4 // uploads transfer payload bytes, keep low
 	bulkPollConcurrency   = 8 // status polls are lightweight, allow more
+)
+
+// A brief blip recovers sooner than the flat one-second retry did, while
+// downloadMaxAttempts still spans roughly the same ~99s.
+// var, not const, so tests can shorten them.
+var (
+	initialDownloadRetryDelay = 250 * time.Millisecond
+	maxDownloadRetryDelay     = time.Second
 )
 
 // backoffDelay returns initial doubled once per 0-based attempt, capped at limit.
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -565,10 +565,7 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 		_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainedErrorBody))
 		_ = resp.Body.Close()
 
-		// Client errors (4xx) will never succeed on retry—except 408 and 429, which ask
-		// for exactly that.
-		retryLater := resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
-		if !retryLater && resp.StatusCode >= http.StatusBadRequest && resp.StatusCode < http.StatusInternalServerError {
+		if utils.IsFatalClientError(resp.StatusCode) {
 			return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
 		}
 
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -570,7 +570,7 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 		// Client errors (4xx) will never succeed on retry—except 408 and 429, which ask
 		// for exactly that. A throttled download otherwise fails on the first 429.
 		retryLater := resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
-		if !retryLater && resp.StatusCode >= 400 && resp.StatusCode < 500 {
+		if !retryLater && resp.StatusCode >= http.StatusBadRequest && resp.StatusCode < http.StatusInternalServerError {
 			return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
 		}
 
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -49,7 +49,7 @@ const (
 // backoffDelay returns initial doubled once per 0-based attempt, capped at limit.
 func backoffDelay(attempt int, initial, limit time.Duration) time.Duration {
 	d := initial
-	for i := 0; i < attempt; i++ {
+	for range attempt {
 		d *= pollBackoffFactor
 		if d >= limit {
 			return limit
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -578,7 +578,7 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 			budget = min(budget, throttledMaxAttempts)
 		}
 		if count >= budget-1 {
-			return 0, fmt.Errorf("download failed after %d attempts (last status: %d)", budget, resp.StatusCode)
+			return 0, fmt.Errorf("download failed after %d attempts (last status: %d)", count+1, resp.StatusCode)
 		}
 		time.Sleep(backoffDelay(count, initialDownloadRetryDelay, maxDownloadRetryDelay))
 	}
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -40,6 +40,10 @@ const (
 	initialDownloadRetryDelay = 250 * time.Millisecond
 	maxDownloadRetryDelay     = time.Second
 
+	// Cap on the error body drained before a retry, so a keep-alive connection can be
+	// reused without reading an unbounded body from a failing server.
+	maxDrainedErrorBody = 64 << 10
+
 	basePollTimeout    = 30 * time.Second
 	perFilePollTimeout = 10 * time.Second
 	perMBPollTimeout   = 5 * time.Second
@@ -558,6 +562,9 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 		if resp.StatusCode == http.StatusOK {
 			break
 		}
+		// Drain before closing: an unread body keeps net/http from reusing the
+		// connection, so every retry would open a new one.
+		_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainedErrorBody))
 		_ = resp.Body.Close()
 
 		// Client errors (4xx) will never succeed on retry—except 408 and 429, which ask
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -206,7 +206,7 @@ func collectConcurrentFailures(count, limit int, fn func(int) string) []string {
 	failuresByIndex := make([]string, count)
 	sem := make(chan struct{}, limit)
 	var wg sync.WaitGroup
-	for i := 0; i < count; i++ {
+	for i := range count {
 		sem <- struct{}{}
 		wg.Add(1)
 		go func(index int) {
@@ -374,7 +374,7 @@ func UploadFile(ac *client.AlpaconClient, src []string, dest, username, groupnam
 	for i, filePath := range src {
 		f, err := os.Open(filePath)
 		if err != nil {
-			for j := 0; j < i; j++ {
+			for j := range i {
 				_ = files[j].Close()
 			}
 			return err
@@ -382,7 +382,7 @@ func UploadFile(ac *client.AlpaconClient, src []string, dest, username, groupnam
 		stat, err := f.Stat()
 		if err != nil {
 			_ = f.Close()
-			for j := 0; j < i; j++ {
+			for j := range i {
 				_ = files[j].Close()
 			}
 			return err
@@ -553,7 +553,7 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 	var resp *http.Response
 	var err error
 
-	for count := 0; count < maxAttempts; count++ {
+	for count := range maxAttempts {
 		resp, err = httpClient.Get(url)
 		if err != nil {
 			return 0, fmt.Errorf("network error while downloading: %w", err)
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -27,21 +27,19 @@ const (
 	downloadBulkAPIURL   = "/api/webftp/downloads/bulk/"
 	downloadStatusURL    = "/api/webftp/downloads/%s/status/"
 
-	// Shared by the poll interval and the download retry delay.
 	backoffFactor = 2
 
 	// Poll interval backs off exponentially up to maxPollInterval.
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
 
-	// Download retries back off to maxDownloadRetryDelay, so a brief blip recovers sooner
-	// while the 100-attempt ceiling still spans roughly the ~99s the flat one-second retry
-	// took.
+	// A brief blip recovers sooner than the flat one-second retry did, while the
+	// 100-attempt ceiling still spans roughly the same ~99s.
 	initialDownloadRetryDelay = 250 * time.Millisecond
 	maxDownloadRetryDelay     = time.Second
 
-	// Cap on the error body drained before a retry, so a keep-alive connection can be
-	// reused without reading an unbounded body from a failing server.
+	// Bounded so a keep-alive connection can be reused without reading an unbounded
+	// body from a failing server.
 	maxDrainedErrorBody = 64 << 10
 
 	basePollTimeout    = 30 * time.Second
@@ -54,7 +52,7 @@ const (
 
 // backoffDelay returns initial doubled once per 0-based attempt, capped at limit.
 func backoffDelay(attempt int, initial, limit time.Duration) time.Duration {
-	d := initial
+	d := min(initial, limit)
 	for range attempt {
 		d *= backoffFactor
 		if d >= limit {
@@ -64,8 +62,7 @@ func backoffDelay(attempt int, initial, limit time.Duration) time.Duration {
 	return d
 }
 
-// nextPollInterval returns the backoff delay for a 0-based poll attempt:
-// initialPollInterval doubled per attempt, capped at maxPollInterval.
+// nextPollInterval returns the backoff delay for a 0-based poll attempt.
 func nextPollInterval(attempt int) time.Duration {
 	return backoffDelay(attempt, initialPollInterval, maxPollInterval)
 }
@@ -562,13 +559,13 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 		if resp.StatusCode == http.StatusOK {
 			break
 		}
-		// Drain before closing: an unread body keeps net/http from reusing the
-		// connection, so every retry would open a new one.
+		// An unread body keeps net/http from reusing the connection, so every retry
+		// would open a new one.
 		_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainedErrorBody))
 		_ = resp.Body.Close()
 
 		// Client errors (4xx) will never succeed on retry—except 408 and 429, which ask
-		// for exactly that. A throttled download otherwise fails on the first 429.
+		// for exactly that.
 		retryLater := resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
 		if !retryLater && resp.StatusCode >= http.StatusBadRequest && resp.StatusCode < http.StatusInternalServerError {
 			return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -34,9 +34,10 @@ const (
 	maxPollInterval     = 2 * time.Second
 
 	// A brief blip recovers sooner than the flat one-second retry did, while the
-	// 100-attempt ceiling still spans roughly the same ~99s.
+	// attempt ceiling still spans roughly the same ~99s.
 	initialDownloadRetryDelay = 250 * time.Millisecond
 	maxDownloadRetryDelay     = time.Second
+	downloadMaxAttempts       = 100
 
 	// Bounded so a keep-alive connection can be reused without reading an unbounded
 	// body from a failing server.
@@ -683,7 +684,7 @@ func downloadSingleFileWithResult(ac *client.AlpaconClient, remotePath, dest, se
 		return DownloadedFile{}, fmt.Errorf("%s", status.Result)
 	}
 
-	localPath, written, err := saveDownloadedURL(ac.HTTPClient, downloadResponse.DownloadURL, dest, remotePath, recursive, 100)
+	localPath, written, err := saveDownloadedURL(ac.HTTPClient, downloadResponse.DownloadURL, dest, remotePath, recursive, downloadMaxAttempts)
 	if err != nil {
 		return DownloadedFile{}, err
 	}
@@ -743,7 +744,7 @@ func downloadBulk(ac *client.AlpaconClient, remotePaths []string, dest, serverID
 		return err
 	}
 	defer func() { _ = utils.DeleteFile(zipPath) }()
-	written, err := fetchFromURLToFile(ac.HTTPClient, response.DownloadURL, zipPath, 100)
+	written, err := fetchFromURLToFile(ac.HTTPClient, response.DownloadURL, zipPath, downloadMaxAttempts)
 	if err != nil {
 		return fmt.Errorf("failed to save downloaded archive: %w", err)
 	}
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -31,25 +31,37 @@ const (
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
 	pollBackoffFactor   = 2
-	basePollTimeout     = 30 * time.Second
-	perFilePollTimeout  = 10 * time.Second
-	perMBPollTimeout    = 5 * time.Second
+
+	// Download retries back off to maxDownloadRetryDelay, chosen so the total wait over
+	// maxAttempts stays what the flat one-second retry gave while a brief blip recovers
+	// sooner.
+	initialDownloadRetryDelay = 250 * time.Millisecond
+	maxDownloadRetryDelay     = time.Second
+
+	basePollTimeout    = 30 * time.Second
+	perFilePollTimeout = 10 * time.Second
+	perMBPollTimeout   = 5 * time.Second
 
 	bulkUploadConcurrency = 4 // uploads transfer payload bytes, keep low
 	bulkPollConcurrency   = 8 // status polls are lightweight, allow more
 )
 
-// nextPollInterval returns the backoff delay for a 0-based poll attempt:
-// initialPollInterval doubled per attempt, capped at maxPollInterval.
-func nextPollInterval(attempt int) time.Duration {
-	d := initialPollInterval
+// backoffDelay returns initial doubled once per 0-based attempt, capped at limit.
+func backoffDelay(attempt int, initial, limit time.Duration) time.Duration {
+	d := initial
 	for i := 0; i < attempt; i++ {
 		d *= pollBackoffFactor
-		if d >= maxPollInterval {
-			return maxPollInterval
+		if d >= limit {
+			return limit
 		}
 	}
 	return d
+}
+
+// nextPollInterval returns the backoff delay for a 0-based poll attempt:
+// initialPollInterval doubled per attempt, capped at maxPollInterval.
+func nextPollInterval(attempt int) time.Duration {
+	return backoffDelay(attempt, initialPollInterval, maxPollInterval)
 }
 
 // alignedPollDelay returns the sleep before the next poll: the backoff for this
@@ -546,15 +558,17 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 		}
 		_ = resp.Body.Close()
 
-		// Client errors (4xx) will never succeed on retry
-		if resp.StatusCode >= 400 && resp.StatusCode < 500 {
+		// Client errors (4xx) will never succeed on retry—except 408 and 429, which ask
+		// for exactly that. A throttled download otherwise fails on the first 429.
+		retryLater := resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
+		if !retryLater && resp.StatusCode >= 400 && resp.StatusCode < 500 {
 			return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
 		}
 
 		if count == maxAttempts-1 {
 			return 0, fmt.Errorf("download failed after %d attempts (last status: %d)", maxAttempts, resp.StatusCode)
 		}
-		time.Sleep(time.Second)
+		time.Sleep(backoffDelay(count, initialDownloadRetryDelay, maxDownloadRetryDelay))
 	}
 
 	defer func() { _ = resp.Body.Close() }()
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -39,6 +39,10 @@ const (
 	maxDownloadRetryDelay     = time.Second
 	downloadMaxAttempts       = 100
 
+	// A server that answered "too many requests" is not persuaded by ninety more of
+	// them. Without Retry-After the throttle window is a guess, so this errs small.
+	throttledMaxAttempts = 10
+
 	// The drain only buys connection reuse, and the shared client sets no timeout,
 	// so this bound is the retry loop's only defense against a stalled error body.
 	maxDrainedErrorBody = 8 << 10
@@ -569,8 +573,12 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 			return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
 		}
 
-		if count == maxAttempts-1 {
-			return 0, fmt.Errorf("download failed after %d attempts (last status: %d)", maxAttempts, resp.StatusCode)
+		budget := maxAttempts
+		if utils.IsRetryLaterStatus(resp.StatusCode) {
+			budget = min(budget, throttledMaxAttempts)
+		}
+		if count >= budget-1 {
+			return 0, fmt.Errorf("download failed after %d attempts (last status: %d)", budget, resp.StatusCode)
 		}
 		time.Sleep(backoffDelay(count, initialDownloadRetryDelay, maxDownloadRetryDelay))
 	}
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -33,14 +33,14 @@ const (
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
 
-	// A brief blip recovers sooner than the flat one-second retry did, while the
-	// attempt ceiling still spans roughly the same ~99s.
+	// A brief blip recovers sooner than the flat one-second retry did, while
+	// downloadMaxAttempts still spans roughly the same ~99s.
 	initialDownloadRetryDelay = 250 * time.Millisecond
 	maxDownloadRetryDelay     = time.Second
 	downloadMaxAttempts       = 100
 
-	// A server that answered "too many requests" is not persuaded by ninety more of
-	// them. Without Retry-After the throttle window is a guess, so this errs small.
+	// A server that answered "too many requests" is not persuaded by more of them,
+	// and without Retry-After the window is a guess—so this errs small.
 	throttledMaxAttempts = 10
 
 	// The drain only buys connection reuse, and the shared client sets no timeout,
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -27,10 +27,12 @@ const (
 	downloadBulkAPIURL   = "/api/webftp/downloads/bulk/"
 	downloadStatusURL    = "/api/webftp/downloads/%s/status/"
 
+	// Shared by the poll interval and the download retry delay.
+	backoffFactor = 2
+
 	// Poll interval backs off exponentially up to maxPollInterval.
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
-	pollBackoffFactor   = 2
 
 	// Download retries back off to maxDownloadRetryDelay, chosen so the total wait over
 	// maxAttempts stays what the flat one-second retry gave while a brief blip recovers
@@ -50,7 +52,7 @@ const (
 func backoffDelay(attempt int, initial, limit time.Duration) time.Duration {
 	d := initial
 	for range attempt {
-		d *= pollBackoffFactor
+		d *= backoffFactor
 		if d >= limit {
 			return limit
 		}
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -39,9 +39,9 @@ const (
 	maxDownloadRetryDelay     = time.Second
 	downloadMaxAttempts       = 100
 
-	// Bounded so a keep-alive connection can be reused without reading an unbounded
-	// body from a failing server.
-	maxDrainedErrorBody = 64 << 10
+	// The drain only buys connection reuse, and the shared client sets no timeout,
+	// so this bound is the retry loop's only defense against a stalled error body.
+	maxDrainedErrorBody = 8 << 10
 
 	basePollTimeout    = 30 * time.Second
 	perFilePollTimeout = 10 * time.Second
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -34,9 +34,9 @@ const (
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
 
-	// Download retries back off to maxDownloadRetryDelay, chosen so the total wait over
-	// maxAttempts stays what the flat one-second retry gave while a brief blip recovers
-	// sooner.
+	// Download retries back off to maxDownloadRetryDelay, so a brief blip recovers sooner
+	// while the 100-attempt ceiling still spans roughly the ~99s the flat one-second retry
+	// took.
 	initialDownloadRetryDelay = 250 * time.Millisecond
 	maxDownloadRetryDelay     = time.Second
 
--- a/api/ftp/ftp_test.go
+++ b/api/ftp/ftp_test.go
@@ -1434,6 +1434,37 @@ func TestFetchFromURLToFile_GivesUpOnAFatalClientError(t *testing.T) {
 	assert.Equal(t, int32(1), calls.Load(), "a 404 will be a 404 again")
 }
 
+func TestFetchFromURLToFile_StopsDrainingAStalledErrorBody(t *testing.T) {
+	stall := make(chan struct{})
+	var releaseOnce sync.Once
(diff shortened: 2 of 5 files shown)


jisung-02:
Carried over from #297 (closed after the branch was renamed to `fix/ftp-retry-backoff`). Copilot's review there, verbatim, so the findings stay attached to the open PR.

**Overall (🟡 Not ready to approve):** The retry loop closes error responses without draining the body, which can prevent HTTP connection reuse and unnecessarily amplify connection churn during repeated retries.

**`api/ftp/ftp.go:559`**
> In the retry loop, the response body is closed without being drained. With net/http, not reading the body to EOF typically prevents connection reuse, so a throttled/download-retry scenario may create a new TCP connection per attempt and amplify load during retries.

**`api/ftp/ftp.go:53`**
> `backoffDelay` is a shared helper (polling + download retries), but it currently hard-codes `pollBackoffFactor` and the name implies it only affects polling. Renaming the factor and updating the adjacent comments makes the shared backoff behavior clearer and avoids misleading documentation.

api/ftp/ftp.go
```diff
@@ -31,27 +31,39 @@ const (
...
-// nextPollInterval returns the backoff delay for a 0-based poll attempt:
-// initialPollInterval doubled per attempt, capped at maxPollInterval.
-func nextPollInterval(attempt int) time.Duration {
-	d := initialPollInterval
-	for i := 0; i < attempt; i++ {
+// backoffDelay returns initial doubled once per 0-based attempt, capped at limit.
+func backoffDelay(attempt int, initial, limit time.Duration) time.Duration {
+	d := initial
+	for range attempt {
 		d *= pollBackoffFactor
```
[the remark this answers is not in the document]
jisung-02: Fixed in 517dfc8. The seed is now clamped (`d := min(initial, limit)`) rather than the result, so attempt 0 respects the cap while the early return inside the loop still keeps a large attempt count from overflowing the doubling. Added `TestBackoffDelay_CapsAnInitialAlreadyOverTheLimit` covering attempt 0 and attempt 3 with `initial > limit`. No caller passes such a pair today, so this changes no runtime behavior — it makes the helper safe for the next reuse.

api/ftp/ftp.go
```diff
@@ -546,15 +558,17 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 		}
 		_ = resp.Body.Close()
```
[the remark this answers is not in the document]
jisung-02: Fixed in 78ec29d. The non-200 path now drains the body before closing it: `io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainedErrorBody))`, with `maxDrainedErrorBody` at 64 KiB so a failing server cannot make the CLI read an unbounded body just to keep the connection reusable.

api/ftp/ftp.go
```diff
@@ -544,17 +559,22 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 		if resp.StatusCode == http.StatusOK {
 			break
 		}
+		// An unread body keeps net/http from reusing the connection, so every retry
+		// would open a new one.
+		_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainedErrorBody))
```
geunwoonoh, line 573: **Warning:** the drain is an unbounded *blocking* read, and this client has no timeout.

`client.NewAlpaconAPIClient` (`client/client.go:50`) constructs `&http.Client{Transport: ...}` with no `Timeout`, and the `Transport` sets no `ResponseHeaderTimeout` or read deadline. Before this PR the non-200 path called `resp.Body.Close()` immediately, so a stalled error body cost nothing. Now the code reads up to 64 KiB off that body first, with no deadline anywhere in the stack.

A server that sends `429` + `Content-Length: 65536` and then dribbles the body hangs the CLI indefinitely — on the error path, inside a loop budgeted for 100 attempts. `maxDrainedErrorBody` caps *bytes*, not *time*, and time is the dimension that matters for a hang. The drain is a pure optimization, so as written it trades a certain small win (connection reuse) for an unbounded worst case.

Either lower the bound (an S3/CloudFront error body is a few hundred bytes; 4–8 KiB is already generous) or — better — put a `Timeout` on the shared client, which would also bound the success-path read in `SaveStreamAtomic`.

---

**Suggestion**, same line: the drain also runs on fatal 4xx, where the connection is about to be abandoned. On a 403/404 the next statement returns an error and the CLI exits, so up to 64 KiB is read to warm a connection nothing will use.

There is a better use for those bytes, though: the code now *has* the S3 error body in hand and discards it. Presigned-URL failures say things like `<Message>Request has expired</Message>`, and `download failed with client error: 403` currently tells the user none of it. Surfacing a short snippet in the error message would turn this read into a diagnostics win instead of dead work.

api/ftp/ftp.go
```diff
@@ -544,17 +559,22 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
...
+		retryLater := resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
+		if !retryLater && resp.StatusCode >= http.StatusBadRequest && resp.StatusCode < http.StatusInternalServerError {
 			return 0, fmt.Errorf("download failed with client error: %d", resp.StatusCode)
 		}
 
 		if count == maxAttempts-1 {
 			return 0, fmt.Errorf("download failed after %d attempts (last status: %d)", maxAttempts, resp.StatusCode)
 		}
-		time.Sleep(time.Second)
+		time.Sleep(backoffDelay(count, initialDownloadRetryDelay, maxDownloadRetryDelay))
```
geunwoonoh, line 587: **Warning:** a 429 now draws ~98 requests at ~1/s, with no `Retry-After` and no jitter.

The PR body reasons about the cap in terms of wall time ("the cap has to stay at one second or the worst-case total wait doubles"), but the variable that changed most is *request volume against a server that just said "too many requests"*: 1 request before this PR, up to 100 after. With the cap at 1s that is ~1 req/s sustained for ~98s.

`Retry-After` is listed under "Not addressed" and the reasoning for deferring it is sound for 408 — but 429 is the one status where the header is the server's actual instruction, and ignoring it while raising the retry count 100× is the combination worth a second look. A server asking for `Retry-After: 60` gets polled ~60 times inside the window it asked to be left alone.

Options, cheapest first: give the `retryLater` class a smaller attempt budget than the 5xx class, add jitter, or honor the header. Note that jitter would be a new convention here — `nextReconnectDelay` (`api/event/wslistener.go:135`) and the gap-fill backoff are both jitterless, so nothing in this repo does it today. That makes the attempt-budget split the smallest change that addresses the volume.

api/ftp/ftp.go
```diff
@@ -544,17 +559,22 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
...
+		// An unread body keeps net/http from reusing the connection, so every retry
+		// would open a new one.
+		_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, maxDrainedErrorBody))
 		_ = resp.Body.Close()
 
-		// Client errors (4xx) will never succeed on retry
-		if resp.StatusCode >= 400 && resp.StatusCode < 500 {
+		// Client errors (4xx) will never succeed on retry—except 408 and 429, which ask
+		// for exactly that.
+		retryLater := resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
```
geunwoonoh: **Warning:** this duplicates `api/event/watcher.go:93-96` verbatim — including the comment prose.

```go
// api/event/watcher.go:93
retryLater := status == http.StatusRequestTimeout || status == http.StatusTooManyRequests
if !retryLater && status >= http.StatusBadRequest && status < http.StatusInternalServerError {
```
```go
// api/ftp/ftp.go:569
retryLater := resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusTooManyRequests
if !retryLater && resp.StatusCode >= http.StatusBadRequest && resp.StatusCode < http.StatusInternalServerError {
```

Both files also carry the same "—except 408 and 429, which ask for exactly that" sentence.

The PR body frames this as intentional ("the same band `671d4bf` settled"), but copying a *policy* is exactly how the two drift: the next status added to the band (503-with-`Retry-After`, 425) gets added in one place and not the other. `utils` already hosts `HTTPStatusCode` and is already imported by both packages, so `utils.IsRetryableHTTPStatus(status int) bool` is a mechanical extraction with no new dependency edge.

This PR deduplicated `nextPollInterval`'s backoff for precisely this reason — the same argument applies one level up.

api/ftp/ftp.go
```diff
@@ -27,31 +27,46 @@ const (
...
 	initialPollInterval = 250 * time.Millisecond
 	maxPollInterval     = 2 * time.Second
-	pollBackoffFactor   = 2
-	basePollTimeout     = 30 * time.Second
-	perFilePollTimeout  = 10 * time.Second
-	perMBPollTimeout    = 5 * time.Second
+
+	// A brief blip recovers sooner than the flat one-second retry did, while the
+	// 100-attempt ceiling still spans roughly the same ~99s.
+	initialDownloadRetryDelay = 250 * time.Millisecond
```
geunwoonoh: **Suggestion:** `maxAttempts=100` is a magic literal at both call sites (`ftp.go:686`, `ftp.go:746`), and the comment right above this line depends on it.

The comment reads "the 100-attempt ceiling still spans roughly the same ~99s", but nothing connects it to the two `100`s 640+ lines away. Naming it puts the constant the PR body reasons about next to the delays it interacts with:

```go
initialDownloadRetryDelay = 250 * time.Millisecond
maxDownloadRetryDelay     = time.Second
downloadMaxAttempts       = 100
```

This is also where a smaller budget for the `retryLater` class would land (see the comment on line 577).

For the record, the `~99s` claim itself checks out: 0.25 + 0.5 + 97×1 = 97.75s, against the old 99×1s.

api/ftp/ftp_test.go
```diff
@@ -1385,5 +1385,73 @@ func TestPollTransferStatus_TimesOut(t *testing.T) {
...
+			written, err := fetchFromURLToFile(ts.Client(), ts.URL, path, 3)
+
+			require.NoError(t, err)
+			assert.Equal(t, int64(len("payload")), written)
+			assert.Equal(t, int32(2), calls.Load(), "the status asks for a retry")
+		})
+	}
+}
+
+func TestFetchFromURLToFile_GivesUpOnAFatalClientError(t *testing.T) {
```
geunwoonoh: **Suggestion:** this is redundant with `TestFetchFromURL_ClientErrorNoRetry` (`api/ftp/ftp_test.go:779`).

The existing test already asserts that a fatal 4xx (403) errors out with `client error: 403` after exactly one request — *and* additionally asserts that no file was created, which this new test drops. The two differ only in status code (404 vs 403) and `maxAttempts` (5 vs 10).

Folding 404 into the existing test as a table case keeps the extra `os.Stat` assertion and removes ~15 lines.

api/ftp/ftp_test.go
```diff
@@ -1385,5 +1385,73 @@ func TestPollTransferStatus_TimesOut(t *testing.T) {
...
+	defer ts.Close()
+
+	_, err := fetchFromURLToFile(ts.Client(), ts.URL, filepath.Join(t.TempDir(), "out"), 5)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "client error: 404")
+	assert.Equal(t, int32(1), calls.Load(), "a 404 will be a 404 again")
+}
+
+func TestBackoffDelay(t *testing.T) {
```
geunwoonoh, line 1534: **Suggestion:** `TestBackoffDelay` skips `t.Run` and the `name` field.

Both adjacent table tests in this file — `TestNextPollInterval` (`:1234`) and `TestAlignedPollDelay` (`:1255`) — use named subtests, and the golang stack rules call for it. As written, a regression at `attempt: 50` reports a bare duration mismatch with no indication of which row failed.

```go
for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
        assert.Equal(t, tt.want, backoffDelay(tt.attempt, initialDownloadRetryDelay, maxDownloadRetryDelay))
    })
}
```

---

**Suggestion**, related coverage gap: nothing in this file drives `fetchFromURLToFile` through a 5xx, and nothing reaches the `download failed after %d attempts` branch at `ftp.go:574` — the new retryable tests use `maxAttempts=3` with a single failure, so they never exhaust the budget.

The gap predates this PR, but this is the PR that rewrites the loop, so it is the natural place to close it. A 429-forever server with `maxAttempts=2` covers exhaustion in ~250ms.

geunwoonoh (commented):
## Review

**Stack**: golang · **Verdict**: COMMENTED — no critical issues

### Summary

`fetchFromURLToFile` rejected the entire 4xx band as unretryable, so a throttled presigned download died on the first 429. This PR carves 408/429 out of that band, replaces the flat `time.Sleep(time.Second)` with an exponential backoff extracted from the existing `nextPollInterval`, drains error bodies so `net/http` can reuse connections across retries, and modernizes counted loops / test counters in the two touched files.

The core change is right, and extracting `backoffDelay(attempt, initial, limit)` instead of writing a second copy is the correct call. `nextPollInterval`'s behavior is provably unchanged (`min(250ms, 2s)` is a no-op).

Verified locally at `ba6f42a` (= PR head):

```
gofmt -l .                    no output
go vet ./...                  exit 0
go build ./...                exit 0
go test -race -count=1 ./...  all packages ok
golangci-lint                 not installed locally — CI claim unverified
```

### Critical issues (0)

None. Nothing here breaks the stated goal.

### Warnings (3)

1. **`api/ftp/ftp.go:564`** — the drain is an unbounded *blocking* read, and this client has no timeout. The 64 KiB bound caps bytes, not time.
2. **`api/ftp/ftp.go:577`** — a 429 now draws ~98 requests at ~1/s, with no `Retry-After` and no jitter. The PR body reasons about wall time; request volume against a throttling server went 1 → 100.
3. **`api/ftp/ftp.go:569`** — duplicates `api/event/watcher.go:93-96` verbatim, including the comment prose.

### Suggestions (5)

1. **`api/ftp/ftp_test.go:1422`** — `TestFetchFromURLToFile_GivesUpOnAFatalClientError` is redundant with the existing `TestFetchFromURL_ClientErrorNoRetry` (`:779`).
2. **`api/ftp/ftp.go:686,746`** — `maxAttempts=100` is a magic literal at both call sites, and a comment 640 lines away depends on it.
3. **`api/ftp/ftp_test.go:1437`** — `TestBackoffDelay` skips `t.Run` and the `name` field, unlike both adjacent table tests.
4. **Coverage gap** — the 5xx retry path and attempt exhaustion (`ftp.go:574`) are both untested.
5. **`api/ftp/ftp.go:564`** — the drain also runs on fatal 4xx, and the S3 error body it reads is discarded rather than surfaced in the error.

Details inline.

### Highlights

- Extracting `backoffDelay` from `nextPollInterval` rather than writing a second doubling loop, and keeping `nextPollInterval` as a thin wrapper — the existing poll tests then prove the refactor is behavior-preserving for free.
- The early `return limit` inside the doubling loop makes overflow structurally impossible, and `ba6f42a` correctly clamps the *seed* rather than the *result* to preserve that property. `TestBackoffDelay_CapsAnInitialAlreadyOverTheLimit` covering both attempt 0 and attempt 3 is the right pair.
- Both Copilot findings were addressed with real fixes and tests, not acknowledgements.
- The `Not addressed` section naming `api/event/watcher.go:91` as having the same `Retry-After` gap — that is the kind of note that keeps a known gap from being rediscovered as a bug.
- The `style(ftp)` modernization is explicitly scoped to the two files the fix already touches, with the rest deferred to #279. Mixing style into a fix PR is normally worth pushing back on; bounding it this way makes it reviewable.

### Checklist

- [x] Code follows stack conventions — one deviation (suggestion 3)
- [x] Tests included and passing — verified `go test -race -count=1 ./...` clean; two gaps noted (suggestions 1, 4)
- [x] No security vulnerabilities — no credential/secret/injection surface; the drain hang (warning 1) is availability-only and needs a misbehaving server
- [x] Documentation updated if needed — the `~99s` claim at `ftp.go:36` checks out (0.25 + 0.5 + 97×1 = 97.75s vs the old 99×1s)

api/ftp/ftp.go
jisung-02: Bound lowered to 8 KiB in `0bd8d1c`, and the comment now says what the bound actually is:

```go
// The drain only buys connection reuse, and the shared client sets no timeout,
// so this bound is the retry loop's only defense against a stalled error body.
maxDrainedErrorBody = 8 << 10
```

You are right that the 64 KiB was reasoning about memory when the exposure is time. `TestFetchFromURLToFile_StopsDrainingAStalledErrorBody` drives a server that flushes one byte past the bound and then stalls; with the `io.LimitReader` removed it hangs instead of returning, so the test fails for the right reason.

A `Timeout` on the shared client is left out of this PR. It is the real fix, but it also bounds the success-path read in `SaveStreamAtomic`, and picking that value is a decision about every API call rather than about this loop.

On the second half: the drain still runs on a fatal 4xx. Skipping it there would zero that path's exposure but leave the 408/429 path — the one that actually loops — untouched, so lowering the bound covers both instead. Surfacing the S3 error body is not in this PR either; it changes user-facing error strings and is a diagnostics improvement rather than a regression this branch introduced.

api/ftp/ftp.go
jisung-02: Fixed in `6e42210` with the attempt-budget split, the option you named as the smallest change.

```go
budget := maxAttempts
if utils.IsRetryLaterStatus(resp.StatusCode) {
    budget = min(budget, throttledMaxAttempts)
}
```

`throttledMaxAttempts` is 10, so a 429 now draws 10 requests over ~8.75s instead of 98 over ~98s. The 5xx class keeps the full budget — S3 signals its own throttling with `503 SlowDown`, so the class being narrowed is what a proxy or WAF in front would produce.

Ten is a guess and the comment says so rather than justifying it with a number nobody measured. Without `Retry-After` the window is unknowable from here, and the cost of guessing low is one rerun of `alpacon cp`. `Retry-After` itself stays deferred, together with `api/event/watcher.go`.

`TestFetchFromURLToFile_StopsAtTheAttemptBudget` counts the requests against the real 100-attempt budget. It costs ~7.8s, which seemed worth paying for an assertion whose whole subject is volume.

The budget change also surfaced a defect of its own, fixed in `e6e5af2`: the exhaustion error printed `budget`, so a run that spent fifty attempts on 5xx and then drew a 429 reported "after 10 attempts". It prints `count+1` now.

api/ftp/ftp.go
jisung-02: Extracted in `14aa221`. Both call sites are one line now and `watcher.go` drops `net/http` entirely.

The shape differs from your suggestion: `utils.IsFatalClientError(status int) bool` rather than `IsRetryableHTTPStatus`. With the retryable form both callers would still spell out `!retryable && status >= 400 && status < 500`, so the band boundaries stay duplicated — which is the thing that drifts. The fatal form is the question both sites actually ask.

`6e42210` then split out `utils.IsRetryLaterStatus` for the throttle budget, and `IsFatalClientError` calls it, so 408 and 429 are named exactly once in the tree.

You were also right that the argument applies one level up from the `backoffDelay` extraction; that framing is what settled it against leaving the copy in place.

api/ftp/ftp.go
jisung-02: Named in `c2a9788` as `downloadMaxAttempts`, next to the delays it interacts with, and both call sites use it.

The comment dropped the number rather than keeping it in sync:

```go
// A brief blip recovers sooner than the flat one-second retry did, while
// downloadMaxAttempts still spans roughly the same ~99s.
```

Worth recording what the archaeology turned up, since it changed how the throttle budget was shaped: the 100 arrived in `ac13e7e` (2023-12) as 100 iterations of `time.Sleep(1 * time.Second)`, next to a warning that read "Transfer may timeout if it exceeds 100 seconds". It was a wall-clock budget expressed as a count because one attempt was one second, and this PR is what broke that identity by making the first two delays 250ms and 500ms. The warning text is gone from the repo, so the only thing explaining the number had already been deleted.

That argues the whole thing should be a deadline rather than a count, which would also drop the `~99s` comment. Not done here — doing it properly means converting `downloadMaxAttempts` too, which changes the 5xx path this review did not touch. It is recorded as a `Directive` trailer on `6e42210` to land with `Retry-After`, since that value would feed the same deadline.

api/ftp/ftp_test.go
jisung-02: Folded in `a517331`. `TestFetchFromURLToFile_GivesUpOnAFatalClientError` is gone and 404 is a row in `TestFetchFromURL_ClientErrorNoRetry`, so the `os.Stat` assertion the new test had dropped now covers both status codes. The surviving `maxAttempts` is 10; both rows return on the first attempt, so the difference was not observable.

api/ftp/ftp_test.go
jisung-02: Both done.

`8970b87` runs `TestBackoffDelay` as named subtests, matching `TestNextPollInterval` and `TestAlignedPollDelay`. `TestBackoffDelay_CapsAnInitialAlreadyOverTheLimit` stays separate — it varies `initial` and `limit` rather than `attempt`, so folding it in would put two more fields on the table that every existing row repeats.

`d4750af` closes the coverage gap. A 502 folds into the retryable table as a third row, which is also why that test is now `RetriesRetryableStatuses` — a 502 is not a client error, and the assertion message follows: 408 and 429 ask for the retry, a 502 merely allows it. Exhaustion gets its own test because it asserts an error, a request count and an absent file rather than a written payload.

`e1d1f88` later merged that exhaustion test with the throttle-budget one from `6e42210` into a single table, since they stand up the same 429-only server and differ only in the budget passed and the count expected.

api/ftp/ftp_test.go
```diff
@@ -1385,5 +1397,135 @@ func TestPollTransferStatus_TimesOut(t *testing.T) {
...
+			written, err := fetchFromURLToFile(ts.Client(), ts.URL, path, 3)
+
+			require.NoError(t, err)
+			assert.Equal(t, int64(len("payload")), written)
+			assert.Equal(t, int32(2), calls.Load(), "the status is retryable")
+		})
+	}
+}
+
+func TestFetchFromURLToFile_StopsAtTheAttemptBudget(t *testing.T) {
```
geunwoonoh, line 1443: **Suggestion:** the `e6e5af2` fix is not covered by the tests that ship with it.

I verified this by experiment rather than by reading: I reverted `count+1` back to `budget` locally and ran `go test -run TestFetchFromURLToFile ./api/ftp/` — **it passes**.

Both rows here use a 429-only server, so `count` never gets ahead of the throttle budget and `count+1` is numerically identical to `budget` in both cases. The defect `e6e5af2` fixed — a run that spends attempts on 5xx and then draws a 429 — is exactly the shape neither row produces.

Distinguishing them needs the 429 to arrive at `count >= throttledMaxAttempts`:

```go
// 10x 502, then 429 -> count is 10 when budget collapses to 10
{name: "throttle budget after other failures", ..., wantCalls: 11,
 wantErr: "download failed after 11 attempts (last status: 429)"},
```

That is also the only path where the `>=` in `count >= budget-1` is observable at all. It is the right form — with `==` that run would fall through, loop to exhaustion, and reach `SaveStreamAtomic` with a closed body — but nothing currently proves it stays `>=`. Two subtle behaviors, zero coverage.

Non-blocking; see the note on `ftp.go:44` for why this pairs with making the delays test-overridable.

api/ftp/ftp.go
```diff
@@ -27,31 +27,51 @@ const (
...
+
+	// A brief blip recovers sooner than the flat one-second retry did, while
+	// downloadMaxAttempts still spans roughly the same ~99s.
+	initialDownloadRetryDelay = 250 * time.Millisecond
+	maxDownloadRetryDelay     = time.Second
+	downloadMaxAttempts       = 100
+
+	// A server that answered "too many requests" is not persuaded by more of them,
+	// and without Retry-After the window is a guess—so this errs small.
+	throttledMaxAttempts = 10
```
geunwoonoh, line 40: **Suggestion:** closing the coverage gap above cheaply argues for making the delay constants test-overridable — which also buys back suite time.

`TestFetchFromURLToFile_StopsAtTheAttemptBudget` costs **8.02s** of `api/ftp`'s 18.2s (the package was 7.46s total before this branch). A mixed-class test would add ~10s more, since it needs ≥10 real attempts before the 429 lands. All of it is paid in `time.Sleep`, not in assertions.

The codebase already has this pattern, documented in the source:

```go
// api/event/event.go:30 — var, not const, so tests can shorten them.
var ( gapFillInitialInterval = 300 * time.Millisecond; ... )
```
```go
// api/event/wslistener_test.go:17 — keeps reconnect assertions off the production 1s backoff.
const testReconnectBaseDelay = 10 * time.Millisecond
```

Making `initialDownloadRetryDelay` / `maxDownloadRetryDelay` `var` would take both the existing budget test and the missing mixed-class one to milliseconds, and follows the local convention rather than inventing one.

---

**Observation**, not for this PR: the class that keeps the full `downloadMaxAttempts` budget includes `503 SlowDown`, which is S3's own throttle signal. So the narrowing lands on the WAF/proxy 429 while the storage backend's throttle still draws ~98 requests at ~1/s. You name that trade-off explicitly and 503-at-full-budget is pre-existing, so it is not a regression — but it is the same problem, and the deadline conversion in the `Directive` trailer on `6e42210` is where it should be solved.

api/ftp/ftp.go
```diff
@@ -27,31 +27,51 @@ const (
...
-	pollBackoffFactor   = 2
-	basePollTimeout     = 30 * time.Second
-	perFilePollTimeout  = 10 * time.Second
-	perMBPollTimeout    = 5 * time.Second
+
+	// A brief blip recovers sooner than the flat one-second retry did, while
+	// downloadMaxAttempts still spans roughly the same ~99s.
+	initialDownloadRetryDelay = 250 * time.Millisecond
+	maxDownloadRetryDelay     = time.Second
+	downloadMaxAttempts       = 100
```
geunwoonoh: **Suggestion:** the PR body no longer describes the branch, and on a public repo it is the durable record.

The commit table stops at `ba6f42a`; the ten commits since are absent. `throttledMaxAttempts`, `utils.IsFatalClientError`, the 8 KiB drain bound, and the newly-deferred client `Timeout` appear nowhere in it, and "Not addressed" still lists `Retry-After` alone — which now understates what was deferred and overstates the 429 behavior the body describes ("Both call sites pass `maxAttempts=100`, so the cap has to stay at one second" no longer holds for the throttled class).

The review threads carry all of it, including the `ac13e7e` archaeology on where the `100` came from, which is genuinely worth keeping. But threads are harder to find six months from now than the body is, and the archaeology is the part that explains why the follow-up should be a deadline rather than a count.

geunwoonoh (dismissed):
## Review (round 2)

**Stack**: golang · **Head**: `e6e5af2` (was `ba6f42a` at my last review) · **Verdict**: APPROVED

### Summary

All six findings from my previous review are addressed in code — verified against `e6e5af2`, not taken from the replies. Two items were consciously deferred with reasoning I accept, and one of the fixes surfaced a defect of its own (`e6e5af2`) that you found and fixed unprompted.

| Finding | Resolution | Verified |
|---|---|---|
| W1 · unbounded blocking drain | `0bd8d1c` — bound 64 KiB → 8 KiB, comment now names *time* as the exposure; `TestFetchFromURLToFile_StopsDrainingAStalledErrorBody` added | ✅ `ftp.go:48` |
| W2 · 429 volume | `6e42210` — `throttledMaxAttempts = 10`; a 429 draws 10 requests over 7.75s instead of 98 over 98s | ✅ `ftp.go:44,576-581` |
| W3 · duplicated retry band | `14aa221` — `utils.IsFatalClientError`; `watcher.go` drops `net/http` entirely and both call sites are one line | ✅ `utils/error.go:75-85` |
| S2 · magic literal `100` | `c2a9788` — `downloadMaxAttempts`; no `, 100)` left in the file | ✅ `ftp.go:40,692,752` |
| S1 · redundant fatal-4xx test | `a517331` — new test deleted, 404 folded as a row; the `os.Stat` assertion now covers both codes | ✅ `ftp_test.go:779` |
| S3 · `t.Run` + coverage gap | `8970b87` named subtests; `d4750af` adds 502 + exhaustion; `e1d1f88` merges the two 429 tables | ✅ `ftp_test.go:1435` |

Verified locally at `e6e5af2`:

```
gofmt -l .                    no output
go vet ./...                  exit 0
go build ./...                exit 0
go test -race -count=1 ./...  all 38 packages ok
golangci-lint                 not installed locally — CI claim still unverified
```

I specifically checked the boundary the budget split could have broken. `count >= budget-1` (not `==`) is the correct form: on a run that draws 5xx for 50 attempts and then a 429, `budget` collapses to 10 while `count` is already 50, so an `==` check would have fallen through and run the loop to exhaustion with a closed body reaching `SaveStreamAtomic`. The `>=` handles it and reports 51.

`utils.IsFatalClientError(0)` also preserves `watcher.go`'s old behavior exactly (status 0 → not fatal → retry window), and `TestClientErrorBand` covers that row — that was the subtle case in the extraction.

### Critical issues (0)

### Warnings (0)

Both remaining deferrals are sound, and I would have made the same calls:

- **Client `Timeout`** — it is the real fix for the drain, but it also bounds the success-path read in `SaveStreamAtomic`, and that value is a decision about every API call in the CLI. Out of scope for a download-retry fix.
- **`Retry-After`** — deferred with `api/event/watcher.go`, and `6e42210` carries a `Directive` trailer saying this budget is what its value should feed. That is the right place for it.

### Suggestions (3, all non-blocking)

1. **`api/ftp/ftp_test.go:1435`** — the `e6e5af2` fix is not covered by the tests that ship with it. Confirmed empirically: reverting `count+1` to `budget` locally still passes `go test -run TestFetchFromURLToFile ./api/ftp/`.
2. **`api/ftp/ftp.go:44`** — closing that gap cheaply argues for making the delay constants test-overridable, which would also buy back the 8.02s the budget test currently costs.
3. **The PR body is stale** — the commit table stops at `ba6f42a` and the ten commits since are absent. On a public repo the body is the durable record.

Details inline.

### Observation (for the follow-up, not this PR)

The class that kept the full 100-attempt budget includes `503 SlowDown`, which is S3's own throttle signal. So the narrowing applies to the WAF/proxy 429 while the storage backend's throttle still draws ~98 requests at ~1/s. You name this trade-off explicitly, and 503-retries-at-full-budget is pre-existing behavior this PR did not introduce — but it is the same problem, and the `Directive` trailer's deadline conversion is where it should land.

### Highlights

- Every finding was verified in code, and every one held. The `count >= budget-1` form in particular shows the boundary was reasoned about rather than pattern-matched.
- `utils.IsFatalClientError` over my suggested `IsRetryableHTTPStatus` is the better shape, and the reply says why: with the retryable form both callers would still spell out `!retryable && status >= 400 && status < 500`, leaving the band boundaries duplicated — which is the part that drifts. Pushing back on a review suggestion with a correct reason is worth more than compliance.
- `TestFetchFromURLToFile_StopsDrainingAStalledErrorBody` fails in 5s via `select` rather than hanging to the 10-minute test timeout, and the `defer release()` / `defer ts.Close()` ordering is correct (LIFO — the handler blocks in `<-stall`, so `ts.Close()` would deadlock if release ran second). The comment on that line explains exactly that.
- `throttledMaxAttempts = 10` is documented as a guess *as a guess* ("without Retry-After the window is a guess—so this errs small") instead of being dressed up with a number nobody measured.
- Finding and fixing `e6e5af2` — a defect introduced by the fix for W2 — without being asked.
- The `Directive` trailer on `6e42210` turns a deferral into a scheduled decision with its trigger recorded.
- The `ac13e7e` archaeology on where the `100` came from (100 iterations of a 1s sleep next to a since-deleted "may timeout if it exceeds 100 seconds" warning — a wall-clock budget expressed as a count, whose identity this PR is what broke) is the kind of finding that changes the shape of the follow-up rather than just explaining the present.

---

**Verdict**: APPROVED — nothing outstanding blocks the merge. Suggestion 1 is the only one I would want to see land eventually, and it pairs naturally with suggestion 2, so a follow-up rather than another round here.

**Checklist**:
- [x] Code follows stack conventions — no deviations remaining
- [x] Tests included and passing — verified `go test -race -count=1 ./...` clean at `e6e5af2`; one coverage gap noted, empirically confirmed
- [x] No security vulnerabilities — no credential/secret/injection surface; the drain exposure is now bounded at 8 KiB
- [x] Documentation updated if needed — code comments are accurate; the PR body is not (suggestion 3)

api/ftp/ftp_test.go
jisung-02: Covered in `ccfc8d4`. The row serves 502 for `throttledMaxAttempts` calls and then 429, so the budget collapses to 10 with `count` already at 10:

```go
statusFor: func(call int32) int {
	if call <= throttledMaxAttempts {
		return http.StatusBadGateway
	}
	return http.StatusTooManyRequests
},
wantCalls: throttledMaxAttempts + 1,
wantErr:   "download failed after 11 attempts (last status: 429)",
```

The fixed 429 handler became a per-row `statusFor func(call int32) int` to make room for it.

I ran both mutations you named, and both fail:

```
// count+1 -> budget
Error: "download failed after 10 attempts (last status: 429)" does not contain
       "download failed after 11 attempts (last status: 429)"

// >= -> ==
--- FAIL: .../throttle_budget_after_other_failures (98.90s)
```

The `==` case is worse than the thread describes. It is not an error after exhaustion — it is `nil`. The run falls through the guard, spends all 100 attempts, exits the loop normally, and hands the closed body to `SaveStreamAtomic`, which writes a zero-byte file and reports success. Silent truncation, not a failure. The 98.9s is the 100 attempts of real backoff on the way there.

After `ddc52dd` those same two mutations fail in 0.486s and 0.779s.

api/ftp/ftp.go
jisung-02: Done in `ddc52dd`, following `api/event/event.go:30` including the comment form:

```go
// A brief blip recovers sooner than the flat one-second retry did, while
// downloadMaxAttempts still spans roughly the same ~99s.
// var, not const, so tests can shorten them.
var (
	initialDownloadRetryDelay = 250 * time.Millisecond
	maxDownloadRetryDelay     = time.Second
)
```

`downloadMaxAttempts` stays `const` — it is a budget, not a delay, and no test needs to move it.

The swap helper is shaped like `swapGapFillVars`:

```go
func swapDownloadRetryDelays(initial, max time.Duration) func() {
	oi, om := initialDownloadRetryDelay, maxDownloadRetryDelay
	initialDownloadRetryDelay, maxDownloadRetryDelay = initial, max
	return func() { initialDownloadRetryDelay, maxDownloadRetryDelay = oi, om }
}
```

| | before | after |
| --- | --- | --- |
| `TestFetchFromURLToFile_StopsAtTheAttemptBudget` | 16.80s | 0.06s |
| `api/ftp` | 24.4s | 7.9s |

That is back under the 7.46s the package took before this branch, with the new mixed-class row included.

One thing you did not ask for: `TestBackoffDelay` was passing these two names into `backoffDelay` while asserting hardcoded 250ms/500ms/1s. That was harmless while they were `const`; as `var` it is a test asserting fixed arithmetic against globals another test can write. It now passes `250*time.Millisecond, time.Second` literally — it is testing `backoffDelay`'s math, not the constants.

On the 503 observation: agreed, and it is now in the PR body's Not addressed section rather than only here, since it is the same problem the deadline conversion has to solve.

api/ftp/ftp.go
jisung-02: Body updated. The commit table now runs to `ddc52dd`, and the claims that no longer held are gone or corrected:

- "Both call sites pass `maxAttempts=100`, so the cap has to stay at one second" — removed. `throttledMaxAttempts` has its own paragraph, including that the budget is evaluated per response and can collapse below the attempts already spent.
- 64 KiB is now 8 KiB, and the reason given is time rather than memory, naming `client/client.go:50` as why no other bound exists.
- `utils.IsFatalClientError` / `IsRetryLaterStatus`, and `watcher.go` dropping `net/http`, were absent entirely.
- The `var` delays and what they cost the suite.

Not addressed grew from one item to three: `Retry-After` (now pointing at the `Directive` trailer on `6e42210`), the unset client `Timeout` with why its value is not a download-retry decision, and `503 SlowDown` keeping the full budget.

The `ac13e7e` archaeology is in that last item — `100` was a 100-second budget written as a count, next to a since-deleted "may timeout if it exceeds 100 seconds" warning, and this branch is what broke the one-attempt-equals-one-second identity that made the count mean a duration. That is the argument for a deadline over a count, so it belongs where the follow-up is described rather than in a thread.

The Review section now records both rounds, including that `6e42210` is what introduced the defect `e6e5af2` fixed.

geunwoonoh (approved):

PR #299, opened 2026-08-03
fix(websh): disarm the interrupt handler when the read-only terminal ends

jisung-02:
## Background

`OpenReadOnlyTerminal` (`api/websh/websh.go:200`) backs `alpacon websh watch`. It arms `os.Interrupt` and `SIGTERM` with `signal.Notify` and never stops them.

This is the only production `signal.Notify` in the repo without a matching `signal.Stop`. The three other sites — `cmd/tunnel/tunnel.go:141`, `cmd/tunnel/run.go:93`, and `cmd/event/watch.go:78` — already pair them. (`cmd/tunnel/run_test.go:271` also calls `Notify` without `Stop`, but it is a test subprocess whose every exit path is `os.Exit`, which skips defers anyway.)

## Changes

`defer signal.Stop(sigChan)` after the `Notify`, plus moving the signal setup above `checkTerminal` so the defer ordering is right.

**This is hygiene, not a user-visible bug fix.** The goroutine watching the channel does `<-sigChan` once and returns, so with the handler still armed, a second `SIGTERM` arriving between the return and the end of the deferred teardown would land in the buffered channel and be dropped rather than terminating the process. That window is microseconds wide, and reaching it takes two external `kill`s — the tty's Ctrl+C is not one of these signals, because `checkTerminal` puts the terminal in raw mode and Ctrl+C arrives as the byte `0x03` for the separate stdin goroutine to handle.

So the justification is cost, not impact: one line, and the site stops being the odd one out.

### Defer ordering

The `Stop` is registered **before** the `term.Restore` defer, so LIFO runs it **after**. Registering it last instead — the obvious placement — would disarm the handler at the top of the unwind, restoring the default terminate disposition while the terminal is still in raw mode. A signal landing in that window would kill the process before `term.Restore` runs and leave the user's shell needing `stty sane`, which is a window the pre-change code did not have: the armed handler absorbed the signal and teardown finished.

`cmd/event/watch.go:78` orders it the other way and says why in a comment — its teardown is `watcher.Stop()`, a network/goroutine unwind that can genuinely hang, so staying killable beats finishing cleanly. Here the teardown is a `tcsetattr` and a socket close; neither hangs, and `term.Restore` is the one that must not be skipped. Same reasoning, opposite conclusion, so the code carries a comment either way.

The signal setup stays **below** the websocket dial: arming earlier would make a hung handshake uninterruptible for the dialer's 45s timeout.

## Testing

`go build ./...`, `go vet ./...`, `gofmt` clean. `go test -race ./api/websh/... ./cmd/websh/...` passes.

No new test. The behavior needs a PTY, a fake websocket server, and a subprocess to receive real signals, and the window involved is the microseconds of teardown — not something a deterministic test can pin down. For a one-line hygiene change that bar is not worth clearing; verified by reading the exit paths instead.

## Checklist

- [x] Existing tests pass
- [ ] No test covers the behavior — see above

### DIFF
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -210,14 +210,12 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	}
 	defer func() { _ = wsClient.conn.Close() }()
 
-	oldState, err := checkTerminal()
-	if err != nil {
-		utils.CliErrorWithExit("failed to set up terminal: %v", err)
-	}
-	defer func() { _ = term.Restore(int(os.Stdin.Fd()), oldState) }()
-
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+	// Registered before the term.Restore defer so LIFO runs it after: a signal
+	// arriving mid-teardown lands in the buffered channel instead of killing the
+	// process with the terminal still in raw mode.
+	defer signal.Stop(sigChan)
 	go func() {
 		<-sigChan
 		select {
@@ -225,6 +223,12 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 		default:
 		}
 	}()
+
+	oldState, err := checkTerminal()
+	if err != nil {
+		utils.CliErrorWithExit("failed to set up terminal: %v", err)
+	}
+	defer func() { _ = term.Restore(int(os.Stdin.Fd()), oldState) }()
 
 	// In raw mode Ctrl+C is 0x03 — detect it to exit
 	go func() {


api/websh/websh.go
```diff
@@ -218,6 +218,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+	defer signal.Stop(sigChan)
 	go func() {
 		<-sigChan
```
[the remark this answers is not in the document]
jisung-02: Not taking this one, for three reasons.

**The goroutine's lifetime is bounded by the process.** `OpenReadOnlyTerminal` has exactly one caller (`cmd/websh/websh_watch.go:47`), and the process exits as soon as it returns. There is no long-lived process in which this goroutine can "remain stuck indefinitely" — it outlives the function by the microseconds of teardown and then the process is gone.

**Closing `sigChan` would not make the teardown symmetric.** The stdin goroutine right below (`api/websh/websh.go:231`) blocks on `os.Stdin.Read` and has the exact same lifetime; no channel close reaches it. Closing only `sigChan` fixes the one leak that is visible and leaves the identical one next to it, which is worse than leaving both to process exit.

**It costs more than the fix.** A closed channel receive returns immediately, so the goroutine would wake and do `wsClient.Done <- nil` after the function has already returned. Harmless, but suppressing it means the closed-channel handling you mention — several lines added to a one-line change.

Worth being clear about what this PR is: hygiene, not a user-visible bug fix. The window it closes is microseconds wide and unreachable from the tty (raw mode delivers Ctrl+C as `0x03`, not a signal). It is in because it costs one line and makes this the fourth of four production `signal.Notify` sites that pair with `Stop` — not because anyone is hitting it. The goroutine lifetime you flag predates this change and is orthogonal to it.


---

_Edited: the original said "sixth of six" and cited `cmd/event/wait.go`, which is not on `main` — that file lives on an unmerged branch. Corrected count per @geunwoonoh's review._

api/websh/websh.go
```diff
@@ -218,6 +218,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+	defer signal.Stop(sigChan)
```
geunwoonoh, line 218: **Warning:** the defer-ordering hazard is identified in the PR body but not mitigated in the code — and the ordering is arguably inverted for *this* function.

LIFO order on return is `signal.Stop` → `term.Restore` → `conn.Close`. So from the instant the function starts unwinding, signals are back to default disposition (kill) **while the tty is still in raw mode**. A `SIGTERM` in that window now kills the process before `term.Restore` runs, leaving the user's shell needing `stty sane`. Before this PR, that signal was absorbed by the buffered channel and teardown completed.

Compare `cmd/event/watch.go:78-80`, which the PR cites as precedent:

```go
signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
// Runs before watcher.Stop, so a second Ctrl+C still kills a hung teardown.
defer signal.Stop(sigChan)
```

That comment earns the ordering: `watcher.Stop()` is a network/goroutine teardown that genuinely can hang, so staying killable is worth more than finishing cleanly. In `OpenReadOnlyTerminal` the teardown is `term.Restore` (a `tcsetattr` syscall) plus `conn.Close` (a socket close) — neither realistically hangs, and `term.Restore` is precisely the one you don't want skipped. The trade-off leans the other way here.

Both windows are microseconds and unreachable from the tty (raw mode delivers Ctrl+C as `0x03`), so this is not a merge blocker. But the PR body says *"Nothing in the code says so — a later edit that hoists this defer would silently undo it"* — and then doesn't add the comment. For a PR whose whole thesis is *match what the other sites do*, the comment is the other half of the pattern and costs one line:

```suggestion
	// Registered after the conn.Close and term.Restore defers, so LIFO disarms first.
	defer signal.Stop(sigChan)
```

geunwoonoh (commented):
## Review: hygiene change is correct — two documentation gaps

**Stack**: golang · **Changes**: +1 -0 in 1 file

Verified locally: `go build ./...`, `go vet`, `gofmt -l`, and `go test -race ./api/websh/... ./cmd/websh/...` all clean.

### Summary

Adds `defer signal.Stop(sigChan)` after the `signal.Notify` in `OpenReadOnlyTerminal` (`api/websh/websh.go:221`), the backing function for `alpacon websh watch`.

**The code change is correct and safe.** I traced all three exit paths (stdin `0x03`, external signal, `readFromServer` error) and none regress.

### Critical issues (0)

None.

### Warnings (2)

**1. `api/websh/websh.go:221` — the defer-ordering hazard is identified in the PR body but not mitigated in the code, and the ordering is arguably inverted for *this* function.** See inline comment.

**2. The PR description cites a file that does not exist on `main`.**

The body claims *"The five other sites — `cmd/tunnel/tunnel.go:142`, `cmd/tunnel/run.go:94`, `cmd/event/watch.go:80`, and both in `cmd/event/wait.go`"*, and the inline reply says this makes it *"the sixth of six"*. Verified against `main`:

```
main:api/websh/websh.go:220
main:cmd/event/watch.go:78
main:cmd/tunnel/run.go:93
main:cmd/tunnel/tunnel.go:141
main:cmd/tunnel/run_test.go:271   (test)
```

`cmd/event/wait.go` exists only on an unmerged branch (`61777a2`, `033f9cb`, …), not in this PR's base. Real count: **3 other production sites**, making this the **fourth of four**, not the sixth of six. The `cmd/event/wait.go:96` reference in the "Not using the goroutine-side shape" paragraph is likewise dangling.

The PR body is the durable artifact here — worth correcting before merge.

### Suggestions (1)

**1. `api/websh/websh.go:224-227, 236-239, 304-307` — the non-blocking send to `Done` appears three times.** See inline comment.

### On the Copilot thread

The rebuttal at `websh.go:223` is correct and I'd close the thread. `OpenReadOnlyTerminal` has exactly one caller (`cmd/websh/websh_watch.go:47`) and the process exits immediately after it returns, so "stuck indefinitely" has no reachable instance. The point that the adjacent stdin goroutine blocked on `os.Stdin.Read` (`websh.go:234`) has identical lifetime and cannot be unblocked by any channel close is the decisive one — closing `sigChan` would fix the visible half of a symmetric pair and leave the other half, which reads as a fix but isn't one.

---

### Out-of-scope observations

Also asked to check conventions, duplication, dead code, and performance. **None of the below is introduced by this PR** — all pre-existing in `api/websh/websh.go`. Listed separately so it doesn't muddy the merge decision, and not anchored inline because these lines are outside the diff.

**Unused code: none found.** Every exported function in the package has a production caller; `BuildSessionRequest` is called at `websh.go:164` and covered by tests.

| # | Location | Finding |
|---|---|---|
| O1 | `websh.go:275` | **Copy-paste bug.** A `checkTerminal()` failure in `runWsClient` reports `"websocket connection failed %v"`. `OpenReadOnlyTerminal:215` gets this right (`"failed to set up terminal: %v"`) — its twin never got the fix. |
| O2 | `websh.go:264-269` | `err = wsClient.runWsClient(); if err != nil { return err }; return nil` → `return wsClient.runWsClient()`. |
| O3 | `websh.go:201-217` vs `252-262`+`273-277` | ~15 duplicated lines: struct init, dial, `defer conn.Close`, `checkTerminal`, `defer term.Restore`. A shared `dialAndPrepare` helper would collapse both — and would have prevented O1. |
| O4 | `websh.go:336` | **Perf.** `case <-time.After(5*time.Millisecond)` inside a `for` loop allocates a fresh `Timer` per iteration — ~200 heap timers/sec for the life of an idle session, each pinned in the runtime heap until it fires. Standard fix: hoist a `time.NewTicker` with `defer ticker.Stop()`. |
| O5 | `websh.go:310` | **Perf.** `fmt.Print(string(message))` copies every frame's bytes into a string, then routes through `fmt`'s reflection path. `os.Stdout.Write(message)` is one syscall, zero copies — this is the hot path for terminal output. |
| O6 | `websh.go:320, 323, 340` | **Correctness risk.** These do *blocking* sends to `Done` while `:225, :237, :305` use the non-blocking `select`/`default` form. `Done` is buffered size 1 — if two goroutines race to report, the blocking sender parks forever. Inconsistent within one file. |
| O7 | `cmd/websh/websh.go:307` | `_ = websh.OpenNewTerminal(...)` discards the error; `cmd/websh/websh_join.go:31` handles it. Divergent handling of the same call. |

O1 and O6 look worth a follow-up issue; O4/O5 are cheap wins if that file is touched again.

### Highlights

- Genuinely minimal — one line, no collateral changes.
- The PR body pre-empts the obvious objections (why not the `os.Exit` goroutine shape; why no test) rather than leaving reviewers to ask.
- The Copilot rebuttal reasons from the actual call graph instead of accepting a plausible-sounding suggestion.
- Correctly declines to add a test that would need a PTY, a fake websocket server, and a subprocess to observe a microsecond window.

---

**Verdict**: COMMENTED

- [x] Code follows stack conventions (gofmt, vet clean; matches existing `Notify`/`Stop` pairing)
- [x] Existing tests pass (`go test -race ./api/websh/... ./cmd/websh/...` → ok)
- [ ] No test for the new behavior — justified in the PR body, agreed
- [x] No security implications
- [ ] PR description references a file absent from `main` (Warning 2)
- [ ] Defer-ordering rationale not captured in a comment (Warning 1)

Nothing here blocks merge. The code side is a single comment line; the rest is a PR-body edit.

api/websh/websh.go
```diff
@@ -218,6 +218,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
...
 	go func() {
 		<-sigChan
 		select {
```
geunwoonoh, line 221: **Suggestion:** the non-blocking send to `Done` appears three times in this file — here, at `:236-239` (stdin goroutine), and at `:304-307` (`readFromServer`).

A small method would remove the repetition and make the mixed usage in this file harder to get wrong — `readUserInput:320,323` and `writeToServer:340` do *blocking* sends to the same size-1 buffered channel, so if two goroutines race to report, the blocking sender parks forever:

```go
func (wsClient *WebsocketClient) done(err error) {
	select {
	case wsClient.Done <- err:
	default:
	}
}
```

Pre-existing, not introduced by this PR — noting it while the file is open.

api/websh/websh.go
jisung-02: Taken, and you talked me out of the suggestion itself.

The comment you proposed would have documented the ordering you just argued against. If the trade-off leans the other way for this function — and it does, for the reason you give: `tcsetattr` and a socket close don't hang, and `term.Restore` is the one step that must not be skipped — then the fix is the ordering, not a comment explaining the wrong ordering.

So `0e1806e` moves the signal setup above `checkTerminal` instead. `defer signal.Stop` is now registered *before* the `term.Restore` defer, so LIFO runs it *after*, and the unwind is `term.Restore` → `signal.Stop` → `conn.Close`. Signals stay armed through the raw-mode restore; a `SIGTERM` in that window lands in the buffered channel and teardown finishes. The `stty sane` window you describe never opens.

```go
	sigChan := make(chan os.Signal, 1)
	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
	// Registered before the term.Restore defer so LIFO runs it after: a signal
	// arriving mid-teardown lands in the buffered channel instead of killing the
	// process with the terminal still in raw mode.
	defer signal.Stop(sigChan)
```

One thing the move buys that wasn't in the original reasoning: before this change, `checkTerminal` ran first, so there was a window between `term.MakeRaw` and `signal.Notify` where a real `SIGINT` hit the default disposition and killed the process with the tty already raw. That window is now closed too — it's the same hazard you identified, one step earlier in the function.

The setup deliberately stays *below* the websocket dial. Arming earlier would make a hung handshake uninterruptible for the dialer's 45s `HandshakeTimeout`.

Separately: you were right about the site count. `cmd/event/wait.go` is not on `main` — it lives on an unmerged branch, and I cited it. It's three other production sites, not five, so this is the fourth of four. Corrected in the PR body and in the reply above.

api/websh/websh.go
jisung-02: Agreed, and the blocking half is the part worth fixing — the helper is what makes it cheap.

Filed as #301 with your `done(err)` shape as the proposed fix, plus the path that actually reaches it: the remote side closes → `readFromServer` sends and returns → `runWsClient` drains the slot and returns → `OpenNewTerminal`'s deferred `conn.Close` makes `writeToServer`'s next `WriteMessage` fail, so it sends and takes the slot → `readUserInput` parks forever on its send when stdin finally yields. It can also wedge one step earlier on `inputChan <- string(char)`, which is size-1 as well.

Today it's bounded — both entry points exit the process right after returning, so the parked goroutines die with it. The reason to fix it is that the shutdown contract only holds because the caller happens to exit, and four sends to one channel currently follow two different rules.

Keeping it out of this PR: it's pre-existing, and it's a behavior change to three call sites this diff doesn't touch. The `inputChan` send needs a different answer anyway — dropping a keystroke silently is worse than not sending it, so that one wants a shutdown signal rather than a `default:` branch.

Also filed #300 while the file was open: `runWsClient` reports a `checkTerminal` failure as `websocket connection failed`, so "stdin is not a tty" prints as a network error.

geunwoonoh (approved):
## Re-review at `0e1806e` — both findings resolved, Warning 1 resolved better than proposed

**Changes since `b2953e6`**: +10 −6 in 1 file

---

### Warning 1 (defer ordering) — fixed by reordering, not by commenting

I had suggested adding a comment documenting the LIFO placement. You correctly pointed out that this would have documented the ordering I'd just argued against, and moved the signal setup above `checkTerminal` instead. Verified in `api/websh/websh.go:213-231`:

| | Registration order | LIFO unwind |
|---|---|---|
| `b2953e6` | `conn.Close` → `term.Restore` → `signal.Stop` | `signal.Stop` → `term.Restore` → `conn.Close` |
| `0e1806e` | `conn.Close` → `signal.Stop` → `term.Restore` | **`term.Restore` → `signal.Stop` → `conn.Close`** |

Signals now stay armed across `term.Restore`, so a `SIGTERM` mid-teardown lands in the buffered channel and the restore completes. The `stty sane` window is closed rather than documented. This is the right call and it's the stronger fix.

I checked the residual window between `signal.Stop` and `conn.Close`: the tty is already restored by then, and a skipped `conn.Close` is harmless since process exit drops the fd. Nothing left there.

Also confirmed the secondary window you identified in the reply — pre-change, `term.MakeRaw` ran before `signal.Notify`, so an external `kill -INT` in that gap died with the tty raw. The move closes it. Accurately described (a tty Ctrl+C isn't in scope, since raw mode delivers `0x03`).

### Warning 2 (PR body accuracy) — fixed

Now reads "three other sites" with `cmd/tunnel/tunnel.go:141`, `cmd/tunnel/run.go:93`, `cmd/event/watch.go:78` — all three line numbers match `main`. The dangling `cmd/event/wait.go:96` paragraph is gone. The added `run_test.go:271` carve-out checks out: every exit path in that helper is `os.Exit` (`exit 7`/`exit 8`), so defers wouldn't run regardless.

### Suggestion 1 (`Done` helper) — correctly deferred to #301

With #300 filed for the `runWsClient` error-message bug. Keeping a three-call-site behavior change out of a diff that touches none of them is the right scoping, and the point that `inputChan` needs a shutdown signal rather than a `default:` branch is correct — a silent keystroke drop would be worse than the current park.

---

### Claims spot-checked

| Claim | Result |
|---|---|
| `DefaultDialer` 45s `HandshakeTimeout` — why signal setup stays below the dial | Confirmed: `gorilla/websocket@v1.5.3/client.go:143` |
| `run_test.go` exits only via `os.Exit` | Confirmed |
| `CliErrorWithExit` skips defers | Confirmed: `utils/log.go:28` → `os.Exit(1)` |
| Follow-ups actually filed | #300 and #301 both open |

### Verification (at `0e1806e`, clean worktree)

```
go build ./...                                  clean
go vet ./...                                    clean
gofmt -l .                                      clean
go test -race ./api/websh/... ./cmd/websh/...   ok (1.447s / 1.847s)
```

`golangci-lint` isn't installed locally, so that gate is CI-only — `go vet` and `gofmt` are clean.

### Remaining nit (non-blocking)

The PR body still says *"the justification is cost, not impact: **one line**"* — the diff is now +10 −6 (a code move plus a 3-line comment and the defer). Purely cosmetic staleness in the description; the reasoning around it is correct.

### Highlights

- Pushed back on my own suggestion with a correct argument and shipped the better fix — the ordering, not a comment rationalizing the wrong ordering.
- The new comment (`:215-217`) states the *mechanism* (registration order → LIFO consequence), which survives refactoring better than a comment restating intent.
- Contrasts with `cmd/event/watch.go` in the PR body instead of blindly copying it — same reasoning, opposite conclusion, both documented.
- Owned the site-count error explicitly rather than quietly editing it.
- Two adjacent pre-existing defects found and filed instead of scope-creeping the diff.

---

**Verdict**: APPROVED

- [x] Both review findings resolved
- [x] Code follows stack conventions (`gofmt`, `go vet` clean; `Notify`/`Stop` now paired at all 4 production sites)
- [x] Existing tests pass under `-race`
- [x] No security implications
- [x] Follow-ups filed for out-of-scope findings (#300, #301)
- [ ] No test for the new behavior — justified (needs PTY + fake ws server + subprocess signals for a microsecond window); agreed

PR #303, opened 2026-08-03
fix: strip terminal control sequences from CLI text output

jisung-02:
## Background

The CLI writes strings received from the API straight to the terminal. When a value carries control characters such as ESC or CR, the terminal reads them as commands rather than text and redraws lines that were already printed, so what appears on screen can differ from the actual data.

These values come from places people type into freely—note bodies, the recommendations an approver leaves, server, group, token and application names—so control characters can find their way in. The web console is unaffected because HTML does not interpret escape sequences, but the CLI hands the bytes to the terminal, which means the same record can look different on the two surfaces.

Table alignment was off for a related reason: a newline inside a value splits one row across several lines and leaves headers reading like `CONTENTPRIVATE`.

## Changes

The stripping happens where values are printed rather than where they are built. Doing it per value was tried before—the helpers have existed for a while, but the approval, audit, log, note and webhook paths added later never picked them up. One pass at the print site covers the tables added from here on as well.

| Location | What changes |
|---|---|
| `utils/output.go` `PrintTable` | Every table cell. All table-rendering commands go through here |
| `cmd/worksession/worksession_output.go` | The values `formatAdjustments` and `formatRecommendations` interpolate (severity, text, scope, server name), and the create, update and use success lines |
| `cmd/worksession/worksession_timeline.go` | The timeline rows and the recordings section |
| `cmd/websh/websh_watch.go` | The four identity fields in the watch header |
| `cmd/server/server_create.go` | The registration token menu and the guide header's platform, server and URL fields |
| `cmd/whoami.go` | Values in the print loop. Not where the line list is built, since it interpolates seventeen fields and the next one added could miss it again |
| `cmd/login.go` | The application name printed once the credential is verified |
| `cmd/username/username_get.go` | The single `username` line |
| `utils/utils.go` | New `SanitizeTerminalText`, with the seven sites that used the same pair routed through it, and a new `StripFormatChars` folded into it |
| `cmd/event/render.go` | Drops its own copy of the format-character pass, now that the shared helper runs it |

The last five rows came out of a second pass. The first sweep was shaped around `fmt.Print` calls, which is why it missed the timeline table—it writes to a `tabwriter` directly and never reaches `PrintTable`—along with the header blocks and the success lines built by helpers. The bundled review is written up in the branch notes.

Order matters. `StripANSIEscapes` has to run first so `ESC [ 2 K` leaves as one unit, and `StripControlChars` then removes the standalone control bytes left behind. Reversed, only the ESC goes and `[2K` stays on screen as text. That ordering was known separately at each call site, so it now lives in one function.

Format characters are removed as well. They carry no control byte, so neither of the two passes above notices them, while a bidirectional override reorders how the rest of the line is drawn—`denied` can reach the screen as `approved`. `cmd/event/render.go` already removed them, but only on that one path, and `PrintTable`, which this change establishes as the single place tables reach the terminal, did not. `StripFormatChars` runs innermost, since one of these buried inside an escape sequence would otherwise stop the sequence from matching and leave `[2K` on screen as text.

Order matters once more inside `formatRecommendations`, and for the same reason in `activeWorkSessionSetMessage`. The `info` fallback used to run on the raw severity, so a value that the pass empties out slipped past the check and printed as `[]`; the description check has the same shape, and an empty result now drops the parentheses with it. Both checks sit on the stripped value, which is the one that reaches the screen.

## JSON output

`encoding/json` escapes only what JSON requires, which is everything below U+0020. DEL, the C1 range and format characters all pass through untouched, and a terminal in 8-bit mode reads a C1 byte the same way it reads `ESC [`. `escapeJSONControls` in `utils/output.go` rewrites them as `\uXXXX` on all four JSON paths (`FormatJSON`, the `PrintTable` JSON branch, and both `PrintJson` branches).

Escaping rather than stripping, because this is the machine-facing path: a decoder reads the escape back as exactly the rune the API sent, so scripts and agents see no change in value. Byte-for-byte the output does differ from before for values carrying any of these—the escape is longer than what it replaces—so a consumer diffing raw bytes rather than decoded values will notice.

Two details the first version of this function got wrong, both raised in review. `json.Indent` does not validate UTF-8, so a body can carry C1 as a bare `0x80`-`0x9f` byte rather than its two-byte encoding; the scan now decodes runes, which is also what keeps the continuation bytes of ordinary multi-byte runes—`가` encodes as `ea b0 80`—from being mistaken for one. And JSON escapes are UTF-16 code units, so the format characters past the BMP need a surrogate pair rather than a single escape.

`PrettyJSON` is left alone. It feeds the temp-file edit flow rather than a terminal.

## Left as they are

Seven places keep their current handling. The reasons are easy to miss on review, so they are written down here.

| Location | Why it stays |
|---|---|
| `cmd/websh/websh_records.go` `sanitizeRecord` | Turns control characters into spaces rather than dropping them, then collapses the run of whitespace. Dropping them would glue neighbouring words together (`reboot` and `db-01` arriving as `rebootdb-01`). Both passes still run, so the value ends up the same |
| `cmd/worksession/worksession_timeline.go` `recordingPreview` | The two passes sit apart on purpose. The code between them reproduces what a `\r` does to a line, and folding them into one call would remove the `\r` before that runs |
| `cmd/worksession/worksession_recording.go` `printRecordingContent` | Only the escape pass runs. `SanitizeTerminalText` also drops `\n`, which would flatten a whole recording onto one line. Keeping the line breaks needs a helper of its own, left for a later change |
| `cmd/server/server_create.go` install and register snippets | Multi-line shell the operator pastes. The pass drops newlines, which would break the commands themselves |
| `cmd/server/server_create.go` new token key | The API generates it rather than a person, and altering it would hand the operator a credential that no longer works |
| `utils/output.go` `PrintJson` default output | The `\n` and `\t` escapes are turned back into real whitespace so a multi-line value stays readable; certificate text would otherwise arrive as one long line. This output is meant to be read, not parsed—`--output json` is the machine-facing path and leaves the escapes alone |
| `utils/log.go` `CliError`, `CliErrorWithExit`, `CliWarning`, `CliInfo`, `CliSuccess` | The largest surface the print-site pass does not reach—some six hundred call sites, over half of them interpolating an `err` that `parseAPIError` builds from the response body, which is the same source as the values above. `cmd/event/` already calls `strip(err.Error())` by hand in four places for exactly that reason. It is not folded in here because `SanitizeTerminalText` drops `\n`, so a multi-line server error would collapse onto one line—the same reason the recording row above stays. It waits on the line-preserving helper that row already asks for |

Going further—one primitive taking a replacement rune and a set to leave alone, behind wrappers named for intent—would turn the first and third rows into a single call each. It is not here on YAGNI grounds: only one call site would want a non-default wrapper today, so it adds more lines than it removes. The point to do it is when the third row is picked up.

Paths that deliberately mirror a terminal—`websh` and the remote command output in `exec`—are out of scope here. Passing escapes through is the feature there.

## Behaviour change

Newlines and tabs inside table cells are removed, so a multi-line value collapses onto one line. That resolves the alignment issue above, but it is a change for scripts parsing table output. `--output json` is where the value comes back whole; `describe` is not, since work-session and approval describe render through the same table path.

On the JSON side, ordinary format characters are escaped along with the rest—a zero-width joiner inside an emoji sequence now comes back as `\u200d`. The decoded value is unchanged.

## Testing

The tests assert on the value rather than on appearance—not "the table looks intact" but "no control byte remains in the cell".

- `utils/output_test.go` — four vectors as sub-tests (7-bit CSI + CR, 8-bit CSI, a lone newline, DEL). With only a 7-bit ESC the regex alone would pass, so the 8-bit CSI and DEL paths that `StripControlChars` covers could drop out and still stay green. The 8-bit vector is written as the escaped rune (`\^[`): a lone `\x9b` byte is not valid UTF-8, so `strings.Map` replaces it with `U+FFFD` and keeps it, which would assert replacement instead of removal. API responses are UTF-8 JSON, so the escaped rune is also the form that actually arrives. Each case also pins the join: the separator drops, it does not become a space
- `utils/output_test.go` — the JSON paths, one case per vector, each decoding the output back and comparing it with what went in, so the escape is proved lossless rather than assumed. A second case feeds `PrintJson` a body carrying the raw bytes, which is what a server may legitimately send
- `utils/utils_test.go` — `StripFormatChars` on its own, and `SanitizeTerminalText` for both the bidirectional override and a format character buried inside an escape sequence, which is the case that pins the ordering
- `utils/output_test.go` — the table path drops format characters, and the JSON path escapes them, including one past the BMP so the surrogate pair is covered. A further case feeds `PrintJson` a bare C1 byte, the form `json.Indent` lets through
- `cmd/worksession/worksession_output_test.go` — four advisory fields (recommendation text and severity, server name, scope name) plus a case that a newline inside a value does not forge an extra advisory line, and the create, update and use success lines with the same check
- `cmd/worksession/worksession_timeline_test.go` — the timeline row and the recordings section, pinning that a newline in a server or user name does not forge a second entry
- `cmd/websh/websh_test.go` — the watch header, pinning that a newline does not forge an identity line
- `cmd/server/server_create_test.go` — the token menu, pinning that a newline in a token name does not forge a numbered choice the operator can pick, and the guide header fields
- `cmd/whoami_test.go` — four fields plus a case that a newline does not forge an identity line the account does not have (`Role: owner`)
- `cmd/username/username_test.go` — drives the whole command through httptest

Each test was confirmed to fail when the stripping is reverted.

The new tests capture stdout, and five near-copies of that capture had grown across `utils`, `cmd`, `cmd/username` and `cmd/worksession`. Three were wrong in ways the others were not: none closed the read end of the pipe, the timeline copy took no `*testing.T` and so could not restore stdout if the body panicked, and it read the pipe only after closing the writer, which would block inside the function under test once output passed the 64KB buffer. `pkg/testutil` now holds one `redirect` primitive with `CaptureStdout` and `CaptureOutput` over it, and the five copies call it. The consolidation is what stops a sixth copy arriving with its own variant.

Verification: `go test -race ./...` no failures, `go vet ./...` no output, `golangci-lint run ./...` `0 issues.`, `gofmt -l` no output.


cmd/whoami_test.go
```diff
@@ -317,3 +322,54 @@ func TestIsWorksessionRequired(t *testing.T) {
...
+		var buf bytes.Buffer
+		_, _ = io.Copy(&buf, r)
+		done <- buf.String()
+	}()
+
+	fn()
+	_ = w.Close()
+	os.Stdout = old
+	return <-done
+}
```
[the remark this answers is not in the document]
jisung-02: Fixed in 236e51a. Checking whether the same gap was elsewhere turned up five near-copies of this capture across `utils`, `cmd`, `cmd/username` and `cmd/worksession`, three of them wrong in ways the others were not, so the helper moved to `pkg/testutil` instead of being patched here.

`redirect` returns a `stop` guarded by `sync.Once`, which is what lets `t.Cleanup` call the same function on the panicking path: it restores the original stdout and closes the writer, and the cleanup then closes the reader. `CaptureStdout` and `CaptureOutput` sit on top of it, and this file now calls `testutil.CaptureStdout`.

cmd/username/username_test.go
```diff
@@ -29,3 +36,35 @@ func TestSetUsernameErrorText(t *testing.T) {
...
+	done := make(chan string, 1)
+	go func() {
+		var buf bytes.Buffer
+		_, _ = io.Copy(&buf, r)
+		done <- buf.String()
+	}()
+	usernameGetCmd.Run(usernameGetCmd, nil)
+	_ = w.Close()
+	os.Stdout = old
+	got := <-done
```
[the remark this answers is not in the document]
jisung-02: Fixed in 236e51a. This block is now a single `testutil.CaptureStdout(t, ...)` call, so the restore and both closes run from `t.Cleanup` whether the body returns or panics.

The same review turned up a worse instance of this in `cmd/worksession/worksession_timeline_test.go`: it took no `*testing.T` at all, and it read the pipe only after closing the writer, which would block inside the function under test once output passed the pipe buffer. The shared helper drains in a goroutine started before the body runs, so that one is gone too.

utils/utils.go
```diff
@@ -498,6 +496,12 @@ func StripControlChars(s string) string {
 	}, s)
 }
 
+// SanitizeTerminalText strips escape sequences before the standalone control
+// bytes; reversed, ESC alone would go and "[2K" would stay on screen as text.
+func SanitizeTerminalText(s string) string {
```
junho226, line 515: Warning: `SanitizeTerminalText` does not remove Unicode format characters, so bidi overrides pass through. `IsControlRune` is `r < 0x20 || (r >= 0x7f && r <= 0x9f)`, and U+202A-U+202E / U+2066-U+2069 are outside that range.

Confirmed with a throwaway test in this package:

```
in  runes="denied\u202e\u2066 approved"
out runes="denied\u202e\u2066 approved"   <- unchanged
bidi control U+202E survived SanitizeTerminalText
bidi control U+2066 survived SanitizeTerminalText
```

What makes this worth fixing here rather than later: `cmd/event/render.go` in this same repo already strips exactly these, with this comment —

> format characters go entirely: a bidi override could render "denied" as "approved".

So the repo currently has it backwards. The one narrow `event watch` path is hardened against bidi, while `PrintTable` — which every table-rendering command in the CLI goes through, and which this PR establishes as the single choke point — is not. The threat this PR opens with ("what appears on screen can differ from the actual data") is precisely what a bidi override does, and the values named as sources are the same free-text fields.

```suggestion
// SanitizeTerminalText strips escape sequences before the standalone control
// bytes; reversed, ESC alone would go and "[2K" would stay on screen as text.
// Format characters go too: a bidi override reorders rendering without carrying
// a control byte, so "denied" can display as "approved".
func SanitizeTerminalText(s string) string {
	return StripControlChars(StripANSIEscapes(StripFormatChars(s)))
}
```

(`StripFormatChars` being a `strings.Map` over `unicode.Is(unicode.Cf, r)`, which would also let `cmd/event/render.go` drop its own copy of that pass.)

On the JSON side the same runes get through `escapeJSONControls`, but there the fix is to escape rather than strip — `\u202e` keeps the value round-tripping losslessly for a decoder while making it inert in a terminal, which is exactly the split this PR already argues for.

utils/output.go
```diff
@@ -55,6 +55,26 @@ func (a NextAction) PlainText() string {
...
+// only below U+0020, so these reach the terminal as raw bytes and 8-bit CSI opens a
+// control sequence there. Valid JSON carries them inside strings alone, so scanning
+// bytes needs no parser, and a decoder reads the escape back as the same rune.
+func escapeJSONControls(b []byte) []byte {
+	out := make([]byte, 0, len(b))
+	for i := 0; i < len(b); i++ {
+		switch {
+		case b[i] == 0x7f:
+			out = append(out, `\u007f`...)
+		case b[i] == 0xc2 && i+1 < len(b) && b[i+1] >= 0x80 && b[i+1] <= 0x9f:
```
junho226: Suggestion: this matches only the UTF-8 encoding of C1 (`0xC2` + `0x80..0x9F`), but `PrintJson` runs `json.Indent` over the raw HTTP response body, and `json.Indent` does not validate UTF-8. Confirmed:

```
body := []byte("{\"name\":\"a\x9bmK\"}")
json.Indent accepted it
bare 0x9b survived escapeJSONControls: "{\n  \"name\": \"a\x9bmK\"\n}"
```

So a non-UTF-8 response body still delivers a bare 8-bit CSI to the terminal on that path. Reachability is low — the server emits UTF-8, so U+009B arrives as `0xC2 0x9B` — and scoping to valid UTF-8 is a defensible decision. The reason I am raising it is that the function comment states the 8-bit CSI hole as closed:

> these reach the terminal as raw bytes and 8-bit CSI opens a control sequence there

Either add the bare-byte case (`b[i] >= 0x80 && b[i] <= 0x9f` when it is not a valid UTF-8 lead byte) or narrow the comment to say the UTF-8-encoded form, so the next reader does not take a stronger guarantee than the code gives.

junho226 (dismissed):
## Summary

Moves the sanitization to the print site rather than the value site, which is the right call: the helpers already existed and the approval/audit/log/note/webhook paths added later never picked them up. One pass at `PrintTable` also covers every table added from here on.

**Verified locally on the PR head:**

```
go build ./...   → clean
go vet ./...     → clean
gofmt -l .       → clean
go test ./utils/... ./cmd/... ./pkg/...  → all pass
```

I also checked the two things most likely to be wrong:

- **`cmd/event/render.go` is a pure refactor, not a regression.** It keeps its own `strings.Map` for the `unicode.Cf` pass and only routes the trailing pair through `SanitizeTerminalText`. The bidi handling that file added is intact.
- **`escapeJSONControls`'s parser-free premise holds.** DEL and the C1 block can appear in valid JSON only inside string literals (structural characters and inter-token whitespace are all ASCII), so a byte scan is sound and inserting `\uXXXX` there stays valid JSON.

## Warnings (1)

**Bidi overrides survive `SanitizeTerminalText`, so they reach every table cell.** Inline comment on `utils/utils.go`.

## Suggestions (1)

**`escapeJSONControls` only matches the UTF-8-encoded C1 form.** Inline comment on `utils/output.go`.

## Highlights

- **The "Left as they are" table is the most useful part of the PR.** Six call sites with the reason each keeps its current handling, including the non-obvious ones: `sanitizeRecord` maps control bytes to spaces so `reboot`+`db-01` does not become `rebootdb-01`; `recordingPreview` keeps the two passes apart because the code between them reproduces what `\r` does to a line. A reviewer would have flagged both as misses.
- **Escaping rather than stripping on the JSON path**, with the byte-level consequence stated plainly (a consumer diffing raw bytes rather than decoded values will see a difference). That is the correct split between the human surface and the machine surface.
- **The test rationale for the 8-bit vector** is the kind of thing that usually goes unwritten: with only a 7-bit ESC the regex alone passes, so `StripControlChars` could silently drop out and stay green. Writing the vector as an escaped rune rather than a lone `\x9b` byte is also right, since `strings.Map` would replace an invalid UTF-8 byte with U+FFFD and the test would assert replacement instead of removal.
- **`pkg/testutil/capture.go`** makes `stop` idempotent and registers it with `t.Cleanup`, so a panicking test cannot leave the pipe installed and swallow every later test's output in the package. Easy to get wrong.
- The YAGNI note on not generalizing into one primitive taking a replacement rune plus a keep-set, with the trigger named (when the third row is picked up), is the right amount of restraint.

---

**Verdict**: APPROVED — neither finding blocks a clear net improvement.

- [x] Build, vet, gofmt, tests clean
- [x] No regression in the existing `cmd/event/render.go` sanitization
- [x] Behavior change (newlines/tabs collapsing in table cells) called out for script consumers
- [x] JSON path is lossless on decode
- [ ] Unicode format characters (bidi) covered — see below

utils/utils.go
jisung-02: Fixed in a04efd2. `StripFormatChars` now runs innermost in `SanitizeTerminalText`, so `PrintTable` and every other call site drop these. Running it first also covers a format character buried inside an escape sequence, which would otherwise stop `ansiEscapeRE` from matching and leave `[2K` on screen as text; that case is pinned in `TestSanitizeTerminalText`.

`cmd/event/render.go` drops its own copy of the pass. Its `strings.Map` stays for the whitespace mapping, which is the opposite of the table rule: event lines turn line breaks into spaces so a multi-line server error does not glue into one word, while a table cell drops them so a value cannot spill into a second row.

On the JSON side these are escaped rather than stripped, as you suggested, in d48b592. Thank you for the throwaway test - it made the gap concrete.

utils/output.go
jisung-02: Fixed in ef49f88 rather than narrowing the comment. The branch rests on no control byte reaching the terminal, so the code seemed the side that should move.

The scan now decodes runes instead of matching byte pairs: a plain 0x80-0x9f test on raw bytes would also catch the continuation bytes of valid multi-byte runes, since `가` encodes as `ea b0 80`. Run against the previous implementation over 400k inputs, the output is byte-identical for every valid UTF-8 case and differs only where the input is not valid UTF-8.

d48b592 then extends the same function to format characters, which needed a surrogate pair for the ones past the BMP.

Thank you for the review.

utils/utils.go
```diff
@@ -498,6 +497,25 @@ func StripControlChars(s string) string {
...
+			return -1
+		}
+		return r
+	}, s)
+}
+
+// SanitizeTerminalText strips escape sequences before the standalone control
+// bytes; reversed, ESC alone would go and "[2K" would stay on screen as text.
+// Format characters go first, so one buried in a sequence cannot break the match.
+func SanitizeTerminalText(s string) string {
```
junho226, line 515: Suggestion: the largest surface this PR's choke point does not reach is missing from the "Left as they are" table.

It is the `CliError` / `CliErrorWithExit` / `CliWarning` / `CliInfo` / `CliSuccess` family in `utils/log.go`. It is in neither the changes table nor the exclusions table.

```
utils.Cli{Error,Warning,Info,Success} call sites (non-test)   634
  of those, ones interpolating err                            376
Does that path go through SanitizeTerminalText                no (log.go unchanged)
```

`err` is what `parseAPIError` builds from the server response body, which is the very source this PR points at in its opening. The evidence is that `cmd/event/` is already doing `strip(err.Error())` by hand in six places. Redrawing `Failed to delete X` as `Deleted X` by winding back with CR makes the error line, if anything, the more valuable target.

**I am not asking you to fix it here.** `SanitizeTerminalText` removes `\n`, so a multi-line server error gets flattened onto one line, and that is the very reason the body has already written down for `printRecordingContent` — *"Keeping the line breaks needs a helper of its own, left for a later change"*. That deferred helper is also the precondition for this side.

So the request is **one row added to the table**. That 376 places are not covered, and that it is handled together once the line-preserving helper exists. This PR creates its value by writing down the boundaries, and the single biggest boundary is the one not written down.

junho226 (approved):
## Re-review — both points fixed, and I ran them

| Last round | Resolution | What I checked |
|---|---|---|
| `SanitizeTerminalText` let Cf through, so bidi overrides survived | `a04efd2` — `StripFormatChars` added **innermost** | `SanitizeTerminalText("denied<U+202E><U+2066> approved")` -> `"denied approved"`, zero Cf left |
| `escapeJSONControls` matched only the UTF-8 encoding of C1, so a bare byte passed | `ef49f88` — decodes runes instead of matching byte pairs | `{"name":"a\x9bmK"}` -> `{"name":"a\u009bmK"}` — the exact input from last round |

I also ran a throwaway test in `utils` (removed afterwards) against the whole
affected rune set:

```
round-tripped 203 runes (76 BMP, 127 astral)   <- surrogate-pair path genuinely exercised
NoRawControlSurvives    PASS   <- no surviving byte in 0x7f-0x9f
BareC1Byte              PASS
SanitizeDropsBidi       PASS
FormatCharInsideEscape  PASS   <- "a\x1b<U+200B>[2Kb" -> "ab"
```

All 203 Cf/C1/DEL runes decode back to the value that went in, so
*"escaping rather than stripping keeps the value a decoder reads back unchanged"*
is measured, not assumed. 127 of them are astral, so the surrogate-pair branch is
covered by real inputs.

`cmd/event/render.go` dropping its own Cf pass in favour of
`utils.SanitizeTerminalText` removes the duplication without regressing the event
path's bidi protection — checked.

Locally: `go build`, `go vet`, `gofmt -l`, and
`go test ./utils/... ./cmd/event/... ./pkg/...` all clean. CI green
(golangci-lint, govulncheck, CodeQL included).

### Suggestions (1)

One inline comment: the largest un-sanitised surface is missing from the
"Left as they are" table. It is a documentation line, not a code request.

### Highlights

- I offered "narrow the comment or fix the code" and you took the code — and then
  went past what I asked by decoding runes rather than widening the byte test.
  Catching that `가` encodes as `ea b0 80`, so a plain `0x80-0x9f` byte check would
  swallow continuation bytes, is the part I had not thought through; diffing 400k
  inputs against the old implementation is the right way to land it.
- `StripFormatChars` running innermost is a reason outside what I raised: a format
  character buried *inside* an escape sequence breaks the `ansiEscapeRE` match and
  leaves `[2K` on screen. I missed that, and it is pinned by a test.
- Keeping `render.go`'s own `strings.Map` and writing down *why* — event lines turn
  breaks into spaces, table cells drop them, opposite rules. That is what stops the
  next person from "finishing" the consolidation.
- The "Left as they are" table itself. `recordingPreview`'s "the two passes sit
  apart on purpose, the code between them reproduces what a `\r` does" is not
  recoverable from reading the code.
- The YAGNI call with its trigger stated: one call site would want a non-default
  wrapper today, so it adds lines; the point to do it is when the third row is
  picked up.
- Finding that three of the five stdout-capture copies were wrong in different
  ways, and that the timeline one would block inside the function under test past
  64KB. That is a real bug found while fixing something else.

---

**Verdict**: APPROVED — nothing outstanding.

utils/utils.go
jisung-02: Added to the "Left as they are" table, and thank you for framing it as the documentation line it is.

You are right that this is the largest surface the print-site pass does not reach, and right about why it cannot be folded in here: `SanitizeTerminalText` drops `\n`, so a multi-line server error would collapse onto one line. That is the same condition the `printRecordingContent` row already records, so the two are waiting on the same line-preserving helper. The new row says so, and points at `cmd/event/` calling `strip(err.Error())` by hand as the evidence that the need is already felt.

One note on the numbers, since the row now carries them. Counting `Cli{Error,ErrorWithExit,Warning,Info,Success}(` across non-test Go files I get 607 call sites, 339 of them interpolating `err`, rather than 634 and 376 — close enough that we are looking at the same thing, but I could not reproduce your exact figures, most likely a difference in how multi-line calls are matched. The row therefore reads "some six hundred call sites, over half of them interpolating an `err`" rather than quoting a precise count I cannot back.

`utils/log.go` itself is untouched on this branch, so nothing in that surface changes behaviour here—only the boundary is now written down.

PR #308, opened 2026-08-05
fix(websh): let every session goroutine leave teardown

### LINKED ISSUE #300: websh: terminal setup failure is reported as a websocket connection error
## What

`runWsClient` reports a terminal setup failure with the websocket dial's error message.

https://github.com/alpacax/alpacon-cli/blob/main/api/websh/websh.go#L275-L279

```go
func (wsClient *WebsocketClient) runWsClient() error {
	oldState, err := checkTerminal()
	if err != nil {
		utils.CliErrorWithExit("websocket connection failed %v", err)
	}
```

`checkTerminal` only calls `term.IsTerminal` and `term.MakeRaw`. It never touches the connection — by the time `runWsClient` runs, the dial in `OpenNewTerminal` has already succeeded. So the two failures a user can actually hit here are reported as a connection problem:

- stdin is not a tty (piped, redirected, or a CI runner): `websh command should be a terminal`
- `tcsetattr` failed

Both print as `websocket connection failed websh command should be a terminal`, which points debugging at the network instead of at the terminal.

The message looks like a copy of the dial error handler a few lines above (`api/websh/websh.go:263`). `OpenReadOnlyTerminal` already has the correct wording for the same call:

https://github.com/alpacax/alpacon-cli/blob/main/api/websh/websh.go#L227-L230

```go
	oldState, err := checkTerminal()
	if err != nil {
		utils.CliErrorWithExit("failed to set up terminal: %v", err)
	}
```

## Fix

Match `OpenReadOnlyTerminal`:

```go
	utils.CliErrorWithExit("failed to set up terminal: %v", err)
```

Note the format verb too — the current string is missing the `:` separator that the rest of the file uses.

## Scope

One line in `api/websh/websh.go`. No behavior change beyond the message; the exit code stays `1`.

## Found by

Out-of-scope finding from the review of #299. Not part of that PR's diff, so it was left for a separate change.

### LINKED ISSUE #301: websh: blocking sends to the size-1 Done channel park goroutines at teardown
## What

Three of the four sends to `WebsocketClient.Done` block forever once the channel's single buffer slot is taken.

`Done` is created with capacity 1 and has exactly one receiver — the `return <-wsClient.Done` that ends `runWsClient` (`api/websh/websh.go:288`) or `OpenReadOnlyTerminal` (`api/websh/websh.go:249`). That receiver runs once. After it fires, the first goroutine to send fills the buffer and every later send blocks with nobody left to drain it.

`readFromServer` guards against this; the other three sites do not.

https://github.com/alpacax/alpacon-cli/blob/main/api/websh/websh.go#L303-L315

```go
func (wsClient *WebsocketClient) readFromServer() {
	for {
		_, message, err := wsClient.conn.ReadMessage()
		if err != nil {
			select {
			case wsClient.Done <- err:
			default:
			}
			return
		}
```

https://github.com/alpacax/alpacon-cli/blob/main/api/websh/websh.go#L317-L331

```go
func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
	reader := bufio.NewReader(os.Stdin)
	for {
		char, _, err := reader.ReadRune()
		if err != nil {
			if err == io.EOF {
				wsClient.Done <- nil      // blocks if the slot is taken
				return
			}
			wsClient.Done <- err          // blocks if the slot is taken
			return
		}
		inputChan <- string(char)
	}
}
```

https://github.com/alpacax/alpacon-cli/blob/main/api/websh/websh.go#L333-L350

```go
		case <-time.After(time.Millisecond * 5):
			if len(inputBuffer) > 0 {
				err := wsClient.conn.WriteMessage(websocket.BinaryMessage, []byte(string(inputBuffer)))
				if err != nil {
					wsClient.Done <- err  // blocks if the slot is taken
					return
				}
```

## Reachable path

Ordinary session teardown reaches it:

1. The remote side closes the session. `readFromServer` sends the error and returns.
2. `runWsClient` receives it, drains the slot, and returns. `OpenNewTerminal`'s deferred `conn.Close` runs.
3. The closed connection makes `writeToServer`'s next `WriteMessage` fail, so it sends to `Done`. The slot is free again, so this one succeeds and it returns.
4. `readUserInput` is still parked in `ReadRune`. When stdin finally yields EOF or an error, its send finds the slot full and blocks permanently.

`readUserInput` can also wedge one step earlier, on `inputChan <- string(char)` (`api/websh/websh.go:329`), once `writeToServer` has returned — `inputChan` is also capacity 1.

## Impact

Today this is bounded: `websh` and `websh watch` exit the process right after these functions return, so the parked goroutines die with it. Nothing hangs from the user's point of view. The problem is that the shutdown contract only holds because the caller happens to exit — the pattern breaks the moment a session is opened from a longer-lived process, and it already leaves the four sends to the same channel following two different rules.

## Fix

Give the three remaining sends the same non-blocking form `readFromServer` uses. A small helper keeps it honest:

```go
func (wsClient *WebsocketClient) done(err error) {
	select {
	case wsClient.Done <- err:
	default:
	}
}
```

and call `wsClient.done(nil)` / `wsClient.done(err)` at all four sites, including `readFromServer`. `OpenReadOnlyTerminal` has two more inline copies of the same `select`/`default` (`api/websh/websh.go:221-224`, `api/websh/websh.go:239-242`) that the helper would also absorb.

The `inputChan` send is a separate decision — it needs a shutdown signal rather than a dropped write, since dropping a keystroke silently is worse than not sending it. Worth handling in the same change, but not with the same mechanism.

## Found by

Out-of-scope finding from the review of #299, plus @geunwoonoh's `done(err)` helper suggestion on that PR. Not part of that PR's diff, so it was left for a separate change.

jisung-02:
## Background

A websh session runs three goroutines—`readFromServer`, `readUserInput`, and `writeToServer` on the interactive path, and the signal watcher, the Ctrl+C reader, and `readFromServer` on the read-only one. They reported the end of the session by sending into a buffered `Done` channel that was received exactly once. Ordinary teardown filled that slot and the receiver returned, so the next send parked its goroutine for good: `readFromServer` reports the remote close, `writeToServer`'s write then fails into the drained slot, and `readUserInput` blocks when stdin finally yields EOF. The signal watcher parked for a different reason—`signal.Stop` disarms delivery without closing `sigChan`, which was all it waited on. The Ctrl+C reader never looked at the outcome at all and went on consuming stdin after its caller had returned.

The goroutines die with the process today, since both entry points exit right after. The contract should not depend on that. Neither should the terminal: only the read-only path watched for signals, so a SIGTERM to an interactive session killed the process straight from the handler, no deferred restore ran, and the user was left in raw mode until they ran `reset`.

The session outcome was wrong at both ends besides. The proxy closes a user channel with 4000 at the end of a session, and any deliberate close reached the caller as a `*websocket.CloseError`, so typing `exit` printed a red `Error` plus the "report on github issues" footer and exited 1; on the main path the outcome was discarded entirely. Three errors on this path also named the wrong thing: `checkTerminal` failures surfaced with the dial handler's wording, `checkTerminal`'s own message named neither the stream nor the fix, and gorilla reports every rejected upgrade as `bad handshake`.

## Changes

Teardown now rests on one channel that only closes, the outcome it carries is reported rather than exited on, and both entry points tear down the same way.

| Location | What changes |
|---|---|
| `types.go` | `Done chan error` becomes an unexported `done chan struct{}` plus `err error` and `finishOnce sync.Once`. `Header` is unexported with them, since nothing outside the package reads it |
| `finish` | Records the first outcome under a `sync.Once`: writes `err`, then closes `done`. A channel that only closes cannot park a sender, so the hazard is structural rather than resting on the `sync.Once`. A deliberate close is normalized to `nil`; every other close code stays an error |
| `sessionEndCloseCode` | 4000—what the proxy actually closes a user channel with (`sendCloseFrame`, proxy-server `internal/ws/channel.go:337`), and what alpamon sends for the same meaning. Without it the normalization would have passed its tests and changed nothing in production, since a websh session never ends with 1000 |
| `readFromServer`, `readUserInput`, `writeToServer`, `watchInterrupt`, `readCtrlC` | Report through `finish` and return. The watcher selects on `done` as well, so it leaves with the rest of teardown. `readUserInput` also selects on `done` while sending to `inputChan`, since the writer is gone by then |
| `readCtrlC` | The anonymous Ctrl+C reader becomes a named method that checks `done` between reads instead of consuming stdin past the end of the session, and guards on `n > 0` so a `(0, nil)` read cannot re-judge the previous byte |
| `notifySignals`, `runWsClient` | The interactive path installs the same signal watcher the read-only path had, so a SIGTERM ends the session through the normal teardown and the deferred restore runs. Raw mode disables ISIG, so Ctrl+C is still a byte for the remote shell rather than a local signal |
| `runWsClient`, `OpenReadOnlyTerminal` | Wait on `<-done` and return `wsClient.err`, read after the close |
| `newWebsocketClient`, `dial` | The client literal and the dial block, both duplicated in the two entry points, become a constructor and a method. `dial` returns its error rather than ending the process, and the message carries the handshake status the response already returned, sanitized |
| `enterRawMode` | The raw-mode setup duplicated across both entry points becomes one function returning the restore to defer. Its failure returns too, so the `signal.Stop` and `conn.Close` defers still run when the terminal cannot be set up |
| `cmd/websh` | `websh SERVER` surfaces the session outcome instead of discarding it. All three sites exit with `utils.ExitCodeGeneralError` and no bug-report footer: a session ending badly is the network, the server, or the terminal, not a CLI defect to report |
| `ctrlC`, `writeFlushInterval` | The raw-mode `0x03` and the 5ms flush tick become named constants |
| `writeToServer` | Flushes on a `time.Ticker` instead of `time.After`, which was reallocated every pass and restarted on every arriving rune—so a steady stream of input deferred the flush until it stopped, and a pasted block went out as a single `WriteMessage` |
| `readFromServer` | `os.Stdout.Write(message)` instead of `fmt.Print(string(message))`, dropping a copy and a trip through reflection on the path server output floods through |

The non-blocking `select`/`default` the issue proposed was rejected: it covers the `Done` sends but leaves `inputChan` needing a second mechanism, and keeps which outcome wins dependent on arrival order.

## Known limits

Two blocking terminal reads cannot be released by teardown, because nothing but closing stdin releases a read parked in the kernel, and that races the `term.Restore` deferred right above it. Both are documented on the functions.

- `readUserInput` stays in `ReadRune` until the next keystroke. `websh SERVER` waits up to three seconds for its sudo listener after the session ends, and a key pressed in that window is eaten by this goroutine instead of reaching the shell—one character, once.
- `readCtrlC` notices `done` only after the read it is already parked in returns.

One race is left unfixed rather than fixed badly: if `writeToServer` is holding buffered input when the remote closes normally, its failing write can reach `finish` before `readFromServer` reports the close, turning a clean exit into code 1. Making the close win needs a priority rule between outcomes, which contradicts the "keep the first outcome" contract `TestFinish_KeepsTheFirstOutcome` pins. The buffer is normally empty at exit, so the window is small.

## Safety

The status line the new message quotes is written by whoever answered the handshake, so it reaches the terminal through `utils.SanitizeTerminalText` like every other server-supplied string this CLI prints.

## Testing

Every goroutine that could park now has a test that fails if it does—`awaitReturn` gives it two seconds and reports the park instead of hanging. One table drives the five teardown cases: `readUserInput` reporting EOF, `readUserInput` sending to `inputChan`, `writeToServer` waiting to flush, `watchInterrupt` waiting for a signal, and `readCtrlC` waiting for the next byte. `readFromServer` is covered separately, since it ends on its own failing read.

Also covered: EOF reported as a clean end, so Ctrl+D exits 0 rather than 1; `readCtrlC` ending the session on the Ctrl+C byte after a byte that is not one; `finish` keeping the first outcome; `finish` normalizing 4000, 1000 and 1001 while keeping 1006 and a transport failure as errors; and `OpenReadOnlyTerminal` leaving no goroutine behind when raw mode fails after a successful dial. That last one polls the goroutine count inline rather than through `assert.Eventually`, which evaluates its condition in a goroutine of its own and so can never watch the count come back down, and it primes `signal.Notify` first because the first call anywhere starts one process-wide goroutine that never exits.

`dial` and `runWsClient` return their setup errors, so both assert in process and the subprocess helpers are gone.

The normal path is covered too, since a loop that stops correctly is still worthless if it does the wrong thing while running: a keystroke reaches the server, server output reaches the screen, and the dial failure carries the handshake status.

Each was checked by mutation: breaking the guard it covers makes that test and no other fail. Removing the close normalization fails only the normalized cases of `TestFinish_NormalizesARemoteCloseToASuccess`; dropping 4000 from it fails only the `session end` case; breaking the EOF normalization fails only `TestReadUserInput_ReportsEOFAsACleanEnd`; moving a `go` statement back above `enterRawMode` fails only `TestOpenReadOnlyTerminal_LeavesNoGoroutineWhenSetupFails`.

The interactive signal path is the one change with no automated test—driving either entry point past `enterRawMode` needs a pty. Deleting its `go` statement does fail the build, since `sigChan` goes unused, but deleting the `notifySignals` call with it would not.

`gofmt -l .` clean, `go vet ./...` clean, `golangci-lint run ./...` clean, `go test -race -count=1 ./...` passing.

Closes #300
Closes #301

### DIFF
--- a/api/websh/types.go
+++ b/api/websh/types.go
@@ -10,7 +10,7 @@ import (
 )
 
 type WebsocketClient struct {
-	Header     http.Header
+	header     http.Header
 	conn       *websocket.Conn
 	done       chan struct{} // closed once the first outcome is recorded
 	err        error
--- a/api/websh/types.go
+++ b/api/websh/types.go
@@ -2,6 +2,7 @@ package websh
 
 import (
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/alpacax/alpacon-cli/api/types"
@@ -9,9 +10,11 @@ import (
 )
 
 type WebsocketClient struct {
-	Header http.Header
-	conn   *websocket.Conn
-	Done   chan error
+	Header    http.Header
+	conn      *websocket.Conn
+	Done      chan error
+	quit      chan struct{} // closed once the first outcome is reported
+	closeOnce sync.Once
 }
 
 type SessionRequest struct {
--- a/api/websh/types.go
+++ b/api/websh/types.go
@@ -10,11 +10,11 @@ import (
 )
 
 type WebsocketClient struct {
-	Header    http.Header
-	conn      *websocket.Conn
-	Done      chan error
-	quit      chan struct{} // closed once the first outcome is reported
-	closeOnce sync.Once
+	Header     http.Header
+	conn       *websocket.Conn
+	done       chan struct{} // closed once the first outcome is recorded
+	err        error
+	finishOnce sync.Once
 }
 
 type SessionRequest struct {
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -241,6 +241,9 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 		return err
 	}
 	defer func() { _ = wsClient.conn.Close() }()
+	// Whatever this returns through, the goroutines below have to be released.
+	// The sync.Once makes it a no-op once an outcome is already recorded.
+	defer wsClient.finish(nil)
 
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
@@ -248,7 +251,6 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	// arriving mid-teardown lands in the buffered channel instead of killing the
 	// process with the terminal still in raw mode.
 	defer signal.Stop(sigChan)
-	go wsClient.watchInterrupt(sigChan)
 
 	restore, err := enterRawMode()
 	if err != nil {
@@ -256,6 +258,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	}
 	defer restore()
 
+	go wsClient.watchInterrupt(sigChan)
 	go wsClient.readCtrlC()
 	go wsClient.readFromServer()
 
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -275,7 +275,7 @@ func OpenNewTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) 
 func (wsClient *WebsocketClient) runWsClient() error {
 	oldState, err := checkTerminal()
 	if err != nil {
-		utils.CliErrorWithExit("websocket connection failed %v", err)
+		utils.CliErrorWithExit("failed to set up terminal: %v", err)
 	}
 	defer func() { _ = term.Restore(int(os.Stdin.Fd()), oldState) }()
 
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -206,9 +206,14 @@ func newWebsocketClient(header http.Header) *WebsocketClient {
 }
 
 func (wsClient *WebsocketClient) dial(websocketURL string) {
-	conn, _, err := websocket.DefaultDialer.Dial(websocketURL, wsClient.Header)
+	conn, resp, err := websocket.DefaultDialer.Dial(websocketURL, wsClient.Header)
 	if err != nil {
-		utils.CliErrorWithExit("websocket connection failed: %v", err)
+		// The handshake response carries the reason a bad handshake alone never names.
+		status := ""
+		if resp != nil {
+			status = fmt.Sprintf(" (status %s)", resp.Status)
+		}
+		utils.CliErrorWithExit("websocket connection failed: %v%s", err, status)
 	}
 	wsClient.conn = conn
 }
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -272,8 +272,8 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	return wsClient.err
 }
 
-// Handles graceful termination of the websh terminal.
-// Exits on error without further error handling.
+// OpenNewTerminal opens an interactive terminal on the session.
+// A failed dial exits the process instead of returning.
 func OpenNewTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) error {
 	wsClient := newWebsocketClient(ac.SetWebsocketHeader())
 	wsClient.dial(sessionResponse.WebsocketURL)
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -30,8 +30,8 @@ const (
 	writeFlushInterval = 5 * time.Millisecond
 
 	// sessionEndCloseCode is how the proxy closes a user channel at the end of a
-	// session (sendCloseFrame in proxy-server internal/ws/channel.go); alpamon spends
-	// the same 4000 on the same meaning. A websh session never ends with 1000, so
+	// session (sendCloseFrame in proxy-server internal/ws/channel.go); alpamon sends
+	// the same 4000 for the same meaning. A websh session never ends with 1000, so
 	// without this every normal end would be reported as a failure.
 	sessionEndCloseCode = 4000
 )
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -201,8 +201,7 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
 func newWebsocketClient(header http.Header) *WebsocketClient {
 	return &WebsocketClient{
 		Header: header,
-		Done:   make(chan error, 1),
-		quit:   make(chan struct{}),
+		done:   make(chan struct{}),
 	}
 }
 
@@ -214,12 +213,12 @@ func (wsClient *WebsocketClient) dial(websocketURL string) {
 	wsClient.conn = conn
 }
 
-// finish takes effect only once: Done holds one value and is received once, so a
-// second send would park its goroutine for good.
+// finish keeps the first outcome. err is written before done closes, so anyone
+// who saw done can read it.
 func (wsClient *WebsocketClient) finish(err error) {
-	wsClient.closeOnce.Do(func() {
-		wsClient.Done <- err
-		close(wsClient.quit)
+	wsClient.finishOnce.Do(func() {
+		wsClient.err = err
+		close(wsClient.done)
 	})
 }
 
@@ -241,7 +240,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 		select {
 		case <-sigChan:
 			wsClient.finish(nil)
-		case <-wsClient.quit:
+		case <-wsClient.done:
 		}
 	}()
 
@@ -264,7 +263,8 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	}()
 
 	go wsClient.readFromServer()
-	return <-wsClient.Done
+	<-wsClient.done
+	return wsClient.err
 }
 
 // Handles graceful termination of the websh terminal.
@@ -290,7 +290,8 @@ func (wsClient *WebsocketClient) runWsClient() error {
 	go wsClient.readUserInput(inputChan)
 	go wsClient.writeToServer(inputChan)
 
-	return <-wsClient.Done
+	<-wsClient.done
+	return wsClient.err
 }
 
 func checkTerminal() (*term.State, error) {
@@ -331,7 +332,7 @@ func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
 		// After teardown writeToServer is gone, so an unguarded send would park here.
 		select {
 		case inputChan <- string(char):
-		case <-wsClient.quit:
+		case <-wsClient.done:
 			return
 		}
 	}
@@ -341,7 +342,7 @@ func (wsClient *WebsocketClient) writeToServer(inputChan <-chan string) {
 	var inputBuffer []rune
 	for {
 		select {
-		case <-wsClient.quit:
+		case <-wsClient.done:
 			return
 		case input := <-inputChan:
 			inputBuffer = append(inputBuffer, []rune(input)...)
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -333,11 +333,11 @@ func (wsClient *WebsocketClient) runWsClient() error {
 // enterRawMode returns the restore for the caller to defer.
 func enterRawMode() (func(), error) {
 	if !term.IsTerminal(int(os.Stdin.Fd())) {
-		return nil, errors.New("failed to set up terminal: stdin is not a terminal")
+		return nil, errors.New("stdin is not a terminal")
 	}
 	oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
 	if err != nil {
-		return nil, fmt.Errorf("failed to set up terminal: %w", err)
+		return nil, fmt.Errorf("failed to enter raw mode: %w", err)
 	}
 
 	return func() { _ = term.Restore(int(os.Stdin.Fd()), oldState) }, nil
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -248,12 +248,8 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	}
 	defer func() { _ = wsClient.conn.Close() }()
 
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
-	// Registered before the raw-mode restore so LIFO runs it after: a signal
-	// arriving mid-teardown lands in the buffered channel instead of killing the
-	// process with the terminal still in raw mode.
-	defer signal.Stop(sigChan)
+	sigChan, stopSignals := notifySignals()
+	defer stopSignals()
 
 	restore, err := enterRawMode()
 	if err != nil {
@@ -299,7 +295,7 @@ func (wsClient *WebsocketClient) readCtrlC() {
 
 // OpenNewTerminal opens an interactive terminal on the session.
 // Input is forwarded to the server. Terminal echo is suppressed via raw mode.
-// Ends cleanly on the remote close or on Ctrl+D.
+// Ends cleanly on the remote close, on Ctrl+D, or on a signal.
 func OpenNewTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) error {
 	wsClient := newWebsocketClient(ac.SetWebsocketHeader())
 	if err := wsClient.dial(sessionResponse.WebsocketURL); err != nil {
@@ -311,6 +307,9 @@ func OpenNewTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) 
 }
 
 func (wsClient *WebsocketClient) runWsClient() error {
+	sigChan, stopSignals := notifySignals()
+	defer stopSignals()
+
 	restore, err := enterRawMode()
 	if err != nil {
 		return err
@@ -319,12 +318,24 @@ func (wsClient *WebsocketClient) runWsClient() error {
 
 	inputChan := make(chan string, 1)
 
+	go wsClient.watchInterrupt(sigChan)
 	go wsClient.readFromServer()
 	go wsClient.readUserInput(inputChan)
 	go wsClient.writeToServer(inputChan)
 
 	<-wsClient.done
 	return wsClient.err
+}
+
+// notifySignals returns the signal channel and the stop for the caller to defer.
+// Defer the stop before the raw-mode restore so LIFO runs it after: a signal
+// arriving mid-teardown lands in the buffered channel instead of killing the
+// process with the terminal still in raw mode.
+func notifySignals() (chan os.Signal, func()) {
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+
+	return sigChan, func() { signal.Stop(sigChan) }
 }
 
 // enterRawMode returns the restore for the caller to defer.
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -358,8 +358,7 @@ func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
 		char, _, err := reader.ReadRune()
 		if err != nil {
 			if errors.Is(err, io.EOF) {
-				wsClient.finish(nil)
-				return
+				err = nil // the user closing stdin ends the session rather than failing it
 			}
 			wsClient.finish(err)
 			return
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -29,10 +29,10 @@ const (
 	ctrlC              = 0x03
 	writeFlushInterval = 5 * time.Millisecond
 
-	// sessionEndCloseCode is what the proxy closes a user channel with when the
-	// session ends normally—see sendCloseFrame in proxy-server internal/ws/channel.go.
-	// alpamon spends the same 4000 on the same meaning. A websh session never ends
-	// with 1000, so leaving this out would leave every normal end reported as a failure.
+	// sessionEndCloseCode is how the proxy closes a user channel at the end of a
+	// session (sendCloseFrame in proxy-server internal/ws/channel.go); alpamon spends
+	// the same 4000 on the same meaning. A websh session never ends with 1000, so
+	// without this every normal end would be reported as a failure.
 	sessionEndCloseCode = 4000
 )
 
@@ -225,9 +225,9 @@ func (wsClient *WebsocketClient) dial(websocketURL string) error {
 	return nil
 }
 
-// finish keeps the first outcome. err is written before done closes, so anyone
-// who saw done can read it. A normal remote close is how a session ends rather
-// than a failure; every other close code stays an error.
+// finish keeps the first outcome. err is written before done closes, so anyone who
+// saw done can read it. A deliberate close ends the session rather than failing it;
+// every other close code stays an error.
 func (wsClient *WebsocketClient) finish(err error) {
 	if websocket.IsCloseError(err, sessionEndCloseCode, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
 		err = nil
@@ -247,6 +247,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 		return err
 	}
 	defer func() { _ = wsClient.conn.Close() }()
+
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
 	// Registered before the raw-mode restore so LIFO runs it after: a signal
@@ -350,8 +351,8 @@ func (wsClient *WebsocketClient) readFromServer() {
 	}
 }
 
-// Teardown cannot release this one mid-read: a goroutine parked in ReadRune
-// stays there until the next keystroke, and only closing stdin would change that.
+// readUserInput cannot be released mid-read: a goroutine parked in ReadRune stays
+// there until the next keystroke, and only closing stdin would change that.
 func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
 	reader := bufio.NewReader(os.Stdin)
 	for {
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -206,6 +206,14 @@ func newWebsocketClient(header http.Header) *WebsocketClient {
 	}
 }
 
+func (wsClient *WebsocketClient) dial(websocketURL string) {
+	conn, _, err := websocket.DefaultDialer.Dial(websocketURL, wsClient.Header)
+	if err != nil {
+		utils.CliErrorWithExit("websocket connection failed: %v", err)
+	}
+	wsClient.conn = conn
+}
+
 // finish takes effect only once: Done holds one value and is received once, so a
 // second send would park its goroutine for good.
 func (wsClient *WebsocketClient) finish(err error) {
@@ -220,12 +228,7 @@ func (wsClient *WebsocketClient) finish(err error) {
 // Exits cleanly on Ctrl+C or SIGTERM.
 func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) error {
 	wsClient := newWebsocketClient(ac.SetWebsocketHeader())
-
-	var err error
-	wsClient.conn, _, err = websocket.DefaultDialer.Dial(sessionResponse.WebsocketURL, wsClient.Header)
-	if err != nil {
-		utils.CliErrorWithExit("websocket connection failed: %v", err)
-	}
+	wsClient.dial(sessionResponse.WebsocketURL)
 	defer func() { _ = wsClient.conn.Close() }()
 
 	sigChan := make(chan os.Signal, 1)
@@ -268,20 +271,10 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 // Exits on error without further error handling.
 func OpenNewTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) error {
 	wsClient := newWebsocketClient(ac.SetWebsocketHeader())
-
-	var err error
-	wsClient.conn, _, err = websocket.DefaultDialer.Dial(sessionResponse.WebsocketURL, wsClient.Header)
-	if err != nil {
-		utils.CliErrorWithExit("websocket connection failed: %v", err)
-	}
+	wsClient.dial(sessionResponse.WebsocketURL)
 	defer func() { _ = wsClient.conn.Close() }()
 
-	err = wsClient.runWsClient()
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return wsClient.runWsClient()
 }
 
 func (wsClient *WebsocketClient) runWsClient() error {
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -25,6 +25,9 @@ import (
 const (
 	sessionsBaseURL     = "/api/websh/sessions/"
 	userChannelsBaseURL = "/api/websh/user-channels/"
+
+	ctrlC              = 0x03
+	writeFlushInterval = 5 * time.Millisecond
 )
 
 func GetSessionList(ac *client.AlpaconClient) ([]SessionListItem, error) {
@@ -203,9 +206,8 @@ func newWebsocketClient(header http.Header) *WebsocketClient {
 	}
 }
 
-// finish reports the first outcome of the session and signals teardown to every
-// goroutine. Later calls are dropped: Done holds one value and is received once,
-// so a second send would park its goroutine for good.
+// finish takes effect only once: Done holds one value and is received once, so a
+// second send would park its goroutine for good.
 func (wsClient *WebsocketClient) finish(err error) {
 	wsClient.closeOnce.Do(func() {
 		wsClient.Done <- err
@@ -246,12 +248,12 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	}
 	defer func() { _ = term.Restore(int(os.Stdin.Fd()), oldState) }()
 
-	// In raw mode Ctrl+C is 0x03 — detect it to exit
+	// Raw mode suppresses SIGINT, so Ctrl+C only ever arrives as a byte
 	go func() {
 		buf := make([]byte, 1)
 		for {
 			_, err := os.Stdin.Read(buf)
-			if err != nil || buf[0] == 0x03 {
+			if err != nil || buf[0] == ctrlC {
 				wsClient.finish(nil)
 				return
 			}
@@ -350,7 +352,7 @@ func (wsClient *WebsocketClient) writeToServer(inputChan <-chan string) {
 			return
 		case input := <-inputChan:
 			inputBuffer = append(inputBuffer, []rune(input)...)
-		case <-time.After(time.Millisecond * 5):
+		case <-time.After(writeFlushInterval):
 			if len(inputBuffer) > 0 {
 				err := wsClient.conn.WriteMessage(websocket.BinaryMessage, []byte(string(inputBuffer)))
 				if err != nil {
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"os/signal"
@@ -194,14 +195,29 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
 	return response, nil
 }
 
+func newWebsocketClient(header http.Header) *WebsocketClient {
+	return &WebsocketClient{
+		Header: header,
+		Done:   make(chan error, 1),
+		quit:   make(chan struct{}),
+	}
+}
+
+// finish reports the first outcome of the session and signals teardown to every
+// goroutine. Later calls are dropped: Done holds one value and is received once,
+// so a second send would park its goroutine for good.
+func (wsClient *WebsocketClient) finish(err error) {
+	wsClient.closeOnce.Do(func() {
+		wsClient.Done <- err
+		close(wsClient.quit)
+	})
+}
+
 // OpenReadOnlyTerminal opens a read-only terminal view for watching another user's session.
 // Input is not forwarded to the server. Terminal echo is suppressed via raw mode.
 // Exits cleanly on Ctrl+C or SIGTERM.
 func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) error {
-	wsClient := &WebsocketClient{
-		Header: ac.SetWebsocketHeader(),
-		Done:   make(chan error, 1),
-	}
+	wsClient := newWebsocketClient(ac.SetWebsocketHeader())
 
 	var err error
 	wsClient.conn, _, err = websocket.DefaultDialer.Dial(sessionResponse.WebsocketURL, wsClient.Header)
@@ -218,10 +234,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 	defer signal.Stop(sigChan)
 	go func() {
 		<-sigChan
-		select {
-		case wsClient.Done <- nil:
-		default:
-		}
+		wsClient.finish(nil)
 	}()
 
 	oldState, err := checkTerminal()
@@ -236,10 +249,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 		for {
 			_, err := os.Stdin.Read(buf)
 			if err != nil || buf[0] == 0x03 {
-				select {
-				case wsClient.Done <- nil:
-				default:
-				}
+				wsClient.finish(nil)
 				return
 			}
 		}
@@ -252,10 +262,7 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 // Handles graceful termination of the websh terminal.
 // Exits on error without further error handling.
 func OpenNewTerminal(ac *client.AlpaconClient, sessionResponse SessionResponse) error {
-	wsClient := &WebsocketClient{
-		Header: ac.SetWebsocketHeader(),
-		Done:   make(chan error, 1),
-	}
+	wsClient := newWebsocketClient(ac.SetWebsocketHeader())
 
 	var err error
 	wsClient.conn, _, err = websocket.DefaultDialer.Dial(sessionResponse.WebsocketURL, wsClient.Header)
@@ -304,10 +311,7 @@ func (wsClient *WebsocketClient) readFromServer() {
 	for {
 		_, message, err := wsClient.conn.ReadMessage()
 		if err != nil {
-			select {
-			case wsClient.Done <- err:
-			default:
-			}
+			wsClient.finish(err)
 			return
 		}
 		fmt.Print(string(message))
@@ -320,13 +324,18 @@ func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
 		char, _, err := reader.ReadRune()
 		if err != nil {
 			if err == io.EOF {
-				wsClient.Done <- nil
+				wsClient.finish(nil)
 				return
 			}
-			wsClient.Done <- err
+			wsClient.finish(err)
 			return
 		}
-		inputChan <- string(char)
+		// After teardown writeToServer is gone, so an unguarded send would park here.
+		select {
+		case inputChan <- string(char):
+		case <-wsClient.quit:
+			return
+		}
 	}
 }
 
@@ -334,13 +343,15 @@ func (wsClient *WebsocketClient) writeToServer(inputChan <-chan string) {
 	var inputBuffer []rune
 	for {
 		select {
+		case <-wsClient.quit:
+			return
 		case input := <-inputChan:
 			inputBuffer = append(inputBuffer, []rune(input)...)
 		case <-time.After(time.Millisecond * 5):
 			if len(inputBuffer) > 0 {
 				err := wsClient.conn.WriteMessage(websocket.BinaryMessage, []byte(string(inputBuffer)))
 				if err != nil {
-					wsClient.Done <- err
+					wsClient.finish(err)
 					return
 				}
 				inputBuffer = []rune{}
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -247,10 +247,6 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
 		return err
 	}
 	defer func() { _ = wsClient.conn.Close() }()
-	// Whatever this returns through, the goroutines below have to be released.
-	// The sync.Once makes it a no-op once an outcome is already recorded.
-	defer wsClient.finish(nil)
-
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
 	// Registered before the raw-mode restore so LIFO runs it after: a signal
--- a/api/websh/websh.go
+++ b/api/websh/websh.go
@@ -200,27 +200,32 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
 
 func newWebsocketClient(header http.Header) *WebsocketClient {
 	return &WebsocketClient{
-		Header: header,
+		header: header,
 		done:   make(chan struct{}),
 	}
 }
 
-func (wsClient *WebsocketClient) dial(websocketURL string) {
-	conn, resp, err := websocket.DefaultDialer.Dial(websocketURL, wsClient.Header)
+func (wsClient *WebsocketClient) dial(websocketURL string) error {
+	conn, resp, err := websocket.DefaultDialer.Dial(websocketURL, wsClient.header)
 	if err != nil {
 		// The handshake response carries the reason a bad handshake alone never names.
-		status := ""
-		if resp != nil {
-			status = fmt.Sprintf(" (status %s)", utils.SanitizeTerminalText(resp.Status))
+		if resp == nil {
+			return fmt.Errorf("websocket connection failed: %w", err)
 		}
-		utils.CliErrorWithExit("websocket connection failed: %v%s", err, status)
+		return fmt.Errorf("websocket connection failed: %w (status %s)", err, utils.SanitizeTerminalText(resp.Status))
 	}
 	wsClient.conn = conn
+
+	return nil
 }
 
 // finish keeps the first outcome. err is written before done closes, so anyone
-// who saw done can read it.
+// who saw done can read it. A normal remote close is how a session ends rather
+// than a failure; every other close code stays an error.
 func (wsClient *WebsocketClient) finish(err error) {
+	if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
+		err = nil
+	}
 	wsClient.finishOnce.Do(func() {
 		wsClient.err = err
 		close(wsClient.done)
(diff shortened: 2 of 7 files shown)


api/websh/websh.go
```diff
@@ -216,66 +241,54 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
...
 		utils.CliErrorWithExit("failed to set up terminal: %v", err)
 	}
 	defer func() { _ = term.Restore(int(os.Stdin.Fd()), oldState) }()
 
-	// In raw mode Ctrl+C is 0x03 — detect it to exit
+	// Raw mode suppresses SIGINT, so Ctrl+C only ever arrives as a byte
 	go func() {
 		buf := make([]byte, 1)
 		for {
 			_, err := os.Stdin.Read(buf)
```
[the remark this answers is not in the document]
jisung-02: Correct that both goroutines stay blocked, but neither block is reachable by the mechanism this PR changes, and neither line is part of the change.

- `os.Stdin.Read` here and `bufio.Reader.ReadRune` in `readUserInput` are blocking reads on a tty. No channel can release them — closing `done` is invisible to a goroutine parked inside the read syscall. Unblocking them would mean closing the fd out from under the process or putting read deadlines on the terminal, both of which race the `term.Restore` deferred right above.
- Both lines are unchanged from `main` (`git diff main...HEAD` shows them as context). What this PR fixes is the goroutines that were parked on a channel send and *could* have left: `readFromServer`, `writeToServer`, `readUserInput`'s send into `inputChan`, and the signal watcher.
- All three call sites — `cmd/websh/websh.go:307`, `websh_join.go:31`, `websh_watch.go:44` — are the last statement of their Cobra `Run`, so the process exits as soon as the terminal function returns. There is no long-lived caller today, and if one is ever added it needs a cancellable stdin reader rather than another channel.

api/websh/websh.go
```diff
@@ -304,10 +318,7 @@ func (wsClient *WebsocketClient) readFromServer() {
 	for {
 		_, message, err := wsClient.conn.ReadMessage()
 		if err != nil {
-			select {
-			case wsClient.Done <- err:
-			default:
-			}
+			wsClient.finish(err)
```
geunwoonoh, line 358: **Warning:** a normal session end is reported as an error.

`readFromServer` passes the server's close frame straight into `finish(err)`. Confirmed directly:

```
err = &websocket.CloseError{Code:1000, Text:"session ended"}
→ "websocket: close 1000 (normal): session ended"
```

Both callers that check the error treat it as a failure (`cmd/websh/websh_join.go:31`, `cmd/websh/websh_watch.go:44`), so typing `exit` in a `websh join` session prints a red `Error:` plus the "report on github issues" footer and exits 1 — on a perfectly normal end.

Pre-existing, but since the abstraction this PR introduces is named "session outcome", `finish` is the natural place to normalize it:

```go
func (wsClient *WebsocketClient) finish(err error) {
	if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
		err = nil // the remote ending the session is how a session ends
	}
	wsClient.finishOnce.Do(func() { ... })
}
```

Related: `cmd/websh/websh.go:322` discards this error entirely (`_ = websh.OpenNewTerminal(...)`), so the main path surfaces nothing. Fix this one first — otherwise surfacing it there makes every normal session exit 1.

api/websh/websh.go
```diff
@@ -216,66 +241,54 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
...
 	go func() {
 		buf := make([]byte, 1)
 		for {
 			_, err := os.Stdin.Read(buf)
-			if err != nil || buf[0] == 0x03 {
-				select {
-				case wsClient.Done <- nil:
-				default:
-				}
+			if err != nil || buf[0] == ctrlC {
```
geunwoonoh: **Warning:** this is the one goroutine of the six that still cannot leave teardown.

It never looks at `done`. When `readFromServer` finishes first and `done` closes, this loop keeps reading stdin and, for any byte that is not Ctrl+C, reads again — it goes on consuming user input after `OpenReadOnlyTerminal` has returned and `term.Restore` has run. That is exactly the dependency the description calls out as one the contract should not rest on, and it is also the only one of the six with no test. (The "Ctrl+C ending the session with no error" coverage is `TestWatchInterruptEndsTheSessionOnSignal`, which is the signal path.)

The missing `n > 0` guard belongs with it: if `Read` returns `(0, nil)`, `buf[0]` is judged again from the previous read.

Giving it a name the way `watchInterrupt` got one also makes it testable:

```go
func (wsClient *WebsocketClient) readCtrlC() {
	buf := make([]byte, 1)
	for {
		n, err := os.Stdin.Read(buf)
		if err != nil || (n > 0 && buf[0] == ctrlC) {
			wsClient.finish(nil)
			return
		}
		select {
		case <-wsClient.done: // stop eating stdin once the session is over
			return
		default:
		}
	}
}
```

api/websh/websh.go
```diff
@@ -320,27 +331,34 @@ func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
...
 			if err == io.EOF {
-				wsClient.Done <- nil
+				wsClient.finish(nil)
 				return
 			}
-			wsClient.Done <- err
+			wsClient.finish(err)
+			return
+		}
+		// After teardown writeToServer is gone, so an unguarded send would park here.
```
geunwoonoh, line 378: **Warning:** the guard covers the send, but the goroutine still parks one line up.

Copilot's suppressed comment is correct here. After teardown this goroutine stays blocked in `reader.ReadRune()` until the user types another byte — the `done` case only helps once a rune has already been read.

Not fixable without closing stdin, so I am not asking for a change. But it does not match "every session goroutine" in the title, so the limit is worth stating in the function's comment.

api/websh/websh.go
```diff
@@ -194,20 +198,41 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
...
+func (wsClient *WebsocketClient) dial(websocketURL string) {
+	conn, resp, err := websocket.DefaultDialer.Dial(websocketURL, wsClient.Header)
 	if err != nil {
-		utils.CliErrorWithExit("websocket connection failed %v", err)
+		// The handshake response carries the reason a bad handshake alone never names.
+		status := ""
+		if resp != nil {
+			status = fmt.Sprintf(" (status %s)", utils.SanitizeTerminalText(resp.Status))
+		}
+		utils.CliErrorWithExit("websocket connection failed: %v%s", err, status)
```
geunwoonoh: **Warning:** `os.Exit` from the `api/` layer.

`dial` does not return an error; it ends the process. Every other function in this file returns an error and lets `cmd/` decide how to exit. Three consequences:

- The tests have to drive two subprocess helpers (~40 lines) just to observe stderr. `dial(url) error` would be testable in process.
- `CliErrorWithExit` calls `reportCLIError()`, appending "report on github issues". A 401 handshake rejection is a user or configuration problem, not a CLI bug — and `utils/log.go:31-33` documents `CliErrorWithExitCode` for precisely this case.
- When terminal setup fails in `OpenReadOnlyTerminal`, the `signal.Stop` and `conn.Close` defers are skipped.

Minimum fix: `CliErrorWithExitCode(1, ...)` to drop the footer. Proper fix: have `dial` return an error.

api/websh/websh.go
```diff
@@ -216,66 +241,54 @@ func OpenReadOnlyTerminal(ac *client.AlpaconClient, sessionResponse SessionRespo
...
-
-	return nil
+	return wsClient.runWsClient()
 }
 
 func (wsClient *WebsocketClient) runWsClient() error {
 	oldState, err := checkTerminal()
 	if err != nil {
-		utils.CliErrorWithExit("websocket connection failed %v", err)
+		utils.CliErrorWithExit("failed to set up terminal: %v", err)
```
geunwoonoh: **Suggestion:** same `os.Exit`-from-`api/` point as `dial` — and these four lines (`checkTerminal` + error handling + `defer term.Restore`) are duplicated verbatim at `OpenReadOnlyTerminal` (lines 246-250).

The dial got deduplicated in this PR; this block did not. An `enterRawMode() (restore func())` would fold them.

api/websh/websh.go
```diff
@@ -320,27 +331,34 @@ func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
...
 func (wsClient *WebsocketClient) writeToServer(inputChan <-chan string) {
 	var inputBuffer []rune
 	for {
 		select {
+		case <-wsClient.done:
+			return
 		case input := <-inputChan:
 			inputBuffer = append(inputBuffer, []rune(input)...)
-		case <-time.After(time.Millisecond * 5):
+		case <-time.After(writeFlushInterval):
```
geunwoonoh: **Suggestion:** a timer allocated per loop iteration, and a flush that keeps getting deferred.

`time.After` builds a fresh `*time.Timer` and channel on every call. Even an idle session wakes this loop every 5 ms — 200 per second, ~720k over an hour-long session. Go 1.23+ collects them, but a ticker is one allocation total.

The behavioral half matters more: every arriving input discards the pending timer and starts a new one, so a flush is deferred for as long as input keeps coming and only lands once it stops. A pasted block accumulates entirely in `inputBuffer` and goes out as a single `WriteMessage`.

A ticker fixes both — declare it above the loop with `defer ticker.Stop()`:

```suggestion
		case <-ticker.C:
```

api/websh/websh.go
```diff
@@ -304,10 +318,7 @@ func (wsClient *WebsocketClient) readFromServer() {
...
 		_, message, err := wsClient.conn.ReadMessage()
 		if err != nil {
-			select {
-			case wsClient.Done <- err:
-			default:
-			}
+			wsClient.finish(err)
 			return
 		}
 		fmt.Print(string(message))
```
geunwoonoh: **Suggestion:** the `[]byte` → `string` conversion copies on every message, and `fmt` goes through reflection. This is the path server output floods through, so it shows.

`testutil.CaptureStdout` swaps `os.Stdout`, so the existing test still passes with a direct write.

```suggestion
		_, _ = os.Stdout.Write(message)
```

geunwoonoh (commented):
## Review

**Stack**: golang (no `alpacax` frontmatter in `CLAUDE.md`, inferred from `.go` files)
**Changes**: +394 -56 across 3 files

### What I verified

| Check | Result |
|---|---|
| `gofmt -l .` | clean |
| `go vet ./...` | clean |
| `go test -race -count=1 ./api/websh/...` | ok (1.574s) |
| `golangci-lint` | not installed locally, not run |

The build and test claims in the description reproduce.

### Summary

This swaps the teardown signal from a buffered `Done chan error` received exactly once to a close-only `done chan struct{}` plus `err` and a `sync.Once`. A channel that only closes cannot park a sender, so the hazard goes away structurally rather than by convention — the right call. Alongside that it fixes the dial and terminal-setup error messages and folds the client literal and dial block, duplicated across both entry points, into `newWebsocketClient` and `dial`.

`err` is written before `close(done)` and read after `<-done`, so the happens-before edge holds. The race detector agrees. No defect in the core design.

### Critical issues (0)

Nothing blocks merge.

### Warnings (5)

**1. `api/websh/websh.go:317-325` — a normal session end is reported as an error** (pre-existing, but this PR is the right place to fix it)

`readFromServer` passes the server's close frame straight into `finish(err)`. Confirmed directly:

```
err = &websocket.CloseError{Code:1000, Text:"session ended"}
→ "websocket: close 1000 (normal): session ended"
```

Both callers that check the error surface it as a failure:

```go
// cmd/websh/websh_join.go:31
if err = websh.OpenNewTerminal(alpaconClient, session); err != nil {
    utils.CliErrorWithExit("Websh session terminated with error: %s.", err)
}
// cmd/websh/websh_watch.go:44 — same shape
```

So typing `exit` in a `websh join` session, or watching a session that ends, prints a red `Error:` plus the "report on github issues" footer and exits 1 — on a perfectly normal end. Since the abstraction this PR introduces is named "session outcome", normalizing it inside `finish` is the natural fix:

```go
func (wsClient *WebsocketClient) finish(err error) {
	if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
		err = nil // the remote ending the session is how a session ends
	}
	wsClient.finishOnce.Do(func() { ... })
}
```

**2. `api/websh/websh.go:253-262` — the read-only Ctrl+C reader is the one goroutine that still cannot leave teardown**

It is the only one of the six that never looks at `done`. When `readFromServer` finishes first and `done` closes, this loop keeps reading stdin and, for any byte that is not Ctrl+C, reads again — it goes on consuming user input after `OpenReadOnlyTerminal` has returned and `term.Restore` has run. That is exactly the dependency the description calls out as one the contract should not rest on, and it is also the only one of the six with no test. (The description's "Ctrl+C ending the session with no error" coverage is `TestWatchInterruptEndsTheSessionOnSignal`, which is the signal path.)

The missing `n > 0` guard belongs with it: if `Read` returns `(0, nil)`, `buf[0]` is judged again from the previous read.

Giving it a name the way `watchInterrupt` got one also makes it testable:

```go
func (wsClient *WebsocketClient) readCtrlC() {
	buf := make([]byte, 1)
	for {
		n, err := os.Stdin.Read(buf)
		if err != nil || (n > 0 && buf[0] == ctrlC) {
			wsClient.finish(nil)
			return
		}
		select {
		case <-wsClient.done: // stop eating stdin once the session is over
			return
		default:
		}
	}
}
```

**3. `api/websh/websh.go:331` — `readUserInput` still parks in `ReadRune`**

Copilot's suppressed comment is correct here. The `done` guard covers only the send to `inputChan`; after teardown the goroutine stays blocked in `reader.ReadRune()` until the user types another byte. That is not fixable without closing stdin, so I am not asking for a change — but it does not match "every session goroutine" in the title. Worth stating the limit in the function's comment.

**4. `api/websh/websh.go:208-219, 291` — `os.Exit` from the `api/` layer**

`dial` does not return an error; it ends the process through `utils.CliErrorWithExit`. Every other function in this file returns an error and lets `cmd/` decide how to exit. Three consequences:

- The tests have to drive two subprocess helpers (~40 lines) just to observe stderr. `dial(url) error` would be testable in process.
- `CliErrorWithExit` calls `reportCLIError()`, appending "report on github issues". A 401 handshake rejection or `stdin is not a terminal` is a user or configuration problem, not a CLI bug. The comment on `CliErrorWithExitCode` in `utils/log.go:31-33` describes precisely this case.
- When terminal setup fails in `OpenReadOnlyTerminal`, the `signal.Stop` and `conn.Close` defers are skipped.

Minimum fix: `CliErrorWithExitCode(1, ...)` to drop the footer. Proper fix: have `dial` return an error.

**5. `cmd/websh/websh.go:322` — the main path discards the outcome**

```go
_ = websh.OpenNewTerminal(alpaconClient, session)
```

The `err` field and `finish` plumbing this PR builds produces nothing on `alpacon websh <server>`, unlike join and watch. **Do not surface it here before fixing warning 1**, or every normal session will exit 1. These two go together.

(Outside the diff, so no inline comment.)

### Suggestions (7)

**6. `api/websh/websh.go:357` — a timer allocated per loop iteration** (performance)

`time.After` builds a fresh `*time.Timer` and channel on every call. Even an idle session wakes the loop every 5 ms, so that is 200 per second, ~720k over an hour-long session. Go 1.23+ collects them, but a ticker is one allocation total.

The behavioral half matters more: every arriving input discards the pending timer and starts a new one, so **a flush is deferred for as long as input keeps coming and only lands once it stops**. A pasted block accumulates entirely in `inputBuffer` and goes out as a single `WriteMessage`. A ticker fixes both:

```go
func (wsClient *WebsocketClient) writeToServer(inputChan <-chan string) {
	ticker := time.NewTicker(writeFlushInterval)
	defer ticker.Stop()

	var inputBuffer []rune
	for {
		select {
		case <-wsClient.done:
			return
		case input := <-inputChan:
			inputBuffer = append(inputBuffer, []rune(input)...)
		case <-ticker.C:
			// ...
		}
	}
}
```

**7. `api/websh/websh.go:324` — `fmt.Print(string(message))`** (performance)

The `[]byte` → `string` conversion copies on every message and `fmt` goes through reflection. This is the path server output floods through, so it shows. `os.Stdout.Write(message)` is a direct write. `testutil.CaptureStdout` swaps `os.Stdout`, so the existing test still passes.

**8. `api/websh/websh.go:333` — `err == io.EOF` → `errors.Is(err, io.EOF)`**

`bufio.Reader` returns `io.EOF` unwrapped today, but this is the idiom that survives wrapping. `errors` is already imported.

**9. `api/websh/types.go:13` — `Header` can be unexported now**

Checked while looking at `Done` → `done`: `Header` is set only by `newWebsocketClient` and read only by `dial`, with no use outside the package. (`SetWebsocketHeader` is used by `api/event` and `pkg/tunnel`, but that is unrelated to this field.)

**10. `api/websh/websh.go:246-250` vs `288-293` — duplication that survived**

The dial got deduplicated, but the four lines of `checkTerminal` + error handling + `defer term.Restore` remain copied across both entry points. An `enterRawMode() (restore func())` would fold them.

**11. Test duplication and naming**

Four tests share one shape — "finish first, then assert the goroutine returns": `TestReadUserInputReturnsWhenOutcomeAlreadyReported`, `TestReadUserInputReturnsWhenWriterIsGone`, `TestWriteToServerReturnsWhenDoneIsClosed`, `TestWatchInterruptReturnsWhenOutcomeAlreadyReported`. A table of `[]struct{name string; setup func(*testing.T); run func(*WebsocketClient)}` would save 20+ lines and match the table-driven convention the golang stack rules ask for.

The two subprocess tests, `TestRunWsClientReportsTerminalSetupFailure` and `TestDialReportsTheHandshakeStatus`, share the same skeleton too — build command, set env, capture stderr, `require.Error` — and fold into `runHelper(t, name, env...) string`.

On naming: this file's existing tests use an underscore separator (`TestJoinWebshSession_InvalidURL`, `TestGetSessionRecords_FollowsCursor`); the new ones do not.

**12. `api/websh/websh.go:230-232` — asymmetric doc comment**

`OpenNewTerminal` gained "A failed dial or terminal setup exits the process instead of returning". `OpenReadOnlyTerminal` behaves identically and says nothing.

### Highlights

- **The close-only channel is the right choice.** As the description argues, a non-blocking `select`/`default` covers only the `Done` sends and leaves `inputChan` wanting a second mechanism. Good call, and good that the rejected alternative is written down.
- **The memory model is respected.** `err` write → `close(done)` → `<-done` → `err` read gives the ordering, and both production readers (`runWsClient`, `OpenReadOnlyTerminal`) read only after `<-done`.
- **`awaitReturn` turns a park into a failure instead of a hang.** Without it, a goroutine-leak regression only ever shows up as a CI timeout.
- **The terminal-escape handling is correct.** `SanitizeTerminalText(resp.Status)` is applied, and I checked the other path too: escape bytes injected into a server-supplied `websocket_url` come back `%q`-escaped by `net/url`, so no raw ESC reaches the terminal.
- **One of Copilot's suppressed comments is wrong.** It claims the helper subprocess may inherit a tty stdin and go flaky, but `exec.Cmd` attaches `os.DevNull` when `Stdin` is nil. The subprocess never inherits the parent's stdin, so the test is deterministic — noting it here so nobody "fixes" it.

---

**Verdict**: COMMENTED

Nothing blocks merge and the design is sound. Warnings 1 and 2 fall inside the scope this PR sets for itself, so they are worth a decision — fix here, or split into a follow-up.

**Checklist**:
- [x] Code follows stack conventions — only minor drift, items 8 and 11
- [x] Tests included and passing — pass under `-race`, though one goroutine is uncovered
- [x] No security vulnerabilities — escape handling verified directly
- [ ] Documentation updated if needed — items 3 and 12 want comment fixes

api/websh/websh.go
```diff
@@ -320,27 +331,34 @@ func (wsClient *WebsocketClient) readUserInput(inputChan chan<- string) {
 		char, _, err := reader.ReadRune()
 		if err != nil {
 			if err == io.EOF {
```
geunwoonoh: **Suggestion:** `bufio.Reader` returns `io.EOF` unwrapped today, but `errors.Is` is the idiom that survives wrapping. `errors` is already imported.

```suggestion
			if errors.Is(err, io.EOF) {
```

api/websh/websh.go
```diff
@@ -194,20 +198,41 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
...
+// finish keeps the first outcome. err is written before done closes, so anyone
+// who saw done can read it.
+func (wsClient *WebsocketClient) finish(err error) {
+	wsClient.finishOnce.Do(func() {
+		wsClient.err = err
+		close(wsClient.done)
+	})
+}
+
+// OpenReadOnlyTerminal opens a read-only terminal view for watching another user's session.
```
geunwoonoh, line 241: **Suggestion:** asymmetric doc comment. `OpenNewTerminal` gained "A failed dial or terminal setup exits the process instead of returning"; `OpenReadOnlyTerminal` behaves identically and says nothing.

api/websh/types.go
```diff
@@ -2,16 +2,19 @@ package websh
...
 
 	"github.com/alpacax/alpacon-cli/api/types"
 	"github.com/gorilla/websocket"
 )
 
 type WebsocketClient struct {
-	Header http.Header
-	conn   *websocket.Conn
-	Done   chan error
+	Header     http.Header
```
geunwoonoh: **Suggestion:** `Header` can be unexported now.

Checked while looking at `Done` → `done`: it is set only by `newWebsocketClient` and read only by `dial`, with no use outside the package. (`SetWebsocketHeader` is used by `api/event` and `pkg/tunnel`, but that is unrelated to this field.)

api/websh/websh_test.go
```diff
@@ -270,3 +284,306 @@ func TestGetSessionRecords_QueryHitsSearchEndpoint(t *testing.T) {
 	assert.Equal(t, "docker", gotQuery)
 	assert.Len(t, records, 1)
 }
+
+func TestReadUserInputReturnsWhenOutcomeAlreadyReported(t *testing.T) {
```
geunwoonoh: **Suggestion:** four tests share one shape — "finish first, then assert the goroutine returns": this one, `TestReadUserInputReturnsWhenWriterIsGone`, `TestWriteToServerReturnsWhenDoneIsClosed`, and `TestWatchInterruptReturnsWhenOutcomeAlreadyReported`.

A table of `[]struct{name string; setup func(*testing.T); run func(*WebsocketClient)}` would save 20+ lines and match the table-driven convention the golang stack rules ask for.

The two subprocess tests (`TestRunWsClientReportsTerminalSetupFailure`, `TestDialReportsTheHandshakeStatus`) share a skeleton too — build command, set env, capture stderr, `require.Error` — and fold into `runHelper(t, name, env...) string`.

Naming nit: this file's existing tests use an underscore separator (`TestJoinWebshSession_InvalidURL`, `TestGetSessionRecords_FollowsCursor`); the new ones do not.

api/websh/websh.go
jisung-02: Fixed in 8dab66e. `finish` normalizes `CloseNormalClosure` and `CloseGoingAway` to `nil` before the `sync.Once`, so every other close code still reaches the caller as an error.

Took the related point with it: `cmd/websh/websh.go` now surfaces the outcome instead of discarding it, which is only safe in that order — the two went in the same commit.

`TestFinish_NormalizesARemoteCloseToASuccess` covers all four cases (normal, going away, abnormal, transport failure). Mutation-checked: dropping the normalization fails the two normalized subtests and nothing else.

api/websh/websh.go
jisung-02: Fixed in 8dab66e, taking your shape almost verbatim — named `readCtrlC`, `n > 0` guard, `done` checked between reads.

It is now the fifth case in the teardown table, and `TestReadCtrlC_EndsTheSessionOnCtrlC` writes `a` before the Ctrl+C byte so the test proves the loop goes around rather than ending on any input. Mutation-checked: removing the `done` select fails only its teardown case.

One scope correction on the doc comment: it stops consuming stdin once the session is over, but not before the read it is already parked in returns. The comment says so rather than claiming more.

api/websh/websh.go
jisung-02: Agreed on all of it — no code change, comment added in c3fa8cc:

```go
// Teardown cannot release this one mid-read: a goroutine parked in ReadRune
// stays there until the next keystroke, and only closing stdin would change that.
```

The PR description now states the same limit instead of implying every goroutine is covered.

api/websh/websh.go
jisung-02: Took the proper fix in 8dab66e rather than the minimum one. `dial` returns its error, `checkTerminal` became `enterRawMode` and returns its own, and both entry points propagate.

All three consequences you listed are gone: the two subprocess helpers collapsed into in-process assertions (`TestDial_ReportsTheHandshakeStatus`, `TestRunWsClient_ReportsTerminalSetupFailure`), and the `signal.Stop` and `conn.Close` defers now run when terminal setup fails in `OpenReadOnlyTerminal`.

On the footer: since `cmd/` now decides, all three `cmd/websh` sites use `CliErrorWithExitCode(1, ...)`. A websh session ending badly is the network, the server, or the terminal — not a CLI defect worth reporting.

api/websh/websh.go
jisung-02: Fixed in 8dab66e — `enterRawMode() (func(), error)`, returning the restore to defer. Both entry points call it, and with the error return the duplication and the `os.Exit` go at once.

api/websh/websh.go
jisung-02: Fixed in c3fa8cc. `time.NewTicker(writeFlushInterval)` declared above the loop with `defer ticker.Stop()`.

The behavioral half was the reason to take it: a pasted block no longer waits for the typing to stop before any of it goes out.

api/websh/websh.go
jisung-02: Fixed in c3fa8cc — `_, _ = os.Stdout.Write(message)`. `TestReadFromServer_PrintsEveryMessage` still passes through `testutil.CaptureStdout` unchanged, as you predicted.

api/websh/websh.go
jisung-02: Fixed in c3fa8cc — `errors.Is(err, io.EOF)`.

api/websh/websh.go
jisung-02: Resolved differently in 8dab66e: the asymmetry is gone because the claim is no longer true. `dial` and `enterRawMode` return their errors now, so neither entry point exits the process.

Both doc comments were rewritten to the same shape — what the terminal is, how input is treated, how the session ends:

```go
// OpenReadOnlyTerminal opens a read-only terminal view for watching another user's session.\n// Input is not forwarded to the server. Terminal echo is suppressed via raw mode.\n// Ends cleanly on the remote close, on Ctrl+C, or on a signal.\n\n// OpenNewTerminal opens an interactive terminal on the session.\n// Input is forwarded to the server. Terminal echo is suppressed via raw mode.\n// Ends cleanly on the remote close or on Ctrl+D.\n```\n\nThe read-only line gained the remote close: it was listing only Ctrl+C and SIGTERM, which is not the common way that session ends.

File: api/websh/types.go (hunk 11 above)
Comment (jisung-02) [reply api/websh/types.go]: Fixed in 8dab66e — `header http.Header`. Confirmed the same way you did: `newWebsocketClient` writes it, `dial` reads it, nothing outside the package touches it.

File: api/websh/websh_test.go (hunk 12 above)
Comment (jisung-02) [reply api/websh/websh_test.go]: Both taken.

The four became `TestSessionGoroutines_ReturnAfterTheOutcomeIsTaken`, a table of `{name, start func(*testing.T, *WebsocketClient) func()}` — `start` does the per-case setup and hands back the call to run, so the shared part (`newWebsocketClient`, `finish`, `awaitReturn`) lives once in the loop. `readCtrlC` joined as a fifth case, which is how it stopped being the one goroutine without coverage.

The subprocess pair did not need `runHelper` in the end: with `dial` and `runWsClient` returning their errors (your `os.Exit` point), both assert in process and the helpers are deleted outright.

Naming fixed in 8298df1 — all seven new tests now use the underscore separator.

File: api/websh/websh.go [hunk 13]
```diff
@@ -194,153 +198,194 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
...
 	// process with the terminal still in raw mode.
 	defer signal.Stop(sigChan)
-	go func() {
-		<-sigChan
-		select {
-		case wsClient.Done <- nil:
-		default:
-		}
-	}()
+	go wsClient.watchInterrupt(sigChan)
```
geunwoonoh: **Warning (new):** fixing the `os.Exit` finding introduced a goroutine leak here.

When `enterRawMode` fails two lines down, the function returns without ever calling `finish`, so `done` never closes. `watchInterrupt` waits on both `sigChan` and `done`, and `defer signal.Stop(sigChan)` **disarms delivery without closing the channel** — the exact mechanism this PR's Background section names as the original reason the signal watcher parked. Neither case fires again.

Reproduced against a real `OpenReadOnlyTerminal` call with stdin pointed at a pipe:

```
returned err = failed to set up terminal: stdin is not a terminal
goroutines before=3 after=6 (delta +3)
LEAK: watchInterrupt is still parked after OpenReadOnlyTerminal returned

goroutine 5 [select]:
  websh.(*WebsocketClient).watchInterrupt(...) websh.go:268
  created by websh.OpenReadOnlyTerminal websh.go:251
```

`CliErrorWithExit` used to hide this by killing the process; `return err` exposed it. No user-visible impact today — `cmd/websh/websh_watch.go:44` calls `os.Exit(1)` right after — but "the contract should not depend on that" is this PR's own thesis, and this is the one path where it no longer holds.

Moving the `go` statement below `enterRawMode` is enough; `sigChan` is buffered, so a signal arriving in the gap is not lost. I applied this and confirmed the leak is gone with `./api/websh/...` and `./cmd/websh/...` still passing:

```go
	defer signal.Stop(sigChan)

	restore, err := enterRawMode()
	if err != nil {
		return err
	}
	defer restore()

	go wsClient.watchInterrupt(sigChan)
	go wsClient.readCtrlC()
	go wsClient.readFromServer()
```

A `defer wsClient.finish(nil)` right after the client is created is the more durable alternative — it covers any early return added later and is a no-op on the normal path thanks to the `sync.Once`.

api/websh/websh.go
```diff
@@ -194,153 +198,194 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
...
+
+	return nil
+}
+
+// finish keeps the first outcome. err is written before done closes, so anyone
+// who saw done can read it. A normal remote close is how a session ends rather
+// than a failure; every other close code stays an error.
+func (wsClient *WebsocketClient) finish(err error) {
+	if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
+		err = nil
```
geunwoonoh, line 233: **Suggestion:** worth confirming the server actually sends a 1000.

This catches `CloseNormalClosure` (1000) and `CloseGoingAway` (1001) only. `1006` (connection dropped with no close frame) and `1005` (close frame with no payload) still surface as errors, so if alpacon-server or alpamon does not send a proper close frame at session end, this fix will not show up in production. Not verifiable from this repo — worth a check on the server side.

Separately, treating 1001 as success means a mid-session server restart drops the user back to their shell with exit 0 and no message. This is the canonical gorilla pairing so I am not arguing against it, just flagging it as a decision rather than a default.

api/websh/websh.go
```diff
@@ -194,153 +198,194 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
...
+	<-wsClient.done
+	return wsClient.err
 }
 
-func checkTerminal() (*term.State, error) {
+// enterRawMode returns the restore for the caller to defer.
+func enterRawMode() (func(), error) {
 	if !term.IsTerminal(int(os.Stdin.Fd())) {
-		return nil, errors.New("websh command should be a terminal")
+		return nil, errors.New("failed to set up terminal: stdin is not a terminal")
```
geunwoonoh: **Suggestion:** the context is baked into the error string rather than wrapped by the caller.

The Go convention is `fmt.Errorf("failed to set up terminal: %w", err)` at the call site. As written, the composed output reads `Websh session terminated with error: failed to set up terminal: stdin is not a terminal.` — doubled. Both paths in this function are consistent and it works, so this is taste, not a defect.

geunwoonoh (commented):
## Follow-up review — `8dab66e`

All 12 items from the previous review are addressed. One new goroutine leak came in with the fix for item 4; the one-line fix is below and I verified it.

### What I verified

| Check | Result |
|---|---|
| `gofmt -l .` | clean |
| `go vet ./...` | clean |
| `go build -o /dev/null .` | ok |
| `go test -race -count=1 ./...` | all packages pass |

### Previous review — 12/12 addressed

| # | Item | Status | How |
|---|---|---|---|
| 1 | A normal end reported as an error | fixed | `finish` normalizes `IsCloseError(1000, 1001)` to `nil`, plus a four-case `TestFinish_NormalizesARemoteCloseToASuccess` table |
| 2 | Ctrl+C reader cannot leave teardown | fixed | Extracted as `readCtrlC` with the `n > 0` guard and a `done` check, plus two tests |
| 3 | `ReadRune` parks | documented | Limit stated in the function comment, as asked |
| 4 | `os.Exit` from the `api/` layer | fixed | `dial` and `enterRawMode` both return errors; all three `cmd/` callers use `CliErrorWithExitCode(1, ...)` |
| 5 | Main path discards the outcome | fixed | `cmd/websh/websh.go:322` |
| 6 | `time.After` per iteration | fixed | `time.NewTicker` with `defer ticker.Stop()` |
| 7 | `fmt.Print(string(...))` | fixed | `_, _ = os.Stdout.Write(message)` |
| 8 | `err == io.EOF` | fixed | `errors.Is` |
| 9 | `Header` exported | fixed | `header` |
| 10 | `checkTerminal` duplication | fixed | `enterRawMode() (func(), error)` |
| 11 | Test duplication and naming | fixed | Folded into a five-case table, underscore naming throughout |
| 12 | Asymmetric doc comments | fixed | Both updated |

Item 4 in particular was fixed at the root rather than the surface. Once `dial` returned an error, the whole subprocess test scaffolding became unnecessary — `TestRunWsClientTerminalHelperProcess`, `TestDialHelperProcess`, and the `bytes` and `os/exec` imports are all gone. The test file gained coverage while getting structurally simpler.

### Critical issues (0)

### Warnings (1, new)

**`api/websh/websh.go:251` — fixing item 4 introduced a goroutine leak**

```go
defer signal.Stop(sigChan)
go wsClient.watchInterrupt(sigChan)   // ← 251

restore, err := enterRawMode()
if err != nil {
    return err                         // done is never closed
}
```

When `enterRawMode` fails, `finish` is never called, so `done` never closes. `watchInterrupt` waits on both `sigChan` and `done`, and `defer signal.Stop(sigChan)` **disarms delivery without closing the channel** — this is the exact mechanism the Background section identifies as the original reason the signal watcher parked. Neither case ever fires.

Reproduced against a real `OpenReadOnlyTerminal` call with stdin pointed at a pipe:

```
returned err = failed to set up terminal: stdin is not a terminal
goroutines before=3 after=6 (delta +3)
LEAK: watchInterrupt is still parked after OpenReadOnlyTerminal returned

goroutine 5 [select]:
  websh.(*WebsocketClient).watchInterrupt(...) websh.go:268
  created by websh.OpenReadOnlyTerminal websh.go:251
```

The old code hid this because `utils.CliErrorWithExit` killed the process; item 4's fix turned that into `return err` and exposed it. There is no user-visible impact today, since `cmd/websh/websh_watch.go:44` calls `os.Exit(1)` right after. But "The goroutines die with the process today ... The contract should not depend on that" is this PR's own thesis, and this is the one path where it no longer holds.

Moving the `go` statement below `enterRawMode` is enough — `sigChan` is buffered, so a signal arriving in the gap is not lost. I applied this and confirmed the leak is gone and `./api/websh/...` and `./cmd/websh/...` still pass:

```go
	defer signal.Stop(sigChan)

	restore, err := enterRawMode()
	if err != nil {
		return err
	}
	defer restore()

	go wsClient.watchInterrupt(sigChan)
	go wsClient.readCtrlC()
	go wsClient.readFromServer()
```

A `defer wsClient.finish(nil)` right after the client is created is the more durable alternative — it covers any early return added later, and is a no-op on the normal path thanks to the `sync.Once`.

### Suggestions (2)

**1. `api/websh/websh.go:227` — worth confirming the server actually sends a 1000**

The normalization catches `CloseNormalClosure` (1000) and `CloseGoingAway` (1001) only. `1006` (connection dropped with no close frame) and `1005` (close frame with no payload) still surface as errors, so if alpacon-server or alpamon does not send a proper close frame at session end, item 1's fix will not show up in production. Not verifiable from this repo — worth a check on the server side.

Separately: treating 1001 as success means a mid-session server restart drops the user back to their shell with exit 0 and no message. This is the canonical gorilla pairing so I am not arguing against it, just flagging that it is a decision.

**2. `api/websh/websh.go:325-331` — context baked into the error string**

`enterRawMode` builds `"failed to set up terminal: ..."` itself, where the Go convention is for the caller to wrap. The composed output reads `Websh session terminated with error: failed to set up terminal: stdin is not a terminal.` — doubled. Both paths are consistent and it works, so this is taste.

### Highlights

- **Item 4 was fixed at the root.** Making `dial` return an error let the subprocess test scaffolding disappear entirely, which is what the earlier review predicted would happen.
- **`readCtrlC`'s comment is honest about what it cannot do** — "though not before the read it is already parked in returns". The unfixable part is documented rather than glossed over.
- **`TestFinish_NormalizesARemoteCloseToASuccess` pins the cases that must *not* be normalized** (1006, transport failure), so widening the condition later gets caught.
- **The new `readCtrlC` case slots into the existing table.** Consolidating the tests did not cost coverage.

---

**Verdict**: COMMENTED — hold approval on one line.

Everything from the last round is addressed, and addressed properly. The only thing left is the new leak, and both the fix and its verification are above. Once that line moves, this is good to merge.

The reason to hold is not user impact — there is none today. It is that this PR exists to establish that session goroutines do not depend on process death for their exit, and it introduces one path where they do. Merging it as-is leaves that as precedent.

**Checklist**:
- [x] Tests included and passing — full suite green under `-race`
- [x] No security vulnerabilities — `SanitizeTerminalText` survives the move to error wrapping
- [x] Documentation updated if needed — items 3 and 12 landed

api/websh/websh.go
jisung-02: Fixed in c41f627 (the `go` statements moved in 18b7f7f; this reply covers both).

You were right, and the repro was exact. Every `go` statement now sits below `enterRawMode`, so nothing is started on the failing return.

One correction to what I did first. I also added the `defer wsClient.finish(nil)` you offered as the more durable option, and then removed it again in f50ff89 after measuring: with the ordering fixed, deleting the defer alone leaves the suite green, while moving a `go` statement back above `enterRawMode` fails `TestOpenReadOnlyTerminal_LeavesNoGoroutineWhenSetupFails`. The test is the durable guard, and the defer was only restating what the test already pins — and its comment ("the goroutines below have to be released") was false, since none is running by then. Happy to put it back if you would rather have belt and braces.

The test drives the real function against a live websocket server with stdin on a pipe. Two things it needs that are worth knowing: `assert.Eventually` cannot be used here because it evaluates its condition in a goroutine of its own, so the count never comes back down; and `signal.Notify` starts one process-wide goroutine on first use anywhere, so the baseline is primed before it is sampled.

Separately, your point about the two entry points led to c41f627: the interactive path had no signal watcher at all, so a SIGTERM there killed the process from the handler with no deferred restore and left the terminal in raw mode. It now installs the same watcher through a shared `notifySignals`.

api/websh/websh.go
jisung-02: Checked it against the counterpart repos, and you were right to ask — the fix did not work in production as written.

The CLI dials the proxy user channel, and the proxy closes it with 4000:

```
alpacon-server  websh/api/serializers.py:233   get_ws_proxy_url() + user_channel.get_user_ws_url()
proxy-server    cmd/alpacon-proxy-server/main.go:149   /ws/websh/{sessionID}/{channelID}/{token}/
                internal/ws/handler.go:157            defer userChannel.Disconnect()
                internal/ws/channel.go:351            sendCloseFrame(u.conn, "user channel")
                internal/ws/channel.go:337            FormatCloseMessage(sessionCloseCode, "")
                internal/ws/channel.go:66             sessionCloseCode = 4000
```

alpamon sends the same 4000 for the same meaning and already treats it as a normal end (`pkg/runner/pty.go:250`). A websh session never produces 1000, so the normalization would have passed its tests and changed nothing for users.

Fixed in f002c97: `sessionEndCloseCode = 4000` joins the set, and the test table gained a `session end` case. Mutation-checked — dropping 4000 fails that case and no other.

On 1001: kept, as you suggested it was a decision rather than a default. All three are deliberate closes, which is the line I drew against 1005/1006. The silent exit 0 on a mid-session restart is real, though — flagging it in the description rather than leaving it implicit.

api/websh/websh.go
jisung-02: Taken, in e5e01d8 — the doubling was worth removing even as taste.

`enterRawMode` no longer builds the phrase into its own message. `stdin is not a terminal` already names the stream and the fix, and the `MakeRaw` failure gets its own words instead of borrowing the setup phrase:

```go
return nil, errors.New("stdin is not a terminal")
...
return nil, fmt.Errorf("failed to enter raw mode: %w", err)
```

Composed output is now `Websh session ended with error: stdin is not a terminal.`

api/websh/websh.go
```diff
@@ -194,153 +204,204 @@ func CreateWebshSession(ac *client.AlpaconClient, serverName, username, groupnam
...
 	if err != nil {
-		utils.CliErrorWithExit("websocket connection failed %v", err)
+		return err
 	}
-	defer func() { _ = term.Restore(int(os.Stdin.Fd()), oldState) }()
+	defer restore()
 
 	inputChan := make(chan string, 1)
 
+	go wsClient.watchInterrupt(sigChan)
```
geunwoonoh, line 321: **Suggestion (not blocking):** the ordering invariant here is not pinned by a test, unlike the one in `OpenReadOnlyTerminal`.

`f50ff89` removed the defensive `defer wsClient.finish(nil)` on the grounds that the test is the durable guard. That reasoning holds — I mutation-checked it — but the guard only covers `OpenReadOnlyTerminal`. Now that `runWsClient` also starts `watchInterrupt`, it carries the same constraint: this `go` statement has to stay below `enterRawMode`, because that return path calls nothing that closes `done`, and `signal.Stop` disarms `sigChan` without closing it.

Mutation-checked:

```
move this go statement above enterRawMode
  → go test ./api/websh/   ...  ok      (nothing catches it)
  → direct probe           ...  LEAK: watchInterrupt parked after runWsClient returned
```

The code as written is correct — the same probe passes against the branch. This is only about the guard being asymmetric. Adding the goroutine-count assertion to the existing `TestRunWsClient_ReportsTerminalSetupFailure` would cover it, since that test already drives the raw-mode failure path.

geunwoonoh (approved):
## Follow-up review — `4927bde`

All three items from the last round are addressed, and the investigation of one of them turned up a production bug the suggestion itself had missed. Approving.

### What I verified

| Check | Result |
|---|---|
| `gofmt -l .` | clean |
| `go vet ./...` | clean |
| `go build -o /dev/null .` | ok |
| `go test -race -count=1 ./...` | all packages pass |

### Last round — 3/3 addressed

**1. The goroutine leak (warning) — fixed, with a regression test**

`notifySignals()` was extracted and the `go` statements moved below `enterRawMode`. `TestOpenReadOnlyTerminal_LeavesNoGoroutineWhenSetupFails` was added on top of that. I mutation-checked that it actually bites — moving the `go` statement back above `enterRawMode`:

```
--- FAIL: TestOpenReadOnlyTerminal_LeavesNoGoroutineWhenSetupFails
    Error: "5" is not less than or equal to "4"
    Messages: a goroutine was started before enterRawMode and outlived
              OpenReadOnlyTerminal: ... signal.Stop disarms sigChan without
              closing it, so the watcher would have no way out
```

The test also handles the two things that make this measurement hard: `signal.Notify` starts one process-wide goroutine on first use that never exits (primed out of the baseline), and `assert.Eventually` runs its condition in a goroutine of its own so it can never see the count come back down (polled inline instead). Both are written down in the comments.

**2. "Worth confirming the server actually sends a 1000" (suggestion) — checked, and it does not**

This is the most valuable change in the round. The proxy sends **4000**:

```go
// sessionEndCloseCode is how the proxy closes a user channel at the end of a
// session (sendCloseFrame in proxy-server internal/ws/channel.go); alpamon sends
// the same 4000 for the same meaning. A websh session never ends with 1000, so
// without this every normal end would be reported as a failure.
sessionEndCloseCode = 4000
```

Which means the previous round's 1000/1001 normalization was a **complete no-op in production** — every normal session end would still have printed an error. Verified end to end against a real 4000 close frame:

```
wsClient.err = <nil>
OK: the proxy's 4000 close normalizes to a clean end
```

Citing `proxy-server internal/ws/channel.go` in the comment is what keeps this magic number maintainable — whoever meets it next can trace the claim.

**3. Error-string context (suggestion) — solved differently, and better**

Rather than wrapping at the call site, each error was made self-sufficient: `stdin is not a terminal` is a complete statement that needs no prefix, while `failed to enter raw mode: %w` adds context where the underlying syscall error is opaque. The doubled prefix is gone.

### Beyond the review

- **`runWsClient` now watches signals too** (`c41f627`). The interactive path had no signal handling at all, so `kill <pid>` killed the process with the terminal still in raw mode. Ctrl+C still reaches the remote shell — raw mode clears ISIG, so it arrives as a byte rather than a signal.
- **`readUserInput`'s EOF collapsed into one `finish` call**, with a test.
- **The redundant `defer finish(nil)` was removed** (`f50ff89`) after measuring that the test catches the ordering regression without it. I confirmed that reasoning above.
- `utils.ExitCodeGeneralError` named, messages aligned with the "Websh" proper-noun rule in `CLAUDE.md`, comments this branch made stale corrected, and a duplicate test case dropped.

### Critical issues (0)
### Warnings (0)

### Suggestions (1, new — not blocking)

**`api/websh/websh.go:309-322` — the same invariant in `runWsClient` is not pinned**

`f50ff89` removed the defensive defer on the grounds that the test is the durable guard. That guard covers `OpenReadOnlyTerminal` only, and `runWsClient` now carries the same ordering constraint since it also starts `watchInterrupt`. Mutation-checked:

```
move runWsClient's go statement above enterRawMode
  → go test ./api/websh/   ...  ok      (nothing catches it)
  → direct probe           ...  LEAK: watchInterrupt parked
```

The current code is correct — the same probe passes against what is on the branch. But with the defer gone, both call sites deserve the same guard. A few lines of goroutine-count assertion on the existing `TestRunWsClient_ReportsTerminalSetupFailure` would do it.

### Highlights

- **The suggestion was verified rather than applied.** That 4000, not 1000, is what actually arrives is only discoverable by reading the server, and without it this PR's central fix would have been inert in production.
- **`f50ff89` made the right call.** Faced with a defensive defer and a test that both guard the same thing, it measured which one actually holds and kept only that.
- **The new case comment in `TestFinish_NormalizesARemoteCloseToASuccess`** states exactly what the other cases are worth: "every other case here is a contract the CLI honors but never meets in production".
- **`d5f8a3f` corrects only the comments this branch's own changes made stale** — scoped exactly right.

---

**Verdict**: APPROVED

Fifteen items across three rounds are addressed, and what remains is one test-symmetry suggestion, not a defect.

The basis for approving is not the response rate but the quality of the responses. Item 3 went a different way than suggested and went better. Item 2 started as verifying a suggestion and ended by finding the real bug that the suggestion had missed. `f50ff89` cut the author's own code on the strength of a measurement. This did not just absorb the review, it went past it.

**Checklist**:
- [x] Code follows stack conventions — `ExitCodeGeneralError` placement, "Websh" proper-noun rule
- [x] Tests included and passing — full suite green under `-race`; the regression test mutation-checked
- [x] No security vulnerabilities — `SanitizeTerminalText` preserved
- [x] Documentation updated if needed — comments reconciled; the exit-code docs already describe `1`, so no change needed

api/websh/websh.go
jisung-02: Thanks — I reproduced this. Moving `runWsClient`'s `go` statement above `enterRawMode` compiles and the suite stays green, so you're right that nothing pins it.

I'd like to leave it for a follow-up. The `OpenReadOnlyTerminal` test pins a regression that actually happened on this branch; the same assertion on `runWsClient` would pin one that hasn't, and I'd rather not add a second copy of that measurement — the priming and the inline poll — on a hypothetical. It's the same standard I used to remove the defensive defer in `f50ff89`, so it seemed fairer to apply it consistently.

Happy to add it now if you'd prefer the symmetry; it's a small change and I don't feel strongly.

Issue #312, opened 2026-08-06
Approval-wait loops poll on a fixed interval and abort on 429


jisung-02:
## Summary

#309 fixed the command-status poll loop in `api/event/event.go`, and explicitly left the approval-wait loops out of scope. Those loops have the same problem in a different shape: they poll on a fixed interval with no backoff, they ignore the `Retry-After` the client now exposes, and — unlike the command poll, which merely starved — a single 429 aborts the wait outright.

## 1. `exec --wait` re-attempt loop (highest impact)

`cmd/exec/run.go:248`

```go
ticker := time.NewTicker(approvalWaitPollInterval)  // 5s, fixed
```

Each tick re-submits the command through `runPresenceStepUp`, so a tick is not one request. `RunCommandStreaming` performs `CreateEventSession`, a WebSocket connect, `SubmitCommand`, `SubscribeEvent`, `getCommandChunks`, and at least one status GET — five or more HTTP requests per tick.

- default `--wait` (5m): ~60 ticks, ~300 requests
- `--wait-approval 30m`: ~360 ticks, ~1,800 requests, past the default `service_token: 1000/hour` quota before the wait ends

The failure mode is worse than a starved deadline. A 429 is neither a `RemoteCommandError` nor an approval denial, so `isApprovalDenial(err)` is false and the loop takes the terminal branch at `cmd/exec/run.go:277`:

```go
spinner.Stop()
return err
```

The wait ends with the throttle error it caused, while the approval request is still perfectly valid.

There is a design constraint here worth naming: as the comment at `cmd/exec/run.go:213` explains, the plugin's denial line carries no approval request id and this credential channel has no approval-status endpoint (ADR 0015 moves approval out of band), so re-running the command is the only poll available. Slowing the tick reduces the cost but does not make it cheap; a status endpoint the CLI could poll instead would.

## 2. `work-session create --wait` (no retry at all)

`cmd/worksession/worksession_create.go:437`

```go
s, err := wsapi.GetWorkSession(ac, id)
if err != nil {
    return nil, fmt.Errorf("polling failed: %w", err)
}
```

One GET per 10s tick (`pollInterval`), so request volume is not a concern — 30 requests over the 5-minute default. The issue is that any transient error ends the wait: a single 429, or a brief network blip, discards a wait the user asked to run for 30 minutes.

## 3. MFA step-up polls (lower priority)

[...]

PR #313, opened 2026-08-06
fix(event): pace the command poll and back off when throttled

### LINKED ISSUE #309: exec/websh polling: fixed 1s tick ignores 429/Retry-After and times out long-running commands under service-token throttling
## Summary

When the event-WebSocket path is unavailable and `websh`/`exec` falls back to polling, the poll loop ticks at a fixed 1s, ignores HTTP 429 (including the server's `Retry-After`), and only resets its 30-minute deadline on a running-status response. Under alpacon-server's default service-token throttle (`service_token: 1000/hour`), any long wait deterministically exhausts the quota and then starves itself: the loop keeps consuming freed slots at 1 req/s, never receives a successful response, and dies with `[client_timeout]` — even when the remote command finished long ago.

## Observed (production CI, 2026-08-05)

Three consecutive `Deploy Alpacon AI` runs in alpacax/iac-onprems failed this way. Every websh in CI first prints:

```
Warning: real-time output unavailable (failed to create event session: You do not have permission to perform this action.); falling back to polling
```

(the CI service token lacks the `event_session` scope), then:

- Run A: `docker compose up` blocked on a large image pull; after ~40 min of 1s polling → `Error: [client_timeout] CLI gave up waiting for the server to report a result`
- Run B: a **no-op** `manage.py migrate` finished server-side in seconds (`No migrations to apply.` was eventually delivered), but its result GET was throttled for 30 straight minutes → `[client_timeout]`, exit 1
- Run C: quota still hot from B → the next websh failed at submission: `Error: failed to execute command on 'r770-1' server: Request was throttled. Expected available in 29 seconds.`
- Collateral: a subsequent `alpacon login` on the same account: `Login failed: Request was throttled. Expected available in 649 seconds.`

## Mechanism

`api/event/event.go`:

- `PollCommandExecution` → `pollCommandExecution(ac, cmdId, execTimeout(), 1*time.Second, false)` — fixed 1s tick, default 30m deadline (`ALPACON_EXEC_TIMEOUT`)
- Inside the loop:

```go
responseBody, err := ac.SendGetRequest(...)
if err != nil {
    continue   // 429 (and any other error) retried on the next 1s tick; Retry-After discarded
}
```

- The deadline timer resets only on `IsRunningStatus(response.Status)`. While throttled there are no successful responses, so the timer can never reset — an already-completed command's result becomes unfetchable and the wait dies at the deadline.

The arithmetic: 1 req/s = 3,600 req/h against a 1,000/h default quota — a single ~17-minute wait exhausts it; after that, the 1s loop eats each freed slot, so recovery never happens while the loop runs.

## Proposed fix (three parts)

1. **Honor 429 properly.** `client.go` already attaches the HTTP status to errors (`withStatus` / `utils.HTTPStatusCode`), so the loop can branch on 429 today; additionally surface the `Retry-After` header on the error and sleep that long (falling back to exponential backoff, e.g. 2s→4s→…cap 60s, when absent). Reset backoff on the first success. Non-429 transient errors get a milder backoff (cap ~15s) instead of a 1s hammer.
2. **Decay the poll tick.** Keep 1s for the first ~10s (snappy short commands), then grow to a cap of 5–10s. A 30-minute wait drops from ~1,800 requests to ~200–360, fitting inside the default quota with headroom.
3. **Treat 429 as liveness for the deadline.** A 429 means the server is alive and pushing back, not that the command is lost; it should not let the deadline starve a command whose result is already available. Keep an absolute overall cap to avoid unbounded waits.

## Notes

- Complementary server-side mitigations exist (per-token `rate_limit` override; granting the `event_session` scope restores WebSocket streaming and removes polling entirely), but the CLI should still behave correctly under throttling — any fallback-to-polling client with a long-running command currently cannot succeed within the default quota.
- `cmd/exec/run.go`'s `approvalWaitPollInterval` (5s) is a separate, already-throttled loop and is not part of this issue.

jisung-02:
## Background

When the event WebSocket is unavailable, `exec` and `websh` fall back to polling the command's status on a fixed 1s tick. That is 3,600 requests an hour against alpacon-server's default `service_token: 1000/hour`, so the quota is gone in about 17 minutes—and once it is gone the loop starves itself, spending every freed slot on a request that is throttled again. #309 has the production CI evidence: a no-op `manage.py migrate` that finished server-side in seconds was reported as `[client_timeout]` after 30 minutes of throttled polling, and the exhausted quota then broke an unrelated `alpacon login` on the same account.

The loop also discarded the server's `Retry-After`, which the client layer never exposed in the first place.

## Changes

The poll gap now follows the command's age and the server's answers, the deadline is no longer spent on requests that were refused, and the streaming path stops waiting for a poll to notice the command is over.

| Location | What changes |
|---|---|
| `withRetryAfter`, `utils.RetryAfter` | `Retry-After` rides the error chain next to the status code, reachable the same way `utils.HTTPStatusCode` already reaches the status. Only the delta-seconds form is parsed—alpacon-server sends seconds through DRF throttling, and a misread HTTP-date would park a poll until the deadline. All three response paths tag it: JSON, non-JSON (a gateway throttle answers with its own HTML), and `SendMultipartStreamRequest`. A value too large for a `time.Duration` is treated as no hint rather than allowed to wrap |
| `nextPollTick` | The gap widens with the poll's age—base tick for the first 10 ticks, 5x through 60, 10x after. A 30-minute wait drops from ~1,800 requests to ~200. Age runs from where the poll starts rather than from submission, so a command resuming after an approval hold gets the fast window again; that is the forgiving direction and is left alone |
| `nextPollBackoff` | Doubles on consecutive failures up to 60 ticks, with the server's `Retry-After` preferred when it sends one. The same cap covers `Retry-After`, so a value the server picks cannot park the loop until the deadline |
| `pollCommandExecution` deadline | Time lost to a 429 extends the deadline instead of spending it: a throttled GET says nothing about the command, which may well have finished already. The extension has a budget of one more timeout, spent whole on the last extension rather than trimmed, so the ceiling is one timeout plus one backoff wait and a token permanently over quota gives up instead of extending forever. Trimming it would let a mandated wait expire the deadline mid-sleep, so the loop would serve the wait and never make the poll it waited for. The budget is capped by count as well—60 grants—because a duration alone converts to whatever request count the server's chosen wait dictates: 30 minutes of extension is 30 polls at the capped 60s wait and 1,800 at a flat `Retry-After: 1`. A running status refreshes both along with the deadline they belong to |
| `pollCommandExecution` warning | One warning per throttled stretch: the first 429 warns, and reported progress clears the flag along with the allowance it belongs to, so a second stretch minutes later is not silent either. With the extension budget the wait can approach two timeouts, and a silent `alpacon exec` is indistinguishable from a hung one |
| `pollCommandExecution` wait, `pollSeams` | The timer and ticker become one computed wait, which also drops the drain-before-`Reset` dance the running-status branch needed. It never runs past the deadline, so `[client_timeout]` is reported at the deadline it names rather than one gap later—10 ticks at the slow pace, 60 once a throttle widened it. The wait is a channel receive rather than a sleep so it can lose to a cancel, and `pollSeams` carries the clock behind it—passed in, not a package var, since a poll can outlive the stream that started it by an in-flight request |
| `errPollCancelled`, `streamSubscribed` cancel | The poll ends when the run does. A run that exits on the fin left its poll behind, and since the command is already over the orphan spends one more request of the very budget this pacing protects—twice over for a process that opens a second stream, which `cmd/exec`'s MFA retry does |
| `EventTypeCommandFin`, `CommandOutputListener.Finished` | A wider poll gap would otherwise show up as tail latency on the streaming path, where the poll is not the output carrier but the only thing that notices the command is over: chunks carry no end marker, so an `alpacon exec` could sit for up to 10 ticks after its output had already arrived. alpacon-server publishes `command_fin` when alpamon reports the command handled, on the same channel and frame shape as the chunks, so the listener reports that too. The subscription is per server, so the listener filters on the command id in the payload |
| `streamSubscribed` | The run ends on the fin. The fin only says it is over, so the exit code still comes from one GET; a read that fails, or that lands on a status the poll would keep waiting on, falls back to the poll rather than reporting a half-finished run. Both paths ask the same `isPollWaitStatus`, so the approval hold and the transient `error` of the approve-to-deliver window mean the same thing on either. The poll stays as the net for everything fin does not cover—a rejected command, an agent that never fins, a dropped WebSocket |

## Behaviour change

Under a permanent throttle the loop now spends up to two timeouts before returning `[client_timeout]`, where before it returned at one—an hour at the default `ALPACON_EXEC_TIMEOUT`. Two timeouts is what the duration budget buys at the default timeout when the server asks for waits at the cap; a throttle asking for shorter ones runs out of grants first and returns sooner. Past a timeout of an hour the count is what binds at the cap instead, so the extension is shorter than a timeout there. The rate-limit warning prints as soon as either begins, so the wait is not silent.

`cp` inherits the new pacing as well. `api/ftp/ftp.go:684` and `api/ftp/ftp.go:739` wait on the exported `PollCommandExecution`, whose signature is unchanged, so a transfer that outlives the fast and medium windows is noticed up to 10s after it actually finished instead of within 1s. The fin exit does not cover that path—it belongs to the streaming subscription, and `cp` only polls—so the tail latency is what buys `cp` the same quota relief.

## Left as they are

Three decisions that read like oversights on review, so the reasoning is written down.

| Decision | Why |
|---|---|
| A `Retry-After` replaces the exponential schedule outright rather than being combined with it | A server that answers every 429 with `Retry-After: 1` then holds the loop at one poll per second, which is the exhaustion pattern this PR is meant to fix. It is still the right default, and the "Retry-After wins even when it is shorter" case in `TestNextPollBackoff` pins it: the server knows when its own window frees up, and second-guessing it with `max(retryAfter, schedule)` means waiting longer than asked on the common path. alpacon-server sends the remaining window through DRF `SimpleRateThrottle.wait()`, which is a real number of seconds, not 1. A proxy throttle in front of it (nginx `limit_req`, a CDN) is where a flat `Retry-After: 1` is plausible, and that path is in range here since the non-JSON branch now tags the header too. What bounds that case is not `nextPollBackoff`'s cap—that bounds the gap, not the request count—but the extension count: 60 grants, so a flat `Retry-After: 1` adds 60 polls to the window instead of a second window's worth of them. Worth switching to `max` if it turns up in practice anyway |
| A fin whose GET fails leaves no trace in the output | The fin is consumed, so the run falls back to a poll that may be up to 10 ticks out, and from outside that is indistinguishable from a fin that never fired. A warning is wrong for a path that recovers by design, and the CLI has no debug level to put it at—`utils` prints errors, info, and warnings only. Adding one is a decision wider than this line |
| The GET the fin triggers runs inline in the streaming select | The loop stops draining `Chunks()` for the length of one round trip, and a full 256-chunk buffer would block the WebSocket read behind it. Nothing is lost—the drain that follows re-reads everything past `lastSeq` over REST—so the cost is that one round trip, and the same select already makes a synchronous REST call for gap filling. Moving it off the loop buys a round trip and pays a goroutine, a channel, and a shutdown race for it |

## Known limits

One race this does not close, and widens: alpamon posts a command's chunks (queue priority 10) before its fin (11), but its reporters run concurrently, so the fin can be persisted first—its own comment says as much, "retries and concurrent reporters can still race". The drain that follows a fin then misses a chunk still in flight, and with `lastSeq` already advanced the buffered `Result` is not printed either, so the tail is dropped silently.

The poll had the same single drain, only 0 to 10 ticks later, which hid it. Nothing in the CLI can tell the two apart: a chunk still inside the agent is absent from the server's `max_seq` too, and `result` is masked where live chunks are raw, so a byte comparison is not sound. The fix belongs in the fin payload, since the agent knows the last seq it sent, and is filed as alpacax/alpacon-server#2885. Until then this is the pre-existing behaviour with a narrower window in front of it.

One narrower limit on the extension count: it bounds the extra requests only while the mandated wait keeps its length. A throttle that answers with long waits until the duration budget is spent and then shortens them can still burn the extended window at the short pace. The extra wall-clock is bounded by one timeout either way, and it takes a server changing its own answer mid-stretch, so it is left as is.

## Testing

```
gofmt -l .                      no output
go vet ./...                    clean
golangci-lint run ./...         0 issues
go test -race ./...             38 packages ok, 0 failures
go test -race -shuffle=on ./... 38 packages ok, 0 failures
```

`api/event` was also run at `-shuffle=1` through `8` individually. The shuffled run is not decoration here: it is what caught the poll goroutine outliving its stream. Left alive it read clock seams a later test had swapped, and the race surfaced in whichever test happened to run next rather than in the one that leaked it—which is why the seams are now passed in instead of swapped.

New cases in `api/event/event_test.go`, `api/event/commandoutput_listener_test.go`, and `client/client_test.go` cover the poll schedule, the throttle backoff and its cap from both sides, the deadline extension and both of its bounds, the allowance reset on progress, `Retry-After` on each response path, the fin filter, and the fin-driven exit. Each one was confirmed to fail before its fix—the duration-bound case hangs until the test timeout without the cap, the pacing case sees only base ticks if the loop measures the gap since the previous poll instead of how long it has been polling, and the fin case never ends at all, since its poll tick is set an hour out.

The two count-bound cases were checked the same way, by breaking the production line and reading the number that moved. Dropping the count from the extension condition grows the window a permanently throttled poll gets from 1.08s to 1.44s, exactly the second budget the duration alone would have granted; dropping the count's reset from the progress branch stops it at 720.1ms, the refreshed deadline with no grants left to spend.

The fin exit is tested on a command that prints nothing, the case #309 opened with: with no output there is nothing but a terminal signal to end the run on. Its fallbacks are pinned as well—a read that races a status the server has not written yet, one that fails outright, one that lands on an approval hold or the transient `error` beside it, and, on the approval-resume path, a failed server lookup that leaves the fin subscription unsent—since each must leave the run to the poll rather than report what it read.

None of the three decides anything on elapsed time. The poll tick reaches `streamSubscribed` as a parameter, so the exit case sets it an hour out and the fin becomes the only thing that can end the run; the fallbacks count reads, where a second one means the first ended nothing. An assertion like "finished within 500ms" would only ever be measuring the distance to the 1s tick it was racing, and would fail on a loaded runner for reasons that have nothing to do with the code.

The throttle cases run on a fake clock rather than the wall clock. Timing real waits made the result depend on the machine: a slow runner stretches every wait, which makes a loop with no backoff look better-behaved than it is and can expire a deadline the loop meant to extend.

Also run against a live workspace, `alpacon exec web-editor "sleep 65; echo done"`, so the command outlives the fast and medium pacing windows:

```
before this PR's fin commits   77.2s
with them                      72.1s
alpacon exec ... "echo x"       5.1s   (same on both, the fixed setup cost)
```

Subtracting the 65s the command itself takes and the setup, the exit went from roughly 7s after the output to roughly 2s.

## Follow-up

This is the first CLI path to subscribe to `command_fin`, and the subscription has no permission branch on the server side: `EventSubscriptionSerializer.validate` covers `sudo`, `work_session`, `command_output`, and `notification`, not `command_fin`. Any authenticated user can subscribe with a server id they have no access to, and an omitted `target_id` wildcard-matches every server, since `get_subscribed_channels` only narrows by target when one is given. What that exposes is metadata—the payload is `{type, target, event, id}`, a command id and the fact that it finished, and `command_output` keeps its per-command requester check.

Filed against the server as a note on alpacax/alpacon-server#2885, which already carries the chunk-seq race below, with `command_reject` included: it shares the same `notify_command_event` publish path and has no branch either. The event type predates this PR, so the CLI is not opening the gap; it is the first consumer to put traffic on the channel.

The approval-wait loops in `cmd/exec/run.go` and `cmd/worksession/worksession_create.go` have the same fixed-tick problem and are deliberately out of scope here—filed as #312.

Closes #309

### DIFF
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path"
 	"time"
@@ -18,6 +19,17 @@ import (
 
 const (
 	getEventURL = "/api/events/commands/"
+
+	// Poll pacing, in multiples of the caller's base tick, widening as the command
+	// ages and again once the server pushes back. A fixed 1s tick burns alpacon-server's
+	// default 1000/hour service-token throttle in ~17 minutes, then starves itself—each
+	// freed slot goes to a request that is throttled again.
+	pollFastWindow     = 10
+	pollMediumWindow   = 60
+	pollMediumTick     = 5
+	pollSlowTick       = 10
+	pollBackoffFactor  = 2
+	pollMaxBackoffTick = 60
 )
 
 // maxGapWidth caps enumerated missing seqs (per gap and cumulatively) so a buggy or
@@ -37,6 +49,12 @@ var (
 
 // gapFillNow is a seam so tests drive backoff timing deterministically.
 var gapFillNow = time.Now
+
+// pollNow and pollSleep are seams so tests run the loop on a fake clock.
+var (
+	pollNow   = time.Now
+	pollSleep = time.Sleep
+)
 
 // gapFillState throttles gap-fill re-fetches while lastSeq is not advancing and
 // records seqs given up on so they can be retried once at command end.
@@ -239,48 +257,79 @@ func execTimeout() time.Duration {
 	return 30 * time.Minute
 }
 
-// waitApproval keeps polling through the awaiting_approval hold (bounded by
-// timeout, no reset) so an approved job resumes streaming; otherwise that hold
-// is terminal and surfaces as a PendingApprovalError via errorFromDetails.
+func nextPollTick(tick, elapsed time.Duration) time.Duration {
+	switch {
+	case elapsed < time.Duration(pollFastWindow)*tick:
+		return tick
+	case elapsed < time.Duration(pollMediumWindow)*tick:
+		return time.Duration(pollMediumTick) * tick
+	default:
+		return time.Duration(pollSlowTick) * tick
+	}
+}
+
+// nextPollBackoff returns the gap after a failed poll; attempt is the failure
+// count so far, 0 for the first. A server-sent retryAfter wins but shares the
+// cap, so a hostile value cannot park the loop until the deadline.
+func nextPollBackoff(tick time.Duration, attempt int, retryAfter time.Duration) time.Duration {
+	maxDelay := time.Duration(pollMaxBackoffTick) * tick
+	if retryAfter > 0 {
+		return min(retryAfter, maxDelay)
+	}
+	delay := tick
+	for range attempt {
+		delay *= time.Duration(pollBackoffFactor)
+		if delay >= maxDelay {
+			return maxDelay
+		}
+	}
+	return delay
+}
+
+// waitApproval polls through the awaiting_approval hold—bounded by timeout, which
+// the hold never resets and only a throttled wait extends—so an approved job
+// resumes streaming. Without it the hold is terminal (PendingApprovalError).
 func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick time.Duration, waitApproval bool) (EventDetails, error) {
 	var response EventDetails
 
-	timer := time.NewTimer(timeout)
-	defer timer.Stop()
-	ticker := time.NewTicker(tick)
-	defer ticker.Stop()
+	started := pollNow()
+	deadline := started.Add(timeout)
+	delay := tick
+	failures := 0
 
 	for {
-		select {
-		case <-timer.C:
+		pollSleep(delay)
+		if pollNow().After(deadline) {
 			return response, &ClientTimeoutError{}
-		case <-ticker.C:
-			responseBody, err := ac.SendGetRequest(utils.BuildURL(getEventURL, cmdId, nil))
-			if err != nil {
-				continue
-			}
-			if err = json.Unmarshal(responseBody, &response); err != nil {
-				return response, err
-			}
-
-			if IsRunningStatus(response.Status) {
-				// Drain timer.C before Reset to prevent a spurious ClientTimeoutError if the timer fires between Stop and Reset.
-				if !timer.Stop() {
-					select {
-					case <-timer.C:
-					default:
-					}
-				}
-				timer.Reset(timeout)
-				continue
-			}
-			// Approval-resume keeps polling through the hold and the transient
-			// "error" compute_status the server emits in the approve→deliver window.
-			if waitApproval && (IsAwaitingApprovalStatus(response.Status) || response.Status == "error") {
-				continue
-			}
-			return response, nil
 		}
+
+		responseBody, err := ac.SendGetRequest(utils.BuildURL(getEventURL, cmdId, nil))
+		if err != nil {
+			delay = nextPollBackoff(tick, failures, utils.RetryAfter(err))
+			failures++
+			if utils.HTTPStatusCode(err) == http.StatusTooManyRequests {
+				// The server is alive: the command may be done, only its result GET refused.
+				deadline = deadline.Add(delay)
+			}
+			continue
+		}
+		failures = 0
+		if err = json.Unmarshal(responseBody, &response); err != nil {
+			return response, err
+		}
+
+		if IsRunningStatus(response.Status) {
+			deadline = pollNow().Add(timeout)
+			delay = nextPollTick(tick, pollNow().Sub(started))
+			continue
+		}
+		// Approval-resume keeps polling through the hold and the transient
+		// "error" compute_status the server emits in the approve→deliver window.
+		if waitApproval && (IsAwaitingApprovalStatus(response.Status) || response.Status == "error") {
+			delay = nextPollTick(tick, pollNow().Sub(started))
+			continue
+		}
+		return response, nil
 	}
 }
 
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -54,11 +54,36 @@ var (
 // gapFillNow is a seam so tests drive backoff timing deterministically.
 var gapFillNow = time.Now
 
-// pollNow and pollSleep are seams so tests run the loop on a fake clock.
-var (
-	pollNow   = time.Now
-	pollSleep = time.Sleep
-)
+// errPollCancelled ends a poll whose caller has already finished—a stream that
+// exited on the fin event—so it stops instead of spending another request of the
+// throttle budget this pacing exists to protect.
+var errPollCancelled = errors.New("poll cancelled")
+
+// pollSeams are the hooks only a stream or a test supplies: cancellation, and a
+// clock in place of real time. The zero value polls uncancellably on the real one.
+// Passed rather than kept in package vars—a poll outlives the stream that started
+// it by an in-flight request, and a test swapping a var would swap it under a
+// loop still running.
+type pollSeams struct {
+	cancel <-chan struct{}
+	now    func() time.Time
+	after  func(time.Duration) <-chan time.Time
+}
+
+func (s pollSeams) Now() time.Time {
+	if s.now == nil {
+		return time.Now()
+	}
+	return s.now()
+}
+
+// After hands back a channel rather than blocking, so a wait can lose to cancel.
+func (s pollSeams) After(d time.Duration) <-chan time.Time {
+	if s.after == nil {
+		return time.After(d)
+	}
+	return s.after(d)
+}
 
 // gapFillState throttles gap-fill re-fetches while lastSeq is not advancing and
 // records seqs given up on so they can be retried once at command end.
@@ -248,7 +273,7 @@ func GetCommandByID(ac *client.AlpaconClient, cmdID string) (EventDetails, error
 
 // PollCommandExecution polls with default timeout/tick; tests use pollCommandExecution directly.
 func PollCommandExecution(ac *client.AlpaconClient, cmdId string) (EventDetails, error) {
-	return pollCommandExecution(ac, cmdId, execTimeout(), 1*time.Second, false)
+	return pollCommandExecution(ac, cmdId, execTimeout(), 1*time.Second, false, pollSeams{})
 }
 
 func execTimeout() time.Duration {
@@ -301,11 +326,12 @@ func isPollWaitStatus(status string, waitApproval bool) bool {
 // waitApproval polls through the awaiting_approval hold—bounded by timeout, which
 // the hold never resets and throttled waits extend by at most one timeout plus one
 // backoff wait—so an approved job resumes streaming. Without it the hold is
-// terminal (PendingApprovalError).
-func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick time.Duration, waitApproval bool) (EventDetails, error) {
+// terminal (PendingApprovalError). A closed seams.cancel ends the poll with
+// errPollCancelled.
+func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick time.Duration, waitApproval bool, seams pollSeams) (EventDetails, error) {
 	var response EventDetails
 
-	started := pollNow()
+	started := seams.Now()
 	deadline := started.Add(timeout)
 	delay := tick
 	failures := 0
@@ -313,13 +339,22 @@ func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick 
 	throttleWarned := false
 
 	for {
+		select {
+		case <-seams.cancel:
+			return response, errPollCancelled
+		default:
+		}
 		// Never sleep past the deadline, or the timeout lands a whole gap late—10
 		// ticks at the slow pace, 60 once throttled. An extended deadline always
 		// leaves the whole wait, so a throttled retry still gets its poll.
-		if wait := min(delay, deadline.Sub(pollNow())); wait > 0 {
-			pollSleep(wait)
+		if wait := min(delay, deadline.Sub(seams.Now())); wait > 0 {
+			select {
+			case <-seams.After(wait):
+			case <-seams.cancel:
+				return response, errPollCancelled
+			}
 		}
-		if !pollNow().Before(deadline) {
+		if !seams.Now().Before(deadline) {
 			return response, &ClientTimeoutError{}
 		}
 
@@ -351,18 +386,18 @@ func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick 
 		}
 
 		if IsRunningStatus(response.Status) {
-			deadline = pollNow().Add(timeout)
+			deadline = seams.Now().Add(timeout)
 			// The refreshed deadline gets its throttle allowance back with it, so a
 			// throttle late in a long command is not paid for by an earlier one—and
 			// the warning with it, so a later stretch is not silent.
 			throttledWait = 0
 			throttleWarned = false
-			delay = nextPollTick(tick, pollNow().Sub(started))
+			delay = nextPollTick(tick, seams.Now().Sub(started))
 			continue
 		}
 		// Running is handled above, so what is left here is the approval hold.
 		if isPollWaitStatus(response.Status, waitApproval) {
-			delay = nextPollTick(tick, pollNow().Sub(started))
+			delay = nextPollTick(tick, seams.Now().Sub(started))
 			continue
 		}
 		return response, nil
@@ -459,13 +494,19 @@ func streamSubscribed(ac *client.AlpaconClient, session *EventSessionResponse, l
 
 	pollResult := make(chan EventDetails, 1)
 	pollErr := make(chan error, 1)
+	// The poll ends when this function does: an exit on the fin event would
+	// otherwise leave it polling a command that is already over.
+	pollCancel := make(chan struct{})
+	defer close(pollCancel)
 	go func() {
-		details, err := pollCommandExecution(ac, cmdID, timeout, tick, waitApproval)
-		if err != nil {
+		details, err := pollCommandExecution(ac, cmdID, timeout, tick, waitApproval, pollSeams{cancel: pollCancel})
+		switch {
+		case errors.Is(err, errPollCancelled): // the stream is gone; nobody reads these
+		case err != nil:
 			pollErr <- err
-			return
+		default:
+			pollResult <- details
 		}
-		pollResult <- details
 	}()
 
 	// Runs on whichever terminal signal arrives first, the fin event or the poll.
@@ -673,7 +714,7 @@ func runCommandFallback(ac *client.AlpaconClient, serverName, command, username,
 // blocking through an awaiting_approval hold so --wait is honored on this path.
 func runCommandFallbackFromID(ac *client.AlpaconClient, cmdID string, out io.Writer, waitApproval bool, cause error) error {
 	utils.CliWarning("real-time output unavailable (%v); falling back to polling", cause)
-	details, err := pollCommandExecution(ac, cmdID, execTimeout(), 1*time.Second, waitApproval)
+	details, err := pollCommandExecution(ac, cmdID, execTimeout(), 1*time.Second, waitApproval, pollSeams{})
 	if err != nil {
 		return err
 	}
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -32,8 +32,7 @@ const (
 	pollMaxBackoffTick = 60
 
 	// Base gap the pacing above multiplies, for the poll running behind a live
-	// stream. Passed to streamSubscribed rather than read there, so a test can
-	// widen it away and leave the fin event as the only thing that ends a run.
+	// stream.
 	streamPollTick = time.Second
 )
 
@@ -338,8 +337,7 @@ func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick 
 				// The server is alive: the command may be done, only its result GET refused.
 				// Budgeted so a token stuck over quota gives up—extending by exactly the
 				// wait would keep the deadline ahead forever. The last extension is granted
-				// whole rather than trimmed, so every mandated wait still buys its poll:
-				// the ceiling is one timeout plus one backoff wait.
+				// whole rather than trimmed, so every mandated wait still buys its poll.
 				if throttledWait < timeout {
 					deadline = deadline.Add(delay)
 					throttledWait += delay
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -287,8 +287,8 @@ func nextPollBackoff(tick time.Duration, attempt int, retryAfter time.Duration) 
 }
 
 // waitApproval polls through the awaiting_approval hold—bounded by timeout, which
-// the hold never resets and throttled waits extend by at most one more timeout—so
-// an approved job resumes streaming. Without it the hold is terminal (PendingApprovalError).
+// the hold never resets and throttled waits extend by at most one timeout plus one
+// backoff wait—so an approved job resumes streaming. Without it the hold is terminal (PendingApprovalError).
 func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick time.Duration, waitApproval bool) (EventDetails, error) {
 	var response EventDetails
 
@@ -317,8 +317,11 @@ func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick 
 					utils.CliWarning("rate limited by the server, retrying in %s", delay)
 				}
 				// The server is alive: the command may be done, only its result GET refused.
-				// Capped at one more timeout, so a token stuck over quota still gives up:
-				// extending by exactly the wait would otherwise keep the deadline ahead forever.
+				// Budgeted so a token stuck over quota gives up: extending by exactly the
+				// wait would keep the deadline ahead of the clock forever. The extension
+				// that exhausts the budget is granted whole rather than trimmed to what is
+				// left, so every wait the server mandates still buys the poll it waited
+				// for—the ceiling is one timeout plus one backoff wait.
 				if throttledWait < timeout {
 					deadline = deadline.Add(delay)
 					throttledWait += delay
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -20,8 +20,8 @@ import (
 const (
 	getEventURL = "/api/events/commands/"
 
-	// Poll pacing, in multiples of the caller's base tick, widening as the command
-	// ages and again once the server pushes back. A fixed 1s tick burns alpacon-server's
+	// Poll pacing, in multiples of the caller's base tick, widening as the poll ages
+	// and again once the server pushes back. A fixed 1s tick burns alpacon-server's
 	// default 1000/hour service-token throttle in ~17 minutes, then starves itself—each
 	// freed slot goes to a request that is throttled again.
 	pollFastWindow     = 10
@@ -35,8 +35,9 @@ const (
 	// duration: a duration budget is spent at whatever pace the server asks for, so a
 	// flat Retry-After: 1 buys a second window of one-second polls. 60 grants at the
 	// capped 60-tick wait is 3,600 ticks—a whole quota window at the default
-	// one-second tick—so the duration budget binds first at the cap and this bound
-	// only catches a throttle asking for less.
+	// one-second tick—so at the default timeout the duration budget binds first at the
+	// cap and this bound only catches a throttle asking for less. Past an hour's
+	// timeout the count binds at the cap too.
 	pollMaxThrottleExtensions = 60
 
 	// Base gap the pacing above multiplies, for the poll running behind a live
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -32,6 +32,11 @@ const (
 	pollMaxBackoffTick = 60
 )
 
+// streamPollTick is the base gap for the status poll behind a live stream, the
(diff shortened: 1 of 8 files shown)


api/event/event.go
```diff
@@ -239,48 +257,95 @@ func execTimeout() time.Duration {
...
+					throttleWarned = true
+					utils.CliWarning("rate limited by the server, retrying in %s", delay)
+				}
+				// The server is alive: the command may be done, only its result GET refused.
+				// Capped at one more timeout, so a token stuck over quota still gives up:
+				// extending by exactly the wait would otherwise keep the deadline ahead forever.
+				if throttledWait < timeout {
+					deadline = deadline.Add(delay)
+					throttledWait += delay
 				}
```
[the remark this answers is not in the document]
jisung-02: Good catch on the arithmetic—you are right that the total can exceed one timeout. The extension that exhausts the budget is added whole rather than trimmed to the allowance left, so the true ceiling is one timeout plus one backoff wait (60 ticks, capped in `nextPollBackoff`).

That overshoot is deliberate, so I corrected the claim rather than the code (866f940):

- Trimming the last extension to what is left would let a wait the server mandated expire the deadline mid-sleep. The loop would serve the full `Retry-After` wait and then time out without ever making the poll it was waiting for—and a throttled GET says nothing about the command, which may well have finished.
- The termination property you would expect from the cap still holds: `throttledWait` only grows, so once it reaches `timeout` no further extension is granted.

The comment and the `pollCommandExecution` doc comment now name the real ceiling, and `TestPollCommandExecution_ThrottleExtensionIsBounded` already pins the exact delay sequence that encodes it. Thank you for the precise reading.

api/event/event.go
```diff
@@ -239,48 +286,121 @@ func execTimeout() time.Duration {
...
+					// The wait can reach two timeouts, and nothing else on this path
+					// prints while it lasts.
+					throttleWarned = true
+					utils.CliWarning("rate limited by the server, retrying in %s", delay)
+				}
+				// The server is alive: the command may be done, only its result GET refused.
+				// Budgeted so a token stuck over quota gives up—extending by exactly the
+				// wait would keep the deadline ahead forever. The last extension is granted
+				// whole rather than trimmed, so every mandated wait still buys its poll.
+				if throttledWait < timeout {
```
junho226: Warning: this extension doubles the request count in the one scenario the "Left as they are" table dismisses.

With a flat `Retry-After: 1`, `nextPollBackoff` returns 1s every time — `attempt` is ignored once `retryAfter > 0` — and this branch then extends the deadline by that same 1s, over and over, until `throttledWait >= timeout`.

At the default 30-minute timeout that is ~1800 polls to spend the budget, then ~1800 more inside the extended window: **≈3600 requests**. Before this PR the same proxy produced ≈1800 on the fixed 1s tick with no extension.

So "the cap in `nextPollBackoff` bounds it either way" is true of the *gap* but not of the *count*, and in that case the PR regresses the metric it exists to improve.

Against DRF's real `Retry-After` this is a large win and the decision is right — the argument against `max()` on the common path holds. The cheaper guard is a different one: **do not extend the deadline for a wait the loop did not choose.** Extending only for schedule-driven backoff, or bounding the number of extensions rather than their total duration, keeps the DRF behaviour intact while closing the pathological case.

Raising it because the non-JSON branch now deliberately brings gateway throttles into range, which is exactly where a flat `Retry-After: 1` is plausible.

api/event/event.go
```diff
@@ -329,17 +449,30 @@ func StreamApprovedCommand(ac *client.AlpaconClient, cmdID string, out io.Writer
...
+func streamSubscribed(ac *client.AlpaconClient, session *EventSessionResponse, listener *CommandOutputListener, cmdID, serverID string, out io.Writer, timeout, tick time.Duration, waitApproval bool) error {
 	if err := SubscribeEvent(ac, session.ChannelID, EventTypeCommandOutput, cmdID); err != nil {
 		listener.Stop()
 		return runCommandFallbackFromID(ac, cmdID, out, waitApproval, err)
 	}
+	// Chunks carry no end marker, so without this the exit waits for the poll below
+	// to notice—up to 10 ticks once the command is a minute old. Best effort: on
+	// failure that poll stays the only terminal signal, as before.
+	if serverID != "" {
+		_ = SubscribeEvent(ac, session.ChannelID, EventTypeCommandFin, serverID)
```
junho226, line 486: Warning: please file the server-side gap as an issue rather than leaving it in the PR body.

I checked it against alpacon-server and your reading is correct. `EventSubscriptionSerializer.validate` (`events/api/serializers.py:695`) has branches for `SUDO`, `WORK_SESSION`, `COMMAND_OUTPUT` and `NOTIFICATION`, and **none for `COMMAND_FIN`**. So both consequences hold:

- any authenticated user may subscribe with a server id they have no access to;
- an omitted `target_id` wildcard-matches every server, since `get_subscribed_channels` only narrows when one is given.

The event type predates this PR — it is in migration `0002` — so the CLI is not opening the hole. But this line makes it the first live consumer: after this ships, every `alpacon exec` subscribes to it. A gap that only ever existed on paper starts carrying traffic.

alpacax/alpacon-server#2885 is filed for the chunk-seq race; this one is not filed anywhere I can find. The `notification` branch's force-scope of a missing target is the shape to copy.

(Also verified the positive half while I was there: `command_fin` **is** a valid subscription type and `notify_command_event` publishes it with `target_id=server_id` and `payload.id` = the command pk, so this subscription and the listener's filter are both correct — the ignored error is not hiding an inert feature.)

api/event/event.go
```diff
@@ -239,48 +286,121 @@ func execTimeout() time.Duration {
...
+// terminal (PendingApprovalError). A closed seams.cancel ends the poll with
+// errPollCancelled.
+func pollCommandExecution(ac *client.AlpaconClient, cmdId string, timeout, tick time.Duration, waitApproval bool, seams pollSeams) (EventDetails, error) {
 	var response EventDetails
 
-	timer := time.NewTimer(timeout)
-	defer timer.Stop()
-	ticker := time.NewTicker(tick)
-	defer ticker.Stop()
+	started := seams.Now()
```
junho226, line 343: Suggestion: `started` is the poll's start, not the command's.

`nextPollTick` is described as widening "with the command's age", but `started := seams.Now()` is set when this function is entered. On `StreamApprovedCommand` and `runCommandFallbackFromID` the poll begins long after submission — an approval hold can last hours — so a command that is objectively old still gets the base tick for its first 10 ticks.

That is the forgiving direction and I would not change the behaviour. The comment and the PR table just claim more than the code measures; "widens with the poll's age" would be accurate.

api/event/event.go
```diff
@@ -361,32 +494,52 @@ func streamSubscribed(ac *client.AlpaconClient, session *EventSessionResponse, l
...
-			// to the buffered Result so output is never silently dropped. On a normal
-			// streamed run lastSeq has advanced and this is skipped.
-			if lastSeq < 0 && details.Result != "" {
-				_, _ = fmt.Fprint(out, details.Result)
+		case <-listener.Finished():
+			// fin says only that it is over; the exit code lives on the command. A
+			// failed read, or one racing a state not yet written, leaves the poll as
+			// the net rather than reporting a half-finished run.
+			details, err := GetCommandByID(ac, cmdID)
+			if err != nil || isPollWaitStatus(details.Status, waitApproval) {
```
junho226, line 549: Suggestion: this failure leaves no trace.

The fin has already been consumed at this point — `finished` is buffered 1 and no second fin arrives — so a failed `GetCommandByID` hands the run back to the poll, which may be up to 10 ticks out. The behaviour is intentional and well argued, but from the outside it is indistinguishable from the fin never having fired.

A debug-level line here would make "why did this run sit for 10 more seconds after its output landed" answerable from the output rather than from the source.

junho226 (commented):
## Review summary

Reviewed the full diff (+927 -98, 8 files) and the reasoning in the PR body.

**Verification I ran myself**

- `gofmt -l .` — no output
- `go vet ./...` — clean
- `go test -race ./api/event/... ./client/... ./utils/...` — all ok (`api/event` 59s)
- **Wire contract checked end-to-end against alpacon-server**: `command_fin` is a valid `EventSubscription.EventType` (present since migration `0002`), `Command.notify_command_event` publishes it with `target_id=server_id` and a payload of `{type, target, event, id}` where `id` is the command pk. So the listener's per-server subscription and its `Payload.ID` filter are both correct, and the feature is not silently inert behind the ignored `SubscribeEvent` error.

No blocking defects. Two things worth a decision before merge.

---

## Warnings

**1. The deadline extension doubles the request count in the exact scenario the "Left as they are" table dismisses** — `api/event/event.go:376`

The Retry-After-wins decision is defended with "the cap in `nextPollBackoff` bounds it either way". The cap bounds the *gap*, but nothing bounds the *count*, and the new extension makes that case worse than it was before this PR:

- A flat `Retry-After: 1` makes `nextPollBackoff` return 1s every time — `attempt` is ignored once `retryAfter > 0` — and the 429 branch then extends the deadline by that same 1s until `throttledWait >= timeout`.
- At the default 30-minute timeout that is roughly 1800 polls to spend the extension budget, then roughly 1800 more inside the extended window: **≈3600 requests**.
- Before this PR the same proxy produced ≈1800 (fixed 1s tick, no extension).

Against DRF's real `Retry-After` this is a large win and the decision is right — the argument that `max()` would wait longer than asked on the common path holds. But the fallback position ("worth switching to `max` if it turns up in practice") is weaker than stated, because in that scenario the PR regresses the very metric it exists to improve, 2x.

The cheaper guard is not `max()`. It is to **not extend the deadline for a wait the loop did not choose** — extend only when the backoff came from the exponential schedule, or bound the number of extensions rather than their total duration. Worth one line, given the non-JSON branch now deliberately brings gateway throttles into range.

**2. The `command_fin` subscription gap belongs in a filed issue, not a PR body** — `api/event/event.go:474`

I confirmed this on the server side. `EventSubscriptionSerializer.validate` (`events/api/serializers.py:695`) branches on `SUDO`, `WORK_SESSION`, `COMMAND_OUTPUT` and `NOTIFICATION` — there is **no branch for `COMMAND_FIN`**. Both consequences you name hold: any authenticated user may subscribe with a server id they cannot see, and an omitted `target_id` wildcard-matches every server through `get_subscribed_channels`.

The event type predates this PR, so the CLI is not creating the hole. But it is the first consumer, and once `alpacon exec` subscribes on every run the channel goes from theoretical to live traffic. alpacax/alpacon-server#2885 is filed for the chunk-seq race; this one is not. Please file it against alpacon-server and link it here — the `notification` branch's force-scope is the model to copy.

## Suggestions

- **`nextPollTick` measures the poll's age, not the command's** (`event.go:334`). `started` is set at poll entry, so on `StreamApprovedCommand` and `runCommandFallbackFromID` — where the poll begins after an approval hold that can last hours — an objectively old command still gets the fast window for its first 10 ticks. That is the forgiving direction and probably fine, but the comment claims more than the code measures.
- **The fin-triggered `GetCommandByID` failure is silent** (`event.go:537`). The fin is consumed (buffered 1) and `continue` hands the run back to the poll, which may be 10 ticks out. Intentional, but a debug-level line would make "why did this take 10 more seconds" answerable from the output rather than from the source.

## Highlights

- `pollSeams` passed rather than kept in package vars, with the reason recorded — a poll outlives the stream that started it by an in-flight request — and `-shuffle` named as what caught it. The right lesson drawn from the right evidence.
- The deadline check placed *after* the wait, so `[client_timeout]` lands at the deadline it names instead of one gap past it.
- `withRetryAfter` bounding at `math.MaxInt64/int64(time.Second)` rather than trusting the header, and parsing only delta-seconds with the reason stated.
- The last extension granted whole rather than trimmed, with the failure mode spelled out: a trimmed wait expires mid-sleep and the loop serves a wait it never gets to use.
- The warning flag cleared together with the allowance it belongs to, so a second throttled stretch minutes later is not silent either.
- A "Known limits" section that names a race this PR **widens**, explains why the CLI cannot tell the two cases apart, and files it (alpacax/alpacon-server#2885) instead of leaving it implicit.
- Tests decided on order and fake clocks, with the argument for why "finished within 500ms" would only ever measure the distance to the tick it was racing.

---

**Verdict**: COMMENTED

Warning 1 deserves a decision before merge but does not block. Warning 2 is a tracking gap, not a code defect. Everything I could verify locally passes.

api/event/event.go
jisung-02: Fixed in 751fea9. The arithmetic held, and so did the reason it matters: a duration budget converts to whatever request count the server's chosen wait dictates, and the wait is the server's to choose.

I could not take the first form of the guard, though. DRF sets `Retry-After` on every 429 it answers—`Throttled` carries `wait()`—so "extend only for schedule-driven backoff" would remove the extension from the only path that has one, which is the path this branch exists for. So the second form: bound the number of grants rather than their total duration, ANDed with the duration budget that was already there.

`pollMaxThrottleExtensions = 60`. Counting grants bounds the number you were measuring: at a steady wait, n grants are n extra requests whatever that wait's length. At a flat `Retry-After: 1` and the default 30-minute timeout that is ~1,800 + 60 requests, against the ~3,600 you calculated and the ~1,800 the fixed tick spent before this branch. 60 is picked so it does not reach into the DRF path: at the capped 60-tick wait, 60 grants is 3,600 ticks—a whole quota window—so the duration budget still binds first, at 30. Progress hands the count back with the deadline, for the same reason the duration allowance goes back with it.

Two cases pin it, both checked by breaking the production line and reading the number that moved. Dropping the count from the condition grows the window a permanently throttled poll gets from 1.08s to 1.44s, exactly the second budget the duration alone would have granted; dropping the count's reset stops the progress case at 720.1ms.

One residual, now in Known limits: the count bounds the extra requests only while the mandated wait keeps its length. A server that grants long waits until the duration budget is spent and then shortens them can still burn the extended window at the short pace. The extra wall-clock is bounded by one timeout either way.

api/event/event.go
jisung-02: Filed. It went onto alpacax/alpacon-server#2885 as a note rather than a separate issue, since that one already deals with the fin payload, and `command_reject` is included with it: it publishes through the same `notify_command_event` call and has no `validate` branch either. The PR body's Follow-up section links it now instead of reading as an open question.

Thank you for confirming the positive half while you were in there—the ignored `SubscribeEvent` error is exactly where an inert feature could have shipped green.

api/event/event.go
jisung-02: Agreed on both halves: the behaviour is the forgiving direction and stays, the wording was wider than the code. The PR body row now says the gap widens with the poll's age and names the approval-resume case, and the test is `TestPollCommandExecution_PacingWidensWithPollAge` as of ad0e56b—its comment already measured the right thing, only the name did not.

api/event/event.go
jisung-02: Left as it is, and written down in the PR body's "Left as they are" table now rather than only in the source.

There is no level to put the line at: `utils` prints errors, info, and warnings, all to stderr, and a warning is the wrong volume for a path that recovers by design—the poll picks the run up. Introducing a debug level is a wider decision than this call site should make. If the question comes up in practice the answer is the level, not a print here.

api/event/event.go
```diff
@@ -18,6 +19,29 @@ import (
 
 const (
 	getEventURL = "/api/events/commands/"
+
+	// Poll pacing, in multiples of the caller's base tick, widening as the command
+	// ages and again once the server pushes back. A fixed 1s tick burns alpacon-server's
```
junho226: Leftover from the old framing. This says the gap widens "as the command ages", but `nextPollTick` is fed `seams.Now().Sub(started)` and `started` is when the poll began, not when the command was submitted — which is why `ad0e56b` renamed the test. On the `StreamApprovedCommand` and `runCommandFallbackFromID` paths the two differ by however long the command sat before the poll started. `widening as the poll ages` closes the last spot.

api/event/event.go
```diff
@@ -18,6 +19,29 @@ import (
...
+	pollMediumTick     = 5
+	pollSlowTick       = 10
+	pollBackoffFactor  = 2
+	pollMaxBackoffTick = 60
+
+	// Grants of extra deadline for time lost to 429s, capped by count as well as
+	// duration: a duration budget is spent at whatever pace the server asks for, so a
+	// flat Retry-After: 1 buys a second window of one-second polls. 60 grants at the
+	// capped 60-tick wait is 3,600 ticks—a whole quota window at the default
+	// one-second tick—so the duration budget binds first at the cap and this bound
```
junho226: The 3,600-tick arithmetic is right, but "the duration budget binds first at the cap" holds only while `timeout` stays under it. True at the default 30m; at an `ALPACON_EXEC_TIMEOUT` over an hour the count is what binds at the cap, and this bound stops being the one that "only catches a throttle asking for less". The gate takes whichever binds first, so behaviour is fine either way and only the claim is affected — qualifying it with "at the default timeout" alongside the existing "at the default one-second tick" would cover it.

junho226 (dismissed):
Re-review of the delta `4dbbaf5..ad0e56b` (4 commits) plus the four findings from my earlier comment. I did not re-read the rest of the PR.

## Prior findings

**1. The deadline extension doubles the request count under a flat `Retry-After: 1` — addressed.**

`751fea9` adds `pollMaxThrottleExtensions = 60` to the extension gate. That is the bound-the-count fix rather than a bound-the-duration one, so the extra requests no longer scale with how short the mandated wait is. Re-derived at the default 30m/1s: `nextPollBackoff` still returns 1s on every 429, but the deadline now grows by 60 x 1s instead of 1800 x 1s, so the run costs roughly 1,860 polls rather than roughly 3,600. The reset sits beside the `throttledWait` reset in the running branch, where the deadline is refreshed anyway, and is correctly absent from the approval-hold branch, which never refreshes it. The gate grants exactly 60, no off-by-one.

I derived both new tests' expected sums from the loop independently and they match: count-bound case `720ms + 60*6ms = 1.08s` where the duration bound alone would give `720ms + 120*6ms = 1.44s`; reset case `100us + 3*360ms` where dropping the reset gives `100us + 2*360ms = 720.1ms`. Those are the same numbers the PR body reports from breaking each production line, so the tests fail for the reason they claim rather than incidentally.

The residual — a server that answers with long waits until the duration budget is spent and then shortens them can still burn a full timeout of extended deadline at the fast pace — is real, and "Known limits" already states it. Wall clock is bounded by one timeout either way; agreed it is not worth more machinery.

**2. `command_fin` subscription scoping should be an alpacon-server issue — addressed, and it already was when I raised it.**

Verified: the gap is written up as an addendum on `alpacax/alpacon-server#2885`, covering `COMMAND_REJECT` as well since it shares the `notify_command_event` publish path, and the PR body's Follow-up section links to it. The issue's last edit predates my review, so this was in place before I asked for it — my note was wrong that it was left in the PR body only.

**3. `nextPollTick` measures the poll's age, not the command's — partially addressed.**

`ad0e56b` renames the test to `TestPollCommandExecution_PacingWidensWithPollAge` and its comment now says "how long the poll has been running". The const-block comment at the top of the file is the one place the old claim survives; inline below.

**4. Silent `GetCommandByID` failure on the fin branch — not done, and the counter-argument is correct.**

I checked `utils/log.go`: it exposes `CliError`, `CliInfo`, `CliWarning`, `CliSuccess` and nothing below warning. There is no debug level to put the line at, a warning is the wrong level for a path that recovers by design, and adding a level is wider than this PR. "Left as they are" is the right resolution — withdrawing the suggestion.

## New findings

None that affect behaviour. Two comment-precision nits inline. The `cmdId` -> `cmdID` rename in `efc585f` is mechanical and matches the spelling the rest of the file already uses.

## Verification

Run in the PR worktree at `ad0e56b`:

```
$ gofmt -l .
(no output)

$ go vet ./...
(no output, exit 0)

$ go test -race ./api/event/... ./client/... ./utils/...
ok  	github.com/alpacax/alpacon-cli/api/event	59.375s
ok  	github.com/alpacax/alpacon-cli/client	(cached)
ok  	github.com/alpacax/alpacon-cli/utils	(cached)
```

I did not re-run `golangci-lint` or the full `-race -shuffle=on` sweep the PR body reports.

api/event/event.go
jisung-02: Fixed in 008674a: `widening as the poll ages`. You are right that this was the last spot—`started` is stamped where the poll begins, so on `StreamApprovedCommand` and `runCommandFallbackFromID` the command is already older than the pacing thinks. Nothing about the behaviour was ever in question; only this line still described the old framing.

api/event/event.go
jisung-02: Fixed in 008674a, and the crossover point is where you put it. The duration budget at the capped wait is `timeout / 60s` grants, so it stays under the count for any timeout below an hour and the count takes over above it. The comment now reads "so at the default timeout the duration budget binds first at the cap ... Past an hour's timeout the count binds at the cap too."

The same unqualified claim was in the PR body's Behaviour change section—"two timeouts is what the duration budget buys"—which is only true while the duration is the binding side, so that is qualified now as well.

junho226 (approved):
## Re-review: `ad0e56b..008674a`

One commit, comments only. It closes both nits from my last pass exactly.

- The const-block header now reads "widening as the poll ages" rather than "as the command ages", which is what `nextPollTick` actually measures — it is fed `seams.Now().Sub(started)` and `started` is set at poll entry.
- The extension-bound rationale is now qualified: "at the default timeout the duration budget binds first at the cap ... Past an hour's timeout the count binds at the cap too." That is the condition the earlier claim was missing.

### Verification

Run in the worktree at `008674a`:

```
$ gofmt -l .
(no output)

$ go vet ./...
(no output, exit 0)

$ go build ./...
(exit 0)
```

I did not re-run `go test -race`: the diff is two comment blocks, with no token of Go code changed. The suite passed on `ad0e56b` in my previous pass.

Nothing further from me.

PR #315, opened 2026-08-07
feat(client): name the token ACL denial behind a 403

geunwoonoh:
## What

Map `api_token_acl_not_allowed` to a clear message when it arrives on a 403.

## Why

alpacon-server is correcting the ACL refusal status from **400 to 403** (alpacax/alpacon-server#2804). An ACL denial is an authorization failure — the request is syntactically complete, the token simply may not do that — and the platform status taxonomy puts that at 403, not 400.

The server body is unchanged: `{"code": "api_token_acl_not_allowed"}`, with no `detail`. That matters here, because a detail-less 401/403 takes `checkAuthStatus` (`client/client.go:128`) rather than the generic `parseAPIError`. `authStatusCodeMessage` knew only `auth_mfa_required`, so the refusal fell through to the catch-all at `client/client.go:162`:

| | today (400) | without this PR (403) |
|---|---|---|
| path | `parseAPIError` field-map branch | `checkAuthStatus` → `authStatusMessage` |
| stderr | `code: api_token_acl_not_allowed` | `permission denied: you do not have the required privileges for this action` |

Every hint that a **token rule** caused the refusal disappears — and this is exactly the audience that struggles with ACL configuration.

The server cannot fix it from its side: `utils/handlers.py:error_code_handler` overwrites the response body with `{"code": ...}` for `PermissionDenied`, and the repo contract forbids attaching a `detail`. The mapping belongs here.

## Nothing else changes

Control flow was already code-based, not status-based, and stays that way:

- `utils.HandleCommonErrors` switches on the code; `api_token_acl_not_allowed` hits `default: return err` under both statuses.
- No re-login, no token clearing, no refresh. `config.DeleteConfig` is reachable only from `alpacon logout`; `RefreshToken` fires only for `auth_mfa_required` and time-based expiry at client construction.
- Exit code stays `1`. The code is still exposed via `ErrorCode()` / `ParseErrorResponse`, so scripts branching on it are unaffected.
- The only status-keyed retry logic (`api/event/watcher.go`, `api/ftp/ftp.go`) treats 400 and 403 identically and is not on this path.

## Safe to merge first

**This is inert until the server flips.** At 400, `checkAuthStatus` returns early (`client/client.go:129-131`) and the new branch is never reached. Merging and releasing this ahead of the server change is a no-op that removes the message regression window entirely.

## Message wording

The message does not interpolate the raw code — that would break the no-raw-code contract documented at `client/client.go:157-159` and guarded by `TestSendRequest_403CodeWithoutDetailKeepsCodeSource`. It names token access control and points at `alpacon token acl`.

It is also worded to stay true across every surface the server raises this code from: real ServerACL/CommandACL/FileACL evaluations on `exec` and `cp`, and the "a programmatic token may not open an interactive channel" refusals on `websh` tunnels and FTP sessions.

## Testing

- `TestSendRequest_403ACLDeniedExplainsTokenAccessControl` — pins the message, the absence of the raw code, the absence of a re-login hint, and that `ParseErrorResponse` still returns the code.
- `go build ./... && go vet ./... && go test ./client/... ./utils/...` pass; `gofmt -l .` is clean.

## Docs

README gains a short paragraph beside the inline-credential refusal note: an ACL denial is **permanent**, not transient, and the fix is to widen the rules rather than retry.

Refs alpacax/alpacon-server#2804


README.md
```diff
@@ -294,6 +294,8 @@ Separately from the work session gate, `exec` (and `websh` when running a comman
 
 Exit code is `1`, and under `--output json` the refusal is an error envelope on stderr with `error_code` `command_inline_credential`. That envelope currently carries no `next_actions`—rewrite the command with `--env="KEY"` as described above (`exec` takes the remote command after `--`, `websh` takes it as one quoted argument).
 
+Also separate from the work session gate: when you authenticate with an API token or a service token, the server checks the request against that token's ACL rules before anything runs. A request outside those rules is refused with `error_code` `api_token_acl_not_allowed` and exit code `1`—nothing reaches the server. This is a permanent refusal, not a transient one: a retry submits the same request. List the rules with `alpacon token acl command ls TOKEN`, `alpacon token acl server ls TOKEN`, or `alpacon token acl file ls TOKEN`, and widen them with the matching `add` subcommand. Deny-by-default applies per ACL type—a token with no rule of a given type has no access of that kind at all.
```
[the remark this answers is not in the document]
geunwoonoh: Valid—and the sentence contradicted itself: its first clause uses "the server" for Alpacon ("the server checks the request against that token's ACL rules"), which is also how the rest of this README uses the term (`the command line the server stores`). So the tail claimed the request never reaches the very thing that returned the 403.

Rather than reword it to "nothing runs", I dropped the clause: the first sentence already says the check happens *before anything runs*, so the tail added no information beyond the misleading part. Fixed in the follow-up commit.

jisung-02 (commented):
## Summary

Maps `api_token_acl_not_allowed` to a message when it arrives on a detail-less 403, ahead of the server's 400→403 correction (alpacax/alpacon-server#2804). One constant in `utils/error.go`, one `switch` case in `authStatusCodeMessage`, one test, one README paragraph.

Verified locally: `go build ./... && go vet ./... && go test ./client/... ./utils/...` pass, `gofmt -l .` is empty.

## Warnings (1)

**README.md:297 — the new paragraph mentions `error_code`, but nothing on this path prints one.**

The sentence reads "refused with `error_code` `api_token_acl_not_allowed` and exit code `1`". Elsewhere in this README, `error_code` names a field of the `--output json` error envelope (`utils/output.go:30`).

An ACL denial does not go through that envelope. `exec` failures reach `utils.CliErrorWithExit("%s", err)` at `cmd/exec/run.go:382`; the envelope branch above it only fires for `isCommandInlineCredentialError` (`cmd/exec/run.go:375-380`). `cmd/ftp/` and the websh path have no `CliErrorEnvelopeWithExit` call at all. So `--output json` still prints a single line on stderr here.

The inline-credential paragraph right above scopes its claim ("under `--output json` the refusal is an error envelope on stderr with `error_code` ..."), and the envelope list at README:291 does not include ACL denials. The new paragraph is the only place that says otherwise, and someone could reasonably try to parse `error_code` out of `--output json` because of it.

Adjusting the docs looks like the right side to fix — emitting a real envelope is outside what this PR sets out to change. Inline suggestion below.

## Suggestions (1)

**client/client_test.go:303 — consider pinning the 400 behavior too.**

The case for merging this ahead of the server is that `checkAuthStatus` returns early on a 400 (`client/client.go:129-131`). A second case asserting that 400 + `{"code": "api_token_acl_not_allowed"}` still renders through `parseAPIError` as `code: api_token_acl_not_allowed` (`client/client.go:523`) would keep that property under test. Optional.

## Highlights

- The message does not interpolate the raw code, which keeps the contract documented at `client/client.go:157-159`, and the test asserts it (`client_test.go:297`).
- The absence of a re-login hint is asserted as well (`client_test.go:299`) — an ACL denial is not a stale session.
- The new case matches the shape of the `AuthMFARequired` case beside it.
- The comment at `utils/error.go:28-35` ("the server sends this on a 403 with no human detail, so `authStatusCodeMessage` renders the message") matches `client/client.go:136-175`.
- Checked the PR description's claim that 400 and 403 behave the same: `utils.IsFatalClientError` (`utils/error.go:126-131`) treats the whole 4xx range alike apart from 408/429, so it holds.

## Verdict

COMMENTED

- [ ] Documentation updated if needed — README:297 needs the `error_code` wording adjusted

README.md
```diff
@@ -294,6 +294,8 @@ Separately from the work session gate, `exec` (and `websh` when running a comman
 
 Exit code is `1`, and under `--output json` the refusal is an error envelope on stderr with `error_code` `command_inline_credential`. That envelope currently carries no `next_actions`—rewrite the command with `--env="KEY"` as described above (`exec` takes the remote command after `--`, `websh` takes it as one quoted argument).
 
+Also separate from the work session gate: when you authenticate with an API token or a service token, the server checks the request against that token's ACL rules before anything runs. A request outside those rules is refused with `error_code` `api_token_acl_not_allowed` and exit code `1`. This is a permanent refusal, not a transient one: a retry submits the same request. List the rules with `alpacon token acl command ls TOKEN`, `alpacon token acl server ls TOKEN`, or `alpacon token acl file ls TOKEN`, and widen them with the matching `add` subcommand. Deny-by-default applies per ACL type—a token with no rule of a given type has no access of that kind at all.
```
jisung-02: Warning: this sentence mentions `error_code`, but nothing on this path prints one.

In this README, `error_code` names a field of the `--output json` error envelope (`utils/output.go:30`). An ACL denial does not go through that envelope: `exec` failures reach `utils.CliErrorWithExit("%s", err)` at `cmd/exec/run.go:382`, and the envelope branch above it only fires for `isCommandInlineCredentialError` (`cmd/exec/run.go:375-380`). `cmd/ftp/` and the websh path have no `CliErrorEnvelopeWithExit` call at all. With `--output json` the user still gets a single line on stderr.

The inline-credential paragraph right above scopes its claim to `--output json`, and the envelope list at line 291 does not include ACL denials — both stay correct. Suggested wording drops the `error_code` reference and keeps everything else:

```suggestion
Also separate from the work session gate: when you authenticate with an API token or a service token, the server checks the request against that token's ACL rules before anything runs. A request outside those rules is refused with exit code `1` and a message naming token access control. This is a permanent refusal, not a transient one: a retry submits the same request. List the rules with `alpacon token acl command ls TOKEN`, `alpacon token acl server ls TOKEN`, or `alpacon token acl file ls TOKEN`, and widen them with the matching `add` subcommand. Deny-by-default applies per ACL type—a token with no rule of a given type has no access of that kind at all.
```

client/client_test.go
```diff
@@ -278,6 +278,30 @@ func TestSendRequest_403CodeWithoutDetailKeepsCodeSource(t *testing.T) {
...
+	assert.ErrorContains(t, err, "token access control")
+	assert.ErrorContains(t, err, "alpacon token acl")
+	// The no-raw-code contract: callers read the code via ErrorCode(), not the message.
+	assert.NotContains(t, err.Error(), "api_token_acl_not_allowed")
+	// A token missing a rule is not a stale session—never suggest re-login.
+	assert.NotContains(t, err.Error(), "alpacon login")
+
+	code, _ := utils.ParseErrorResponse(err)
+	assert.Equal(t, utils.APITokenACLNotAllowed, code)
+}
```
jisung-02, line 303: Suggestion: consider a companion case pinning the 400 behavior.

The case for merging this ahead of the server change is that `checkAuthStatus` returns early on a 400 (`client/client.go:129-131`). A second test asserting that 400 + `{"code": "api_token_acl_not_allowed"}` still renders through `parseAPIError` as `code: api_token_acl_not_allowed` (`client/client.go:523`) would keep that property under test if the status condition is ever widened. Optional.

README.md
geunwoonoh: Fixed in 6a9305e — took the suggested wording as-is.

Confirmed the claim before changing it: `CliErrorEnvelopeWithExit` is called only from `cmd/worksession/`, `cmd/event/`, `cmd/username/`, and the one inline-credential branch at `cmd/exec/run.go:376`. Every ACL-gated surface is outside that set — `exec` falls through to `CliErrorWithExit` at `run.go:383`, and `cmd/ftp/` and websh have no envelope call at all. So the field the paragraph named could never appear on this path.

Worth noting the code is not unreachable everywhere: `buildCliErrorEnvelopeFromErr` pulls it off the error via `ParseErrorResponse`, so a token-ACL failure on a work-session or event subcommand would print `error_code`. That does not rescue the sentence, though — it was unscoped, and the surfaces it describes (command, server, file ACL) are exactly the ones with no envelope.

client/client_test.go
geunwoonoh: Added in aa57ebd — `TestSendRequest_400ACLDeniedKeepsCodeWithoutAuthStatusMessage`.

Took it as more than optional: the 400 is the live path until alpacax/alpacon-server#2804 ships, so it is the shape users hit today, untested until now.

Two assertions rather than one. `ParseErrorResponse` still yields `api_token_acl_not_allowed`, as suggested — and the 403 mapping's message must *not* appear on the 400. Without that second one the test passes whether or not `checkAuthStatus` still returns early, which is the property you were pointing at.

Worth recording what the 400 renders today: no `detail`, so `parseAPIErrorPayload` reaches its field loop and the message becomes `code: api_token_acl_not_allowed` — the raw code, which is exactly what the 403 mapping exists to replace. Left as-is; changing it belongs with the server correction, not here.

jisung-02 (approved):
## Summary

Maps `api_token_acl_not_allowed` to a readable message when it arrives on a detail-less 403. One constant, one `authStatusCodeMessage` case, two tests, one README paragraph.

Verified locally: `go build ./... && go vet ./... && go test ./client/... ./utils/...` pass, `gofmt -l .` is empty.

## Warnings (0)

Both findings from the previous round are resolved in 6a9305e and aa57ebd.

## Suggestions (0)

None.

## Highlights

- `TestSendRequest_400ACLDeniedKeepsCodeWithoutAuthStatusMessage` (aa57ebd) goes past what was suggested. `NotContains(err.Error(), "token access control")` is not a vacuous assertion: if `checkAuthStatus`'s condition were ever widened to 400, the body carries no `detail`, so `hasDetail` stays false, `authStatusCodeMessage` runs, and that substring does appear — the test catches exactly the regression its comment names.
- That 400 path is the one users actually hit until alpacax/alpacon-server#2804 ships, and it had no test until now.
- The message does not interpolate the raw code, keeping the contract documented at `client/client.go:157-159`, and the test asserts it.
- The absence of a re-login hint is asserted too — an ACL denial is not a stale session.
- The new case matches the shape of the `AuthMFARequired` case beside it.
- Both new tests use `newTestClient`, `httptest.NewServer`, and `utils.ParseErrorResponse` the same way as their neighbours, and are placed by feature grouping rather than status-code order, matching this file's existing layout.
- The `alpacon token acl {command,server,file} ls TOKEN` syntax in the README matches the real Cobra `Use:` strings in `cmd/token/`.
- Checked the description's claim that 400 and 403 behave identically for retries: `utils.IsFatalClientError` (`utils/error.go:126-131`) treats the whole 4xx range alike apart from 408/429, so it holds.

## Verdict

APPROVED

- [x] Documentation updated if needed

Issue #302, opened 2026-08-03
fix(utils): usage errors exit 1 while the README documents 2


jisung-02:
## Background

The README "Exit codes" table documents `2` as the usage-error code, and `CLAUDE.md` and `.github/copilot-instructions.md` repeat it. All three describe the exit codes as a stable public contract that scripts, CI, and AI agents branch on.

Local validation failures across the CLI currently exit `1`, so this part of the documented contract does not seem to be reachable in practice today. It may be worth deciding which of the two the project wants to keep.

## What the code does today

- `utils/error.go` defines `ExitCodeWorkSessionDenied` (3), `ExitCodePendingApproval` (4), `ExitCodeServerBusy` (5), and `ExitCodeNotApproved` (6). There is no constant for a usage error.
- `utils.CliUsageErrorEnvelopeWithExit`, the helper used for local validation failures, exits `1`. Its own doc comment states this: `// ... error_code is fixed to UsageErrorCode. Exits(1).` It has roughly 35 production call sites.
- `os.Exit(2)` does not appear anywhere in production code—the only occurrences are in test-harness subprocesses.
- The README row appears to date from `fa29159 docs: document alpacon exit code convention`, which introduced the convention without a corresponding constant.

Under `--output json` the situation is a little different: those call sites do emit a structured `{"error_code": "usage_error", "exit_code": 1, ...}` envelope, so a JSON consumer can already tell a usage error apart. Only the exit code alone is ambiguous.

## Why it might matter

A consumer that reads the exit code alone cannot currently distinguish "you passed a bad flag" from "the network failed" or "the server returned an error". Both are `1`. For the lightweight agents the exit-code contract was designed for, that difference is the difference between "fix the invocation and retry" and "retry unchanged", which suggests the distinction is worth having in some form.

## Possible directions

Either seems defensible; the choice is mostly about how much churn is acceptable.

1. Introduce `ExitCodeUsageError = 2` and point `CliUsageErrorEnvelopeWithExit` at it. This would honor what the docs already promise. It is a breaking change for anyone currently treating `1` as "any failure", since ~35 call sites would move at once, so it would likely want its own PR and a release-notes callout. Since nothing returns `2` today, no consumer can be relying on the current absence.
2. Relax the documentation instead—drop or reword the `2` row so the docs match the implementation. Lower risk, though it gives up a distinction the contract had already advertised.

[...]

PR #316, opened 2026-08-10
docs(utils): record why the CLI omits the credential override

### LINKED ISSUE #305: Correct the --env guidance and document the inline-credential gate
## Background

This issue originally proposed adding `--env` guidance for the ADR 0037 inline-credential gate (alpacax/alpacon-server#2745). That proposal was already merged as #304 (`2d58f11`, landed 30 minutes before this issue was opened): `utils.CommandInlineCredential` at `utils/error.go:22`, detection in `cmd/exec/run.go` printing `Error:` plus `Hint:` in table mode and an envelope carrying `error_code` under `--output json`, bypassing `utils.CliErrorWithExit` so the misleading issue-tracker line (`utils/log.go:13`) is not appended, and exit code 1 kept as the original open question concluded.

The issue was then rescoped to what surfaced while checking that work against the server.

## What the gate actually inspects

Verified by reading `alpacon-server` (`origin/main`, after #2745), since everything below depends on it.

- `check_inline_credential` in `execution_gate/stages.py` reads only `line = inp.command_line`. `credential_context` in `events/api/serializers.py` passes only `attrs.get('line')`. The `env` field is never inspected.
- The CLI sends `--env` values in a separate field, not on the command line (`api/event/types.go:68-79`, `api/event/event.go:191-200`).
- `env` is not returned in API responses. It is absent from `CommandSerializer.Meta.fields`, and the help text for `expose_command_env_in_audit` (`workspaces/models.py:790-799`) states "Env is never returned in API responses regardless of this setting". That setting defaults to false, and while it is off `mask_command_output` (`events/models.py:138-149`) masks env values appearing in command output.

So the gate rejects credentials on the remote command line. `--env="KEY=VALUE"` passes it and is not an audit-log leak; the only exposure left is local.

## Resolved in #307

1. The `--env` guidance disagreed with the server. The help discouraged `--env="KEY=VALUE"` as "recorded verbatim in the audit log" and recommended `--env="KEY"` for staying "out of the audit log", but the audit log does not distinguish the two forms. The real difference is that a literal value stays on the local alpacon command line, reaching shell history, `ps` output, and CI job logs. Fixed in `cmd/exec/exec.go`, `cmd/websh/websh.go`, and the hint in `cmd/exec/run.go`. The README's only `--env` example used the discouraged form and carried a quoting bug, where `"echo $KEY"` was expanded by the local shell before alpacon saw it.

2. The gate was undocumented. Nothing in the README or either help text mentioned it, so the first anyone heard of the rule was a failed command — wrong order for CI pipelines and AI agents, which compose invocations from `--help` and the README. Now described in both help texts, in the README body, and in the exit code table's row for `1`.

3. The hint always named `alpacon exec`. websh command mode delegates to the exec runner (`cmd/websh/websh.go`), so websh users were handed an exec example. A name swap was not enough: the `--` separator exists only in `cmd/exec/parse.go:58`, while websh takes the remote command as a single quoted argument. `reRunHint` still names exec on purpose, since websh has no `--wait` and a rerun after approval has to go through exec (`README.md` exit code table, row 4).

## Moved to #306

The error-code mapping. `alpacon-server` defines 262 error codes and `utils/error.go` knows 11; when one of the rest arrives without a `detail`, `parseAPIErrorPayload`'s field-listing fallback turns the identifier itself into the user-facing message and `utils.CliErrorWithExit` appends the issue-tracker line. #304 handled one code by hand, so the pattern repeats for every code the server adds. Structural rather than a wording fix, so it is tracked separately.

## Still open: `credential_exposure_acknowledged`

The server accepts a per-command override, and its behavior is narrower than the name suggests. `validate` in `events/api/serializers.py` recomputes the flag rather than storing what the client sent, so an acknowledgement counts only when the line actually carries a credential. `execution_gate/stages.py` describes it as an "audited per-command action, never a per-environment auto-pass" and as "hygiene, not authorization" — no role gate, with the audit trail as the control. Even when acknowledged, `mask_command_line` in `events/models.py` redacts the value at write time (ADR 0042).

The CLI does not expose it. #304 left it out deliberately. Not exposing it makes `--env` the only escape path from the CLI, which matches the intent of the gate. Needs a decision.

## Verification scope

The findings and the gate behavior come from reading the `alpacon-cli` and `alpacon-server` (`origin/main`) sources. The #304 behavior was confirmed by `go build ./...`, `go test -race ./cmd/exec/ ./utils/ ./client/`, and by `cmd/exec/credential_hint_test.go`, which drives the real `alpacon exec` binary as a child process against an httptest server returning 400 `{"code": "command_inline_credential"}`. It was later also confirmed against a live server.

Refs alpacax/alpacon-server#2745, #304, #306, #307

jisung-02:
## Background

alpacon-server accepts a per-command `credential_exposure_acknowledged`. Sending it passes the inline-credential gate (`execution_gate/stages.py:274-276`). The CLI has the field nowhere, and no note saying why, so the omission read as an oversight. #305 asked whether to expose it; this PR records the answer next to the constant.

The answer is not to expose it, for three reasons.

First, the override records the exposure rather than preventing it. The server frames the gate as hygiene, not authorization, and deliberately gave the override no role gate—the control is the audit trail, not a permission. alpacon-server commit `fd71f7e8f` states this explicitly: "Deliberately still no role gate on the override: ADR 0037 makes this gate hygiene, not authorization, so the control is the audit trail." The same commit put the override on the detail and list serializers and made it filterable, precisely so auditors can enumerate overrides. Its current role is an audited exception at the API level, not a routine path a CLI flag should pave.

Second, the stored rendering is masked either way (`mask_command_line`, ADR 0042, keyed on the same predicate as the gate). Acknowledging buys nothing but the command running—the secret still lands in `Command.line` verbatim, which is exactly what the gate exists to prevent.

Third, the override's original justification is gone. It entered the server as a false-positive escape hatch for the monitor-to-enforce rollout (alpacax/alpacon-server#2745, `COMMAND_CREDENTIAL_MODE`). That flag no longer exists: commit `fd71f7e8f` removed it after an adversarial review executed the patterns, produced the full false-positive and false-negative set, and both were closed (upstream alpacax/alpacon-ai-server#293, then the re-vendoring commit). The known false positives were fixed in the detector itself—`_is_indirect` in `utils/credentials.py` exempts shell variable references and command substitutions (`--password="$DB_PASSWORD"`), pinned by `RejectSemanticsTest`, and the key-boundary and prompt-mode `--password` cases moved upstream (ai-server#300, #302). The gaps still documented in that module are all false negatives (bare high-entropy secrets, uncommon key names, secrets in prose). A false positive today is a detector bug to report and fix, not an expected condition that needs a bypass flag.

Given all three, exposing the override would add a one-flag way past the rewrite the gate exists to force, while rewriting with `--env="KEY"` costs the caller the same keystroke count. There is no reason to build the bypass.

This is a judgment call recorded, not a door closed: if you read the trade-off differently—for example, a workflow where `--env` genuinely cannot carry the secret—please raise it here. The comment (and the decision) can be revised in this PR.

## Changes

Six comment lines on `CommandInlineCredential` in `utils/error.go`, recording the decision and its rationale. No executable line changed.

## Testing

`gofmt -l .` clean, `go build ./...` succeeds, `go vet ./utils/` clean. Comments only, so there is no behavior to verify.

Closes #305

### DIFF
--- a/utils/error.go
+++ b/utils/error.go
@@ -23,6 +23,12 @@ const (
 	// The rejected-forms list above is repeated in the exec and websh help text
 	// and the README's "When a command is denied" section—when the server-side
 	// gate changes, update all of them together.
+	//
+	// The server also accepts a per-command credential_exposure_acknowledged
+	// override, which the CLI deliberately does not expose: it records the
+	// exposure rather than preventing it (hygiene, not authorization—it carries
+	// no role gate), so exposing it would offer a one-flag way past the gate
+	// where rewriting with --env costs the same. --env stays the only path.
 	CommandInlineCredential = "command_inline_credential"
 
 	// WorkSession gate codes (returned by alpacon-server work_sessions/services.py)


junho226 (approved):
Checked the recorded rationale against both sides rather than taking it on trust, and it holds.

Server side: `credential_exposure_acknowledged` is a real per-command field (`events/models.py:407`), accepted on the write serializers (`events/api/serializers.py:228,331`) and filterable (`events/api/filters.py:129`), and `check_inline_credential` passes the stage on it with `reason='exposure_acknowledged'` (`execution_gate/stages.py:274-276`) — so "records the exposure rather than preventing it" is exact. "No role gate" holds on inspection, not only on the server docstring that asserts it: the stage reads the bare boolean, and the serializer's entire handling of the field is a `bool()` coercion plus a re-run of `contains_inline_credential` (`serializers.py:423-426`), with no permission check anywhere on that path. CLI side: the override appears nowhere in this repo except the new comment, and `--env` is real on both surfaces (`cmd/exec/parse.go:111`, `cmd/websh/websh.go:54`) with no competing bypass flag in either parser, so "rewriting with --env costs the same" is not aspirational. Every identifier the comment names exists as spelled, so there is nothing here that rots silently.

Placement fits: it extends the `CommandInlineCredential` doc comment that already serves as this repo's anchor for the gate, so "why is there no flag?" is answered where it gets asked, and the wording matches the block above it (sentence case, unspaced em dash).

Verified locally at `ce665d84`: `go build ./...`, `go vet ./...` and `gofmt -l .` clean, `go test -race ./...` passing; CI green on all six checks. Two notes on what I could not fully pin down, neither of them about this diff: `golangci-lint` is not installed here, so for that check I relied on the PR's own green run; and one cold full-suite run hung `api/log` into the 10m timeout, which did not reproduce for that package alone (4.7s) or on a cold full run of `main`, and cannot come from a comment-only change.

PR #317, opened 2026-08-10
fix(utils): exit 2 on usage errors

### LINKED ISSUE #302: fix(utils): usage errors exit 1 while the README documents 2

jisung-02:
Closes #302.

## Problem

The README documented `2` as the usage-error exit code, but `utils.CliUsageErrorEnvelopeWithExit` exited 1—and its JSON envelope carried `exit_code: 1` to match, since `buildCliErrorEnvelope` hardcodes it. The documented contract and the shipped behavior disagreed, so a script reading the exit code alone could not tell "the request failed" from "the command line was wrong".

## Change

- `utils.ExitCodeUsageError = 2`, documented alongside the other exit-code constants in `utils/error.go`.
- `buildCliUsageErrorEnvelope` overrides the 1 that `buildCliErrorEnvelope` hardcodes, and `CliUsageErrorEnvelopeWithExit` exits 2 in both output modes.
- `utils.RequirePositiveInt` exits 2 as well—its docstring already called this a usage error while it exited 1.
- README, `CLAUDE.md`, `.github/copilot-instructions.md`, and the `event wait` help text describe code 2, naming the paths that actually emit it rather than promising it for every command's own validation.
- The README also scopes the JSON envelope claim: only the `work-session` and `event` paths emit an envelope carrying `error_code` `usage_error`; `RequirePositiveInt` reaches `CliErrorWithExitCode` directly and prints plain text to stderr.
- `TestExtendTableUsageError_ExitsTwo` covers the table branch of `CliUsageErrorEnvelopeWithExit`—the default output mode, the branch whose exit code actually changed, and the one exit-2 path that had no test.
- The work-session error test's subprocess helper fails loudly on a misconfigured harness instead of silently running in table mode, and the file is renamed `worksession_error_test.go` now that nothing in it is JSON-only.

Errors Cobra itself rejects while parsing (an unknown flag or subcommand) keep exiting 1—Cobra owns that path, and changing it would mean intercepting Cobra's own error handling.

## Breaking change

These paths move from exit 1 to exit 2:

- every `CliUsageErrorEnvelopeWithExit` call site (`work-session` subcommands, `event wait` / `event watch`)
- `--valid-days` (`csr create`), `--limit` (`websh records`), and `--tail` (`webftp-log`, `exec ls`, `audit`, `log`) when given a non-positive value

## Not in this PR

Most local validation still exits 1 (about 27 call sites), including the manual flag parsing in `exec` and `websh`. Moving those to 2—and widening the docs back once they are—is tracked in #318, along with removing the build-then-override shape in `utils/error_envelope.go`. The docs here stay narrow so the documented contract is never wider than the code.

## Verification

Run at `a833baf`:

```
go build ./...          exit 0
go vet ./...            exit 0
gofmt -l .              clean
go test -race ./...     exit 0, 38 packages ok, 0 FAIL
golangci-lint run ./... 0 issues.
```

Neither test guard was assumed to work. Transposing `runWorkSessionHelper`'s first two arguments fails at the call site:

```
--- FAIL: TestExtendTableUsageError_ExitsTwo
    Error: []string{"table", "json"} does not contain "http://127.0.0.1:1"
```

and running the helper process with `GO_WORKSESSION_HELPER_OUTPUT` unset prints `unset or unknown GO_WORKSESSION_HELPER_OUTPUT: ""` and exits 99.

`TestExtendTableUsageError_ExitsTwo` was likewise checked against a revert rather than assumed: putting `CliErrorWithExit` back on `utils/error_envelope.go:74` fails it with `expected: 2, actual: 1`.


README.md
```diff
@@ -300,7 +300,7 @@ Exit code is `1`, and under `--output json` the refusal is an error envelope on
 |------|---------|
 | `0`  | Success |
 | `1`  | General error (network failure, server error, etc.), and permanent refusals such as a credential on the command line (`command_inline_credential`)—see "When a command is denied" |
-| `2`  | Usage error (invalid flags or arguments) |
+| `2`  | Usage error—a flag or argument rejected by a `work-session` subcommand, by `event wait` / `event watch`, or by the positive-value check behind `--tail`, `--limit`, and `--valid-days` (`utils.ExitCodeUsageError`). Under `--output json` the envelope carries `error_code` `usage_error`. Every other command still exits `1` when it rejects a flag or argument, as do errors Cobra itself rejects while parsing (an unknown flag or subcommand) |
```
junho226: This cell lists three sources of exit 2 — `work-session`, `event wait`/`watch`, and the positive-value check — and then says "Under `--output json` the envelope carries `error_code` `usage_error`." That reads as applying to all three, but the third one emits no envelope at all:

```
$ alpacon log h --tail=0 --output json
Error: --tail must be a positive integer.
$ echo $?
2
```

Plain text on stderr, no `error_code`, no `exit_code` field. `RequirePositiveInt` calls `CliErrorWithExitCode` directly (`utils/utils.go:624`) and never reaches `CliUsageErrorEnvelopeWithExit`, so the JSON branch is simply not on that path.

The PR body sets the standard here — "the docs here stay narrow so the documented contract is never wider than the code" — and this is the one line that goes wider. A script that switches on `error_code == "usage_error"` under `--output json` gets an unparseable stderr line for `--tail=0`.

Smallest fix is to scope the sentence to the paths that actually build an envelope, e.g.:

```suggestion
| `2`  | Usage error—a flag or argument rejected by a `work-session` subcommand, by `event wait` / `event watch`, or by the positive-value check behind `--tail`, `--limit`, and `--valid-days` (`utils.ExitCodeUsageError`). Under `--output json` the `work-session` and `event` paths emit an error envelope carrying `error_code` `usage_error`; the positive-value check prints a plain message to stderr. Every other command still exits `1` when it rejects a flag or argument, as do errors Cobra itself rejects while parsing (an unknown flag or subcommand) |
```

The alternative — routing `RequirePositiveInt` through the envelope helper — is the better end state but needs an operation name the helper doesn't currently take; see the comment on `utils/utils.go`.

utils/utils.go
```diff
@@ -621,7 +621,7 @@ func SplitAndTrim(s, sep string) []string {
 // RequirePositiveInt exits with a usage error when an integer flag is not positive.
 func RequirePositiveInt(flagName string, value int) {
 	if value <= 0 {
-		CliErrorWithExit("--%s must be a positive integer.", flagName)
+		CliErrorWithExitCode(ExitCodeUsageError, "--%s must be a positive integer.", flagName)
```
junho226, line 624: The exit code is right, and dropping `reportCLIError()` by moving from `CliErrorWithExit` to `CliErrorWithExitCode` is a real improvement — a mistyped `--tail=0` no longer prints "report on github.com/…" as if the CLI had crashed.

Worth noting for the docs, though: this stays outside the envelope path, so under `--output json` all six call sites print bare text while every other exit-2 source prints a JSON envelope. Verified:

```
$ alpacon log h --tail=0 --output json
Error: --tail must be a positive integer.
```

Two ways to close it, both fine by me:

1. Narrow the README sentence (my suggestion on `README.md:303`) — zero code change, keeps this PR's scope.
2. Give `RequirePositiveInt` an operation and route it through `CliUsageErrorEnvelopeWithExit`, making all exit-2 paths envelope-consistent. Six call sites would each need to pass their op name, which is more than this PR set out to do — good candidate for #318 alongside the rest of the local-validation sweep.

Not blocking on its own; it only matters because the README currently promises option 2's behaviour.

utils/error.go
```diff
@@ -45,6 +45,13 @@ const (
 	// so the general case reads like the specific ones beside it.
 	ExitCodeGeneralError = 1
 
+	// ExitCodeUsageError is the process exit code for a flag or argument a command
```
junho226, line 57: The doc comment is the part of this PR that will age best — "1 means the request failed (network, server), 2 means the invocation itself was wrong" is the rule a future contributor needs when deciding which helper to call.

One class of call site doesn't fit it cleanly today, both in `cmd/worksession/worksession_create.go` (outside this diff, so noting here rather than inline):

- `:136` — `--use requires an active workspace. Run 'alpacon login' first.` The invocation was fine; the environment wasn't.
- `:230` — `--use requires the session to be active…`. The session has already been created server-side by this point, so "the invocation was wrong" is not what happened; the create succeeded and only the local attach didn't. Two blocks below, that same pending state exits `ExitCodePendingApproval` (4) with a comment explaining that a machine consumer should branch on "wait or check later" — which is exactly the right reading here too.

Both already used `CliUsageErrorEnvelopeWithExit` before this PR, so the classification isn't yours; this change is just what makes it observable. Worth a look when #318 revisits the taxonomy.

junho226 (commented):
## PR review: #317 — fix(utils): exit 2 on usage errors

**Author**: @jisung-02 · **Branch**: `302-usage-error-exit-code` → `main` · **Stack**: golang (cobra CLI) · **Changes**: +69 −12 across 11 files · **HEAD**: `076a6541e`

---

### Summary

The core change is right and I verified it end to end. `CliUsageErrorEnvelopeWithExit` printed `exit_code: 2` while the process exited 1; it now exits 2 in both output modes, and `buildCliUsageErrorEnvelope` overrides the 1 that `buildCliErrorEnvelope` hardcodes so the envelope and the process finally agree. `RequirePositiveInt` moves to 2 as well. Since this is a wire contract every calling script reads, I built the binary and ran it rather than taking the diff's word for it.

**Observed exit codes** (binary built from `076a6541e`):

| Invocation | Exit |
|---|---|
| `alpacon log somehost --tail=0` | **2** |
| `alpacon audit --tail=-1` | **2** |
| `alpacon websh records SESS --limit 0` | **2** |
| `alpacon csr create --domain example.com --valid-days 0` | **2** |
| `alpacon event wait` (no `--type`) | **2** |
| `alpacon event wait --type work_session --timeout 0` | **2** |
| `alpacon event wait --type totally_unknown_type` | **2** |
| `alpacon event watch` (no `--type`) | **2** |
| `alpacon work-session use` / `use --unset abc123` | **2** |
| `alpacon work-session list --requester-type bogus` | **2** |
| `alpacon work-session extend ses-1` (no expiry) | **2** |
| `alpacon --nope` (unknown flag) | 1 |
| `alpacon bogus-subcommand` | 1 |
| `alpacon websh records` (missing `ExactArgs`) | 1 |
| `alpacon --output=bogus server ls` | 1 |
| `alpacon exec --output=bogus srv -- ls` | 1 |
| `alpacon csr create --valid-days 5` (no `--domain`) | 1 |

JSON mode agrees where an envelope is emitted:

```
$ alpacon work-session use --output json ; echo $?
{
  "ok": false,
  "exit_code": 2,
  "error_code": "usage_error",
  "message": "SESSION_ID argument is required (or pass --unset)",
  "context": { "operation": "use" }
}
2
```

---

### Findings

#### 1. README's JSON-envelope sentence is wider than the code (inline on `README.md`)

The new exit-code cell covers three sources in one breath — `work-session`, `event wait`/`watch`, **and the positive-value check behind `--tail`, `--limit`, `--valid-days`** — then says "Under `--output json` the envelope carries `error_code` `usage_error`." That last clause is not true for the third source:

```
$ alpacon log h --tail=0 --output json
Error: --tail must be a positive integer.
$ echo $?
2
```

Plain text on stderr, no envelope, no `error_code`. `RequirePositiveInt` reaches `CliErrorWithExitCode` directly and never goes through `CliUsageErrorEnvelopeWithExit`. The PR body's own principle — "the docs here stay narrow so the documented contract is never wider than the code" — is what this line breaks, and it is the one thing I'd fix before merge. Details and both possible fixes are in the inline comments.

#### 2. `csr create` now answers two different codes for the same flag

`--valid-days 0 --domain example.com` exits 2, but `--valid-days 5` without `--domain` exits 1 — `cmd/csr/csr_create.go:59` guards "`--valid-days`, `--key`, and `--out` require `--domain` or `--ip`" with `CliErrorWithExit`, fifteen lines above the `RequirePositiveInt` that now exits 2. Both reject the same flag on the same command. I understand this is deliberately #318's scope and the docs are narrowed to match, so it is not a defect in this PR — but it is the sharpest visible seam the split leaves behind, and worth naming in #318 as the first site to convert.

#### 3. Two converted sites are state outcomes, not bad command lines

Both already routed through `CliUsageErrorEnvelopeWithExit`, so this PR inherits rather than introduces the classification — but it is what promotes them from 1 to 2, which is when it starts to matter:

- `cmd/worksession/worksession_create.go:136` — `--use requires an active workspace. Run 'alpacon login' first.` Not being logged in is an environment problem; the command line is fine. A script that reads 2 as "fix the flags" will rewrite flags that were never wrong.
- `cmd/worksession/worksession_create.go:230` — `--use requires the session to be active…`. By this point the session **already exists server-side**; the create succeeded and only the local attach failed. Two blocks below, the same pending-approval state exits **4** via `PrintPendingApproval` with the comment that a machine consumer should branch on "wait or check later". So one underlying state now yields 2 on the `--use` path and 4 without it. 4 looks like the right code for both.

Neither line is in this diff, so I can't comment inline — flagging for a follow-up.

#### 4. PR body attributes `--tail` to `cp`

The `--tail` site is `cmd/webftp/webftp.go:41`, whose command is `Use: "webftp-log"`. `cp` lives in `cmd/ftp/cp.go` and calls `RequirePositiveInt` nowhere. Body text only — the README and `CLAUDE.md` don't name commands there, so the shipped docs are unaffected.

---

### Checked and correct

Things I went looking for and did not find, stated so you know they were covered:

- **Cobra does not bypass the new path.** `SilenceUsage`/`SilenceErrors` are set nowhere in the repo, so every cobra error — unknown flag, unknown subcommand, `ExactArgs` — returns from `RootCmd.Execute()` into `cmd/root.go:Execute()` → `CliErrorWithExit` → 1. Observed on all three. The PR body's claim that cobra keeps 1 is accurate, and the docs say so.
- **`--output=bogus` exits 1** through the root `PersistentPreRunE`, and `cmd/exec/exec.go:138` / `cmd/websh/websh.go:264` repeat that check locally, also at 1. This is the CLI's own validation rather than a cobra parse error, but "every other command still exits 1 when it rejects a flag or argument" covers it, so the docs hold.
- **`cmd/event/wait.go:92` is not a demoted runtime error.** It looked like one, but `resolveWaitOptions` (`cmd/event/wait_adapter.go:60-80`) only fails on the local `default` branch — "`--until` is required for X: this CLI has no built-in end condition for it" — and never on a server response. So exit 2 is right, and the new help text's "`1` … the subscription was rejected" still maps to `wait.go:123`, which is untouched. Nice that the help block and the code stayed in sync.
- **The "report an issue" footer is gone from the positive-int path.** Moving to `CliErrorWithExitCode` drops `reportCLIError()`, so `--tail=0` no longer looks to the user like a CLI bug. Unadvertised improvement.
- **The test-helper exit-code collision is handled.** `runWorkSessionHelper`'s own marker failure also exits 2 (`worksession_error_json_test.go:141`), but it writes non-JSON to stderr, so `require.NoError(json.Unmarshal(...))` fails first; the added `env.ExitCode` assert then pins envelope↔process agreement rather than just the process code. The comment explaining this is the right call.
- **Nothing branches on the old code.** No `.sh`, Makefile, or workflow in this repo tests `$?`, `-eq 1`, or `exit_code`. No CHANGELOG file exists, so README + `CLAUDE.md` + `.github/copilot-instructions.md` + the `event wait` help text is the full documentation surface, and all four were updated.

### Sweep completeness

Complete for what it claims. All 27 `CliUsageErrorEnvelopeWithExit` call sites (`cmd/worksession/`, `cmd/event/`) and all 6 `RequirePositiveInt` call sites (`csr create`, `websh records`, `webftp-log`, `audit`, `exec list`, `log`) move to 2 through the two helpers. 388 other `CliErrorWithExit(` calls remain at 1 across `cmd/`; the large majority are genuine runtime failures, and the local-validation subset — `cmd/token/acl_server_add.go:38,41,71`, `cmd/token/acl_file_add.go:45`, `cmd/exec/exec.go:120,125,138`, `cmd/websh/websh.go:232,236,264`, `cmd/websh/websh_invite.go:25` and the rest — is deferred to #318 with the docs narrowed to match. Splitting the wire-contract change from the long tail is the right call; the narrowed docs are what makes it safe.

### Verification

Run in a clean worktree at `076a6541e`:

```
go build -v .                exit 0
go vet ./...                 exit 0, no output
gofmt -l .                   clean
go test -race ./...          exit 0, 38 packages ok, 0 failures
go test -race -run 'UsageError|RequirePositiveInt|ExitCodeUsageError|BuildCliUsageErrorEnvelope' -v
                             6 tests PASS
```

The full `-race` suite completed without the `api/log` stall seen on an earlier cold run of this repo. CI on the PR is green across all six checks (Build and Test, golangci-lint, govulncheck, CodeQL, Analyze go/actions).

**Not verified**: `golangci-lint` is not installed on this machine, so I relied on the PR's green CI job for it; same for `govulncheck`. I did not exercise any path that needs a live Alpacon workspace, so every observed exit code above is from local validation reached before the first network call.

---

**Verdict**: COMMENTED — the exit-code change itself is correct, consistent, and demonstrated; the one thing to fix is the README sentence promising a JSON envelope the `--tail`/`--limit`/`--valid-days` path does not emit.

- [x] No security concerns
- [ ] Documentation accurate — see finding 1

README.md
jisung-02: Applied your suggestion verbatim in c026eda. The row now splits the two behaviours:

> Under `--output json` the `work-session` and `event` paths emit an error envelope carrying `error_code` `usage_error`; the positive-value check prints a plain message to stderr.

I took option 1 rather than routing `RequirePositiveInt` through the envelope helper, for the reason you gave on `utils/utils.go`: the six call sites would each have to pass an operation name, which is wider than this PR. Noted in #318 so the envelope-consistency change lands with the rest of the sweep.

The paragraph above the table (README.md:291) was already scoped to `work-session`, `event`, and the inline-credential refusal, so it needed no change.

utils/utils.go
jisung-02: Agreed on both counts. Going with option 1 for now (README scoped in c026eda), and option 2 is the better end state — recorded in #318 as part of the sweep, since giving each of the six call sites an operation name is more than this PR set out to do.

Good catch on `reportCLIError()`: that was a side effect of moving to `CliErrorWithExitCode`, not something I set out to do, but it is the right outcome — a mistyped `--tail=0` should not read like a CLI crash.

utils/error.go
jisung-02: Both sites are now in #318 under a "Reclassify, not convert" section, kept separate from the 27 call sites that move 1 → 2 — they need the opposite treatment.

I agree `:230` reads as 4: the session exists server-side by then, the create succeeded, and only the local attach failed, so the same underlying state should not yield 2 on the `--use` path and 4 without it. `:136` is an environment problem with a fine command line, which is exactly what the doc comment's rule excludes from 2.

Also added your `csr create` observation to #318 as the first site to convert — `--valid-days 0` exiting 2 while `--valid-days 5` without `--domain` exits 1 is the sharpest seam this split leaves behind.

junho226 (commented):
## PR review: #317 - fix(utils): exit 2 on usage errors

**Author**: @jisung-02
**Branch**: `302-usage-error-exit-code` → `main`
**Stack**: golang
**Changes**: +69 -12 across 11 files

---

### Summary

`CliUsageErrorEnvelopeWithExit` printed an envelope claiming `exit_code: 2` while the process exited 1, so the envelope contradicted the process and a caller reading only the exit code could not separate "the request failed" from "the command line was wrong". This adds `utils.ExitCodeUsageError = 2` and makes both output branches and `RequirePositiveInt` actually exit 2, then narrows the docs to the paths that really do.

The strongest thing about this PR is the discipline in the last three commits: the docs were walked *back* twice (`076a654`, `c026eda`) so the documented contract stays narrower than the code rather than wider. That is the right direction for a public exit-code contract, and I verified the final wording holds.

### Verification performed

Run locally in a worktree at `c026eda`:

| Check | Result |
|---|---|
| `go build ./...` | pass (no output) |
| `go vet ./...` | pass (no output) |
| `gofmt -l .` | pass (no files listed) |
| `go test -race ./...` | pass — 38 packages `ok`, 0 `FAIL` |
| `go test -count=1 -race ./utils/... ./cmd/worksession/... ./cmd/event/...` | pass (uncached, to defeat the test cache) |

```
ok  	github.com/alpacax/alpacon-cli/utils	2.552s
ok  	github.com/alpacax/alpacon-cli/cmd/worksession	3.064s
ok  	github.com/alpacax/alpacon-cli/cmd/event	2.983s
```

**Could not run locally:** `golangci-lint` is not installed in this environment (`command not found`), so I did not independently reproduce the PR's `golangci-lint run ./...` claim. CI does run it, and `gh pr checks 317` reports `golangci-lint pass`, along with `Build and Test`, `Analyze (go)`, `Analyze (actions)`, `CodeQL`, and `govulncheck` — all passing.

### Claims I checked against the code

Every documentation claim in this PR is accurate as far as I can verify:

- **The three exit-2 paths are exactly the three the docs name.** The only non-test sites reaching exit 2 are `utils/error_envelope.go:72` (JSON branch), `:74` (table branch), and `utils/utils.go:624` (`RequirePositiveInt`). No fourth path.
- **`--tail` / `--limit` / `--valid-days` is the complete `RequirePositiveInt` surface.** Six call sites: `cmd/webftp/webftp.go:41`, `cmd/audit/audit.go:41`, `cmd/log/log.go:28`, `cmd/exec/list.go:50` (all `tail`), `cmd/websh/websh_records.go:28` (`limit`), `cmd/csr/csr_create.go:74` (`valid-days`).
- **"The positive-value check prints a plain message to stderr"** (README) is right, and it is the non-obvious part. `RequirePositiveInt` calls `CliErrorWithExitCode`, and `utils/log.go:34-38` writes plain text to stderr without consulting `OutputFormat` — so `--tail=0 --output json` emits no envelope. Documenting that asymmetry rather than glossing it is exactly the right call.
- **"Cobra parse errors exit 1"** holds: `cmd/root.go:72-76` routes `RootCmd.Execute()` errors through `CliErrorWithExit` → `os.Exit(1)`.
- **The `event wait` help text is correct.** I checked this specifically because `cmd/event/wait.go:92` routes a `resolveWaitOptions` error through the usage helper, which would be wrong if that function could fail on a network or subscription error. It cannot — `cmd/event/wait_adapter.go:60-80` returns exactly one error, the `--until is required for <type>` usage case. So splitting "connection failed or the subscription was rejected" (1) from "a flag or argument was rejected" (2) is accurate.
- **`event watch` needs no help-text update** — its `Long` block documents no exit codes, so nothing went stale there.
- **The new test comment is accurate.** `worksession_error_json_test.go:141` really does `os.Exit(2)` on a missing marker, so the added `assert.Equal(t, utils.ExitCodeUsageError, env.ExitCode)` is doing real disambiguation work rather than restating the exit-code assertion above it.

### Findings

**1. (Warning) The `BREAKING CHANGE` footer on commit `dade758` names the wrong commands.**

The footer reads:

> BREAKING CHANGE: `--valid-days` (csr create), `--limit` (websh records), and `--tail` (**cp**, exec list, audit, log) now exit 2 instead of 1 when given a non-positive value.

`cp` (`cmd/ftp/cp.go`) never calls `RequirePositiveInt` — its behavior does not change in this PR. The command that does is `webftp-log` (`cmd/webftp/webftp.go:41`, registered as `Use: "webftp-log"`), which the footer omits entirely.

The PR body already has the correct list (`webftp-log`, `exec ls`, `audit`, `log`), so this is only wrong in the commit trailer — but the trailer is the machine-readable half. Anything generating release notes or a breaking-change digest from commit footers would tell users that `cp` changed (it did not) and never tell them `webftp-log` did, which is the one an operator scripting log tailing would actually trip over. Worth an amend before merge, or a correction in the release notes if the history is already settled.

**2. (Suggestion) No test covers the table-mode branch of `CliUsageErrorEnvelopeWithExit`.** See the inline comment on `utils/error_envelope.go:74`.

**3. (Nit, optional) The build-then-override pattern is now duplicated.** See the inline comment on `utils/error_envelope.go:27`.

### Highlights

- `ExitCodeUsageError`'s doc comment explains *why* 2 is distinct from 1 ("1 means the request failed, 2 means the invocation itself was wrong") rather than restating the value. That is the comment style the rest of `utils/error.go` already uses, and it is what makes the constant usable by someone who did not read this PR.
- `TestExitCodeUsageError_IsTwoAndDistinct` pins 2 against every other exit constant, matching the existing `TestExitCodeNotApproved_IsSixAndDistinct`. Pinning the literal `2` is correct here — the value is a public contract, so a test that read the constant on both sides would pass even if someone renumbered it.
- `TestRequirePositiveIntExitsWithUsageErrorCode` uses the subprocess-helper pattern to assert a real process exit code instead of refactoring `os.Exit` away behind an injectable seam. The right trade for a CLI.
- Scoping the docs to the paths that actually emit 2, and filing #318 for the remaining ~27 call sites instead of widening the promise now, keeps the contract honest.

---

**Verdict**: COMMENTED

**Checklist**:
- [x] Code follows stack conventions (`gofmt` clean, `go vet` clean)
- [x] Tests included and passing (envelope builder, constant distinctness, and a real-exit-code subprocess test)
- [x] No security vulnerabilities (`CodeQL`, `govulncheck` pass in CI)
- [x] Documentation updated (README, `CLAUDE.md`, copilot-instructions, `event wait` help — all verified accurate)
- [ ] Commit trailer corrected (finding 1)

utils/error_envelope.go
```diff
@@ -57,11 +65,11 @@ func CliErrorEnvelopeWithExitCode(code int, operation string, err error, format
...
+// CliUsageErrorEnvelopeWithExit reports a bad flag or argument and exits(2); error_code is fixed to UsageErrorCode.
 func CliUsageErrorEnvelopeWithExit(operation string, format string, args ...any) {
 	if OutputFormat == OutputFormatJSON {
-		PrintJSONError(os.Stderr, buildCliErrorEnvelope(operation, UsageErrorCode, fmt.Sprintf(format, args...)))
-		os.Exit(1)
+		PrintJSONError(os.Stderr, buildCliUsageErrorEnvelope(operation, fmt.Sprintf(format, args...)))
+		os.Exit(ExitCodeUsageError)
 	}
-	CliErrorWithExit(format, args...)
+	CliErrorWithExitCode(ExitCodeUsageError, format, args...)
```
junho226, line 74: Suggestion: this line is the behavior change most users will actually hit, and it is the one branch with no test.

`--output json` is opt-in, so the table branch is the default path for every human and most scripts. It changed from `CliErrorWithExit` (exit 1) to `CliErrorWithExitCode(ExitCodeUsageError, ...)` (exit 2), which is precisely the breaking part of this PR.

What the added tests cover:

- `TestExtendJSONErrorEnvelope_UsageError` — the JSON branch (line 72), via the subprocess helper that sets `utils.OutputFormat = utils.OutputFormatJSON`.
- `TestBuildCliUsageErrorEnvelope_CarriesUsageCodeAndExitTwo` — the envelope builder only, no process exit.
- `TestRequirePositiveIntExitsWithUsageErrorCode` — reaches `CliErrorWithExitCode` from `RequirePositiveInt`, a different caller.

So nothing asserts that `alpacon work-session use` (no args, default output) exits 2. If someone later "simplifies" this line back to `CliErrorWithExit`, the whole suite still passes.

The cheap fix: `TestWorkSessionJSONErrorHelperProcess` already forks a real process, and the only reason it lands in JSON mode is the assignment on `worksession_error_json_test.go:143`. Dropping that assignment for one case gives you a table-mode run of the same `extend` path, where you assert the exit code and that stderr is *not* JSON:

```go
func TestExtendTableUsageError_ExitsTwo(t *testing.T) {
	_, stderr, exitCode := runWorkSessionTableHelper(t, "http://127.0.0.1:1", "extend", "ses-1")

	assert.Equal(t, utils.ExitCodeUsageError, exitCode)
	assert.NotContains(t, stderr, `"error_code"`)
}
```

Not a merge blocker — the code is correct as written. But given that #318 will move ~27 more call sites onto this contract, a regression test on the default branch seems worth having before that lands.

utils/error_envelope.go
```diff
@@ -24,6 +24,14 @@ func buildCliErrorEnvelope(operation, errorCode, message string) JSONErrorEnvelo
 	}
 }
 
+// buildCliUsageErrorEnvelope overrides the exit code buildCliErrorEnvelope
```
junho226, line 27: Nit (optional, take it or leave it): this is now the second helper that builds an envelope and then overrides the `ExitCode: 1` that `buildCliErrorEnvelope` hardcodes — `CliErrorEnvelopeWithExitCode` (line 57-64) does the same thing, and each carries its own comment explaining the same trap.

I am *not* suggesting you change it in this PR. The current shape is deliberate and tested (`TestBuildCliErrorEnvelope_DefaultsExitCodeToOne` pins the default precisely so the override stays visible), and threading an exit code through `buildCliErrorEnvelope` would touch call sites unrelated to #302 — outside this PR's scope.

Just flagging it for whoever picks up #318: at a third refusal code, the "remember to override or the envelope lies about the process" footgun is worth designing out rather than commenting again. Adding the exit code as a parameter to `buildCliErrorEnvelope` would make the correct thing the only thing.

utils/error_envelope.go
jisung-02: Added in `4bf9197`. I parametrized the existing helper rather than adding a second `runWorkSessionTableHelper`, so both modes share one subprocess setup:

```go
func TestExtendTableUsageError_ExitsTwo(t *testing.T) {
	// Table is the default output mode, so this is the branch of
	// CliUsageErrorEnvelopeWithExit most callers reach.
	stdout, stderr, exitCode := runWorkSessionHelper(t, utils.OutputFormatTable, "http://127.0.0.1:1",
		"extend", "ses-1")

	assert.Equal(t, utils.ExitCodeUsageError, exitCode)
	assert.Empty(t, stdout)
	assert.NotContains(t, stderr, `"error_code"`)
	assert.Contains(t, stderr, "--expires-in or --expires-at")
}
```

`runWorkSessionHelper` now takes the output format instead of hardcoding JSON at the assignment you pointed at, and the helper process reads it from `GO_WORKSESSION_HELPER_OUTPUT`. The JSON-named identifiers followed, since none of them is JSON-only any more: `TestWorkSessionJSONErrorHelperProcess` to `TestWorkSessionErrorHelperProcess`, and the marker `worksession-json-helper` to `worksession-error-helper`.

I checked it catches the regression you described rather than assuming it would. Reverting line 74 to `CliErrorWithExit`:

```
--- FAIL: TestExtendTableUsageError_ExitsTwo (0.02s)
    	Error:      	Not equal:
    	            	expected: 2
    	            	actual  : 1
```

Restored, and the suite is green at `4bf9197`: `go test -race ./...` exit 0 with 38 packages ok and 0 FAIL, `golangci-lint run ./...` 0 issues.

utils/error_envelope.go
jisung-02: Leaving it here, as you suggested, and recorded in #318 under a new "Design out the exit-code override" section so it does not get lost with the rest of the sweep. It names both override sites (`utils/error_envelope.go:29` and `:56`), the reason this PR is the wrong place — the current shape is pinned by `TestBuildCliErrorEnvelope_DefaultsExitCodeToOne`, and threading a parameter through would touch call sites unrelated to #302 — and your conclusion that taking the exit code as a parameter of `buildCliErrorEnvelope` makes the correct thing the only thing.

A correction I owe you, found while writing that section up. The PR body's `## Problem` paragraph claimed the old code printed `exit_code: 2` while the process exited 1. It did not. On `main`, `CliUsageErrorEnvelopeWithExit` called `buildCliErrorEnvelope` with no override, so the envelope carried `exit_code: 1` and the process exited 1 — consistent with each other, and inconsistent only with the README, which already documented 2. Commit `f4b82de` states it correctly ("exited 1 while the README documented 2"); the PR body did not, and both of your review summaries carried that wrong framing forward from it. Mine to fix, and the body is now corrected.

It does not weaken the nit. The override is still a trap the next caller has to remember, and `CliErrorEnvelopeWithExitCode` needing the same comment twice is the evidence for that. But it has not actually fired yet, and I did not want #318 recording a motivation that never happened.

cmd/worksession/worksession_error_test.go
```diff
@@ -99,20 +113,21 @@ func TestUseJSONErrorEnvelope_LocalStateError(t *testing.T) {
 }
 
 // runWorkSessionHelper re-runs the test binary as a subprocess executing the given
-// work-session subcommand in json output mode; returns (stdout, stderr, exitCode).
-func runWorkSessionHelper(t *testing.T, serverURL string, args ...string) (string, string, int) {
+// work-session subcommand in the given output mode; returns (stdout, stderr, exitCode).
+func runWorkSessionHelper(t *testing.T, outputFormat, serverURL string, args ...string) (string, string, int) {
```
junho226, line 119: Nit: `outputFormat` and `serverURL` are adjacent `string` parameters, so transposing them at a call site still compiles. In the two usage-error tests nothing reaches the network, so a swap would set `OutputFormat` to a URL — which falls through to the table branch — and the test would pass for the wrong reason. Not worth a struct, but the helper could guard it cheaply:

```go
require.Contains(t, []string{utils.OutputFormatTable, utils.OutputFormatJSON}, outputFormat)
```

Same line of thinking applies to `utils.OutputFormat = os.Getenv("GO_WORKSESSION_HELPER_OUTPUT")` at line 156 — an empty value silently means table rather than failing the helper.

junho226 (dismissed):
Follow-up review of the delta from `c026eda` (my previous review) to `4bf9197`. All three findings are resolved. Approving.

## Previous findings

**1. `BREAKING CHANGE` footer named the wrong commands — FIXED.** `dade758` said `--tail (cp, exec list, audit, log)`; the rewritten `220801a` says `--tail (webftp-log, exec ls, audit, log)`. Checked against every `RequirePositiveInt` call site rather than against the footer:

| call site | command |
|---|---|
| `cmd/webftp/webftp.go:41` | `Use: "webftp-log"` (`webftp.go:11`) |
| `cmd/exec/list.go:50` | `Use: "ls"`, alias `list` (`list.go:11`) |
| `cmd/audit/audit.go:41` | `Use: "audit"` (`audit.go:11`) |
| `cmd/log/log.go:28` | `Use: "log SERVER"` (`log.go:11`) |
| `cmd/csr/csr_create.go:74` | `csr create` (`--valid-days`) |
| `cmd/websh/websh_records.go:28` | `websh records` (`--limit`) |

That is all six non-test call sites, `webftp-log` is now named, and the phantom `cp` is gone — `cmd/ftp/cp.go` calls `RequirePositiveInt` nowhere. Picking `ls` over the `list` alias matches the canonical `Use:`.

**2. No test for the table branch — FIXED.** `TestExtendTableUsageError_ExitsTwo` (`cmd/worksession/worksession_error_json_test.go:76`) covers `utils/error_envelope.go:74`. I did not take the commit message's word for it — reverting line 74 to `CliErrorWithExit(format, args...)` and re-running gives:

```
--- FAIL: TestExtendTableUsageError_ExitsTwo (0.03s)
    worksession_error_json_test.go:82:
        	Error:      	Not equal:
        	            	expected: 2
        	            	actual  : 1
```

The assertions also separate this from the neighbouring paths: `NotContains(stderr, "\"error_code\"")` rules out the JSON branch, and `Contains(stderr, "--expires-in or --expires-at")` rules out the helper's own missing-marker `os.Exit(2)`. Threading the format through `runWorkSessionHelper` instead of adding a second helper was the right call — the two modes now differ by one argument.

**3. Build-then-override duplication — deliberately deferred, correctly.** Still present at `utils/error_envelope.go:27-33`, and the same shape at `:56-66`. #318 is open and its "Design out the exit-code override" section names `buildCliUsageErrorEnvelope` (`utils/error_envelope.go:29`) directly, so the PR body's claim that it is tracked holds up. Fine to leave for a change that already has `error_envelope.go` open.

## New findings

Nothing blocking. One nit inline on the helper signature; separately, the file is still named `worksession_error_json_test.go` while the JSON framing inside it was deliberately dropped (`TestWorkSessionJSONErrorHelperProcess` → `TestWorkSessionErrorHelperProcess`, `GO_WANT_WORKSESSION_JSON_HELPER` → `GO_WANT_WORKSESSION_ERROR_HELPER`). The filename is the last JSON-named thing left, and it now holds a table-mode test. Rename to `worksession_error_test.go` whenever this file is next touched.

## Verification

Run in a worktree at `4bf9197`:

```
go build ./...        exit 0
go vet ./...          exit 0
gofmt -l .            no output
go test -race ./...   exit 0, 38 ok, 0 FAIL
```

`golangci-lint` is not installed on this machine (`command -v golangci-lint` returns nothing), so I did not run it — I am relying on CI for that one. `gh pr checks 317`: `golangci-lint` pass, `Build and Test` pass, `Analyze (go)` pass, `Analyze (actions)` pass, `CodeQL` pass, `govulncheck` pass.

cmd/worksession/worksession_error_test.go
jisung-02: Both taken, in `a833baf`, along with the rename.

**The call-site guard.** Your line exactly:

```go
require.Contains(t, []string{utils.OutputFormatTable, utils.OutputFormatJSON}, outputFormat)
```

Confirmed it fires rather than assuming it: transposing the two arguments in `TestExtendTableUsageError_ExitsTwo` gives

```
--- FAIL: TestExtendTableUsageError_ExitsTwo
    Error: []string{"table", "json"} does not contain "http://127.0.0.1:1"
```

**The env var at line 156.** Same check in the helper process, and the value is now read into a local before it reaches `utils.OutputFormat`:

```go
outputFormat := os.Getenv("GO_WORKSESSION_HELPER_OUTPUT")
if outputFormat != utils.OutputFormatTable && outputFormat != utils.OutputFormatJSON {
    fmt.Fprintf(os.Stderr, "unset or unknown GO_WORKSESSION_HELPER_OUTPUT: %q\n", outputFormat)
    os.Exit(helperSetupExitCode)
}
utils.OutputFormat = outputFormat
```

Running the helper with the variable unset:

```
unset or unknown GO_WORKSESSION_HELPER_OUTPUT: ""
exit 99
```

**Why 99 and not 2.** Writing the guard as `os.Exit(2)` would have made it useless here: 2 is `ExitCodeUsageError`, which is the value every one of these tests asserts, so a broken harness would have been read as the command reporting a usage error. The missing-marker path above it already had that problem — `TestExtendJSONErrorEnvelope_UsageError` carried a comment saying the parsed envelope is what tells the two exit-2 paths apart. Both now exit `helperSetupExitCode = 99`, and that comment is gone with the ambiguity it worked around.

**Filename.** `worksession_error_json_test.go` → `worksession_error_test.go`, in the same commit. Splitting the rename out would have meant reverting these edits and reapplying them on top of it, so it rides along; `git log --follow` and the diff both still show it as a rename.

Verification at `a833baf`: `go build ./...`, `go vet ./...` exit 0, `gofmt -l .` clean, `go test -race ./...` exit 0 with 38 ok and 0 FAIL, `golangci-lint run ./...` 0 issues.

junho226 (approved):
Round 3 — approving again. The delta is one test-only commit plus a merge of `main`. It fixes both round-2 nits and changes nothing about exit-code behaviour.

## Scope

Delta from my round-2 commit `4bf9197` to head `5dd346a`:

```
$ git diff --stat 4bf9197 a833baf
 ...ion_error_json_test.go => worksession_error_test.go} | 17 ++++++++++++++---
 1 file changed, 14 insertions(+), 3 deletions(-)
```

One authored commit (`a833baf`) touching one test file, then a merge of `origin/main` (`5dd346a`). `utils/error_envelope.go`, `utils/utils.go` (`RequirePositiveInt`), and `cmd/event/wait.go` are byte-identical to what I approved in round 2, so the round-1 findings stay resolved without re-checking.

## Round-2 leftovers — both fixed

**1. The filename was the last JSON-named thing in the file — FIXED.** Renamed to `worksession_error_test.go` at `similarity index 89%`, so `git log --follow` still reaches the file's history.

**2. (inline) Transposable `outputFormat` / `serverURL`, and an empty `GO_WORKSESSION_HELPER_OUTPUT` silently meaning table — FIXED, both halves.** The parent-side guard is the `require.Contains` I suggested (`cmd/worksession/worksession_error_test.go:123`), and the child process now rejects an unset or unknown value (`:161-165`). I checked both fire rather than reading them and assuming:

```
$ GO_WANT_WORKSESSION_ERROR_HELPER=1 go test -count=1 \
    -run '^TestWorkSessionErrorHelperProcess$' ./cmd/worksession/ \
    -args worksession-error-helper extend ses-1
unset or unknown GO_WORKSESSION_HELPER_OUTPUT: ""
FAIL	github.com/alpacax/alpacon-cli/cmd/worksession	0.990s

$ GO_WANT_WORKSESSION_ERROR_HELPER=1 GO_WORKSESSION_HELPER_OUTPUT=http://127.0.0.1:1 \
    go test -count=1 -run '^TestWorkSessionErrorHelperProcess$' ./cmd/worksession/ \
    -args worksession-error-helper extend ses-1
unset or unknown GO_WORKSESSION_HELPER_OUTPUT: "http://127.0.0.1:1"
FAIL	github.com/alpacax/alpacon-cli/cmd/worksession	0.673s
```

The second run is the transposition case itself: a swapped call site puts the server URL where the output format goes, and both layers now refuse it instead of falling through to the table branch and passing for the wrong reason.

Moving the missing-marker path from `os.Exit(2)` to `helperSetupExitCode` (99) is the right consequence of that, and dropping the comment in `TestExtendJSONErrorEnvelope_UsageError` that explained how to tell the two exit-2 paths apart is correct — there is no second exit-2 path left to tell apart.

## The merge

`5dd346a` merges `origin/main`. It touches two files this PR also owns, and both resolve cleanly:

- `cmd/worksession/worksession_error_test.go` — main's #321 rewrote `require.True(t, errors.As(err, &exitErr), ...)` as `require.ErrorAs(t, err, &exitErr)` and dropped the then-unused `errors` import. It landed on the renamed file, not on a stale copy left behind by the rename.
- `utils/error.go` — main's `APITokenACLNotAllowed` const lands above the exit-code block; `ExitCodeGeneralError` and `ExitCodeUsageError` survive intact at `:55` and `:62`.

## New findings

None.

## Verification

Run in a worktree at `5dd346a`:

```
go build ./...        exit 0
go vet ./...          exit 0
gofmt -l .            no output
go test -race ./...   exit 0, 38 ok, 0 FAIL
```

`golangci-lint` is not installed on this machine (`command -v golangci-lint` returns nothing), so I did not run it and am relying on CI for that one. `gh pr checks 317`: `golangci-lint` pass, `Build and Test` pass, `Analyze (go)` pass, `Analyze (actions)` pass, `CodeQL` pass, `govulncheck` pass.

Issue #320, opened 2026-08-11
test: run independent tests in parallel and finish testifylint adoption


jisung-02:
## Background

#279 normalized roundabout assertions to dedicated testify helpers (Contains/NotContains, ErrorAs/NotErrorAs, Len, NoError, JSONEq) and enabled `testifylint` in `.golangci.yml` to keep them from coming back. Three checkers were left disabled there because their findings need work of their own. A full-suite sweep with golangci-lint v2.12.2 (output caps removed) also surfaced test patterns that predate current Go and tooling.

While those were being fixed, a second problem with the suite showed up: Go runs test packages concurrently, but the tests inside one package run one after another unless each calls `t.Parallel()`. Nothing in this repository did, so `go test -race ./...` took the sum of every sleep in `api/event`—76 s on a 10-core machine, 74 s of it in that one package. Both batches land together in #417, which closes this issue.

Counts below are as of #417's head (807 of 1067 top-level tests parallel) and drift as tests change.

## Scope—run independent tests in parallel

`t.Parallel()` becomes the first statement of every top-level test that owns no process-wide state. A test stays serial when it:

- calls `t.Setenv` or `t.Chdir`—Go panics when either is combined with `t.Parallel()`, in whichever order;
- swaps `os.Stdout` or `os.Stderr`, including through `testutil.CaptureOutput`, or sets `utils.OutputFormat`;
- reassigns a package-level seam (`runPresenceStepUp`, `tunnelFlags`, `refreshAccessToken` and about fifteen more); what decides is whether the assignment outlives the test, not the name;
- reads a package-level Cobra command—`Commands()` sorts the child list in place and `Find()` builds the merged flag set lazily, so a read-only test still races with its neighbour;
- re-executes itself as a helper process, since the child reaches an `os.Exit`;
- watches a timing window that a parallel neighbour sharing the core would narrow.

Serial tests run to completion before the parallel ones start, within one package's test binary, so a serial test may still reassign a global as long as it restores it. The rule and the reasons go into `CLAUDE.md` under "Test patterns" so the next test file follows it.

Bare table loops in `cmd/approval` and `cmd/worksession` run their cases under `t.Run`, with the empty filter named `empty` (Go would call it `#00`) and a parenthesised case renamed so `-run 'TestX/case'` can select it.

[...]

PR #278, opened 2026-07-22
feat(server): add reboot, shutdown, upgrade, and refresh system-action commands

jisung-02:
## Background

The CLI had no way to trigger server system-level actions (reboot, shutdown, OS package upgrade, system information refresh). The server already exposes these through the actions endpoint (`POST /api/servers/servers/{id}/actions/`, body `{action, force}`), but there was no CLI entry point, so engineers and AI agents had to fall back to the web console. This PR adds four subcommands that wrap that endpoint.

## Changes

Adds four commands under `alpacon server`.

- `reboot`, `shutdown`, `upgrade` are disruptive actions. Each prompts for confirmation before running (skippable with `-y`) and runs the MFA retry flow when the server requires it. If the target has active user work (an open Websh/WebFTP session or an in-flight command), the server-side busy guard refuses the action, and `--force` overrides it.
- `refresh` is a non-disruptive action (`update_information`). It has no confirmation, `--force`, or MFA path and simply issues the API call.

The API layer is a single `RequestServerAction(ac, serverName, action, force)` in `api/server/server.go`, which resolves the server ID by name and POSTs to the actions endpoint. The shared logic of the three disruptive commands (confirmation, MFA retry, busy-guard handling) is centralized in `runDisruptiveServerAction` in `cmd/server/server_action.go`, and the shared `-y`/`--force` flag registration lives in `addDisruptiveActionFlags`.

Command help (`Long`) and `Example` follow the existing `cmd/server` subcommand conventions, and the confirmation prompts are unified on the `<verb> server '%s'?` pattern. This branch also modernizes the error parsing in `utils/error.go` to Go 1.25 string-iterator idioms (`SplitSeq`/`CutPrefix`).

## Permissions

The busy guard and MFA gating are both decided server-side. The CLI only forwards `--force` (the intent to override the busy guard); it never relaxes or bypasses authorization on the client. Disruptive actions may require MFA per the server-side `mfa_rules`, while `refresh` is non-disruptive and is not subject to MFA.

## Testing

`TestRequestServerAction` in `api/server/server_test.go` verifies the transmitted `action` and `force` values and the POST path (`/api/servers/servers/{id}/actions/`) across the `reboot_system` (force=true) and `update_information` (force=false) paths. `go test -race ./...`, `go vet`, and `golangci-lint run` all pass.

## Checklist

- [x] `go build ./...` / `go vet ./...` pass
- [x] `go test -race ./...` passes
- [x] `golangci-lint run` passes (0 issues)
[...]

PR #321, opened 2026-08-11
test: modernize assertion syntax and remove mixed testify usage

### LINKED ISSUE #279: test: modernize assertion syntax and remove mixed testify usage

jisung-02:
## Background

Review of #278 flagged `assert.False(t, strings.Contains(...))` in `cmd/server/server_action_test.go:35`. That form collapses both operands into one bool, so a failure prints only "Should be false", while `assert.NotContains` prints the actual string together with the substring. This PR sweeps the whole suite for the same roundabout patterns, normalizes them to the dedicated helpers, and enables testifylint so they stay out.

## Changes

- Convert the 15 `assert.True/False`-wrapping-`strings.Contains` sites to `Contains`/`NotContains` and drop the 4 `strings` imports left unused (49ea7c7)
- Rewrite the newline checks in `cmd/exec/run_test.go` from `len(line) > 0 && line[len(line)-1] == '\n'` to `strings.HasSuffix`; testify has no suffix helper, so `assert.True` stays (e668afc)
- Convert the 19 assertions wrapping `errors.As`—the same anti-pattern—to `ErrorAs`/`NotErrorAs`. require/assert strength is preserved at every site, and messages that merely restated the type are dropped since the failure output already prints the expected type and the error chain (2cb8ba4)
- Replace the 2 `Equal`-on-`len` sites with `Len` and the 1 `Nil` on an error value with `NoError`, which compares `err != nil` exactly as callers do and therefore also catches typed-nil regressions (66d262e)
- Restructure the two workspace passthrough tests from per-field asserts to a single `JSONEq` over the whole payload. The per-field form had drifted to checking 6 of expected's 8 fields, and the `float64()` casts that existed only to match Unmarshal's number decoding go away with it (3315dbc)
- Enable testifylint in `.golangci.yml`. The three checkers with remaining findings—`require-error` (149), `empty` (29), `go-require` (11)—stay disabled and are tracked in #320 (2702cd1)
- Record the resulting assertion conventions in CLAUDE.md: dedicated helpers over wrapped predicates, the require/assert selection rule and the ban on `require` off the test goroutine, no assertion-restating messages, JSONEq for passthrough tests (57cf678)

## Review follow-ups

- Upgrade to `require` at every site whose next line consumes the result: the four workspace tests that feed `body` into `JSONEq`/`Unmarshal`, where an API error otherwise surfaced as "needs to be valid json" instead of the actual error (386904d), and the six `worksession_create_test.go` sites followed by `err.Error()`, where a nil error panicked and took the package's remaining test results with it (05d1b9d)
- Drop the two remaining messages that restated their assertion (3ca47cf) and the four `NotEmpty` checks that `Contains` already covers (1ab7744)
- Convert the last guarded newline check in `cmd/exec/logs_test.go`; the `if len(...) > 0` guard skipped the assertion on empty output, which is what a dropped newline produces (decc56b)
- Carry the operand in the three `HasSuffix` messages—`assert.True` prints only "Should be true", so this is the one place a message is not redundant with the helper (6c8ec24)
- Correct the CLAUDE.md helper list: `Empty` was documented as enforced while its checker is disabled, and `NotErrorAs` was missing (0f7cdae); refresh the `require-error` count the conversions above changed (e734405)

## Testing

- `go test -race ./...` — all packages ok
- `golangci-lint run ./...` — 0 issues with testifylint active
- Full-suite testifylint count 226 → 189; the checkers this PR addresses (contains, error-is-as, len, error-nil, float-compare) are all at 0
- Mutation check that `JSONEq` guards the previously unchecked fields: mutating only `front_url` in the handler fails the test; reverted with no trace
- Enforcement check: temporarily reintroducing the original pattern is caught immediately as `contains: use assert.NotContains`

Closes #279

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -159,7 +159,7 @@ _ = json.NewEncoder(w).Encode(resp)
 
 - Dedicated helper over a predicate wrapped in `True/False`: `Contains`/`NotContains`, `ErrorAs`/`NotErrorAs`, `Len`, `NoError`, `Empty`. The wrapper collapses both operands to one bool and failure output loses them. `testifylint` enforces all of these but `Empty`—its `empty` checker is still disabled in `.golangci.yml` (#320), so apply that one by hand. Suffix/prefix checks have no helper, so `assert.True(t, strings.HasSuffix(...))` stays
 - `require` when later lines depend on the assertion, `assert` when checks are independent. Never `require` inside an httptest handler or goroutine—`FailNow` off the test goroutine is undefined
-- No message that restates the assertion (helpers print operands and error chains themselves); keep only a reason the code cannot say
+- No message that restates the assertion (helpers print operands and error chains themselves); keep only a reason the code cannot say. `True`/`False` are the exception—they print `Should be true` and nothing else, so their message carries the operand: `"stderr line must end with a newline: %q", line`
 - JSON passthrough tests: one `JSONEq` over the whole body, not per-field asserts—field lists drift and silently drop fields
 
 ### Comments
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -157,7 +157,7 @@ _ = json.NewEncoder(w).Encode(resp)
 
 ### Assertion conventions (testify)
 
-- Dedicated helper over a predicate wrapped in `True/False`: `Contains`/`NotContains`, `ErrorAs`, `Len`, `NoError`, `Empty`. The wrapper collapses both operands to one bool and failure output loses them. Enforced by `testifylint`; suffix/prefix checks have no helper, so `assert.True(t, strings.HasSuffix(...))` stays
+- Dedicated helper over a predicate wrapped in `True/False`: `Contains`/`NotContains`, `ErrorAs`/`NotErrorAs`, `Len`, `NoError`, `Empty`. The wrapper collapses both operands to one bool and failure output loses them. `testifylint` enforces all of these but `Empty`—its `empty` checker is still disabled in `.golangci.yml` (#320), so apply that one by hand. Suffix/prefix checks have no helper, so `assert.True(t, strings.HasSuffix(...))` stays
 - `require` when later lines depend on the assertion, `assert` when checks are independent. Never `require` inside an httptest handler or goroutine—`FailNow` off the test goroutine is undefined
 - No message that restates the assertion (helpers print operands and error chains themselves); keep only a reason the code cannot say
 - JSON passthrough tests: one `JSONEq` over the whole body, not per-field asserts—field lists drift and silently drop fields
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,6 +155,13 @@ _ = json.NewEncoder(w).Encode(resp)
 - API tests use `httptest.NewServer` with a minimal `*client.AlpaconClient` pointing at `ts.URL`
 - Command logic is extracted to unexported helpers (e.g., `parseExecArgs`) for direct unit testing
 
+### Assertion conventions (testify)
+
+- Dedicated helper over a predicate wrapped in `True/False`: `Contains`/`NotContains`, `ErrorAs`, `Len`, `NoError`, `Empty`. The wrapper collapses both operands to one bool and failure output loses them. Enforced by `testifylint`; suffix/prefix checks have no helper, so `assert.True(t, strings.HasSuffix(...))` stays
+- `require` when later lines depend on the assertion, `assert` when checks are independent. Never `require` inside an httptest handler or goroutine—`FailNow` off the test goroutine is undefined
+- No message that restates the assertion (helpers print operands and error chains themselves); keep only a reason the code cannot say
+- JSON passthrough tests: one `JSONEq` over the whole body, not per-field asserts—field lists drift and silently drop fields
+
 ### Comments
 
 - Always write comments in English
--- a/api/event/event_test.go
+++ b/api/event/event_test.go
@@ -1920,7 +1920,7 @@ func TestGiveUpGap_BoundsCumulativeSkipped(t *testing.T) {
 		}
 	})
 
-	assert.Equal(t, maxGapWidth, len(g.skipped), "cumulative skipped capped at maxGapWidth")
+	assert.Len(t, g.skipped, maxGapWidth, "cumulative skipped capped at maxGapWidth")
 	// Gap 1 is fully recorded (retryable), gap 2 fills the cap mid-gap (partial
 	// retry), and later gaps must not claim a retry they won't get.
 	assert.Contains(t, stderr, "will retry at command end")
--- a/api/event/event_test.go
+++ b/api/event/event_test.go
@@ -3,7 +3,6 @@ package event
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -309,7 +308,7 @@ func TestErrorFromDetails_PropagatesExitCode(t *testing.T) {
 			require.Error(t, err)
 
 			var remoteErr *RemoteCommandError
-			require.True(t, errors.As(err, &remoteErr), "err must be *RemoteCommandError")
+			require.ErrorAs(t, err, &remoteErr)
 			assert.Equal(t, tt.wantOutput, remoteErr.Output)
 			assert.Equal(t, tt.wantExitCode, remoteErr.ExitCode)
 			assert.Equal(t, tt.wantErrorPhase, remoteErr.ErrorPhase)
@@ -320,7 +319,7 @@ func TestErrorFromDetails_PropagatesExitCode(t *testing.T) {
 func TestErrorFromDetails_AwaitingApprovalReturnsPendingApprovalError(t *testing.T) {
 	err := errorFromDetails(EventDetails{ID: "cmd-9", Status: "awaiting_approval"})
 	var pending *PendingApprovalError
-	require.True(t, errors.As(err, &pending), "err must be *PendingApprovalError")
+	require.ErrorAs(t, err, &pending)
 	assert.Equal(t, "cmd-9", pending.CommandID)
 }
 
@@ -430,8 +429,7 @@ func TestPollCommandExecution_ClientTimeout(t *testing.T) {
 	require.Error(t, err)
 
 	var clientTimeout *ClientTimeoutError
-	require.True(t, errors.As(err, &clientTimeout),
-		"deadline expiry must surface a *ClientTimeoutError, got %T: %v", err, err)
+	require.ErrorAs(t, err, &clientTimeout, "deadline expiry must surface a client timeout")
 }
 
 func TestNextPollTick(t *testing.T) {
@@ -587,8 +585,7 @@ func TestPollCommandExecution_ThrottleExtensionIsBoundedByDuration(t *testing.T)
 	_, err := pollCommandExecution(ac, "cmd-1", 100*time.Millisecond, 5*time.Millisecond, false, seams)
 
 	var clientTimeout *ClientTimeoutError
-	require.True(t, errors.As(err, &clientTimeout),
-		"a permanently throttled poll must surface a *ClientTimeoutError, got %T: %v", err, err)
+	require.ErrorAs(t, err, &clientTimeout, "a permanently throttled poll must surface a client timeout")
 	// One 300ms extension spends the 100ms budget whole—the ceiling is the timeout plus
 	// one backoff wait, not the timeout. The second wait gets no extension and is cut
 	// to the 95ms left, so the timeout is reported at the deadline, not 300ms past it.
@@ -613,8 +610,7 @@ func TestPollCommandExecution_ThrottleExtensionIsBoundedByCount(t *testing.T) {
 	_, err := pollCommandExecution(ac, "cmd-1", timeout, tick, false, seams)
 
 	var clientTimeout *ClientTimeoutError
-	require.True(t, errors.As(err, &clientTimeout),
-		"a permanently throttled poll must surface a *ClientTimeoutError, got %T: %v", err, err)
+	require.ErrorAs(t, err, &clientTimeout, "a permanently throttled poll must surface a client timeout")
 	var elapsed time.Duration
 	for _, d := range delays() {
 		elapsed += d
@@ -683,8 +679,7 @@ func TestPollCommandExecution_ProgressRestoresThrottleExtensions(t *testing.T) {
 	_, err := pollCommandExecution(ac, "cmd-1", timeout, tick, false, seams)
 
 	var clientTimeout *ClientTimeoutError
-	require.True(t, errors.As(err, &clientTimeout),
-		"a permanently throttled poll must surface a *ClientTimeoutError, got %T: %v", err, err)
+	require.ErrorAs(t, err, &clientTimeout, "a permanently throttled poll must surface a client timeout")
 	var elapsed time.Duration
 	for _, d := range delays() {
 		elapsed += d
@@ -1127,7 +1122,7 @@ func TestRunCommandStreaming_FailedStatusPropagatesExitCode(t *testing.T) {
 	err := runCommandStreamingWithWriter(ac, "srv", "exit 23", "", "", nil, "", stdoutBuf)
 	require.Error(t, err)
 	var remoteErr *RemoteCommandError
-	require.True(t, errors.As(err, &remoteErr), "failed status must yield *RemoteCommandError")
+	require.ErrorAs(t, err, &remoteErr, "failed status must yield a remote command error")
 	assert.Equal(t, 23, remoteErr.ExitCode)
 	assert.Equal(t, "before-exit\n", stdoutBuf.String())
 }
--- a/api/event/event_test.go
+++ b/api/event/event_test.go
@@ -1122,7 +1122,7 @@ func TestRunCommandStreaming_FailedStatusPropagatesExitCode(t *testing.T) {
 	err := runCommandStreamingWithWriter(ac, "srv", "exit 23", "", "", nil, "", stdoutBuf)
 	require.Error(t, err)
 	var remoteErr *RemoteCommandError
-	require.ErrorAs(t, err, &remoteErr, "failed status must yield a remote command error")
+	require.ErrorAs(t, err, &remoteErr)
 	assert.Equal(t, 23, remoteErr.ExitCode)
 	assert.Equal(t, "before-exit\n", stdoutBuf.String())
 }
@@ -1920,7 +1920,7 @@ func TestGiveUpGap_BoundsCumulativeSkipped(t *testing.T) {
 		}
 	})
 
-	assert.Len(t, g.skipped, maxGapWidth, "cumulative skipped capped at maxGapWidth")
+	assert.Len(t, g.skipped, maxGapWidth)
 	// Gap 1 is fully recorded (retryable), gap 2 fills the cap mid-gap (partial
 	// retry), and later gaps must not claim a retry they won't get.
 	assert.Contains(t, stderr, "will retry at command end")
--- a/api/workspace/authentication_test.go
+++ b/api/workspace/authentication_test.go
@@ -31,7 +31,7 @@ func TestGetAuthentication(t *testing.T) {
 
 	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
 	body, err := GetAuthentication(ac)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	expectedJSON, err := json.Marshal(expected)
 	require.NoError(t, err)
@@ -59,7 +59,7 @@ func TestGetAuthentication_EmptyResponse(t *testing.T) {
 
 	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
 	body, err := GetAuthentication(ac)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	var got map[string]any
 	err = json.Unmarshal(body, &got)
--- a/api/workspace/authentication_test.go
+++ b/api/workspace/authentication_test.go
@@ -8,12 +8,13 @@ import (
 
 	"github.com/alpacax/alpacon-cli/client"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestGetAuthentication(t *testing.T) {
 	expected := map[string]any{
 		"mfa_required":         true,
-		"mfa_timeout":          float64(300),
+		"mfa_timeout":          300,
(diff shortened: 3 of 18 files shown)


api/workspace/authentication_test.go
```diff
@@ -32,14 +33,9 @@ func TestGetAuthentication(t *testing.T) {
 	body, err := GetAuthentication(ac)
 	assert.NoError(t, err)
```
geunwoonoh: Warning: assertion strength is inverted relative to the convention this PR just wrote.

The new `require.NoError` on line 37 guards `json.Marshal` over a `map[string]any` of strings/bools/ints/nil, which cannot fail. Meanwhile the call whose output line 38 consumes stays on `assert`. If `GetAuthentication` errors, `body` is nil and `JSONEq` reports *"expected value needs to be valid json"* instead of the actual API error.

Per the new CLAUDE.md rule ("`require` when later lines depend on the assertion"), this is the site that needs it. Same shape in `api/workspace/preferences_test.go:37`.

```suggestion
	require.NoError(t, err)
```

api/workspace/preferences_test.go
```diff
@@ -35,15 +36,9 @@ func TestGetPreferences(t *testing.T) {
 	body, err := GetPreferences(ac)
 	assert.NoError(t, err)
```
geunwoonoh: Warning: same inversion as `authentication_test.go:34` — the impossible-to-fail `Marshal` gets `require` while the API call that line 41 depends on stays on `assert`.

```suggestion
	require.NoError(t, err)
```

cmd/worksession/worksession_create_test.go
```diff
@@ -128,13 +126,13 @@ func TestParseExpiryFlag_ExpiresAt(t *testing.T) {
 func TestParseExpiryFlag_BothProvided(t *testing.T) {
 	_, err := parseExpiryFlag("2h", "2026-12-31T23:59:59Z")
 	assert.Error(t, err)
-	assert.True(t, strings.Contains(err.Error(), "mutually exclusive"))
+	assert.Contains(t, err.Error(), "mutually exclusive")
```
geunwoonoh, line 129: Warning: `assert.Error(t, err)` on line 128 followed by `err.Error()` here. A nil error records a failure and then panics on this line, taking down the package's whole test binary instead of reporting a clean failure. Same shape at lines 134, 145, and 151.

**Pre-existing** — line 128 is not in this diff and the old `assert.True(t, strings.Contains(err.Error(), ...))` form dereferenced identically. But this PR rewrote the dereferencing line in each block, and `require-error` is disabled so testifylint will not flag it. Line 43 in this same file already does it right with `require.Error`.

Reasonable to fold in here or leave to #320 — either way worth a deliberate call.

api/event/event_test.go
```diff
@@ -1127,7 +1122,7 @@ func TestRunCommandStreaming_FailedStatusPropagatesExitCode(t *testing.T) {
 	err := runCommandStreamingWithWriter(ac, "srv", "exit 23", "", "", nil, "", stdoutBuf)
 	require.Error(t, err)
 	var remoteErr *RemoteCommandError
-	require.True(t, errors.As(err, &remoteErr), "failed status must yield *RemoteCommandError")
+	require.ErrorAs(t, err, &remoteErr, "failed status must yield a remote command error")
```
geunwoonoh: Suggestion: this message restates the assertion, which the PR's own new rule bans. `ErrorAs` already prints the expected type and the actual error chain, and the test name `TestRunCommandStreaming_FailedStatusPropagatesExitCode` already says "failed status".

Compare `worksession_create_test.go:46` — `"a settled status must be typed so the caller can exit 6"` — which says *why*, not *what*. The `:432/:588/:613/:682` messages sit in between and are defensible: "deadline expiry" / "permanently throttled" is context the assertion cannot express.

```suggestion
	require.ErrorAs(t, err, &remoteErr)
```

api/event/event_test.go
```diff
@@ -1925,7 +1920,7 @@ func TestGiveUpGap_BoundsCumulativeSkipped(t *testing.T) {
 		}
 	})
 
-	assert.Equal(t, maxGapWidth, len(g.skipped), "cumulative skipped capped at maxGapWidth")
+	assert.Len(t, g.skipped, maxGapWidth, "cumulative skipped capped at maxGapWidth")
```
geunwoonoh: Suggestion: pure restatement — `Len` prints both operands on failure and the expected value is literally named `maxGapWidth`. If there is a reason the cap matters here that the code cannot say, that is what belongs in the message; otherwise drop it.

```suggestion
	assert.Len(t, g.skipped, maxGapWidth)
```

cmd/exec/sudo_hint_test.go
```diff
@@ -15,31 +14,27 @@ func TestSudoDenialHint(t *testing.T) {
 		out := "Alpacon denied this sudo command (SUDO_NO_WORKSESSION_POLICY).\n"
 		hint := sudoDenialHint(out)
 		assert.NotEmpty(t, hint)
-		assert.True(t, strings.Contains(hint, "work-session update"),
-			"hint should point to the work-session update command")
+		assert.Contains(t, hint, "work-session update")
```
geunwoonoh, line 16: Suggestion: `assert.NotEmpty(t, hint)` on the line above is now dead weight. `Contains` cannot pass on an empty haystack with a non-empty needle, and its failure output is strictly more informative than `NotEmpty`'s. Same redundancy at lines 22, 28, and 34 — four lines to delete.

Pre-existing, but this PR rewrote the very next line in each block, so it is in scope if you want it.

CLAUDE.md
```diff
@@ -155,6 +155,13 @@ _ = json.NewEncoder(w).Encode(resp)
 - API tests use `httptest.NewServer` with a minimal `*client.AlpaconClient` pointing at `ts.URL`
 - Command logic is extracted to unexported helpers (e.g., `parseExecArgs`) for direct unit testing
 
+### Assertion conventions (testify)
+
+- Dedicated helper over a predicate wrapped in `True/False`: `Contains`/`NotContains`, `ErrorAs`, `Len`, `NoError`, `Empty`. The wrapper collapses both operands to one bool and failure output loses them. Enforced by `testifylint`; suffix/prefix checks have no helper, so `assert.True(t, strings.HasSuffix(...))` stays
```
geunwoonoh: Suggestion: this bullet lists `Empty` among the dedicated helpers and closes with "Enforced by `testifylint`", but the `empty` checker is in the disable list with 29 open sites — so that one helper is documented as enforced while it is not. `NotErrorAs` is used in this PR but missing from the list.

Consider naming the disabled checkers inline, or scoping the "enforced" clause to the checkers actually on.

geunwoonoh (dismissed):
## Summary

A pure test-quality sweep: replaces `assert.True/False`-wrapped predicates with the dedicated testify helpers, converts two workspace passthrough tests to a single `JSONEq`, turns on `testifylint` with the three still-failing checkers explicitly disabled and issue-linked, and records the resulting conventions in `CLAUDE.md`. No production code is touched.

**Local verification:**

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -race ./...` | all packages `ok` |
| `golangci-lint run ./...` | **not verified** — binary not installed in my workspace |

Sweep completeness (grep across all `*_test.go`): zero remaining `True/False` wrapping `strings.Contains`, `errors.As`, or `errors.Is`; zero `Nil`/`NotNil` on an `err`; zero `True` on a `len(...)` comparison. The 7 surviving `Equal` + `len()` sites are byte-count comparisons (`int64(len("hello world"))`), not collection lengths — correctly left alone. The counts in the description (15 Contains, 19 ErrorAs) match the diff exactly.

## Critical issues (0)

None.

## Warnings (2)

1. **[api/workspace/authentication_test.go:34, api/workspace/preferences_test.go:37]** Assertion strength is inverted relative to the convention this PR just wrote. The new `require.NoError` guards `json.Marshal` on a `map[string]any` of strings/bools/ints/nil — which cannot fail — while the call whose output the next line consumes stays on `assert`. On a `GetPreferences` error, `body` is nil and `JSONEq` reports *"expected value needs to be valid json"* instead of the actual API error.

2. **[cmd/worksession/worksession_create_test.go:128, 134, 145, 151]** `assert.Error(t, err)` immediately followed by `err.Error()`. A nil error records a failure and then panics on the next line, taking down the package's whole test binary rather than reporting a clean failure. **Pre-existing** — the `assert.Error` lines aren't in the diff and the old `assert.True(t, strings.Contains(err.Error(), …))` form dereferenced identically — but the PR rewrote the dereferencing line in each block, and `require-error` is disabled so testifylint won't flag it. Line 43 in the same file already does it right with `require.Error`. Reasonable to fold in here or leave to #320.

## Suggestions (4)

3. **[api/event/event_test.go:1125, 1923]** The "no message that restates the assertion" rule is applied unevenly. Two retained messages are pure restatements. `worksession_create_test.go:46` (`"a settled status must be typed so the caller can exit 6"`) is the model — it says *why*, not *what*. The four `:432/:588/:613/:682` messages sit in between; "deadline expiry"/"permanently throttled" is context the assertion cannot express, so those are defensible. Dropping the `got %T: %v` args throughout is right — `ErrorAs` prints the actual error and its chain.

4. **[cmd/exec/logs_test.go:183-185]** The pre-PR newline idiom survives here, in the same pattern class this PR sweeps:

   ```go
   if len(stderrLine) > 0 {
       assert.Equal(t, '\n', rune(stderrLine[len(stderrLine)-1]), "stderr should end with newline")
   }
   ```

   Beyond consistency with the `run_test.go` rewrite, the `if` guard makes the assertion vacuous on empty output — exactly the case a dropped-newline bug produces. `assert.True(t, strings.HasSuffix(stderrLine, "\n"), …)` fails there. testifylint cannot see this one, so leaving it is the one hole in "the suite is normalized." (Not inline-commentable — the file is outside this diff.)

5. **[cmd/exec/sudo_hint_test.go:16, 22, 28, 34]** `assert.NotEmpty(t, hint)` directly above `assert.Contains(t, hint, "…")` is dead weight — `Contains` cannot pass on an empty haystack with a non-empty needle, and its failure output is strictly more informative. Four lines to delete.

6. **[CLAUDE.md:160]** The convention bullet lists `Empty` among the enforced helpers and closes with "Enforced by `testifylint`", but the `empty` checker is in the disable list with 29 open sites — so that one helper is documented as enforced while it is not. `NotErrorAs` is used in this PR but missing from the list.

## Performance

No runtime code changed; nothing in the diff reaches the shipped binary. Test-side cost is a wash — the `JSONEq` rewrite adds one `Marshal` plus two `Unmarshal`s per test and removes one `Unmarshal` plus six map lookups. `go test -race ./...` wall time is unchanged.

## Duplicate / unused code

- Import hygiene is exact: `strings` and `errors` were dropped only where they went fully unused, and correctly kept in `api/websh/websh_test.go` (5 remaining uses) and `cmd/login_test.go:732` (`errors.New`). No over-removal, no orphans — confirmed by a clean `go build`.
- The only redundancy found is suggestion 5 above.

## Highlights

- **Landing the linter in the same PR as the sweep**, with the three failing checkers disabled, counted, and issue-linked, is the right sequencing — no regression window and no silent debt.
- **The `run_test.go` newline rewrite is a real strengthening, not cosmetics.** `strings.HasSuffix(line, "\n")` fails on empty input where `len(line) > 0 && line[len(line)-1] == '\n'` passed vacuously.
- **`assert.NoError` over `assert.Nil` at `run_test.go:44` is correct and stricter.** `asPhasedError(err error) (error, bool)` at `cmd/exec/run.go:401` returns the `error` interface with a literal `nil` in the false branch, so there is no typed-nil boxing hazard, and `NoError` would catch a future regression that `Nil`'s reflection would wave through. The description reasons this out accurately.
- **The `JSONEq` rewrite closed a real gap**, not just a stylistic one: `preferences_test.go` was asserting 6 of 8 fields, leaving `package_proxy` and `front_url` entirely unchecked. Dropping the `float64()` casts falls out naturally once the map is only ever marshaled instead of compared against `Unmarshal` output.
- **One pattern class per commit** makes each conversion independently reviewable and revertable.

---

**Verdict**: **APPROVED** — nothing blocking. Warnings 1 and 2 are worth a look before merge; suggestions 3–6 are polish and can ride along or go to #320.

**Checklist**:
- [x] Code follows stack conventions (gofmt/goimports clean, table-driven tests preserved)
- [x] Tests included and passing — `go test -race ./...` all `ok`
- [x] No security vulnerabilities (test-only diff; the sudo-hint disclosure assertion is preserved as `NotContains`)
- [x] Documentation updated (`CLAUDE.md` assertion conventions, see suggestion 6)
- [ ] `golangci-lint run ./...` — not verified locally, relying on CI

api/workspace/authentication_test.go
jisung-02: Fixed in 386904d — `require.NoError` on the `GetAuthentication` call. The same shape in `TestGetAuthentication_EmptyResponse` (line 61) feeds `body` into `json.Unmarshal`, so it is converted too.

The `require` on `json.Marshal` stays. It cannot fail here, as you note, but downgrading it to `assert` would let a nil `expectedJSON` reach `JSONEq` and produce exactly the kind of masked failure this comment is about — no reason to weaken it.

api/workspace/preferences_test.go
jisung-02: Fixed in 386904d, together with the `TestGetPreferences_EmptyResponse` site at line 64 that has the same shape.

cmd/worksession/worksession_create_test.go
jisung-02: Fixed in 05d1b9d. Reproduced the failure mode first: with `assert`, a nil error prints `An error is expected but got nil.` and then panics on the next line with `invalid memory address or nil pointer dereference`, which ends the test binary — the other tests in the package never report at all.

Converted six sites, not four: lines 192 and 251 dereference the same way (the second inside a table-driven subtest, still on the test goroutine, so `FailNow` is well-defined there). Lines 140 and 181 dereference nothing and stay on `assert`.

api/event/event_test.go
jisung-02: Dropped in 3ca47cf. Confirmed against the real failure output — with and without the message, the `Error:`/`expected:`/`in chain:` lines are identical; the message only appended a `Messages:` line saying the same thing again.

The `:432/:588/:613/:682` messages are kept for the reason you give.

api/event/event_test.go
jisung-02: Dropped in 3ca47cf. There is no reason the cap matters here that `Len`'s own output does not already carry.

cmd/exec/sudo_hint_test.go
jisung-02: Fixed in 1ab7744 — four lines deleted. The `NotContains` disclosure assertion and its comment are untouched.

CLAUDE.md
jisung-02: Fixed in 0f7cdae: `NotErrorAs` added, and the enforcement clause now scopes itself — `testifylint` enforces all of the listed helpers but `Empty`, whose checker is disabled with 29 open sites, so that one is applied by hand.

Your suggestion 3 exposed a gap in the same block, fixed in 6c8ec24: the no-restatement rule cannot apply to `True`/`False`, which print `Should be true` and nothing else. The rule now names that exception and requires the message to carry the operand instead.

CLAUDE.md
```diff
@@ -155,6 +155,13 @@ _ = json.NewEncoder(w).Encode(resp)
 - API tests use `httptest.NewServer` with a minimal `*client.AlpaconClient` pointing at `ts.URL`
 - Command logic is extracted to unexported helpers (e.g., `parseExecArgs`) for direct unit testing
 
+### Assertion conventions (testify)
+
+- Dedicated helper over a predicate wrapped in `True/False`: `Contains`/`NotContains`, `ErrorAs`/`NotErrorAs`, `Len`, `NoError`, `Empty`. The wrapper collapses both operands to one bool and failure output loses them. `testifylint` enforces all of these but `Empty`—its `empty` checker is still disabled in `.golangci.yml` (#320), so apply that one by hand. Suffix/prefix checks have no helper, so `assert.True(t, strings.HasSuffix(...))` stays
+- `require` when later lines depend on the assertion, `assert` when checks are independent. Never `require` inside an httptest handler or goroutine—`FailNow` off the test goroutine is undefined
+- No message that restates the assertion (helpers print operands and error chains themselves); keep only a reason the code cannot say. `True`/`False` are the exception—they print `Should be true` and nothing else, so their message carries the operand: `"stderr line must end with a newline: %q", line`
```
geunwoonoh, line 162: Suggestion (follow-up, not for this PR): this exception is a good catch — it closes a real gap in my earlier suggestion 3, and I verified the premise holds (`True`'s failure output is `Should be true` and nothing more).

Worth noting that the rule lands with ~22 pre-existing sites that do not follow it — bare `assert.True(t, strings.HasSuffix(r.URL.Path, "..."))` with no message at all, in `api/worksession/worksession_test.go` (12), `api/approval/approval_test.go` (5), and `api/websh/websh_test.go` (2), plus a few others. These sit inside httptest handlers, so a failure prints `Should be true` with no indication of the path actually received — exactly the pain this rule exists to remove.

`Empty` got an explicit caveat in the bullet above; this one currently reads as fully applied. Either a similar caveat here or a line in #320 would keep the doc honest. No change needed in this PR.

geunwoonoh (approved):
Re-review of `e734405` (8 commits since `57cf678`). All six findings from my earlier review are addressed, two of them beyond what I asked for. Re-approving — my previous approval was auto-dismissed when the new commits landed.

## Findings status

| # | Finding | Commit | Result |
|---|---|---|---|
| W1 | `assert.NoError` → `require` on the consumed API call | `386904d` | Fixed — **plus 2 sites I missed** (`*_EmptyResponse` at `authentication_test.go:61` and `preferences_test.go:64`, same shape feeding `Unmarshal`) |
| W2 | `assert.Error` + `err.Error()` panic risk | `05d1b9d` | Fixed — **6 sites, not the 4 I flagged**; `:192` and `:251` found on top |
| S3 | Assertion-restating messages | `3ca47cf` | Both dropped; the 4 defensible `:432/:588/:613/:682` messages correctly kept |
| S4 | `logs_test.go` newline idiom | `decc56b` | Fixed — the vacuous `if len(...) > 0` guard is gone |
| S5 | Redundant `NotEmpty` before `Contains` | `1ab7744` | 4 lines deleted; the `NotContains` disclosure assertion untouched |
| S6 | CLAUDE.md enforced-helper list | `0f7cdae` | `NotErrorAs` added; the "enforced" clause now scopes itself around `Empty` |

Two commits go past what I raised. `6c8ec24` documents and applies a `True`/`False` exception to the no-restatement rule — those print `Should be true` and nothing else, so the message has to carry the operand. That was a real gap in my suggestion 3, not something I caught. `e734405` refreshes the `require-error` count after the conversions.

## Verification

| Check | Result |
|---|---|
| `go build ./...` / `go vet ./...` | clean |
| `go test -race ./...` | all packages `ok` |
| **`golangci-lint run ./...` (v2.9.0, the version pinned in `lint.yaml`)** | **0 issues** |
| `require-error` / `empty` / `go-require` site counts | **149 / 29 / 11 — exact** |
| Enforcement probe, 7 anti-patterns | 6 caught, `Empty` not — matches the revised doc |

Three notes on the above:

- **The lint claim is now verified.** It was the one unchecked box on my earlier review — I had no `golangci-lint` binary then. Installed the CI-pinned v2.9.0 and ran it against `e734405`.
- **The `.golangci.yml` counts are exact, not approximate.** golangci-lint truncates by default (`max-issues-per-linter: 50`, `max-same-issues: 3`); uncapped, each disabled checker reports exactly the number in its comment, including the refreshed 159 → 149.
- **The revised doc line was checked empirically, not just read.** I injected all seven patterns into a throwaway test file: `Contains`, `NotContains`, `ErrorAs`, `NotErrorAs`, `Len`, and `NoError` were each flagged; `Empty` was not. That is precisely what the new wording claims.

The judgment on what to *leave alone* is also right. `worksession_create_test.go:140` and `:181` keep `assert.Error` because neither dereferences afterward (`:181` returns immediately) — exactly the distinction the CLAUDE.md rule draws.

## Follow-up (non-blocking, not for this PR)

The new `True`/`False`-carries-the-operand rule has ~22 pre-existing sites that do not follow it — bare `assert.True(t, strings.HasSuffix(r.URL.Path, "..."))` in `api/worksession`, `api/approval`, and `api/websh` handlers. Unlike `Empty`, which got an explicit caveat, this one lands as silent doc debt. Left an inline note; #320 is the natural home for it.

---

**Verdict**: **APPROVED**. Every finding addressed, two over-delivered, and the previously unverifiable lint claim now checks out. No new issues found.

**Checklist**:
- [x] Code follows stack conventions (gofmt/goimports clean, table-driven tests preserved)
- [x] Tests included and passing — `go test -race ./...` all `ok`
- [x] No security vulnerabilities (test-only diff; the sudo-hint disclosure assertion survives as `NotContains`)
- [x] Documentation updated and accurate — enforcement claims verified against the linter
- [x] `golangci-lint run ./...` — 0 issues on v2.9.0, the CI-pinned version

CLAUDE.md
jisung-02: Moved to #320 as item 4, so nothing changes in this PR.

Counted them at this branch head: 18 bare sites — `api/worksession/worksession_test.go` (11), `api/approval/approval_test.go` (5), `api/websh/websh_test.go` (2). The four other `HasSuffix` assertions in the repo already carry an operand (`cmd/exec/run_test.go:21`, `:112`, `cmd/exec/logs_test.go:184`, `api/event/subscribe_test.go:165`), so 18 is the whole gap.

One thing the issue records beyond the count: every one of these compares against a fully known path, so adding a message is the weaker fix. `assert.Equal(t, "/api/work-sessions/sessions/ses-abc/", r.URL.Path)` prints both operands and pins the whole path instead of its tail — that is what the sweep should reach for, with the message form kept for the cases where only the tail is meaningful.

I left the CLAUDE.md bullet without a caveat deliberately: unlike `Empty`, whose checker is off and will stay off until #320 decides, this rule is the one we want applied everywhere — the 18 sites are debt to clear, not an exception to document.

PR #323, opened 2026-08-11
fix: strip the Cf runes the control-char passes miss on audit-review output

### LINKED ISSUE #294: fix(utils): control-char stripping misses Cf runes, so a recorded session can spoof the audit timeline
Found while reviewing #292.

## Problem

`utils.IsControlRune` only covers C0, DEL, and C1:

```go
// utils/utils.go:487-490
// IsControlRune reports whether r is a C0, DEL, or C1 control character (the last can introduce CSI/OSC on 8-bit terminals).
func IsControlRune(r rune) bool {
	return r < 0x20 || (r >= 0x7f && r <= 0x9f)
}
```

Unicode `Cf` (format) runes pass through untouched. The ones that matter here are the bidirectional controls—U+202A–U+202E (embedding and override) and U+2066–U+2069 (isolates)—which reorder how a terminal *displays* a string without changing its bytes. This is the Trojan Source class of attack, applied to output rather than source.

## Why it is reachable

Every consumer of this helper renders text that originated on a remote host or from another user:

| Call site | Input |
|---|---|
| `cmd/worksession/worksession_timeline.go:151` | recorded terminal output of a session |
| `cmd/worksession/worksession_timeline.go:267` | recorded session lines |
| `cmd/worksession/worksession_timeline.go:287,290` | uploaded and downloaded file names |
| `cmd/worksession/worksession_timeline.go:295` | commands run in the session |
| `cmd/event/render.go:76` | event channel strings |
| `cmd/websh/websh_records.go:61` | recorded websh output |

`alpacon work-session timeline` exists so a reviewer can read back what someone did. Anyone with websh access to a server can emit arbitrary bytes into that recording. A bidi override embedded in a command line makes the timeline display something other than what was recorded—in a product whose value is that every command is recorded and reviewable, that is the audit trail lying to the reviewer.

Not verified: whether the server sanitizes any of these fields before storing or serving them. If it does, the CLI-side gap is defence in depth rather than the only line.

## Proposal

Extend the predicate to cover `unicode.Cf`, or at minimum the bidi ranges, keeping the existing C0/DEL/C1 checks. `unicode.Is(unicode.Cf, r)` is the whole test; the tradeoff is that it also drops U+00AD (soft hyphen) and U+200B–U+200D (ZWSP/ZWNJ/ZWJ), which legitimate text in some scripts uses. Since every call site here writes to a terminal for human review rather than reproducing text faithfully, dropping them is likely the right default, but it should be a deliberate choice.

## Also noticed

`cmd/worksession/worksession_timeline.go:151` calls `StripControlChars` without `StripANSIEscapes`, unlike the four other call sites in the same file. ESC is `0x1b`, so the escape introducer is stripped and the rest of the sequence renders as literal text (`[31m`). Cosmetic, not a security issue, but the asymmetry looks unintended.

## Scope

- `utils/utils.go:487-490`
- Tests for the bidi ranges
- Decide the U+00AD / U+200B–U+200D question

jisung-02:
## Background

No Cf (Unicode format) rune carries a control byte, so `IsControlRune` misses all of them. Among them are the bidi controls U+202A-202E and U+2066-2069, which reorder what a terminal renders without changing a byte: a recorded `denied` can reach the reviewer's screen as `approved`. That makes it a Trojan Source attack aimed at the audit trail rather than at a compiler.

`a04efd2` added `utils.StripFormatChars` and `utils.SanitizeTerminalText` and covered most call sites. Three paths on the audit-review surface were still running the old `StripANSIEscapes` + `StripControlChars` pair, and a fourth bypassed the shared JSON writer entirely.

## Changes

- `utils.StripFormatAndANSI` names the format-then-escape order in one place (fb8396c). The order is load-bearing, not stylistic: a Cf rune buried in an escape sequence breaks the regex match and leaves the sequence's tail on screen as text. `SanitizeTerminalText` now composes it instead of spelling the order out again.
- `websh records`: `sanitizeRecord` let bidi overrides straight into the table (8572ae2). They are deleted rather than mapped to a space like the control bytes are, because a space would be indistinguishable from whitespace that was actually recorded and no Cf rune carries anything a reviewer needs. The trade is that legitimate non-Latin and emoji text comes out altered — a ZWJ emoji splits into its parts, Arabic joining breaks (c6f5766 corrects a comment that claimed width preservation instead).
- `work-session timeline`: `recordingPreview` stripped escapes, handled `\r` overwrites, then stripped control bytes, and a bidi override survived all three (73fb179). The format strip goes ahead of the escape strip rather than into the control pass at the end, because the `\r` overwrite handling in between needs `\r` alive.
- `work-session recording`: `printRecordingContent` stripped escape sequences only (2c6a5b1, 40579f5). This path shows the recording as it was recorded, so the C0 pass stays out and `\r` and the line endings survive; a bidi override has no such claim on the output. DEL and the C1 block do go (8499328) — U+009B is an 8-bit CSI, so a recorded `reboot` followed by U+009B `2K` erases itself and `rm -rf /` is what the reviewer reads. That is the same rune set `escapeJSONControls` already singles out. Both print helpers take an `io.Writer` so they can be tested at all.
- `work-session recording`: the header line took no strip of any kind (ff22534). `formatTimestamp` returns the server's string verbatim when it does not parse, so a stored `ESC[2J ESC[H` plus a bidi override cleared the reviewer's screen and reordered the header. Every sibling call site already wrapped this value; this one did not.
- `work-session timeline --output json`: `outputTimelineJSON` called `json.MarshalIndent` directly, bypassing the shared writer's `escapeJSONControls` (963d44e). `encoding/json` emits a bidi override verbatim, so the JSON path handed the terminal the reordering the table path already blocked. `PrintJSONValue` escapes rather than strips, so a decoder still reads the recorded value back unchanged. One side effect on this machine-readable surface: `FormatJSON` sets `SetEscapeHTML(false)` where `json.MarshalIndent` escapes HTML by default, so `&`, `<`, and `>` now come out literal instead of as `\u0026`, `\u003c`, `\u003e`. Decoder-equivalent, but a consumer grepping raw bytes would notice — and it does not make the CLI uniform: `PrintTable`'s JSON branch still marshals through `json.MarshalIndent`, so every list command keeps emitting `&`, `<`, `>`. Converging that one is its own change.

### Follow-ups from review

Findings from two review passes over the branch, each fixed here rather than deferred.

- `work-session recording`: `printRecordingContent` still let a bare ESC through (d970456). `ansiEscapeRE` ends an ESC-led form at `\x40-\x7e`, so a sequence whose final byte falls in `\x30-\x3f` never matches, and this is the one call site that follows `StripFormatAndANSI` with no control pass to sweep up the unmatched introducer. `ESC ( 0` switched the charset for the rest of the screen and `ESC 7` / `ESC 8` saved and restored the cursor, which is worse than the `\r` this path deliberately keeps: `\r` overwrites inside its own line, a restored cursor overwrites a line the reviewer already read. ESC is stripped after the regex rather than by widening the regex, so the change stays local — by that point anything the regex does match is gone, and leaving the tail as text is the trade `StripControlChars` already makes.
- `utils.IsC1OrDEL` names the DEL-and-C1 predicate that `escapeJSONControls` and `printRecordingContent` had each spelled out inline (8643b7c). Same rule, same reason, and widening one copy alone would have left a gap that reads as covered.
- `work-session timeline`: `formatDetails` wrapped every field it renders in `SanitizeTerminalText` except `websh_record`, which passed `MaskedRecord` to `TruncateString` raw (7c58b0b). MaskedRecord is the only verbatim terminal capture in the struct, so it was the wrong one to skip. Nothing leaked — `printTimelineTableTo` sanitizes the cell again and `escapeJSONControls` covers the JSON path — but the truncation order was visible: escape bytes spent the 60-rune budget, so a recording row showed less of the command than the command row beside it.
- `work-session recording`: `printRecordingContent` let SO and SI (`\x0e` / `\x0f`) through as well. SO invokes G1 into GL and the switch holds until SI or a reset — the same charset switch `ESC ( 0` performs, reached without an ESC. Both are C0 bytes, so the deliberate absence of a C0 pass on this path passed them straight to the screen. xterm in UTF-8 mode ignores the shift functions, but the Linux console, VTE, and PuTTY honor them, which is reviewer environment enough to close rather than document. BS, BEL, VT, and FF stay: like the `\r` this path keeps on purpose, they act inside their own line.
- The C1 case pinned only the two-byte UTF-8 form `C2 9B`. A recording of a program emitting 8-bit output carries the raw byte `0x9b`, which is invalid UTF-8, so `IsC1OrDEL` never decodes it — what keeps it off the terminal is `strings.Map` substituting U+FFFD for the invalid byte. That is an implicit stdlib property rather than a visible predicate, and a later rewrite into a byte loop would reopen it silently, so the raw form now has its own case. The same substitution is a fidelity change on the one path whose contract is "as it was recorded": invalid bytes used to reach the screen verbatim.

## Safety

The threat model is the reviewer's terminal, not the stored record. Nothing here rewrites what the server holds; each change is the last step before bytes reach a screen. Whether the server sanitizes these fields before serving them was not verified, so treat the CLI side as defence in depth rather than the only line.

`--output json` escapes instead of stripping, which keeps that path lossless for machine consumers while still safe to `cat`.

## Out of scope

U+2028 and U+2029 (Zl/Zp) are reached by no pass here — they are neither Cc nor Cf. The JSON path is safe either way, since both encoders escape them unconditionally, and xterm-family terminals do not honor them as line breaks, so the table path's practical risk is near zero. Anything downstream that does honor them (a web viewer, a log pipeline) would see a cell spill into a second row, which `utils/output.go:162-164` states as an invariant. Left for its own change rather than widened into this one.

The plain-text error path is the same harm class on a different surface: `CliError*` prints the message raw to stderr, and `parseAPIError` puts the server's `detail` — or the truncated body verbatim when it is not JSON — into that message, so a hostile or compromised server can land a bidi override or an escape sequence on the operator's terminal through any command's error output. The `--output json` error envelope already goes through `escapeJSONControls`; only the default text form is exposed. That is every command's stderr, not the audit-review surface this PR is scoped to, so it goes to its own issue rather than into this diff.

`ansiEscapeRE` itself keeps the narrow `\x40-\x7e` final-byte range; ECMA-48 allows `\x30-\x7e` for the nF forms. Every other caller runs a control pass afterwards, which removes the unmatched ESC and leaves the tail inert, so the gap has no other reachable consequence. Widening the regex would change what every sanitized string in the CLI looks like and belongs in its own change.

## Testing

- `go test -race -count=1 ./...` — 38 packages ok, no failures
- `golangci-lint run ./...` — 0 issues
- Every commit was checked out into a detached worktree and built and tested on its own; all pass
- `TestStripFormatChars` now pins all twelve members of `Bidi_Control` individually — the nine embedding, override, and isolate runes plus U+061C, U+200E, and U+200F, which the first version left out — and that the pass leaves control bytes to `StripControlChars`
- `TestRecordingPreview_StripsFormatCharBuriedInSequence` and the matching `sanitizeRecord` case pin the ordering: with the strips reversed, `\x1b[2<ZWJ>Kls` renders as `2Kls`
- `TestPrintRecordingHeader_SanitizesTimestamp` fails on the old code with `Recording 1/1 — \x1b[2J\x1b[H\u202eSPOOFED`
- `TestPrintRecordingContent_StripsC1Controls` fails on the old code with `reboot\^[2Krm -rf /`, and `TestPrintRecordingContent_KeepsControlBytes` still pins that `\r` survives
- `TestPrintRecordingContent_StripsUnmatchedEscapeIntroducers` covers the three forms `ansiEscapeRE` cannot match: `ESC ( 0`, `ESC 7` / `ESC 8`, `ESC # 3`
- `TestPrintRecordingContent_StripsShiftFunctions` fails on the old code with `a\x0eb\x0fc`
- `TestPrintRecordingContent_StripsRawC1Byte` pins the raw `0x9b` form down to the U+FFFD it becomes, so a rewrite away from `strings.Map` breaks the test rather than the protection
- `TestFormatDetails_WebshRecord_SanitizesBeforeTruncating` pins that the escape goes before the 60-rune cut

Closes #294

### DIFF
--- a/cmd/websh/websh_records.go
+++ b/cmd/websh/websh_records.go
@@ -55,7 +55,9 @@ Use --query to search records by command text (fuzzy match).`,
 }
 
 func sanitizeRecord(record string, width int) string {
-	record = utils.StripANSIEscapes(record)
+	// Format chars are deleted rather than spaced: they are zero-width, so a space
+	// in their place would widen the row beyond what was recorded.
+	record = utils.StripFormatAndANSI(record)
 	// Map control chars to spaces so they separate tokens instead of merging them.
 	record = strings.Map(func(r rune) rune {
 		if utils.IsControlRune(r) {
--- a/cmd/websh/websh_records.go
+++ b/cmd/websh/websh_records.go
@@ -55,8 +55,10 @@ Use --query to search records by command text (fuzzy match).`,
 }
 
 func sanitizeRecord(record string, width int) string {
-	// Format chars are deleted rather than spaced: they are zero-width, so a space
-	// in their place would widen the row beyond what was recorded.
+	// Format chars are deleted rather than spaced: a space would be
+	// indistinguishable from whitespace that was actually recorded, and no Cf rune
+	// carries anything a reviewer needs. The tradeoff is that legitimate text
+	// changes shape — a ZWJ emoji splits into its parts, Arabic joining breaks.
 	record = utils.StripFormatAndANSI(record)
 	// Map control chars to spaces so they separate tokens instead of merging them.
 	record = strings.Map(func(r rune) rune {
--- a/cmd/worksession/worksession_recording.go
+++ b/cmd/worksession/worksession_recording.go
@@ -81,11 +81,14 @@ func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total in
 
 func printRecordingContent(w io.Writer, raw string) {
 	// No full control pass: this path shows the recording as it was, so \r and the
-	// line endings have to survive. DEL and the C1 block are the exception — U+009B
-	// is an 8-bit CSI, so it opens a control sequence on a terminal that decodes
-	// them, and no C1 code point is content a recording can carry.
+	// line endings have to survive. DEL, the C1 block, and a bare ESC are the
+	// exception. ansiEscapeRE matches ESC-led forms ending in \x40-\x7e alone, so
+	// "ESC ( 0" and "ESC 7" get past it and still reach the reviewer's terminal,
+	// where they switch the charset and move the cursor off the line. Every matched
+	// sequence is already gone by here, so a leftover ESC is an introducer that
+	// failed to match and its tail belongs on screen as text.
 	content := strings.Map(func(r rune) rune {
-		if r == 0x7f || (r >= 0x80 && r <= 0x9f) {
+		if r == 0x1b || utils.IsC1OrDEL(r) {
 			return -1
 		}
 		return r
--- a/cmd/worksession/worksession_recording.go
+++ b/cmd/worksession/worksession_recording.go
@@ -81,14 +81,15 @@ func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total in
 
 func printRecordingContent(w io.Writer, raw string) {
 	// No full control pass: this path shows the recording as it was, so \r and the
-	// line endings have to survive. DEL, the C1 block, and a bare ESC are the
-	// exception. ansiEscapeRE matches ESC-led forms ending in \x40-\x7e alone, so
-	// "ESC ( 0" and "ESC 7" get past it and still reach the reviewer's terminal,
-	// where they switch the charset and move the cursor off the line. Every matched
-	// sequence is already gone by here, so a leftover ESC is an introducer that
-	// failed to match and its tail belongs on screen as text.
+	// line endings have to survive. ansiEscapeRE matches ESC-led forms ending in
+	// \x40-\x7e alone, so "ESC ( 0" and "ESC 7" get past it and switch the charset
+	// or move the cursor off the line. By here every matched sequence is gone, so a
+	// leftover ESC is an introducer that failed to match and its tail belongs on
+	// screen as text. SO and SI reach that same charset switch without an ESC and
+	// hold it until a reset; xterm in UTF-8 mode ignores them, the Linux console
+	// and VTE do not.
 	content := strings.Map(func(r rune) rune {
-		if r == 0x1b || utils.IsC1OrDEL(r) {
+		if r == 0x1b || r == 0x0e || r == 0x0f || utils.IsC1OrDEL(r) {
 			return -1
 		}
 		return r
--- a/cmd/worksession/worksession_recording.go
+++ b/cmd/worksession/worksession_recording.go
@@ -77,7 +77,9 @@ func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total in
 }
 
 func printRecordingContent(w io.Writer, raw string) {
-	content := utils.StripANSIEscapes(raw)
+	// No control pass: this path shows the recording as it was, so \r and the
+	// line endings have to survive.
+	content := utils.StripFormatAndANSI(raw)
 	_, _ = fmt.Fprint(w, content)
 	if len(content) > 0 && content[len(content)-1] != '\n' {
 		_, _ = fmt.Fprintln(w)
--- a/cmd/worksession/worksession_recording.go
+++ b/cmd/worksession/worksession_recording.go
@@ -71,7 +71,7 @@ func findRecording(recordings []wsapi.TimelineItem, index int) (*wsapi.TimelineI
 func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total int) {
 	header := fmt.Sprintf("Recording %d/%d", idx, total)
 	// formatTimestamp hands back the server's string whenever it fails to parse, so
-	// this line is a sink for whatever was stored, same as every other timeline cell.
+	// this line is a sink for whatever was stored.
 	if ts := utils.SanitizeTerminalText(resolveTimestamp(target.Timestamp)); ts != "" {
 		header += " — " + ts
 	}
--- a/cmd/worksession/worksession_recording.go
+++ b/cmd/worksession/worksession_recording.go
@@ -69,7 +69,9 @@ func findRecording(recordings []wsapi.TimelineItem, index int) (*wsapi.TimelineI
 
 func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total int) {
 	header := fmt.Sprintf("Recording %d/%d", idx, total)
-	if ts := resolveTimestamp(target.Timestamp); ts != "" {
+	// formatTimestamp hands back the server's string whenever it fails to parse, so
+	// this line is a sink for whatever was stored, same as every other timeline cell.
+	if ts := utils.SanitizeTerminalText(resolveTimestamp(target.Timestamp)); ts != "" {
 		header += " — " + ts
 	}
 	_, _ = fmt.Fprintln(w, header)
--- a/cmd/worksession/worksession_recording.go
+++ b/cmd/worksession/worksession_recording.go
@@ -2,6 +2,8 @@ package worksession
 
 import (
 	"fmt"
+	"io"
+	"os"
 
 	wsapi "github.com/alpacax/alpacon-cli/api/worksession"
 	"github.com/alpacax/alpacon-cli/client"
@@ -49,8 +51,8 @@ var workSessionRecordingCmd = &cobra.Command{
 			utils.CliUsageErrorEnvelopeWithExit(opRecording, "Recording index %d out of range (session has %d recording(s)).", recordingIndex, len(recordings))
 		}
 
-		printRecordingHeader(target, idx, len(recordings))
-		printRecordingContent(target.MaskedRecord)
+		printRecordingHeader(os.Stdout, target, idx, len(recordings))
+		printRecordingContent(os.Stdout, target.MaskedRecord)
 	},
 }
 
@@ -65,19 +67,19 @@ func findRecording(recordings []wsapi.TimelineItem, index int) (*wsapi.TimelineI
 	return &recordings[index-1], index
 }
 
-func printRecordingHeader(target *wsapi.TimelineItem, idx int, total int) {
+func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total int) {
 	header := fmt.Sprintf("Recording %d/%d", idx, total)
 	if ts := resolveTimestamp(target.Timestamp); ts != "" {
 		header += " — " + ts
 	}
-	fmt.Println(header)
-	fmt.Println()
+	_, _ = fmt.Fprintln(w, header)
+	_, _ = fmt.Fprintln(w)
 }
 
-func printRecordingContent(raw string) {
+func printRecordingContent(w io.Writer, raw string) {
 	content := utils.StripANSIEscapes(raw)
-	fmt.Print(content)
+	_, _ = fmt.Fprint(w, content)
 	if len(content) > 0 && content[len(content)-1] != '\n' {
-		fmt.Println()
+		_, _ = fmt.Fprintln(w)
 	}
 }
--- a/cmd/worksession/worksession_recording.go
+++ b/cmd/worksession/worksession_recording.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	wsapi "github.com/alpacax/alpacon-cli/api/worksession"
 	"github.com/alpacax/alpacon-cli/client"
@@ -79,9 +80,16 @@ func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total in
 }
 
 func printRecordingContent(w io.Writer, raw string) {
-	// No control pass: this path shows the recording as it was, so \r and the
-	// line endings have to survive.
-	content := utils.StripFormatAndANSI(raw)
+	// No full control pass: this path shows the recording as it was, so \r and the
+	// line endings have to survive. DEL and the C1 block are the exception — U+009B
+	// is an 8-bit CSI, so it opens a control sequence on a terminal that decodes
+	// them, and no C1 code point is content a recording can carry.
+	content := strings.Map(func(r rune) rune {
+		if r == 0x7f || (r >= 0x80 && r <= 0x9f) {
+			return -1
+		}
+		return r
+	}, utils.StripFormatAndANSI(raw))
 	_, _ = fmt.Fprint(w, content)
 	if len(content) > 0 && content[len(content)-1] != '\n' {
 		_, _ = fmt.Fprintln(w)
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -136,6 +136,17 @@ func TestRecordingPreview_StripsFormatCharBuriedInSequence(t *testing.T) {
 	assert.Equal(t, "ls", recordingPreview("\x1b[2\u200dKls"))
 }
 
+// printRecordingHeader
+
+func TestPrintRecordingHeader_SanitizesTimestamp(t *testing.T) {
+	// formatTimestamp returns the server's string as-is when it does not parse, so
+	// the header is a sink for an escape sequence that clears the reviewer's screen.
+	ts := "\x1b[2J\x1b[H\u202eSPOOFED"
+	var buf bytes.Buffer
+	printRecordingHeader(&buf, &wsapi.TimelineItem{Timestamp: &ts}, 1, 1)
+	assert.Equal(t, "Recording 1/1 — SPOOFED\n\n", buf.String())
+}
+
 // printRecordingContent
 
 func TestPrintRecordingContent_StripsFormatChars(t *testing.T) {
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -121,3 +121,16 @@ func TestRecordingPreview_EmptyRaw(t *testing.T) {
 func TestRecordingPreview_OnlyANSI(t *testing.T) {
 	assert.Equal(t, "", recordingPreview("\x1b[?2004h\x1b[2J\x1b[H"))
 }
+
+func TestRecordingPreview_StripsBidiOverride(t *testing.T) {
+	// A bidi override carries no control byte, so the control pass alone leaves
+	// it free to reorder the preview a reviewer reads back.
+	raw := "[user@host:~]$ echo ‮safe"
+	assert.Equal(t, "[user@host:~]$ echo safe", recordingPreview(raw))
+}
+
+func TestRecordingPreview_StripsFormatCharBuriedInSequence(t *testing.T) {
+	// Format chars go before the escape strip, or the match breaks and the
+	// sequence's tail lands on screen as text.
+	assert.Equal(t, "ls", recordingPreview("\x1b[2‍Kls"))
+}
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -163,6 +163,14 @@ func TestPrintRecordingContent_StripsC1Controls(t *testing.T) {
 	assert.Equal(t, "reboot2Krm -rf /\n", buf.String())
 }
 
+func TestPrintRecordingContent_StripsShiftFunctions(t *testing.T) {
+	// SO invokes G1 into GL and holds until SI or a reset: the same charset switch
+	// as "ESC ( 0", reached without an ESC.
+	var buf bytes.Buffer
+	printRecordingContent(&buf, "a\x0eb\x0fc\n")
+	assert.Equal(t, "abc\n", buf.String())
+}
+
 func TestPrintRecordingContent_StripsUnmatchedEscapeIntroducers(t *testing.T) {
 	// ansiEscapeRE ends an ESC-led form at \x40-\x7e, so these three get past it.
 	// Left in, "ESC ( 0" turns the rest of the reviewer's screen into line-drawing
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -1,6 +1,7 @@
 package worksession
 
 import (
+	"bytes"
 	"strings"
 	"testing"
 
@@ -125,12 +126,28 @@ func TestRecordingPreview_OnlyANSI(t *testing.T) {
 func TestRecordingPreview_StripsBidiOverride(t *testing.T) {
 	// A bidi override carries no control byte, so the control pass alone leaves
 	// it free to reorder the preview a reviewer reads back.
-	raw := "[user@host:~]$ echo ‮safe"
+	raw := "[user@host:~]$ echo \u202esafe"
 	assert.Equal(t, "[user@host:~]$ echo safe", recordingPreview(raw))
 }
 
 func TestRecordingPreview_StripsFormatCharBuriedInSequence(t *testing.T) {
 	// Format chars go before the escape strip, or the match breaks and the
 	// sequence's tail lands on screen as text.
-	assert.Equal(t, "ls", recordingPreview("\x1b[2‍Kls"))
+	assert.Equal(t, "ls", recordingPreview("\x1b[2\u200dKls"))
+}
+
+// printRecordingContent
+
+func TestPrintRecordingContent_StripsFormatChars(t *testing.T) {
+	var buf bytes.Buffer
+	printRecordingContent(&buf, "echo \u202esafe\n")
+	assert.Equal(t, "echo safe\n", buf.String())
+}
+
+func TestPrintRecordingContent_KeepsControlBytes(t *testing.T) {
+	// The control pass is deliberately absent here: a recording shown as it was
+	// keeps \r and its line endings.
+	var buf bytes.Buffer
+	printRecordingContent(&buf, "a\rb")
+	assert.Equal(t, "a\rb\n", buf.String())
 }
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -163,6 +163,15 @@ func TestPrintRecordingContent_StripsC1Controls(t *testing.T) {
 	assert.Equal(t, "reboot2Krm -rf /\n", buf.String())
 }
 
+func TestPrintRecordingContent_StripsRawC1Byte(t *testing.T) {
+	// A recording of an 8-bit program carries CSI as the raw byte 0x9b, which is
+	// invalid UTF-8, so IsC1OrDEL never sees it. What keeps it off the terminal is
+	// strings.Map substituting U+FFFD, so a rewrite into a byte loop reopens this.
+	var buf bytes.Buffer
+	printRecordingContent(&buf, "reboot\x9b2Krm -rf /\n")
+	assert.Equal(t, "reboot�2Krm -rf /\n", buf.String())
+}
+
 func TestPrintRecordingContent_StripsShiftFunctions(t *testing.T) {
 	// SO invokes G1 into GL and holds until SI or a reset: the same charset switch
 	// as "ESC ( 0", reached without an ESC.
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -155,9 +155,17 @@ func TestPrintRecordingContent_StripsFormatChars(t *testing.T) {
 	assert.Equal(t, "echo safe\n", buf.String())
 }
 
+func TestPrintRecordingContent_StripsC1Controls(t *testing.T) {
+	// U+009B is an 8-bit CSI: left in, it erases the recorded line and the reviewer
+	// reads what follows instead.
+	var buf bytes.Buffer
+	printRecordingContent(&buf, "reboot\u009b2Krm -rf /\n")
+	assert.Equal(t, "reboot2Krm -rf /\n", buf.String())
+}
+
 func TestPrintRecordingContent_KeepsControlBytes(t *testing.T) {
-	// The control pass is deliberately absent here: a recording shown as it was
-	// keeps \r and its line endings.
+	// The C0 pass is deliberately absent here: a recording shown as it was keeps
+	// \r and its line endings.
 	var buf bytes.Buffer
 	printRecordingContent(&buf, "a\rb")
 	assert.Equal(t, "a\rb\n", buf.String())
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -182,9 +182,8 @@ func TestPrintRecordingContent_StripsShiftFunctions(t *testing.T) {
 
 func TestPrintRecordingContent_StripsUnmatchedEscapeIntroducers(t *testing.T) {
 	// ansiEscapeRE ends an ESC-led form at \x40-\x7e, so these three get past it.
-	// Left in, "ESC ( 0" turns the rest of the reviewer's screen into line-drawing
-	// glyphs and "ESC 7"/"ESC 8" restore the cursor onto an earlier line, which lets
-	// the recording overwrite what was already read.
+	// "ESC 7"/"ESC 8" restore the cursor onto an earlier line, which lets the
+	// recording overwrite what the reviewer already read.
 	for _, raw := range []string{"ab\x1b(0", "ab\x1b7\x1b8", "ab\x1b#3"} {
 		var buf bytes.Buffer
 		printRecordingContent(&buf, raw)
--- a/cmd/worksession/worksession_recording_test.go
+++ b/cmd/worksession/worksession_recording_test.go
@@ -163,6 +163,18 @@ func TestPrintRecordingContent_StripsC1Controls(t *testing.T) {
 	assert.Equal(t, "reboot2Krm -rf /\n", buf.String())
 }
 
+func TestPrintRecordingContent_StripsUnmatchedEscapeIntroducers(t *testing.T) {
+	// ansiEscapeRE ends an ESC-led form at \x40-\x7e, so these three get past it.
+	// Left in, "ESC ( 0" turns the rest of the reviewer's screen into line-drawing
+	// glyphs and "ESC 7"/"ESC 8" restore the cursor onto an earlier line, which lets
+	// the recording overwrite what was already read.
+	for _, raw := range []string{"ab\x1b(0", "ab\x1b7\x1b8", "ab\x1b#3"} {
+		var buf bytes.Buffer
+		printRecordingContent(&buf, raw)
+		assert.NotContains(t, buf.String(), "\x1b", "raw %q", raw)
+	}
+}
+
 func TestPrintRecordingContent_KeepsControlBytes(t *testing.T) {
 	// The C0 pass is deliberately absent here: a recording shown as it was keeps
 	// \r and its line endings.
--- a/cmd/worksession/worksession_timeline.go
+++ b/cmd/worksession/worksession_timeline.go
@@ -313,7 +313,10 @@ func formatDetails(item *wsapi.TimelineItem) string {
 		return detail
 
 	case "websh_record":
-		return utils.TruncateString(item.MaskedRecord, 60)
+		// Sanitize before truncating, as every case above does: otherwise the 60-char
+		// budget goes on escape bytes and the cell shows less of the command than the
+		// rows beside it.
+		return utils.TruncateString(utils.SanitizeTerminalText(item.MaskedRecord), 60)
 
 	default:
 		return ""
--- a/cmd/worksession/worksession_timeline.go
+++ b/cmd/worksession/worksession_timeline.go
@@ -134,7 +134,9 @@ func recordingPreview(raw string) string {
 	scanner := bufio.NewScanner(strings.NewReader(raw))
 	scanner.Buffer(make([]byte, 64*1024), 1024*1024)
 	for i := 0; i < 50 && scanner.Scan(); i++ {
-		line := utils.StripANSIEscapes(scanner.Text())
+		// Not SanitizeTerminalText: its control pass would take \r before the
+		// overwrite handling below gets to use it.
+		line := utils.StripFormatAndANSI(scanner.Text())
 		// Strip trailing CR from CRLF line endings before overwrite handling.
 		line = strings.TrimRight(line, "\r")
 		// \r moves cursor to line start; take only the last overwritten segment
--- a/cmd/worksession/worksession_timeline.go
+++ b/cmd/worksession/worksession_timeline.go
@@ -2,9 +2,9 @@ package worksession
 
 import (
 	"bufio"
-	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 	"sync"
 	"text/tabwriter"
@@ -214,11 +214,11 @@ func outputTimelineJSON(rows []wsapi.TimelineAttributes, recordings []wsapi.Time
 		"timeline":   rows,
 		"recordings": recEntries,
 	}
-	b, err := json.MarshalIndent(out, "", "  ")
-	if err != nil {
+	// Not a plain Marshal: the shared writer escapes the format and C1 runes that
+	// encoding/json leaves alone, and this output is read in a terminal too.
+	if err := utils.PrintJSONValue(os.Stdout, out); err != nil {
 		utils.CliErrorEnvelopeWithExit(opTimeline, err, "Failed to serialize timeline: %s.", err)
 	}
-	fmt.Println(string(b))
 }
 
 func projectTimelineAttributes(item *wsapi.TimelineItem, serverMap map[string]string) wsapi.TimelineAttributes {
--- a/utils/output.go
+++ b/utils/output.go
@@ -70,9 +70,9 @@ func escapeJSONControls(b []byte) []byte {
 	for i := 0; i < len(b); {
 		r, size := utf8.DecodeRune(b[i:])
 		switch {
-		case r == 0x7f || (r >= 0x80 && r <= 0x9f) || unicode.Is(unicode.Cf, r):
+		case IsC1OrDEL(r) || unicode.Is(unicode.Cf, r):
 			out = appendUnicodeEscape(out, r)
-		case r == utf8.RuneError && size == 1 && b[i] >= 0x80 && b[i] <= 0x9f:
+		case r == utf8.RuneError && size == 1 && IsC1OrDEL(rune(b[i])):
 			out = appendUnicodeEscape(out, rune(b[i]))
 		default:
 			out = append(out, b[i:i+size]...)
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -488,6 +488,13 @@ func IsControlRune(r rune) bool {
 	return r < 0x20 || (r >= 0x7f && r <= 0x9f)
 }
 
+// IsC1OrDEL reports whether r is DEL or a C1 control. A terminal decodes U+009B as
+// an 8-bit CSI, so these open a control sequence that the escape pass never sees:
+// it matches ESC-led forms alone.
+func IsC1OrDEL(r rune) bool {
+	return r == 0x7f || (r >= 0x80 && r <= 0x9f)
+}
+
 func StripControlChars(s string) string {
 	return strings.Map(func(r rune) rune {
 		if IsControlRune(r) {
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -497,9 +497,9 @@ func StripControlChars(s string) string {
 	}, s)
 }
 
-// StripFormatChars removes Unicode format characters. They carry no control byte,
-// so the control passes miss them, while a bidi override reorders what is rendered:
-// "denied" can reach the screen as "approved".
+// StripFormatChars removes the Unicode format (Cf) characters. No Cf rune is a
+// control byte, so IsControlRune misses them all, while the bidi controls among
+// them reorder what a terminal renders: "denied" can display as "approved".
 func StripFormatChars(s string) string {
 	return strings.Map(func(r rune) rune {
 		if unicode.Is(unicode.Cf, r) {
@@ -509,11 +509,16 @@ func StripFormatChars(s string) string {
 	}, s)
 }
 
+// StripFormatAndANSI removes format characters before escape sequences. Reversed,
+// one buried in a sequence would break the match and leave its tail on screen.
+func StripFormatAndANSI(s string) string {
+	return StripANSIEscapes(StripFormatChars(s))
+}
+
 // SanitizeTerminalText strips escape sequences before the standalone control
 // bytes; reversed, ESC alone would go and "[2K" would stay on screen as text.
-// Format characters go first, so one buried in a sequence cannot break the match.
 func SanitizeTerminalText(s string) string {
-	return StripControlChars(StripANSIEscapes(StripFormatChars(s)))
+	return StripControlChars(StripFormatAndANSI(s))
 }
 
 // ProcessEditedData facilitates user modifications to original data,
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -376,10 +376,11 @@ func TestStripFormatChars(t *testing.T) {
 	assert.Equal(t, "abé", StripFormatChars("a\u200dbé"))
 
 	// The bidi controls are the ones that matter: they reorder what a terminal
-	// renders without changing a byte. Every one of them goes.
+	// renders without changing a byte. All twelve members of Bidi_Control go.
 	for _, r := range []rune{
 		0x202a, 0x202b, 0x202c, 0x202d, 0x202e, // embedding and override
 		0x2066, 0x2067, 0x2068, 0x2069, // isolates
+		0x061c, 0x200e, 0x200f, // marks
 	} {
 		assert.Equal(t, "ab", StripFormatChars("a"+string(r)+"b"), "U+%04X", r)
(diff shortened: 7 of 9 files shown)


cmd/worksession/worksession_recording.go
```diff
@@ -65,19 +67,21 @@ func findRecording(recordings []wsapi.TimelineItem, index int) (*wsapi.TimelineI
...
-func printRecordingHeader(target *wsapi.TimelineItem, idx int, total int) {
+func printRecordingHeader(w io.Writer, target *wsapi.TimelineItem, idx, total int) {
 	header := fmt.Sprintf("Recording %d/%d", idx, total)
 	if ts := resolveTimestamp(target.Timestamp); ts != "" {
 		header += " — " + ts
 	}
-	fmt.Println(header)
-	fmt.Println()
+	_, _ = fmt.Fprintln(w, header)
+	_, _ = fmt.Fprintln(w)
```
junho226, line 79: **This is the one I'd fold in before merge.** The header takes no strip at all, and `resolveTimestamp` → `formatTimestamp` returns the server's string verbatim when it does not parse (`worksession_timeline.go:240-241`, `return ts`).

Measured on this branch:

```go
ts := "\x1b[2J\x1b[H\u202eSPOOFED"
printRecordingHeader(&buf, &wsapi.TimelineItem{Timestamp: &ts}, 1, 1)
// "Recording 1/1 — \x1b[2J\x1b[H\u202eSPOOFED\n\n"
```

So `work-session recording` clears the reviewer's screen and then reorders the line — the exact threat this PR exists to close, on the command it is closing it for. Every sibling call site already wraps the same value (`worksession_timeline.go:179`, `:192`); only this one doesn't.

The body defers it as "prints an unparsable timestamp back verbatim," which reads as a formatting nit rather than a control-sequence sink. One call fixes it:

```go
if ts := utils.SanitizeTerminalText(resolveTimestamp(target.Timestamp)); ts != "" {
```

And the `io.Writer` you just injected makes it a two-line test — which is worth adding, since `printRecordingHeader` currently has none.

cmd/worksession/worksession_recording.go
```diff
@@ -65,19 +67,21 @@ func findRecording(recordings []wsapi.TimelineItem, index int) (*wsapi.TimelineI
...
 }
 
-func printRecordingContent(raw string) {
-	content := utils.StripANSIEscapes(raw)
-	fmt.Print(content)
+func printRecordingContent(w io.Writer, raw string) {
+	// No control pass: this path shows the recording as it was, so \r and the
+	// line endings have to survive.
+	content := utils.StripFormatAndANSI(raw)
```
junho226: The reasoning for skipping the control pass is sound but the choice isn't binary. Measured here:

```go
printRecordingContent(&buf, "reboot\u009b2Krm -rf /\n")
// -> "reboot\u009b2Krm -rf /\n"
```

U+009B is 8-bit CSI — `ESC [ 2 K` on a terminal decoding 8-bit controls, so the recorded `reboot` is erased and `rm -rf /` is what the reviewer reads. Same spoof as the bidi one, different encoding.

Stripping only DEL and U+0080–U+009F preserves every C0 char, so `\r` and the line endings survive exactly as the comment requires:

```go
content = strings.Map(func(r rune) rune {
    if r == 0x7f || (r >= 0x80 && r <= 0x9f) {
        return -1
    }
    return r
}, content)
```

That's the same rune set `escapeJSONControls` (`utils/output.go:73`) already singles out for this reason, so it isn't a new policy — and no C1 code point is legitimate content in a masked recording. `TestPrintRecordingContent_KeepsControlBytes` would still pass unchanged.

Your call whether this rides along or gets the issue the body promises, but it's the only path left in the PR that hands the terminal a live control introducer.

cmd/websh/websh_records.go
```diff
@@ -55,7 +55,9 @@ Use --query to search records by command text (fuzzy match).`,
 }
 
 func sanitizeRecord(record string, width int) string {
-	record = utils.StripANSIEscapes(record)
+	// Format chars are deleted rather than spaced: they are zero-width, so a space
+	// in their place would widen the row beyond what was recorded.
+	record = utils.StripFormatAndANSI(record)
```
junho226: The strip is right; the stated reason isn't quite. U+200D ZWJ is Cf and is not zero-width in effect — it fuses adjacent emoji into one grapheme, so deleting it *widens* the row rather than preserving it:

```go
recordingPreview("deploy \U0001F468\u200D\U0001F4BB now") // -> "deploy \U0001F468\U0001F4BB now"
// one grapheme, 2 columns  ->  two graphemes, 4 columns
```

Same class: ZWNJ changes Arabic/Persian shaping, U+0600–U+0605 legitimately precede Arabic digits, and emoji tag sequences collapse to a bare flag.

Stripping is still the right call for an audit surface, and `TruncateString` counts runes rather than display columns so the row was never column-exact anyway. But the comment promises width-preservation the code doesn't deliver. Something closer to the real reason reads better — e.g. that a replacement char would be indistinguishable from recorded whitespace, and Cf runes carry nothing a reviewer needs — with a note that legitimate non-Latin and emoji text is altered as a deliberate tradeoff.

cmd/worksession/worksession_timeline.go
```diff
@@ -212,11 +214,11 @@ func outputTimelineJSON(rows []wsapi.TimelineAttributes, recordings []wsapi.Time
 		"timeline":   rows,
 		"recordings": recEntries,
 	}
-	b, err := json.MarshalIndent(out, "", "  ")
-	if err != nil {
+	// Not a plain Marshal: the shared writer escapes the format and C1 runes that
+	// encoding/json leaves alone, and this output is read in a terminal too.
+	if err := utils.PrintJSONValue(os.Stdout, out); err != nil {
```
junho226, line 219: Right fix for the hole — this was the genuine half-fix in the old code, and catching it is the most valuable commit in the PR.

One side effect worth a line in the body: `PrintJSONValue` → `FormatJSON` sets `SetEscapeHTML(false)`, while `json.MarshalIndent` escapes HTML by default. Verified on Go 1.25.12:

```
old: "v": "a \u0026\u0026 b \u003c c"
new: "v": "a && b < c"
```

Decoder-equivalent, and it aligns this path with every other `--output json` path in the CLI, so it's an improvement — but it is a user-visible change on a machine-readable surface and a consumer grepping raw bytes for `\u0026` would notice. Not a blocker.

(I checked the other direction too: U+2028/U+2029 are escaped unconditionally by both encoders, so nothing regresses there.)

utils/utils.go
```diff
@@ -509,11 +509,16 @@ func StripFormatChars(s string) string {
...
+func StripFormatAndANSI(s string) string {
+	return StripANSIEscapes(StripFormatChars(s))
+}
+
 // SanitizeTerminalText strips escape sequences before the standalone control
 // bytes; reversed, ESC alone would go and "[2K" would stay on screen as text.
-// Format characters go first, so one buried in a sequence cannot break the match.
 func SanitizeTerminalText(s string) string {
-	return StripControlChars(StripANSIEscapes(StripFormatChars(s)))
+	return StripControlChars(StripFormatAndANSI(s))
```
junho226, line 528: Composing `StripFormatAndANSI` here is the right call — one source of truth for the order.

Informational, on what these three passes still don't reach. I measured every adjacent category against Go 1.25.12's tables:

- **Tags U+E0000–U+E007F: fully covered.** Every *assigned* rune in the block (U+E0001, U+E0020–U+E007F) is Cf. The 31 that `unicode.Is(unicode.Cf, ·)` misses are exactly the unassigned ones. No gap — worth knowing, since the block is a common blind spot.
- **Zl/Zp (U+2028, U+2029): not reached by any pass.** They are neither Cc nor Cf. JSON is safe (both encoders escape them unconditionally), but the table path lets them through — `printTimelineTableTo` with `Server: "prod\u2028db"` emits it intact. `utils/output.go:162-164` states "a cell must not spill into a second row" as an invariant; xterm-family terminals don't honor U+2028 as a break so practical risk is near zero, but any downstream renderer that does breaks it. Two runes would close the gap, if you think the invariant is worth holding literally.
- Co (private use) and Mn (variation selectors, combining marks) are also untouched — both cosmetic-only, correctly out of scope.

Nothing here needs to change in this PR; recording it so the next person doesn't re-derive it.

junho226 (commented):
## What this does

Closes the Cf (Unicode format) hole on the audit-review surface. `IsControlRune` covers C0/DEL/C1 only, so the bidi controls (U+202A–202E, U+2066–2069) walked straight through the control passes and could render a recorded `denied` as `approved` on the reviewer's screen — Trojan Source aimed at the audit trail. The PR extracts `StripFormatAndANSI` to name the load-bearing format-then-escape order once, then applies it on the three paths that were still running the old pair (`websh records`, `work-session timeline` preview, `work-session recording`) and routes `timeline --output json` through the shared JSON writer instead of a bare `MarshalIndent`.

**Verdict: comment.** The core change is correct, the ordering rationale is right, and the tests pin behavior rather than implementation. Two things keep me from a plain approve: one unsanitized sink inside a function this PR already rewrites (#1), and a stated-but-unfixed C1 gap on the same command (#2). Everything below #3 is informational.

I verified the category coverage empirically against Go 1.25.12's Unicode tables rather than by reading — results in #4. **The Tags block is fully covered; there is no gap there.**

---

## 1. `printRecordingHeader` still hands the terminal a raw control sequence (should fix)

`work-session recording` is the surface this PR hardens, and the header line above the recording gets no strip of any kind. `resolveTimestamp` → `formatTimestamp` returns the server's string verbatim when it does not parse (`worksession_timeline.go:240-241`, `return ts`), and `printRecordingHeader` prints it straight through.

Measured on this branch:

```go
ts := "\x1b[2J\x1b[H\u202eSPOOFED"
printRecordingHeader(&buf, &wsapi.TimelineItem{Timestamp: &ts}, 1, 1)
// buf == "Recording 1/1 — \x1b[2J\x1b[H\u202eSPOOFED\n\n"
```

That clears the reviewer's screen (`ESC[2J ESC[H`) and then reorders the line — both the escape class the old code already handled elsewhere and the Cf class this PR adds. Every sibling call site already wraps this same value: `printRecordingsSectionTo` (`worksession_timeline.go:192`) and `printTimelineTableTo` (`:179`). This one does not.

The PR body lists it as out of scope, described as "prints an unparsable timestamp back verbatim" — that reads as a formatting nit, but the measured behavior is a control-sequence sink squarely inside the PR's own threat model. The fix is one `utils.SanitizeTerminalText(...)` call in a function this PR is already rewriting, and the `io.Writer` added by 2c6a5b1 makes it directly testable. I'd fold it in rather than open an issue.

## 2. `printRecordingContent` leaves 8-bit C1 controls through (reconsider the framing)

```go
printRecordingContent(&buf, "reboot\u009b2Krm -rf /\n")
// -> "reboot\u009b2Krm -rf /\n"
```

U+009B is 8-bit CSI; on a terminal decoding 8-bit controls that is `ESC [ 2 K` (erase line), so the recorded `reboot` is wiped and `rm -rf /` is what the reviewer reads. Same spoof, different encoding.

The PR is right that a blanket `StripControlChars` would take the `\r` and line endings this path must preserve. But the choice isn't binary — stripping only DEL and U+0080–U+009F leaves every C0 char intact, and that is precisely the rune set `escapeJSONControls` (`utils/output.go:73`) already singles out for the same reason, so it is not a new policy. No C1 code point is legitimate content in a masked recording.

Deferring is defensible if the issue actually gets filed, but this is the only path in the PR that still hands the terminal a live control introducer, under a title that claims the audit-review output.

## 3. The deletion is not width-neutral, contrary to the comment

`websh_records.go:58-59` says format chars are deleted rather than spaced "because Cf runes are zero-width and a space in their place would widen the row past what was recorded." U+200D ZWJ is Cf and is not zero-width in effect — it fuses adjacent emoji into one grapheme, so deleting it *widens* the row:

```go
recordingPreview("deploy \U0001F468\u200D\U0001F4BB now") // -> "deploy \U0001F468\U0001F4BB now"
// one grapheme, 2 columns  ->  two graphemes, 4 columns
```

Same class: ZWNJ (U+200C) changes Arabic/Persian shaping, U+0600–U+0605 are Arabic number signs that legitimately precede digits, and emoji tag sequences collapse (the Scotland flag goes 7 runes → 1, rendering as a plain black flag).

None of this argues against stripping — bidi safety over faithful rendering is the right call for an audit surface, and `TruncateString` counts runes rather than display columns anyway, so the row was never column-exact. But the comment asserts a width guarantee the code does not provide, and the real tradeoff (legitimate non-Latin and emoji text is altered) is recorded nowhere. Suggest rewording to the actual reason.

## 4. Category coverage — measured, two residual gaps

Ran every adjacent category against Go 1.25.12's tables on this branch:

| Category | Covered by | Status |
|---|---|---|
| Cc (C0, DEL, C1) | `StripControlChars` | covered everywhere **except** `printRecordingContent` (#2) |
| Cf incl. bidi U+202A–202E, U+2066–2069 | `StripFormatChars` | covered |
| **Tags U+E0000–U+E007F** | `StripFormatChars` | **fully covered.** Every *assigned* rune in the block (U+E0001, U+E0020–U+E007F) is Cf. The 31 runes `unicode.Is(unicode.Cf, ·)` does not match are exactly the unassigned ones (U+E0000, U+E0002–U+E001F) — no gap |
| Cs (surrogates) | n/a | unreachable in a valid Go string; `escapeJSONControls` already handles bare 0x80–0x9F bytes that `json.Indent` can emit |
| Co (private use) | nothing | not stripped. Renders as a font-specific glyph, cannot move the cursor — cosmetic, fine to leave |
| **Zl / Zp (U+2028, U+2029)** | nothing | **not stripped by any pass.** JSON is safe (both `MarshalIndent` and `FormatJSON` escape them unconditionally — verified). The table path passes them through: `printTimelineTableTo` with `Server: "prod\u2028db"` emits `"prod\u2028db"`. `utils/output.go:162-164` states "a cell must not spill into a second row" as an invariant; xterm-family terminals don't honor U+2028 as a break so practical risk is near zero, but anything downstream that does (web viewer, log pipeline) breaks it. Two runes would close it |
| Mn (U+FE0F, combining marks) | nothing | combining-mark stacking can overflow a row vertically — inherent limit of any strip-based approach, out of scope |

So: no gap in what the PR set out to cover. Zl/Zp is the only category with a real, if small, argument for inclusion.

## 5. `--output json` silently stops HTML-escaping (unmentioned behavior change)

Switching `json.MarshalIndent` → `utils.PrintJSONValue` also switches `SetEscapeHTML(true)` → `(false)`. Verified on Go 1.25.12:

```
old: "v": "a \u0026\u0026 b \u003c c"
new: "v": "a && b < c"
```

Decoder-equivalent, and it makes `timeline --output json` consistent with every other `--output json` path in the CLI (all of which go through `PrintJson` / `PrintTable`), so this is an improvement. But it is a user-visible change on a machine-readable surface and isn't in the PR body — a consumer grepping raw JSON bytes for `\u0026` sees a break. Worth a line under Changes. (U+2028/U+2029 are escaped by both encoders, so no regression there.)

## 6. Test gap where it matters most

2c6a5b1 is "print the recording to an injected writer" and the body says both helpers take an `io.Writer` "so they can be tested at all" — but only `printRecordingContent` gained tests. `grep printRecordingHeader` returns the definition and the single call site, nothing else, on this branch or `main`. Given #1, that is exactly where a test belongs.

## Nits

- `websh_records.go:40` (outside the diff): "JSON keeps records verbatim" is now inaccurate — `PrintTable`'s JSON branch runs `escapeJSONControls`, so Cf and C1 come out escaped. Pre-existing, but this PR is what makes it misleading.
- `SanitizeTerminalText`'s doc comment handed the format-ordering sentence to `StripFormatAndANSI`. Correct placement, but the remaining comment now describes two of three steps. Optional.

## What's good

- The ordering rationale is genuinely load-bearing and is now stated once instead of three times. `StripFormatAndANSI` is the right extraction, and `SanitizeTerminalText` composing it keeps a single source of truth.
- `TestRecordingPreview_StripsFormatCharBuriedInSequence` and its `sanitizeRecord` twin pin the ordering by observable behavior — reversing the strips actually fails them, which is what makes the comment trustworthy.
- `TestStripFormatChars` enumerating all nine bidi controls individually, with `U+%04X` in the message, is the right shape for a security invariant: a table-driven loop that names each rune beats one composite string.
- `TestPrintRecordingContent_KeepsControlBytes` pins a deliberate omission so nobody "fixes" it into a regression. Good instinct — it is why #2 is a conversation and not a defect.
- Escaping rather than stripping on the JSON path is the right split: `PrintJSONValue` keeps machine consumers lossless while still safe to `cat`.
- `outputTimelineJSON` was a real hole. Hardening the table path while the JSON path handed over the same bytes verbatim is the classic half-fix, and catching it is the most valuable commit here.

---

## Checks run

Run in a clean worktree at `963d44e` (`gh pr checkout 323`), Go 1.25.12 darwin/amd64.

| Command | Result |
|---|---|
| `go build ./...` | pass, exit 0, no output |
| `gofmt -l .` | pass, no files listed |
| `go vet ./...` | pass, exit 0, no output |
| `go test -count=1 -timeout 8m ./...` | **pass**, exit 0, no failures across all packages |
| `go test -race -count=1 ./utils/... ./cmd/websh/... ./cmd/worksession/...` | **pass** — `ok utils 2.432s`, `ok cmd/websh 2.425s`, `ok cmd/worksession 3.402s` |
| `golangci-lint run ./...` | **pass** — `0 issues.` Note: `golangci-lint` was not installed here, so I installed v2.12.2 via `go install …/v2/cmd/golangci-lint@latest` against the repo's own `.golangci.yml`. CI may pin a different v2 patch |
| `go test -race -count=1 ./...` (whole repo, race on) | **not run to completion** — exceeded my 10-minute command budget and was killed (SIGTERM, 143). The race-enabled run above covers all three packages this PR touches; the non-race full run covers the rest. The PR author reports the full race run green |

Findings #1–#3 and #5 were measured, not inferred: a throwaway `_test.go` in `cmd/worksession` calling `printRecordingHeader`, `printRecordingContent`, `printTimelineTableTo`, and `recordingPreview` with the attack strings, plus a standalone program comparing `json.MarshalIndent` against a `SetEscapeHTML(false)` encoder and probing `unicode.Is` across Cc/Cf/Co/Zl/Zp/Mn and the full Tags block. The probe file was deleted afterwards; the worktree is clean.

cmd/worksession/worksession_recording.go
jisung-02: Folded in as `ff22534`, exactly as you wrote it.

You are right that the body's wording undersold it. "Prints an unparsable timestamp back verbatim" is a
description of the mechanism, not of the consequence, and the consequence is a control-sequence sink on the
one command this branch names in its title. I had triaged it as a different defect class (escape sequences,
not Cf) and cut it on that basis, which was the wrong axis to cut on — the axis that matters is whether the
path hands a terminal something live.

`TestPrintRecordingHeader_SanitizesTimestamp` came with it, and it fails on the old code with
`Recording 1/1 - \x1b[2J\x1b[H\u202eSPOOFED`. That is the first test `printRecordingHeader` has had.

cmd/worksession/worksession_recording.go
jisung-02: Taken, in `8499328`. The framing was wrong: I had treated it as strip-everything or strip-nothing, and
`KeepsControlBytes` only ever pinned the C0 range.

Implemented as you suggested — DEL and U+0080-009F, nothing else — with the reasoning you gave, that this is
the same rune set `escapeJSONControls` (`utils/output.go:73`) already singles out, so it is not a new policy.
`KeepsControlBytes` passes unchanged, as you predicted.

One consequence worth recording in the thread: the escape regex cannot match a sequence introduced by 8-bit
CSI, so its tail survives as text — `reboot\u009b2Krm -rf /` comes out as `reboot2Krm -rf /`. Inert text on
screen against a live control introducer is the trade I took, and it is the same one `sanitizeRecord` already
makes. `TestPrintRecordingContent_StripsC1Controls` pins that literal output so the trade is visible rather
than implied.

cmd/websh/websh_records.go
jisung-02: Correct, and the ZWJ case makes it unambiguous — deleting widens the row rather than preserving it, which is
the opposite of what the comment claimed. Rewritten in `c6f5766` along the lines you suggested: a space would
be indistinguishable from whitespace that was actually recorded, no Cf rune carries anything a reviewer needs,
and the trade is that legitimate non-Latin and emoji text comes out altered.

The `TruncateString` point is a good one to have on the record — the row was never column-exact, so the width
argument could not have been load-bearing even if it had been true.

cmd/worksession/worksession_timeline.go
jisung-02: Added to the body under Changes. The `SetEscapeHTML` difference is real and I had not noticed it: `&`, `<`,
and `>` now come out literal where `json.MarshalIndent` emitted `\u0026`, `\u003c`, `\u003e`.

I agree it is an improvement rather than a regression, since it makes `timeline --output json` consistent with
every other `--output json` path, but it is a user-visible change on a machine-readable surface and belonged
in the body. Thanks for checking the U+2028/U+2029 direction too.

utils/utils.go
jisung-02: Thank you for measuring this rather than reasoning about it — the Tags block result in particular is worth
having written down, since "Cf misses the Tags block" is the kind of claim that would otherwise get
re-derived incorrectly.

Zl/Zp is now the only item under Out of scope in the body, with your reasoning: JSON is safe either way, the
table path lets them through, xterm-family terminals do not honor them so practical risk is near zero, but
`utils/output.go:162-164` states the no-spill invariant literally and a downstream renderer that does honor
them would break it. Left for its own change rather than widened into this one.

cmd/worksession/worksession_recording.go
```diff
@@ -65,19 +68,33 @@ func findRecording(recordings []wsapi.TimelineItem, index int) (*wsapi.TimelineI
...
+func printRecordingContent(w io.Writer, raw string) {
+	// No full control pass: this path shows the recording as it was, so \r and the
+	// line endings have to survive. DEL, the C1 block, and a bare ESC are the
+	// exception. ansiEscapeRE matches ESC-led forms ending in \x40-\x7e alone, so
+	// "ESC ( 0" and "ESC 7" get past it and still reach the reviewer's terminal,
+	// where they switch the charset and move the cursor off the line. Every matched
+	// sequence is already gone by here, so a leftover ESC is an introducer that
+	// failed to match and its tail belongs on screen as text.
+	content := strings.Map(func(r rune) rune {
+		if r == 0x1b || utils.IsC1OrDEL(r) {
```
junho226: **Warning:** `\x0e` (SO) and `\x0f` (SI) get past this predicate, and they do the thing the comment above cites as the reason to strip a bare ESC. SO invokes G1 into GL, so the rest of the reviewer's screen renders in the alternate charset until SI or a reset — the same harm as `ESC ( 0`, reached through a C0 byte instead of an ESC-led form. Measured on this branch: `printRecordingContent("a\x0eb")` returns the SO byte unchanged.

Caveat on reach: xterm in UTF-8 mode ignores the shift functions. The Linux console, VTE, and PuTTY honor them — a binary dumped to a Linux console garbling everything until `reset` is exactly this.

It is a residual gap rather than a regression, so it need not block. But it is a decision, and Out of scope currently names only U+2028/U+2029 and the regex width. Either widen the predicate, which costs nothing against the "keep `\r` and the line endings" contract since no reviewer needs SO:

```go
if r == 0x1b || r == 0x0e || r == 0x0f || utils.IsC1OrDEL(r) {
```

or list SO/SI in Out of scope with the xterm caveat. BS, BEL, VT, and FF I would leave alone — BS overwrites inside its own line, the trade already accepted for `\r`.

cmd/worksession/worksession_recording_test.go
```diff
@@ -121,3 +122,63 @@ func TestRecordingPreview_EmptyRaw(t *testing.T) {
...
+	var buf bytes.Buffer
+	printRecordingContent(&buf, "echo \u202esafe\n")
+	assert.Equal(t, "echo safe\n", buf.String())
+}
+
+func TestPrintRecordingContent_StripsC1Controls(t *testing.T) {
+	// U+009B is an 8-bit CSI: left in, it erases the recorded line and the reviewer
+	// reads what follows instead.
+	var buf bytes.Buffer
+	printRecordingContent(&buf, "reboot\u009b2Krm -rf /\n")
```
junho226, line 162: **Suggestion:** this pins the two-byte UTF-8 form (`C2 9B`), which `IsC1OrDEL` matches after decoding. The form a real 8-bit recording carries is the raw single byte `0x9b`, and `utf8.DecodeRune([]byte{0x9b})` returns `(RuneError, 1)` — the predicate never sees `0x9b`.

It is still safe today, but only because `strings.Map` replaces invalid bytes with U+FFFD. Verified: `printRecordingContent("a\x9bb")` yields `a` + U+FFFD + `b` with no raw `0x9b`, where the old `StripANSIEscapes`-only path let the byte through. `escapeJSONControls` covers the same case with an explicit branch and a comment; here the protection is an implicit stdlib property, so a refactor of this `strings.Map` into a byte loop or a regexp would silently reopen it.

Worth a sibling case built from the raw byte rather than the escape.

utils/output.go
```diff
@@ -70,9 +70,9 @@ func escapeJSONControls(b []byte) []byte {
 	for i := 0; i < len(b); {
 		r, size := utf8.DecodeRune(b[i:])
 		switch {
-		case r == 0x7f || (r >= 0x80 && r <= 0x9f) || unicode.Is(unicode.Cf, r):
+		case IsC1OrDEL(r) || unicode.Is(unicode.Cf, r):
 			out = appendUnicodeEscape(out, r)
-		case r == utf8.RuneError && size == 1 && b[i] >= 0x80 && b[i] <= 0x9f:
+		case r == utf8.RuneError && size == 1 && IsC1OrDEL(rune(b[i])):
```
junho226, line 75: **Nit:** the substitution is inert on this line. `IsC1OrDEL` adds `r == 0x7f` over the `b[i] >= 0x80 && b[i] <= 0x9f` it replaces, but `utf8.DecodeRune([]byte{0x7f})` returns `(U+007F, 1)` and never `RuneError`, so the DEL arm is unreachable inside this branch. Verified.

No change needed — the named predicate reads better than the inline range. Noting only that line 73 is the exact swap and this one is a widening that cannot fire, in case a reader later takes the two as equivalent.

utils/utils_test.go
```diff
@@ -374,6 +374,18 @@ func TestStripFormatChars(t *testing.T) {
 	// Cf carries no control byte, so the control passes leave it behind.
 	assert.Equal(t, "denied approved", StripFormatChars("denied\u202e\u2066 approved"))
 	assert.Equal(t, "abé", StripFormatChars("a\u200dbé"))
+
+	// The bidi controls are the ones that matter: they reorder what a terminal
+	// renders without changing a byte. Every one of them goes.
+	for _, r := range []rune{
+		0x202a, 0x202b, 0x202c, 0x202d, 0x202e, // embedding and override
+		0x2066, 0x2067, 0x2068, 0x2069, // isolates
```
junho226, line 382: **Nit:** the comment above says "Every one of them goes", but Unicode's `Bidi_Control` property has twelve members, not nine — U+061C (ALM), U+200E (LRM), and U+200F (RLM) are the others. All twelve are Cf (verified), so `StripFormatChars` already removes them and the implementation is complete; only the claim is under-inclusive. Adding the three runes to this slice makes the comment true and costs nothing.

cmd/worksession/worksession_timeline.go
junho226, line 219: **Note on the description, not the code:** routing through the shared writer is right, and the escape-rather-than-strip split is the correct call for a machine-readable surface.

One correction to the PR body's "it matches every other `--output json` path in the CLI": `utils.PrintTable`'s JSON branch (`utils/output.go:130`) still uses `json.MarshalIndent`, which escapes HTML by default. So after this change the CLI has two behaviors on `--output json` — this path emits ampersand, less-than and greater-than literally, `PrintTable` emits them escaped. The direction is right; converging `PrintTable` is its own change.

junho226 (commented):
## Summary

This closes the remaining audit-review paths where `IsControlRune` alone let Cf runes through, and it splits the treatment the right way: strip on the way to a terminal, escape on the way to a machine reader. I re-derived the load-bearing claims rather than taking them from the description, and they hold — the ordering claim, the `ansiEscapeRE` final-byte gap, and that the new `websh` cases genuinely fail against the base implementation.

Nothing here blocks merge. What follows is one residual gap in the same harm class the PR just closed, one test that pins the safe encoding rather than the dangerous one, and three nits.

(Every rune below is named in `U+XXXX` form on purpose. A review about bidi overrides should not carry one.)

## Critical issues

None.

## Warnings

### 1. SO / SI (U+000E / U+000F) still reach the terminal from `printRecordingContent`

`cmd/worksession/worksession_recording.go:90-95`

The bare-ESC strip is justified in the comment by `ESC ( 0` switching the charset "for the rest of the screen". `\x0e` (Shift Out) does the same thing by a different route: it invokes G1 into GL, and the effect persists until SI or a reset. It is a C0 control, so the deliberate absence of a C0 pass on this path lets it straight through. Measured on this branch:

```
printRecordingContent("a\x0eb") -> "a\x0eb"   // SO survives
printRecordingContent("a\x0fb") -> "a\x0fb"   // SI survives
```

An honest caveat on reach: xterm in UTF-8 mode ignores the shift functions, so this is not universal. The Linux console, VTE (gnome-terminal), and PuTTY do honor them — the familiar "terminal is garbled after `cat`-ing a binary, fix it with `reset`" is precisely SO leaving the console in the alternate charset. For a CLI whose stated threat model is the reviewer's terminal, that is a plausible enough reviewer environment to decide about rather than leave implicit.

This is a residual gap, not a regression — the old code let it through too. But the PR's own reasoning implies a decision here, and the Out of scope section covers only U+2028/U+2029 and the regex width. Either add the two runes to the predicate, which does not conflict with the "keep `\r` and the line endings" contract (no reviewer needs SO), or name them in Out of scope with the xterm caveat:

```go
if r == 0x1b || r == 0x0e || r == 0x0f || utils.IsC1OrDEL(r) {
```

The other survivors on this path — BS, BEL, VT, FF — I would leave alone. BS overwrites inside its own line, which is the trade already accepted for `\r`.

## Suggestions

### 2. The C1 test pins the encoding that is not the dangerous one

`cmd/worksession/worksession_recording_test.go:158-163`

The C1 rune in that test is written as a Go string escape, which in source is the two-byte UTF-8 form `C2 9B`, and `IsC1OrDEL` matches it after decoding. A recording of a program emitting 8-bit output carries the raw single byte `0x9b`, which is invalid UTF-8: `utf8.DecodeRune([]byte{0x9b})` returns `(RuneError, 1)`, so the predicate never sees `0x9b` at all.

It is still neutralized today, but only because `strings.Map` replaces every invalid byte with U+FFFD:

```
printRecordingContent("a\x9bb")    -> "a" + U+FFFD + "b"   // no raw 0x9b
old path (StripANSIEscapes only)   -> "a\x9bb"             // raw byte survived
```

So the protection on the raw form is an implicit stdlib property, not the visible predicate. `escapeJSONControls` handles the same case with an explicit branch and a comment; this path does not. A later refactor of the `strings.Map` into a byte loop or a regexp would silently reopen it. Worth a case with the raw byte.

Two related notes, neither a defect. That same U+FFFD substitution is a fidelity change on the one path whose stated contract is "shows the recording as it was recorded" — invalid bytes used to pass through verbatim. And it is what makes the raw-byte case safe, so it is load-bearing enough to state somewhere.

### 3. "matches every other `--output json` path" is not quite right

`cmd/worksession/worksession_timeline.go:214-221`

Moving to the shared writer is the right call and the escape-rather-than-strip reasoning is sound. But `utils.PrintTable`'s JSON branch (`utils/output.go:130`) still uses `json.MarshalIndent`, which escapes HTML by default. After this change the CLI carries two HTML-escaping behaviors on `--output json`: `FormatJSON` / `PrintJSONValue` emit ampersand, less-than and greater-than literally, while `PrintTable` still emits them as `&`, `<`, `>`. The direction here is right; the description just overstates where the rest of the CLI already is. Converging `PrintTable` belongs in its own change.

### 4. `IsC1OrDEL` in the RuneError branch widens nothing

`utils/output.go:75`

`IsC1OrDEL` adds `r == 0x7f` over the `b[i] >= 0x80 && b[i] <= 0x9f` it replaces, but `utf8.DecodeRune([]byte{0x7f})` returns `(U+007F, 1)` and never `RuneError`, so the DEL arm is unreachable in that branch. Verified. The named predicate reads better than the inline range so I would keep it — noting only that the two substitutions are not the same kind, and line 73 is the exact one.

### 5. "all nine bidi controls" is twelve

`utils/utils_test.go:378-383`

Unicode's `Bidi_Control` property has twelve members: the nine pinned here plus U+061C (ALM), U+200E (LRM), U+200F (RLM). All twelve are Cf — verified — so `StripFormatChars` already removes them and the implementation is complete. Only the comment's "Every one of them goes" is under-inclusive. Three more runes in the slice makes it true, and it costs nothing.

## Highlights

- The ordering is load-bearing, exactly as claimed. Measured: `StripANSIEscapes(StripFormatChars(...))` on `\x1b[2` + U+200D + `Kls` gives `ls`, and reversed it gives `2Kls`. Naming it once as `StripFormatAndANSI` is the right way to stop it drifting back.
- The `ansiEscapeRE` gap is real and the fix is correctly scoped. Measured: `ESC ( 0`, `ESC 7` / `ESC 8`, and `ESC # 3` all survive the regex, and the ESC strip removes the introducer while leaving the tail as inert text — the same trade `StripControlChars` already makes.
- Coverage of the surface is complete: no non-test call site outside `utils/utils.go` still calls `StripANSIEscapes` directly. Every remaining one goes through a format-aware helper.
- The new `websh` cases are genuine regression tests, not decoration. Against the base sources `TestSanitizeRecord` fails, its actual values still carrying the U+202E override and the U+200D joiner that this branch removes.
- Escaping rather than stripping on `--output json` keeps that surface lossless for decoders while still safe to `cat`. That split is the right instinct and it earns the extra explanation the description gives it.
- The Out of scope section characterizes U+2028/U+2029 correctly — verified `Zl` / `Zp`, neither `Cf` nor `Cc`, so no pass in this PR could reach them.
- Threading `io.Writer` through both print helpers is what makes the header and content paths testable at all, and it is done without disturbing the callers.

## Verification

Run in a clean worktree at `7c58b0b`, Go 1.25.12 darwin/amd64:

| Command | Result |
|---|---|
| `go build ./...` | pass, no output |
| `go vet ./...` | pass, no output |
| `gofmt -l .` | pass, no files listed |
| `go test -race -count=1 ./...` | pass, 38 packages ok, 0 failures |
| base-revert check (base sources + branch tests, `go test ./cmd/... ./utils/...`) | `TestSanitizeRecord` fails as expected. `cmd/worksession` does not compile against the base sources because the tests use the new `io.Writer` signatures, so those cases were confirmed by probe and analysis instead. Worktree restored clean afterwards. |
| standalone stdlib probe with `ansiEscapeRE` copied verbatim | confirmed the ordering claim, the regex final-byte gap, the surviving C0 set, the invalid-UTF-8 to U+FFFD behavior, and the twelve-member `Bidi_Control` set |

Could not run: `golangci-lint run ./...` — not installed on this machine (`which golangci-lint` finds nothing). The description's "0 issues" is unverified here; `go vet` is clean, which is the weaker check.

## Verdict

**COMMENT.** The change is a clear net improvement and I found nothing that must block merge. Leaving it short of an approval for two reasons only: the SO/SI gap is a live decision the author should make explicitly — strip or document — since it is the same harm class this PR just closed, and I could not run the lint the description reports clean. Both are cheap to settle.

## Checklist

- [x] Security: threat model stated and matched — strip toward the terminal, escape toward the decoder
- [x] Security: bidi coverage verified against the full `Bidi_Control` set, not just the nine pinned
- [x] Security: no path left running the old `StripANSIEscapes` + `StripControlChars` pair
- [ ] Security: the SO/SI decision on the C0-preserving path is open
- [x] Go: errors handled, no unchecked type assertions, `_, _ =` on `Fprint*` matches the file's existing style
- [x] Go: `strings.Map` and `unicode.Is(unicode.Cf, ...)` used correctly; category used where a category is right, no hand-rolled table
- [x] Tests: genuine regression tests, verified failing against base where the package still builds
- [ ] Tests: the raw-byte C1 form is not covered
- [x] Build, vet, gofmt, and race tests all pass
- [ ] Lint not runnable in this environment
- [x] Comments carry the reasoning rather than restating the code

cmd/worksession/worksession_recording.go
jisung-02: Widened the predicate — 935347a. SO/SI now go the way of the bare ESC, and the comment carries the xterm-vs-Linux-console caveat so the reach question does not get re-litigated later. TestPrintRecordingContent_StripsShiftFunctions pins it (fails on the old code with the SO/SI bytes intact). BS, BEL, VT, and FF stay, per your read — they act inside their own line, the trade already accepted for `\r`. The plain-text error path turned out to be the same harm class on a different surface while checking this, filed as #324.

cmd/worksession/worksession_recording_test.go
jisung-02: Added — a24939d. TestPrintRecordingContent_StripsRawC1Byte feeds the raw `0x9b` and pins the output down to the U+FFFD it becomes, so a rewrite of the strings.Map into a byte loop breaks the test instead of the protection. The commit body also states the fidelity change you noted: the U+FFFD substitution means invalid bytes no longer reach the screen verbatim on the one path whose contract is "as it was recorded".

utils/output.go
jisung-02: Agreed, and keeping it as is — the DEL arm cannot fire inside the RuneError branch since `utf8.DecodeRune([]byte{0x7f})` never returns RuneError. The named predicate stays for readability; treating line 73 as the exact swap and this one as an unreachable widening is the right reading.

utils/utils_test.go
jisung-02: Fixed — bbf708c. U+061C, U+200E, and U+200F join the slice and the comment now says "All twelve members of Bidi_Control go", which is exactly as wide as the test.

cmd/worksession/worksession_timeline.go
jisung-02: Right — the PR body overstated it. Rewrote that sentence to say the opposite: `PrintTable`'s JSON branch still goes through `json.MarshalIndent`, so the CLI carries two HTML-escaping behaviors on `--output json` until that converges. Filed the convergence (together with the `PrintJson` `\n`/`\t` un-escape, which turned up in the same sweep) as #325.

junho226 (approved):
## Round 2

Re-reviewed at `cc633bb` against my round-1 comment at `7c58b0b` (review 4914913775). Every finding from that round is settled, and the four new commits introduce nothing that blocks.

| # | Round-1 finding | Status |
|---|---|---|
| 1 | SO / SI (U+000E / U+000F) still reach the terminal from `printRecordingContent` | **Fixed** — 935347a |
| 2 | The C1 test pins the two-byte form, not the raw `0x9b` | **Fixed** — a24939d |
| 3 | "matches every other `--output json` path" overstates where the CLI is | **Fixed** in the description; convergence filed as #325 |
| 4 | `IsC1OrDEL`'s DEL arm is unreachable in the `RuneError` branch | **Withdrawn** — we read it the same way, no change wanted |
| 5 | "all nine bidi controls" is twelve | **Fixed** — bbf708c |

**1 — fixed.** The predicate now reads `r == 0x1b || r == 0x0e || r == 0x0f || utils.IsC1OrDEL(r)`, and the comment carries the xterm-versus-Linux-console caveat so the reach question is settled in the source instead of being re-litigated later. Measured on this head: `printRecordingContent("a\x0eb\x0fc\n")` gives `"abc\n"`, while `\r`, BS, BEL, VT and FF all still round-trip unchanged (`"a\bb\ac\vd\fe\n"`) — exactly the split we agreed on, with `TestPrintRecordingContent_StripsShiftFunctions` pinning it. The stderr path you turned up while checking this is the same harm class on a different surface, and #324 is the right home for it rather than this diff.

**2 — fixed.** `TestPrintRecordingContent_StripsRawC1Byte` feeds the raw byte and pins the output down to the U+FFFD it becomes, which is the shape I asked for: a rewrite now breaks the test rather than the protection. One correction to where its comment sends the next reader — nit below.

**3 — fixed.** The description now says the opposite of what it did, and the new sentence is accurate: `PrintTable`'s JSON branch still runs `json.MarshalIndent`, so two HTML-escaping behaviors coexist until that converges. Bundling it with the `PrintJson` escape mismatch in #325 is the right scope.

**4 — withdrawn.** Nothing to do. Keeping the named predicate over the inline range is the better trade, and treating line 73 as the exact swap and line 75 as an unreachable widening is the reading I meant.

**5 — fixed.** Verified against Go 1.25.12's tables rather than by reading: `unicode.Is(unicode.Bidi_Control, ·)` matches exactly twelve runes across the entire code-point space — U+061C, U+200E, U+200F, U+202A–U+202E, U+2066–U+2069 — and the test's slice now pins all twelve with zero misses. "All twelve members of `Bidi_Control` go" is now exactly as wide as the test underneath it.

## New findings

### Nit — the tripwire comment names the wrong `strings.Map`

`cmd/worksession/worksession_recording_test.go:167-169`

The comment says the U+FFFD substitution is what keeps the raw byte off the terminal, "so a rewrite into a byte loop reopens this". The property is real, but the `strings.Map` that provides it is not the one directly below the comment. It is `StripFormatChars` (`utils/utils.go:510`), which runs first inside `StripFormatAndANSI` and has already made the substitution before `printRecordingContent`'s own map sees anything.

Measured on this head:

```
StripFormatAndANSI("reboot\x9b2Krm -rf /\n")     -> "reboot�2Krm -rf /\n"   // already substituted
printRecordingContent's map rewritten as a byte loop -> "reboot�2Krm -rf /\n"   // still safe
```

So the refactor the comment warns against would pass, and the one that genuinely reopens the gap — a byte loop replacing `StripFormatChars`'s `strings.Map` — is the one not named. The test itself is correct and does guard the property end to end: it fails if `StripFormatChars` stops going through `strings.Map`. This is only about where the comment points the reader, which for a comment whose entire job is to be a tripwire is the part that has to be right. Naming `StripFormatChars` would fix it.

While you are on the line: the name says `Strips`, but the byte is substituted rather than stripped. The assertion spells out the U+FFFD so nobody is actually misled — worth folding in only if you touch it anyway.

## Highlights

- The SO/SI fix landed as the predicate change rather than an Out of scope paragraph, which is the harder and better of the two options I offered. The comment now explains the C0-preserving contract, the `ansiEscapeRE` final-byte gap, and the shift functions in one place, while leaving DEL and the C1 block to `IsC1OrDEL`'s own doc comment instead of restating it — the right call now that the predicate is named.
- `TestPrintRecordingContent_StripsRawC1Byte` pins the byte-exact output rather than asserting "no `0x9b` present". That is what turns an implicit stdlib property into something a future change trips over.
- Both new tests fail against the pre-fix code for the stated reason, not incidentally, which is what makes them regression tests rather than decoration.
- cc633bb trimming the comments back after the fixes landed is easy to skip and worth doing; the surviving text carries reasoning only, with no line restating its own code.
- The follow-ups that did not belong here went to #324 and #325 with the scope boundary stated, instead of quietly widening this diff.

## Verification

Worktree at `cc633bb` via `gh pr checkout 323`, Go 1.25.12 darwin/amd64.

| Command | Result |
|---|---|
| `go build ./...` | pass, exit 0, no output |
| `go vet ./...` | pass, exit 0, no output |
| `gofmt -l .` | pass, no files listed |
| `go test -race -count=1 -timeout 9m ./...` | **pass** — exit 0, 38 packages `ok`, no `FAIL`, no `DATA RACE` |
| `golangci-lint run ./...` | **pass** — `0 issues.`, exit 0 |
| standalone stdlib probe | confirmed the U+FFFD origin and the byte-loop counter-case above, the twelve-member `Bidi_Control` set with zero misses against the test's slice, and the surviving C0 set (`\r`, BS, BEL, VT, FF) |

On the lint: last round I reported it not runnable here. It was still not installed, so I installed v2.12.2 (`go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest`, built with go1.25.12) and ran it against the repo's own `.golangci.yml`. `.github/workflows/lint.yaml` uses `golangci-lint-action@v9` without pinning a version, so CI may resolve a different v2 patch — but the description's "0 issues" now reproduces here rather than being taken on trust. Nothing I could not run.

The probe lived outside the repository and the worktree is clean afterwards (`git status --short` empty).

## Verdict

**APPROVE.** Both items that held the last round back are closed: the SO/SI decision was made explicitly in the source with the reach caveat recorded, and the lint I could not run then now reproduces clean here. The one new item is a comment pointing at the wrong function — worth fixing on the way past, not worth another round.

## Checklist

- [x] Security: SO/SI closed on the C0-preserving path; the kept set (`\r`, BS, BEL, VT, FF) is deliberate and pinned
- [x] Security: bidi coverage measured against the full twelve-member `Bidi_Control` property, zero misses
- [x] Security: raw `0x9b` now covered by a test rather than by an implicit stdlib property
- [x] Go: errors handled, no unchecked type assertions, `strings.Map` and `unicode.Is` used correctly
- [x] Tests: new cases fail against the pre-fix code for the stated reason
- [x] Build, vet, gofmt, and the full race suite pass
- [x] Lint runs clean here (installed this round; CI patch version may differ)
- [ ] One comment names the wrong `strings.Map` as the load-bearing one (nit)

PR #337, opened 2026-08-13
fix(utils): make SplitPath total instead of panicking without a colon

### LINKED ISSUE #329: fix(utils): SplitPath panics on input without a colon
## Background

`utils.SplitPath` panics with an index-out-of-range on any input that contains no colon:

```go
func SplitPath(path string) (string, string) {
	parts := strings.SplitN(path, ":", 2)
	return parts[0], parts[1]
}
```

(`utils/utils.go:593-596`)

`strings.SplitN` returns a one-element slice when the separator is absent, so `parts[1]` crashes the process instead of returning an error.

Safety currently rests on an undocumented cross-package contract. The cmd layer pre-validates—`cmd/ftp/cp.go:78` rewrites `user@host:path` into `host:path`, and `validatePaths` (`cmd/ftp/cp.go:243`) rejects a remote path with no colon, or with an empty server name or path. Every `SplitPath` call site also sits behind an `isRemotePath` branch, which is a colon check in disguise (`utils.IsRemoteTarget`). So the panic is **not reachable through the CLI today**—it is a latent contract failure, not a live crash.

`api/ftp/ftp.go` calls `SplitPath` directly with no local guard at four sites. Any future or direct caller of `UploadFile`, `UploadFolder`, or `DownloadFile` with an unvalidated path crashes the process instead of getting an error.

## Proposal

Make the function total: return `(string, string, error)` (or an `ok bool`) and update the eight call sites. Add a regression test for the no-colon input. Purely mechanical—no behavior change for valid input.

## References

- `utils/utils.go:593-596`—the panicking implementation
- `cmd/ftp/cp.go:78,243`—the implicit pre-validation the contract relies on
- Call sites (8):
  - `cmd/ftp/cp.go:119,148,269,308`—guarded by the cmd-layer validation above
  - `api/ftp/ftp.go:338,483,786,792`—unguarded

Line numbers are as of 02aa990.

jisung-02:
## Background

`utils.SplitPath` read `parts[1]` out of `strings.SplitN(path, ":", 2)` without checking the result. When the colon is absent the slice holds one element, so the index is out of range and the process dies.

What kept the function safe was pre-validation in another package. `cmd/ftp/cp.go` rewrites `user@host:path` into `host:path`, `validatePaths` rejects a colonless remote path, and every call site sits behind an `isRemotePath` branch — which is a colon check in disguise. So the panic is not reachable through the CLI today. It is an undocumented cross-package contract that has broken, not a live crash.

The four call sites in `api/ftp/ftp.go` carry no such guard. A direct caller of `UploadFile`, `UploadFolder`, or `DownloadFile` with an unvalidated path gets a panic where an error belongs.

## Changes

- `utils.SplitPath` now returns `(string, string, error)` and fails with `invalid remote path %q: missing ':' separator` when there is no colon. An `error` rather than an `ok bool` because the `api/ftp` functions already return `error`, which keeps those call sites at a one-line propagation.
- All eight call sites updated. The four in `api/ftp` propagate the error; the two in `cmd/ftp` that only need the server name for a message exit through `CliErrorWithExit`.
- A follow-up commit makes `uploadObject` and `downloadObject` return the server name. The cmd layer used to split the remote target twice per transfer: once in the command body for the name the error handlers report against, and again inside each helper for its own message. Returning the name drops the command body's copy. Passing `serverName` and `remotePath` down instead would have put five adjacent string parameters on both helpers, and `cmd/ftp/cp_test.go` covers neither of them — a swapped pair would compile and pass.
- `api/ftp` still parses `dest` on its own. That exported boundary is what this fix was about, so it keeps its own parse rather than trusting a caller.

Behavior for valid input is unchanged. Only colonless input moves from a panic to an error.

## Testing

`TestSplitPath` in `utils/utils_test.go` covers six cases: server and path, a colon inside the path, an empty path after the colon, an empty server before the colon, no colon, and empty input. The last two are the regression guards. The error cases assert that both the fixed `missing ':' separator` phrase and the offending input appear in the message.

- `go test -race ./... -count=1` — no failures
- `gofmt -l .` — no output

Closes #329

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -590,9 +590,13 @@ func CreateAndEditTempFile(data []byte) (string, error) {
 	return tmpl.Name(), nil
 }
 
-func SplitPath(path string) (string, string) {
+// SplitPath splits a "server:/path" target into its server name and remote path.
+func SplitPath(path string) (string, string, error) {
 	parts := strings.SplitN(path, ":", 2)
-	return parts[0], parts[1]
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("invalid remote path %q: missing ':' separator", path)
+	}
+	return parts[0], parts[1], nil
 }
 
 // IsInteractiveShell checks if the current program is running in an interactive terminal.
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -591,12 +591,17 @@ func CreateAndEditTempFile(data []byte) (string, error) {
 }
 
 // SplitPath splits a "server:/path" target into its server name and remote path.
+// It returns an error when path carries no ':' separator or no server name. An
+// empty remote path is valid and means the user's home directory.
 func SplitPath(path string) (string, string, error) {
-	parts := strings.SplitN(path, ":", 2)
-	if len(parts) != 2 {
+	server, remotePath, found := strings.Cut(path, ":")
+	if !found {
 		return "", "", fmt.Errorf("invalid remote path %q: missing ':' separator", path)
 	}
-	return parts[0], parts[1], nil
+	if server == "" {
+		return "", "", fmt.Errorf("invalid remote path %q: missing server name", path)
+	}
+	return server, remotePath, nil
 }
 
 // IsInteractiveShell checks if the current program is running in an interactive terminal.
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -591,8 +591,8 @@ func CreateAndEditTempFile(data []byte) (string, error) {
 }
 
 // SplitPath splits a "server:/path" target into its server name and remote path.
-// It returns an error when path carries no ':' separator or no server name. An
-// empty remote path is valid and means the user's home directory.
+// It returns an error when path carries no ':' separator or no server name.
+// An empty remote path is accepted; callers decide whether it is meaningful.
 func SplitPath(path string) (string, string, error) {
 	server, remotePath, found := strings.Cut(path, ":")
 	if !found {
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -351,22 +351,21 @@ func TestSplitPath(t *testing.T) {
 		input      string
 		wantServer string
 		wantPath   string
-		wantErr    bool
+		wantErrMsg string // empty means no error expected
 	}{
-		{"server and path", "myserver:/home/user/file.txt", "myserver", "/home/user/file.txt", false},
-		{"path with colon", "myserver:/tmp/a:b", "myserver", "/tmp/a:b", false},
-		{"empty path after colon", "myserver:", "myserver", "", false},
-		{"empty server before colon", ":/tmp/file", "", "/tmp/file", false},
-		{"no colon", "localfile.txt", "", "", true},
-		{"empty input", "", "", "", true},
+		{"server and path", "myserver:/home/user/file.txt", "myserver", "/home/user/file.txt", ""},
+		{"path with colon", "myserver:/tmp/a:b", "myserver", "/tmp/a:b", ""},
+		{"empty path after colon", "myserver:", "myserver", "", ""},
+		{"empty server before colon", ":/tmp/file", "", "", "missing server name"},
+		{"no colon", "localfile.txt", "", "", "missing ':' separator"},
+		{"empty input", "", "", "", "missing ':' separator"},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			server, path, err := SplitPath(tt.input)
-			if tt.wantErr {
-				require.Error(t, err)
-				assert.Contains(t, err.Error(), "missing ':' separator")
-				assert.Contains(t, err.Error(), fmt.Sprintf("%q", tt.input))
+			if tt.wantErrMsg != "" {
+				assert.ErrorContains(t, err, tt.wantErrMsg)
+				assert.ErrorContains(t, err, fmt.Sprintf("%q", tt.input))
 			} else {
 				require.NoError(t, err)
 			}
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -344,6 +344,37 @@ func TestSplitAndTrim(t *testing.T) {
 	}
 }
 
+func TestSplitPath(t *testing.T) {
+	tests := []struct {
+		name       string
+		input      string
+		wantServer string
+		wantPath   string
+		wantErr    bool
+	}{
+		{"server and path", "myserver:/home/user/file.txt", "myserver", "/home/user/file.txt", false},
+		{"path with colon", "myserver:/tmp/a:b", "myserver", "/tmp/a:b", false},
+		{"empty path after colon", "myserver:", "myserver", "", false},
+		{"empty server before colon", ":/tmp/file", "", "/tmp/file", false},
+		{"no colon", "localfile.txt", "", "", true},
+		{"empty input", "", "", "", true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			server, path, err := SplitPath(tt.input)
+			if tt.wantErr {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), "missing ':' separator")
+				assert.Contains(t, err.Error(), tt.input)
+			} else {
+				require.NoError(t, err)
+			}
+			assert.Equal(t, tt.wantServer, server)
+			assert.Equal(t, tt.wantPath, path)
+		})
+	}
+}
+
 func TestStripANSIEscapes(t *testing.T) {
 	tests := []struct {
 		name  string
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	osexec "os/exec"
@@ -365,7 +366,7 @@ func TestSplitPath(t *testing.T) {
 			if tt.wantErr {
 				require.Error(t, err)
 				assert.Contains(t, err.Error(), "missing ':' separator")
-				assert.Contains(t, err.Error(), tt.input)
+				assert.Contains(t, err.Error(), fmt.Sprintf("%q", tt.input))
 			} else {
 				require.NoError(t, err)
 			}
--- a/api/ftp/ftp.go
+++ b/api/ftp/ftp.go
@@ -335,7 +335,10 @@ func executeBulkUpload(ac *client.AlpaconClient, request *BulkUploadRequest, fil
 // Uses the single upload API for one file, or the bulk API for multiple files.
 // workSessionID is optional; when non-empty it is attached to the request body.
 func UploadFile(ac *client.AlpaconClient, src []string, dest, username, groupname string, allowOverwrite bool, workSessionID string) error {
-	serverName, remotePath := utils.SplitPath(dest)
+	serverName, remotePath, err := utils.SplitPath(dest)
+	if err != nil {
+		return err
+	}
 
 	serverID, err := server.GetServerIDByName(ac, serverName)
 	if err != nil {
@@ -480,7 +483,10 @@ func createFolderZipTempFile(folderPath string) (*os.File, int64, error) {
 // Uses the single upload API for one folder, or the bulk API for multiple folders.
 // workSessionID is optional; when non-empty it is attached to the request body.
 func UploadFolder(ac *client.AlpaconClient, src []string, dest, username, groupname string, allowOverwrite bool, workSessionID string) error {
-	serverName, remotePath := utils.SplitPath(dest)
+	serverName, remotePath, err := utils.SplitPath(dest)
+	if err != nil {
+		return err
+	}
 
 	// Folder uploads always target a directory; ensure trailing slash so the
 	// server recognises the path as a directory.
@@ -783,13 +789,19 @@ func DownloadFile(ac *client.AlpaconClient, sources []string, dest, username, gr
 		return fmt.Errorf("no source paths provided")
 	}
 
-	serverName, firstPath := utils.SplitPath(sources[0])
+	serverName, firstPath, err := utils.SplitPath(sources[0])
+	if err != nil {
+		return err
+	}
 
 	// Extract remote paths and validate all sources are on the same server
 	remotePaths := make([]string, 0, len(sources))
 	remotePaths = append(remotePaths, strings.Trim(firstPath, "\""))
 	for _, src := range sources[1:] {
-		name, p := utils.SplitPath(src)
+		name, p, err := utils.SplitPath(src)
+		if err != nil {
+			return err
+		}
 		if name != serverName {
 			return fmt.Errorf("all sources must be on the same server (got %q and %q)", serverName, name)
 		}
--- a/cmd/ftp/cp.go
+++ b/cmd/ftp/cp.go
@@ -116,8 +116,12 @@ Requires an active WorkSession when using Browser login (Auth0); Token auth (API
 		}
 
 		if isLocalPaths(sources) && isRemotePath(dest) {
-			serverName, _ := utils.SplitPath(dest)
-			err := uploadObject(alpaconClient, sources, dest, username, groupname, recursive, allowOverwrite, workSessionID)
+			serverName, _, err := utils.SplitPath(dest)
+			if err != nil {
+				utils.CliErrorWithExit("%s", err)
+				return
+			}
+			err = uploadObject(alpaconClient, sources, dest, username, groupname, recursive, allowOverwrite, workSessionID)
 			if err != nil {
 				err = utils.HandleCommonErrors(err, serverName, utils.ErrorHandlerCallbacks{
 					OnMFARequired: func(srv string) error {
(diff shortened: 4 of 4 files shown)


utils/utils_test.go
```diff
@@ -344,6 +344,37 @@ func TestSplitAndTrim(t *testing.T) {
...
+		{"empty input", "", "", "", true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			server, path, err := SplitPath(tt.input)
+			if tt.wantErr {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), "missing ':' separator")
+				assert.Contains(t, err.Error(), tt.input)
+			} else {
```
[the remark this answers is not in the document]
jisung-02: Fixed in 3ee9099. The error cases now assert `fmt.Sprintf("%q", tt.input)` instead of the raw input, so the "empty input" case has a real subject: the message must contain `""`. Dropping the `%q` from the format string would now fail that case.

utils/utils.go
```diff
@@ -590,9 +590,13 @@ func CreateAndEditTempFile(data []byte) (string, error) {
 	return tmpl.Name(), nil
 }
 
-func SplitPath(path string) (string, string) {
+// SplitPath splits a "server:/path" target into its server name and remote path.
+func SplitPath(path string) (string, string, error) {
```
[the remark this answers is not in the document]
jisung-02: Leaving the wording as is. The comment names what the function splits, not what it validates -- there is no accuracy gap to close: `":/tmp/file"` and `"myserver:"` are both split into a server name and a remote path exactly as stated, one of them just being empty.

Spelling out the missing-separator error would repeat two things already visible at the call site: the `error` return in the signature, and the message itself, which names the condition verbatim. An earlier review pass on this branch flagged that longer wording for exactly that reason, and the one-line form is what it settled on.

utils/utils.go
```diff
@@ -590,9 +590,13 @@ func CreateAndEditTempFile(data []byte) (string, error) {
...
 	parts := strings.SplitN(path, ":", 2)
-	return parts[0], parts[1]
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("invalid remote path %q: missing ':' separator", path)
+	}
+	return parts[0], parts[1], nil
 }
```
geunwoonoh, line 605: **Warning**: the missing colon is rejected now, but an empty server name still passes — and that one resolves to an arbitrary server.

`":/tmp/file"` returns `("", "/tmp/file")` with no error. Feed it to `api/ftp.UploadFile` and it calls `server.GetServerIDByName(ac, "")`. `utils.BuildURL` puts empty values into the query (`q.Set("name", "")`), so `?name=` goes out, and `GetServerIDByName` returns `response.Results[0].ID` (`api/server/server.go:119`). If the server ignores an empty filter — django-filter's default — the transfer lands on whichever server comes first in the list. If the filter does apply, the caller gets a misleading "no server found with the given name". Neither tells the user the path was malformed, and the first is worse than the panic this PR removes. I could not verify the server-side filter behaviour from this repo, so please confirm.

`validatePaths` already rejects an empty server name in the cmd layer (`cmd/ftp/cp.go:252`), which makes this the remaining half of the gap this PR closes for direct `api/ftp` callers. An empty *path* is worth keeping valid — scp reads `host:` as the remote home directory.

**Suggestion** — this bundles three things: the empty-server guard, `strings.Cut` in place of `SplitN` + `len` (the repo already uses it in `cmd/worksession/worksession_timeline.go:239` and `cmd/exec/parse.go:182`), and the failure condition in the doc comment. Happy to split them if you only want part.

```suggestion
// SplitPath splits a "server:/path" target into its server name and remote path.
// It returns an error when path carries no ':' separator or no server name. An
// empty remote path is valid and means the user's home directory.
func SplitPath(path string) (string, string, error) {
	server, remotePath, found := strings.Cut(path, ":")
	if !found {
		return "", "", fmt.Errorf("invalid remote path %q: missing ':' separator", path)
	}
	if server == "" {
		return "", "", fmt.Errorf("invalid remote path %q: missing server name", path)
	}
	return server, remotePath, nil
}
```

utils/utils_test.go
```diff
@@ -344,6 +345,37 @@ func TestSplitAndTrim(t *testing.T) {
...
+		name       string
+		input      string
+		wantServer string
+		wantPath   string
+		wantErr    bool
+	}{
+		{"server and path", "myserver:/home/user/file.txt", "myserver", "/home/user/file.txt", false},
+		{"path with colon", "myserver:/tmp/a:b", "myserver", "/tmp/a:b", false},
+		{"empty path after colon", "myserver:", "myserver", "", false},
+		{"empty server before colon", ":/tmp/file", "", "/tmp/file", false},
```
geunwoonoh: **Warning** (follow-up to the `utils.go` comment): this case is what pins the empty server name down as valid, so it is the line to flip if you take that guard.

No mechanical suggestion here because the two error paths would then carry different messages, and the assertions below assert one fixed phrase for every error case. The smallest way through is a per-case expected substring:

```go
tests := []struct {
	name       string
	input      string
	wantServer string
	wantPath   string
	wantErrMsg string // empty means no error expected
}{
	{"server and path", "myserver:/home/user/file.txt", "myserver", "/home/user/file.txt", ""},
	{"path with colon", "myserver:/tmp/a:b", "myserver", "/tmp/a:b", ""},
	{"empty path after colon", "myserver:", "myserver", "", ""},
	{"empty server before colon", ":/tmp/file", "", "", "missing server name"},
	{"no colon", "localfile.txt", "", "", "missing ':' separator"},
	{"empty input", "", "", "", "missing ':' separator"},
}
```

Note that `wantPath` for the empty-server case also flips to `""`, since the function would now return before it has anything to hand back.

utils/utils_test.go
```diff
@@ -344,6 +345,37 @@ func TestSplitAndTrim(t *testing.T) {
...
+		{"empty input", "", "", "", true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			server, path, err := SplitPath(tt.input)
+			if tt.wantErr {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), "missing ':' separator")
+				assert.Contains(t, err.Error(), fmt.Sprintf("%q", tt.input))
+			} else {
```
geunwoonoh: **Suggestion**: `assert.ErrorContains` over `assert.Contains(t, err.Error(), ...)`.

`CLAUDE.md` asks for the dedicated helper rather than a predicate, and this is the same trade — `err.Error()` flattens the chain, so a failure prints a bare string instead of the error. `cmd/login_test.go` already uses `assert.ErrorContains` in more than ten places. `ErrorContains` also fails on a nil error, so the `require.Error` above it becomes redundant.

```suggestion
			if tt.wantErr {
				assert.ErrorContains(t, err, "missing ':' separator")
				assert.ErrorContains(t, err, fmt.Sprintf("%q", tt.input))
			} else {
```

If you take the empty-server guard, the first substring comes from the per-case field instead — see the comment on the table above.

geunwoonoh (commented):
## Review

**Stack**: golang (no `alpacax` frontmatter in `CLAUDE.md` — inferred from the changed files)
**Changes**: +79 −25 in 4 files

### Verification (run locally on `3ee9099`)

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l .` | no output |
| `go test -race ./... -count=1` | every package passes, 0 failures |
| `golangci-lint run ./...` | not verified — the binary is not installed in my environment; relying on CI and the PR description |

### Summary

`utils.SplitPath` read `parts[1]` out of `strings.SplitN(path, ":", 2)` without checking the result, so colonless input killed the process. The function now returns `(string, string, error)` and all eight call sites are updated: the four in `api/ftp` propagate, the two in `cmd/ftp` exit through `CliErrorWithExit`. A follow-up commit has `uploadObject`/`downloadObject` return the server name, which drops the command body's duplicate parse.

The design choices are sound and the PR description carries their rationale — `error` over an `ok bool` because the `api/ftp` functions already return `error`, and returning the name rather than passing `serverName`/`remotePath` down to avoid five adjacent string parameters on both helpers. It is also worth noting that the description states plainly that the panic is unreachable through the CLI today instead of overselling the fix.

### Warnings (1)

**1. `utils/utils.go:594` — the missing colon is now rejected, but an empty server name still passes**

The stated purpose of this PR is to make the exported `api/ftp` boundary safe for callers that do not go through the cmd layer's validation. `":/tmp/file"` still returns `("", "/tmp/file")` with no error, and `utils/utils_test.go:359` pins that down as a valid case.

Feed that value to `api/ftp.UploadFile` and it calls `server.GetServerIDByName(ac, "")`. `utils.BuildURL` puts empty values into the query via `q.Set("name", "")`, so `?name=` goes out, and `GetServerIDByName` returns `response.Results[0].ID` (`api/server/server.go:119`). If the server ignores an empty filter — django-filter's default — the transfer targets whichever server comes first in the list. If the filter does apply, the caller gets a misleading "no server found with the given name". Either way the real problem, a malformed path, never reaches the user, and the first outcome is worse than a panic. I could not check the server-side filter behaviour from this repo, so this is worth confirming.

`validatePaths` in the cmd layer already rejects an empty server name (`cmd/ftp/cp.go:252`), which makes this the remaining half of the gap this PR set out to close. An empty *path* is fine to keep accepting — scp reads `host:` as the remote home directory.

Details and a suggested patch are in the inline comment.

### Suggestions (4)

**1. `utils/utils.go:595` — `strings.Cut` instead of `strings.SplitN`**

`SplitN(_, _, 2)` followed by a `len(parts) != 2` check is what `strings.Cut` expresses with a single `found`, and it drops the slice allocation. The call is not hot enough for that to matter as performance; the gain is that the intent reads off the line. The repo already uses `strings.Cut`/`CutPrefix` in `cmd/worksession/worksession_timeline.go:239`, `cmd/exec/parse.go:182`, and `utils/error.go:170`. Folded into the inline suggestion for Warning 1.

**2. `utils/utils_test.go:367-369` — use `assert.ErrorContains`**

`CLAUDE.md` asks for the dedicated helper over a predicate, and `assert.Contains(t, err.Error(), ...)` loses the error chain in the failure output for the same reason. `cmd/login_test.go` already uses `assert.ErrorContains` in more than ten places. Since `ErrorContains` also fails on a nil error, the preceding `require.Error` can go with it. Inline suggestion attached.

**3. `utils/utils.go:593` — the doc comment does not state the new contract**

The one thing callers now have to handle because of this PR is the error, and the comment describes only the success path. A line on the failure condition would match how the other exported functions here document their conditions (`api/ftp/ftp.go:335`: "workSessionID is optional; when non-empty it is attached to the request body"). Folded into the inline suggestion for Warning 1.

**4. `cmd/ftp/cp.go:243-249` — unreachable branch (pre-existing, outside this PR's scope)**

Not part of this change, but flagging it since it sits next to the code being fixed. Inside `if isRemotePath(path)`, the `if !strings.Contains(path, ":")` branch cannot run. `isRemotePath` → `IsRemoteTarget` strips a `user@` prefix before checking for a colon (`utils/ssh_parser.go:39`), and that strip only removes a prefix ending before the colon, so the full string always retains it. The `len(parts) != 2` check on line 252 is dead for the same reason. Now that `SplitPath` owns the missing-colon case, `validatePaths` could keep just `parts[0] == "" || parts[1] == ""` and drop the dead branch along with its error text. Better as a separate issue.

### Checked, no problem found

- **Behavioural equivalence**: the cmd layer used to compute `serverName` itself and now takes it from the helper's return, but the input (`dest` / `sources[0]`) and the function are the same, so the value is identical. The `_, err := uploadObject(...)` inside the `RetryOperation` closures is a valid redeclaration because the closure is a new scope — confirmed by the build.
- **Reachability of the new error**: on the `cmd/ftp` path, `isRemotePath` plus `validatePaths` sit in front, so `return "", err` in `uploadObject`/`downloadObject` is not reachable in practice. Keeping it as defensive code is right.
- **Remaining duplicate parse**: `dest` is still parsed twice per transfer (`cmd/ftp/cp.go:268`, `api/ftp/ftp.go:338`/`:486`). The description gives the reason — the exported boundary validates its own input rather than trusting a caller — and it holds.
- **Performance**: one `SplitN` and one slice allocation per transfer removed from the cmd layer. Nothing else touches a hot path, and `DownloadFile`'s loop preallocates capacity. No regression.
- **Security**: no hardcoded credentials, no injection surface, no sensitive data in logs. `%q` in the error message escaping the offending path is the right call.

### Highlights

- The rationale for returning `error` — the `api/ftp` call sites already return `error`, so propagation stays one line — matches what the diff actually does.
- The table test covers both regression cases (no colon, empty input) plus the colon-inside-path case.
- `3ee9099` catches a self-inflicted weak assertion: the `%q`-less version passed while asserting nothing on the empty-input case.

---

**Verdict**: COMMENTED

Nothing here blocks a merge. Warning 1 is the remaining half of the gap this change targets and the fix is a few lines, so closing it in this PR seems natural. Suggestions 1-3 touch the same function and test, so they fit in one commit.

**Checklist**
- [x] Follows Go idiom and repo conventions (`strings.Cut` and `ErrorContains` are the two places to improve)
- [x] Tests included, full suite verified passing
- [x] No documentation update needed — internal signature change, CLI interface unchanged

utils/utils.go
jisung-02: Fixed in c1e56a5. Took the bundled suggestion whole—empty-server guard, `strings.Cut`, and the doc comment line on the failure conditions.

Also confirmed the server-side behaviour you couldn't check from this repo: `ServerFilter` in alpacon-server declares `name` as a plain `Meta` field, i.e. a default `CharFilter`, and django-filter skips empty values. So `?name=` returns the unfiltered list and `GetServerIDByName` hands back the first server's ID—the worse-than-panic outcome was the real one.

utils/utils_test.go
jisung-02: Fixed in c1e56a5. Took the per-case `wantErrMsg` table as laid out, including `wantPath` flipping to "" for the empty-server case.

utils/utils_test.go
jisung-02: Fixed in c1e56a5. `assert.ErrorContains` with the per-case substring from the table above, and the now-redundant `require.Error` is gone.

utils/utils.go
```diff
@@ -590,9 +590,18 @@ func CreateAndEditTempFile(data []byte) (string, error) {
 	return tmpl.Name(), nil
 }
 
-func SplitPath(path string) (string, string) {
-	parts := strings.SplitN(path, ":", 2)
-	return parts[0], parts[1]
+// SplitPath splits a "server:/path" target into its server name and remote path.
+// It returns an error when path carries no ':' separator or no server name. An
+// empty remote path is valid and means the user's home directory.
```
geunwoonoh: **Nit** (non-blocking, and this is me correcting my own wording — you took the sentence from my suggestion verbatim).

The first sentence states what the function guarantees, which is accurate. The second one, "means the user's home directory", does not hold up: it is scp convention rather than anything this repo establishes, and `validatePaths` rejects `myserver:` (`cmd/ftp/cp.go:252`), so the form is unreachable through `alpacon cp`. As written it could read to a later maintainer as a supported CLI input.

Keeping the comment to what the function itself promises:

```suggestion
// SplitPath splits a "server:/path" target into its server name and remote path.
// It returns an error when path carries no ':' separator or no server name.
// An empty remote path is accepted; callers decide whether it is meaningful.
```

Take it or leave it — not worth another review round by itself.

geunwoonoh (dismissed):
## Re-review of `c1e56a5`

All three inline comments from my previous review are resolved. Approving.

### Verification (run locally on `c1e56a5`)

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l .` | no output |
| `go test -race ./... -count=1` | every package passes, 0 failures |
| `TestSplitPath` | all 6 subtests pass, including the flipped `empty_server_before_colon` |
| `golangci-lint run ./...` | still not verified — the binary is not installed in my environment; please confirm the CI lint job is green before merging |

### Feedback status

| # | Item | Status |
|---|---|---|
| Warning 1 | Empty server name accepted (`utils.go:594`) | **Applied** — `server == ""` guard, as suggested |
| Suggestion 1 | `strings.Cut` over `SplitN` + `len` | **Applied** |
| Suggestion 2 | `assert.ErrorContains`, redundant `require.Error` dropped | **Applied** |
| Suggestion 3 | Document the failure conditions | **Applied** (one nit below) |
| Suggestion 4 | Dead branch in `validatePaths` (out of scope) | **Split into #340** |

Thanks for chasing down the server-side behaviour I could not check from this repo. `ServerFilter` declaring `name` as a default `CharFilter` and django-filter skipping empty values means `?name=` returns the unfiltered list and `GetServerIDByName` hands back the first server's ID — so the arbitrary-server outcome was the real one, not the hypothetical branch. Good to have that closed rather than guessed at.

### Regression check

I checked whether the new guard breaks anything that used to work. `cmd/ftp/cp.go` is the only caller of the four `api/ftp` entrypoints (`:274`, `:276`, `:317`), and `validatePaths` already rejects an empty server name in front of them (`cmd/ftp/cp.go:252`), so no CLI-reachable input newly errors. `strings.Cut` and `SplitN(_, _, 2)` agree on valid input (`"a:b:c"` → `("a", "b:c")`), so the success path is unchanged too.

### Nit (1, non-blocking)

**`utils/utils.go:594-595`** — the second sentence of the doc comment asserts semantics this repo does not establish. That wording came from my own suggestion, so this is me correcting myself: "means the user's home directory" is scp convention, not something verified here, and `validatePaths` actually rejects `myserver:` (`cmd/ftp/cp.go:252`), so the form is unreachable through `alpacon cp`. A future reader could take it as a supported CLI input. Suggested trim in the inline comment — fine either way, and not worth another round on its own.

### Highlights

- Each of the three comments got a reply naming the commit and what was taken, and the diff matches those replies.
- The unverifiable server-side question was researched and answered rather than left open.
- The out-of-scope finding was neither ignored nor smuggled into this PR — #340 carries the background, the cause, and the suggested fix.
- The commit message leads with why the empty server name was dangerous, not with what changed.

---

**Verdict**: APPROVED

The warning is closed, there is no regression path, and the suite passes. What remains is one comment-wording nit and the dead branch already tracked in #340 — neither blocks a merge.

**Checklist**
- [x] Follows Go idiom and repo conventions
- [x] Tests included, full suite verified passing
- [x] No documentation update needed — internal signature change, CLI interface unchanged

utils/utils.go
jisung-02: Fixed in 9c3d4cd. Took the trimmed wording as suggested—the comment now stops at what the function itself promises.

geunwoonoh (approved):

PR #342, opened 2026-08-13
fix(ftp): create cp downloads owner-only

### LINKED ISSUE #326: fix(ftp): cp-downloaded files land world-readable while edit hardens to 0600
## Background

Files downloaded with `alpacon cp` are written through `utils.SaveFile`/`SaveStreamAtomic`, whose default permission is `0666` before umask—typically `0644` on disk (`utils/file_save.go:52`, reached from `api/ftp/ftp.go:592`). The `edit` command already treats the same class of content as sensitive: it downloads into a `0700` temp dir and explicitly chmods the file to `0600` (`cmd/edit/edit.go:256`).

For a PAM client, files pulled from a remote server routinely include private keys, credentials, and other secrets. With the current default, any other local account on the machine can read what `cp` just downloaded.

## Proposal

Create `cp`-downloaded files with `0600` by default, matching the `edit` command's hardening.

One decision to make: `cert download`/`crl download` (`api/cert/cert.go:236-331`) go through the same `utils.SaveFile` path but produce public artifacts where `0644` is fine and arguably preferable. Either keep those on the permissive default and tighten only the ftp path, or add a permission parameter to the save helpers so each caller states its intent.

## References

- `utils/file_save.go:52`—default `0666` perm in the shared save helper
- `api/ftp/ftp.go:592`—`cp` download path using `SaveStreamAtomic`
- `cmd/edit/edit.go:256`—the existing `0600` hardening to match
- `api/cert/cert.go:236-331`—public-artifact downloads sharing the same helper

jisung-02:
## Background

A file downloaded by `alpacon cp` was created with `0666` before the umask, so on the usual `022` it landed at `0644` and every other local account could read it. The files this CLI pulls off a remote host routinely carry key material and credentials, and their protection was left to whatever umask the caller happened to have set.

`edit` already draws this line for the same class of content: it downloads into a `0700` directory and chmods the file to `0600` (`cmd/edit/edit.go:256`). `cp` was the one path without it.

## Changes

`utils.SaveStreamAtomic` is shared, and the certificate, CRL, and package paths that use it write public artifacts that belong at `0644`, so the mode could not be tightened inside the helper. It now takes the mode as a parameter and each caller states its own intent:

```go
func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (int64, error)
```

| Caller | Mode | Effect |
|---|---|---|
| `api/ftp/ftp.go:592` (`fetchFromURLToFile`) | `0600` | the fix |
| `api/packages/packages.go:171` | `0666` | none — the value `os.Create` already used |

`newFilePerm` applies only when the destination does not exist yet. An existing local file keeps its own mode, so a download never re-permissions a file the user configured. That branch predates this change and is unchanged; `TestSaveStreamAtomic_PreservesExistingFileMode` now passes `0666` as the requested mode so the test proves the existing mode wins rather than merely agreeing with it, and asserts the byte count and content so the case cannot pass on a write that never happened.

`fetchFromURLToFile` keeps its signature because all three call sites want `0600`. `ftp.go:639` and `ftp.go:756` write into a temp zip that `os.CreateTemp` already created at `0600`, so they take the existing-mode branch; `ftp.go:655` is the single-file `cp` path this fixes.

A new file takes its mode from `os.OpenFile` at creation rather than a chmod after the write, so it never exists world-readable.

A replacement is now staged the same way. The temp file used to be chmod-ed to the destination's existing mode before the copy, so overwriting a `0644` file held the incoming content at `0644` for the length of the download and a stream that died mid-way left that partial content readable to every local account. It is created at `0600` and widened to the destination's mode once the copy returns, through the open handle so no one can swap the path underneath the chmod, and only then renamed.

## Safety

The umask can only clear bits, so `0600` survives any umask a user would plausibly set. It cannot be widened by one — a caller with `umask 077` gets `0600` here, same as before.

This bounds a local-account read, not root or a workspace superuser: root reads any mode, and a superuser on the Alpacon side already controls what the server serves. The change closes the case where an unprivileged account on the same machine reads a downloaded key.

## Out of scope

Recursive (`cp -r`) and multi-source downloads arrive as a zip and are extracted by `utils.Unzip` with the mode the archive records (`utils/utils.go:428`). That path is left alone here, but it is not already safe: the archive the agent produces carries MS-DOS attributes and no Unix mode, so `zip.File.Mode()` yields `0666` for every entry and each extracted file lands at `0666 & ~umask`, whatever the remote mode was. Measured against a live workspace on this branch:

```
remote /home/e2e/e2e-dir/id_rsa  -rw-------  ->  local  -rw-r--r--
remote /home/e2e/e2e-dir/b.sh    -rwxr-xr-x  ->  local  -rw-r--r--
$ unzip -Z -l <downloaded archive>
-rw----     2.0 fat        4 bl       10 defN 80-000-00 00:00 e2e-dir/a.txt
-rw----     2.0 fat       19 bl       25 defN 80-000-00 00:00 e2e-dir/b.sh
```

So `cp -r` and multi-source `cp` still expose a remote `0600` key at `0644` locally, and the executable bit is lost either way. Closing that means changing what the archive records (agent/server side) or overriding the mode on extraction, both of which affect every extracted file — it belongs in its own issue, filed as #345.

`utils.SaveFile` (the non-atomic path) keeps its signature. Every caller writes a public artifact — certificates, CRLs, CSRs — and the RSA private key goes out through `savePrivateKey`, which already specifies `0600` explicitly (`pkg/cert/cert.go:114`).

The `os.Chmod(result.TempPath, 0600)` in `cmd/edit/edit.go:256` stays. `edit` does download through this path — `DownloadFileToPath` → `fetchFromURLToFile` — so its temp file already arrives at `0600` before the umask. The chmod still earns its place: it is the only thing that restores owner *write* when a umask narrow enough to strip it (`0177`) is set, and without write permission the editor cannot save.

## Testing

- `go test -race -count=1 ./...` — 38 packages ok, no failures
- `gofmt -l .` and `go vet ./...` — no output
- `TestFetchFromURLToFile_CreatesOwnerOnlyFile` covers the `cp` download path end to end; on the old code the file lands at `0644` and the assertion fails on the group and other bits
- `TestSaveStreamAtomic_AppliesRequestedModeToNewFile` pins the helper's new-file contract
- `TestSaveStreamAtomic_KeepsPartialReplacementUnreadable` reads through a stream that dies part-way and stats the temp file from inside that reader, so it sees the mode the partial file actually had rather than the mode at the end; on the previous code it fails with `Should be zero, but was ----r--r--`
- Both new cases assert that no group or other bit is set rather than an exact `0600`, since the umask may narrow the mode further and an exact match would bind the test to the environment. The owner side is covered by the read-back, which a `0000` file would fail
- Both assert the byte count and the file content alongside the mode, so a truncated write cannot pass them
- Both skip on Windows, where Unix modes are not enforced
- Manual `alpacon cp` against a live workspace (self-hosted server on localhost, alpamon agent in a lima VM, remote account `e2e`), all under `umask 022` unless noted:

| Case | Command | Result |
|---|---|---|
| New file | `cp e2e@suse-leap15:/home/e2e/e2e-secret.txt ./case1-new.txt` | `-rw-------`, content intact |
| Same case on `main` | same, binary built from `main` | `-rw-r--r--` — the exposure this PR closes |
| Existing file | destination pre-created at `0644`, then the same `cp` | `-rw-r--r--` kept, content replaced |
| `umask 077` | same download into a fresh path | `-rw-------` |
| `cp -r` | `cp -r e2e@suse-leap15:/home/e2e/e2e-sec ./case5/` | remote `0600` key extracted as `-rw-r--r--` — the zip path, see Out of scope |
| Multi-source | `cp e2e@…/id_rsa e2e@…/b.sh ./case6/` | both `-rw-r--r--` — same zip path |
| Package download | `package python download example-0.1.0-py3-none-any.whl ./pkgs/` | `-rw-r--r--`, unchanged |

The new-file and existing-file rows were measured again after the chmod moved onto the fd, with the same results.

Closes #326


utils/utils_test.go
```diff
@@ -200,11 +200,35 @@ func TestSaveStreamAtomic_RetainsExistingFileOnReadError(t *testing.T) {
...
 	dest := filepath.Join(t.TempDir(), "file.txt")
 	require.NoError(t, os.WriteFile(dest, []byte("existing"), 0600))
 
-	_, err := SaveStreamAtomic(dest, strings.NewReader("replacement"))
+	_, err := SaveStreamAtomic(dest, strings.NewReader("replacement"), 0666)
+	require.NoError(t, err)
+
+	info, err := os.Stat(dest)
 	require.NoError(t, err)
+	assert.Equal(t, os.FileMode(0600), info.Mode().Perm())
```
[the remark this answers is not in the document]
jisung-02: Fixed in 3621dae. The test now captures the returned byte count and reads the file back, so a SaveStreamAtomic that short-circuited on an existing destination would fail on the content assertion rather than pass on the unchanged mode.

README.md
```diff
@@ -158,7 +158,7 @@ $ alpacon cp -u admin -g developers <SOURCE> <DESTINATION>
...
 ```
 
-`<server>:<path>` denotes a remote target. Saving in `edit` overwrites the remote file; ownership and permissions may be reset by server policy. `edit` only opens existing remote files—it downloads first, so it won't create a new one. `--editor` is tokenized without a shell (the file path is appended as the last argument), so shell syntax such as pipes (`|`), redirections (`>>`), or `&&` won't work.
+`<server>:<path>` denotes a remote target. A file `cp` downloads is created owner-only (`0600`), since remote files routinely carry secrets; a local file that already exists keeps its current mode. Recursive downloads and downloads of two or more sources arrive as an archive, and each file is extracted with whatever mode the archive records for it. Saving in `edit` overwrites the remote file; ownership and permissions may be reset by server policy. `edit` only opens existing remote files—it downloads first, so it won't create a new one. `--editor` is tokenized without a shell (the file path is appended as the last argument), so shell syntax such as pipes (`|`), redirections (`>>`), or `&&` won't work.
```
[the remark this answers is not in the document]
jisung-02: Fixed in 400aba6. The sentence now reads "0600 before the umask, which can only narrow it further", so it states the bound that holds rather than an exact mode.

utils/file_save.go
```diff
@@ -89,6 +96,14 @@ func SaveStreamAtomic(fileName string, r io.Reader) (int64, error) {
 		return written, fmt.Errorf("failed to close file: %w", closeErr)
 	}
 
+	// The umask stripped bits from what O_CREATE requested, so an existing mode
+	// needs a chmod to survive the replacement intact.
+	if replacing {
+		if err := os.Chmod(tempName, perm); err != nil {
```
geunwoonoh: Critical-adjacent warning: `os.Chmod` on a path follows symlinks. This was `file.Chmod(existingPerm)` before the change — an `fchmod(2)` on the open fd.

If the destination directory is writable by another local account and carries no sticky bit, that account can unlink `tempName` and drop a symlink of the same name between `file.Close()` on line 91 and this `os.Chmod`. The chmod then lands on whatever the symlink points at. The temp name is fairly guessable (`.alpacon-<pid>-<unixnano>-<i>.tmp`), and `perm` is the destination's mode, so a wide value like `0644` can get stamped onto someone else's file.

The threat model this PR is built around is an unprivileged account on the same machine. Handing that account a fresh path-based operation cuts against the point.

The property the fix needs is "widen after the copy completes", not "after close". Widening through the fd drops the path lookup entirely:

```go
	written, copyErr := io.Copy(file, r)
	// The temp file was staged at 0600 to hide partial content, so the
	// destination's own mode has to be restored before the rename. Widen
	// through the fd so nothing at tempName can be swapped in first.
	var chmodErr error
	if copyErr == nil && replacing {
		chmodErr = file.Chmod(perm)
	}
	closeErr := file.Close()
	if copyErr != nil {
		return written, fmt.Errorf("failed to write file: %w", copyErr)
	}
	if chmodErr != nil {
		return written, fmt.Errorf("failed to set temp file permissions: %w", chmodErr)
	}
	if closeErr != nil {
		return written, fmt.Errorf("failed to close file: %w", closeErr)
	}
```

That replaces lines 90-105, which crosses a gap in the diff, so it is written out rather than offered as a `suggestion` block.

I ran this exact form against `go test -race -count=1 -run 'SaveStreamAtomic|FetchFromURLToFile' ./utils/... ./api/ftp/...` and both packages are ok. `TestSaveStreamAtomic_KeepsPartialReplacementUnreadable` still holds, since it samples the mode while `io.Copy` is running, and the `0644`/`0600` assertions in `PreservesExistingFileMode` pass unchanged. My working tree is back to the branch state.

utils/file_save.go
```diff
@@ -89,6 +96,14 @@ func SaveStreamAtomic(fileName string, r io.Reader) (int64, error) {
 		return written, fmt.Errorf("failed to close file: %w", closeErr)
 	}
 
+	// The umask stripped bits from what O_CREATE requested, so an existing mode
+	// needs a chmod to survive the replacement intact.
```
geunwoonoh: Warning: this was accurate before the change — `O_CREATE` requested `existingPerm` and the umask cut it down. Now `createPerm` is deliberately `0600` (line 74), so the chmod exists because the temp file was staged narrow, not because of the umask. It is needed whether the umask is `0022` or `0000`.

Left as is, the next reader can conclude "the umask is permissive here, so this chmod is dead" and delete it.

The wording in the block proposed on line 102 already covers this. If that change is not taken, at minimum:

```suggestion
	// The temp file was staged at 0600 to hide partial content, so the
	// destination's own mode has to be restored before the rename.
```

utils/utils_test.go
```diff
@@ -200,17 +200,94 @@ func TestSaveStreamAtomic_RetainsExistingFileOnReadError(t *testing.T) {
...
+		info, err := os.Stat(match)
+		if err != nil {
+			return 0, err
+		}
+		r.modes = append(r.modes, info.Mode().Perm())
+	}
+	return r.reader.Read(p)
+}
+
+func TestSaveStreamAtomic_KeepsPartialReplacementUnreadable(t *testing.T) {
```
geunwoonoh, line 251: Suggestion: this test samples the temp file's mode while the write succeeds. The partial case — a stream that dies mid-way — never happens here. It works as a proxy, but the name overstates the coverage.

Either lower the name to `TestSaveStreamAtomic_StagesReplacementOwnerOnly`, or wrap the `failingReader` already in this file with `tempModeReader` so the stream actually aborts. For the second option, `TestSaveStreamAtomic_RetainsExistingFileOnReadError` (line 184) already sets up the abort, so adding the mode assertion there gets the named coverage for free.

utils/file_save.go
```diff
@@ -49,19 +54,27 @@ func SaveStreamAtomic(fileName string, r io.Reader) (int64, error) {
 		return 0, fmt.Errorf("failed to create directories: %w", err)
 	}
 
-	perm := os.FileMode(0666)
-	var existingPerm os.FileMode
+	perm := newFilePerm
```
geunwoonoh: Suggestion: `newFilePerm`, `perm`, and `createPerm` now share one function, and `perm` is the only one whose name does not say its role. It means "the mode the destination ends up with", so `finalPerm` or `destPerm` reads better.

The three-way split itself is right and should stay. It cannot collapse into "always create at 0600, then chmod", because the new-file path has to let the umask apply to `0666` for package downloads to keep behaving as they did. That constraint is not visible in the code — one line of comment would stop someone from trying the collapse later.

utils/file_save.go
```diff
@@ -49,19 +54,27 @@ func SaveStreamAtomic(fileName string, r io.Reader) (int64, error) {
...
 		return 0, fmt.Errorf("failed to access file: %w", err)
 	}
 
-	file, err := createReplacementTempFile(dir, perm)
+	// A replacement is written owner-only and widened to the destination's mode
+	// only once the content is complete, so a download that dies mid-stream
+	// never leaves its partial content readable to anyone else.
+	createPerm := perm
+	if replacing {
+		createPerm = 0600
```
geunwoonoh: Suggestion: each call site states its intent for the mode it passes, but this literal inside the helper is the one bare number left. A named `const stagingPerm = 0600` would make it clear at the declaration that this is a value used only during the write, not a final mode.

utils/utils_test.go
```diff
@@ -200,17 +200,94 @@ func TestSaveStreamAtomic_RetainsExistingFileOnReadError(t *testing.T) {
 }
 
 func TestSaveStreamAtomic_PreservesExistingFileMode(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
geunwoonoh, line 204: Suggestion: this file now has five copies of the same skip — lines 203, 247, 270, 292, 320 — three of them added here. Lines 292 and 320 skip for symlink privileges rather than file modes, so a `requireUnixModes(t)` helper would cover the three mode ones. Optional for this PR under the surgical-changes rule.

api/ftp/ftp.go
```diff
@@ -589,7 +589,8 @@ func fetchFromURLToFile(httpClient *http.Client, url, filePath string, maxAttemp
 
 	defer func() { _ = resp.Body.Close() }()
 
-	return utils.SaveStreamAtomic(filePath, resp.Body)
+	// Remote files routinely carry secrets, so a new download lands owner-only.
+	return utils.SaveStreamAtomic(filePath, resp.Body, 0600)
```
geunwoonoh, line 593: Suggestion (security follow-up, not a blocker): the fix covers file creation, so a `cp` of a key into a fresh path lands at `0600`. An existing destination keeps its mode, so `touch ./id_rsa` under umask `022` followed by this download still lands the key at `0644`.

The PR body and `README.md:161` both state this, and the reasoning is right — narrowing across the board would be wrong, since `alpacon cp server:/etc/nginx/nginx.conf /etc/nginx/nginx.conf` clamped to `0600` breaks nginx.

Still, a README sentence is thin for the one case the issue was about. A single stderr warning when the destination already exists with a group or other read bit would cost almost nothing and would tell the user what just happened. Worth its own issue.

README.md
```diff
@@ -158,7 +158,7 @@ $ alpacon cp -u admin -g developers <SOURCE> <DESTINATION>
...
 ```
 
-`<server>:<path>` denotes a remote target. Saving in `edit` overwrites the remote file; ownership and permissions may be reset by server policy. `edit` only opens existing remote files—it downloads first, so it won't create a new one. `--editor` is tokenized without a shell (the file path is appended as the last argument), so shell syntax such as pipes (`|`), redirections (`>>`), or `&&` won't work.
+`<server>:<path>` denotes a remote target. A file `cp` downloads is created owner-only—`0600` before the umask, which can only narrow it further—since remote files routinely carry secrets; a local file that already exists keeps its current mode. Recursive downloads and downloads of two or more sources arrive as an archive, and each file is extracted with whatever mode the archive records for it. Saving in `edit` overwrites the remote file; ownership and permissions may be reset by server policy. `edit` only opens existing remote files—it downloads first, so it won't create a new one. `--editor` is tokenized without a shell (the file path is appended as the last argument), so shell syntax such as pipes (`|`), redirections (`>>`), or `&&` won't work.
```
geunwoonoh: Suggestion: the sentence runs about 70 words with three em-dash clauses. Splitting the `cp` download mode and the archive extraction mode into two sentences would read better.

One accuracy detail: "each file is extracted with whatever mode the archive records for it" holds for files, but directories created during extraction use `os.ModePerm` at `utils/utils.go:409`, so they get `0777 & ^umask` rather than the archive's recorded mode. You wrote "file", so it is not wrong, but a reader could take it as covering directories too.

geunwoonoh (changes requested):
## Review

`alpacon cp` downloads were created at `0666` before the umask, so on the usual `022` they landed at `0644`. This adds a `newFilePerm` parameter to `utils.SaveStreamAtomic`, passes `0600` from the `cp` path and the pre-existing `0666` from package downloads, and stages a replacement at `0600` so partial content is not exposed mid-download.

The approach is right. Not tightening the mode inside the shared helper and making each caller state its own intent is the correct call; taking the mode from `os.OpenFile` at creation rather than a chmod after the write is the correct mechanism; and closing the partial-replacement window is a real fix. What that last change did, though, was turn an fd-based chmod into a path-based one, and that reopens a door to the exact attacker this PR is bounding.

### Verified locally

- `go test -race -count=1 ./...` — 38 packages ok, no failures
- `go build ./...`, `go vet ./...`, `gofmt -l .` — all clean (golangci-lint not installed locally)
- Checked the `README.md:161` claim about archive extraction modes against `extractFile` at `utils/utils.go:428` (`os.OpenFile(fpath, ..., f.Mode())`) — accurate

### Critical issues (0)

None.

### Warnings (2)

1. **`utils/file_save.go:102`** — `os.Chmod` on a path follows symlinks; the previous code used `file.Chmod` on the open fd. Inline comment has the replacement.
2. **`utils/file_save.go:99`** — the comment attributes the chmod to the umask, but the reason is now the deliberate `0600` staging.

Both sit in the same block, so one edit covers them.

### Suggestions (4)

1. **`utils/utils_test.go:246`** — `TestSaveStreamAtomic_KeepsPartialReplacementUnreadable` never aborts the stream.
2. **`utils/file_save.go:57`** — `perm` is the only one of three mode variables whose name does not say its role.
3. **`utils/file_save.go:74`** — the `0600` literal inside the helper is the one mode left without a stated intent.
4. **`utils/utils_test.go:203`** — five copies of the same Windows skip block.

### Go conventions

No problems. Checked:

- Declaration order `const -> var -> type -> func`. `tempModeReader` sits in its own place at `utils/utils_test.go:225` with its method right after, not wedged between functions
- Doc comment opens with the function name
- errcheck satisfied; deferred closes keep the `defer func() { _ = file.Close() }()` form
- Error strings lowercase, no trailing punctuation, `%w` wrapping
- testify: `require` where later lines depend on it, `assert` where they do not; no `require` inside an httptest handler. `assert.Zero` over `assert.True(t, mode&0077 == 0)`, so a failure prints `----r--r--` instead of `Should be true`
- Import order stdlib -> third-party; `runtime` inserted alphabetically in `api/ftp/ftp_test.go`
- All comments in English

`0600` over `0o600` matches the rest of the repo.

### Duplicate / unused code

Nothing left over from this change. `existingPerm` is fully gone, no unused imports, no dead branch.

Two pre-existing items, noted only as follow-up candidates — not this PR's job:

- `utils/file_save.go:18` — `saveStream` has exactly one caller, `SaveFile`. A single-use indirection
- The non-atomic `SaveFile` coexists with `SaveStreamAtomic`, and the four `api/cert/cert.go` call sites plus `pkg/cert/cert.go:73` still use the non-atomic one. Not urgent, since certs, CRLs, and CSRs are public artifacts

### Performance

Nothing measurable.

- Replacement path: one `fchmod(2)` becomes one `chmod(2)`. The path lookup makes it marginally more expensive in theory, but it runs once per download. Applying warning 1 returns it to `fchmod` and erases even that
- New-file path: no change in syscall count. It never chmods, it only sets the mode at creation
- The `io.Copy` path, buffers, and allocations are untouched

One test-side note: `tempModeReader.Read` runs a `filepath.Glob` plus a `Stat` on every `Read`. With `strings.NewReader("replacement")` that is two calls, so it is harmless now, but swapping in a large stream would run the Glob per 32KB chunk. Sampling only on the first call would keep that from biting later.

One pre-existing inefficiency sits in the archive path this PR marks out of scope: `reserveDownloadArchiveTempPath` (`api/ftp/ftp.go:610`) creates a file with `os.CreateTemp`, then `fetchFromURLToFile` calls `SaveStreamAtomic` on that path, so every recursive or multi-source download creates two temp files and does one extra rename. Existing behavior, unrelated to this change.

### Security

- No hardcoded credentials. `"private key material"` in the test is a literal fixture
- No new log output; neither the mode nor the path is printed anywhere
- Symlink handling is unchanged: `resolveWritePath` walks to the final target and replaces in place. A dangling symlink is treated as a new file and gets `0600`, covered by `utils/utils_test.go:319`
- Remaining hole: an existing destination keeps its own mode, so `touch ./id_rsa` (umask `022` -> `0644`) followed by a `cp` of a key lands it at `0644`. The PR body and `README.md:161` both state this and the reasoning holds — see the inline comment on `api/ftp/ftp.go:593` for the cheap mitigation worth filing

### Documentation

`README.md:161` matches the code. Inline comment covers the wording and one accuracy detail about extracted directories.

---

**Verdict**: CHANGES REQUESTED

Warnings 1 and 2 are one edit. Nothing else blocks the merge. The reason this is CHANGES REQUESTED rather than an approval with a follow-up: warning 1 is not a style point. It sits inside the threat model this PR defines, and the three-line fix is already confirmed to pass the suite.

**Checklist**

- [x] Tests included and passing (38 packages, `-race`)
- [ ] No security vulnerabilities — path-based chmod at `utils/file_save.go:102`
- [x] Documentation updated (splitting the README sentence is advisory)

utils/file_save.go
jisung-02: Applied in 3b74fbe, in the form you wrote out.

The widening now runs on the open fd between `io.Copy` and `Close`, so `tempName` is never resolved a second time and the swap window is gone. Confirmed `file.Chmod` is not Unix-only: `internal/poll` implements `Fchmod` on Windows too (`fd_windows.go:1145`, read-only attribute), so the Windows path behaves as it did before the staging change.

`go test -race -count=1 ./...` is green and `golangci-lint run ./...` reports `0 issues.`. I also re-ran the two live-workspace cases the PR body lists — new file `-rw-------`, existing `0644` file kept at `-rw-r--r--` with its content replaced.

utils/file_save.go
jisung-02: Right, and the failure mode you describe is the one that worried me too. Replaced in 3b74fbe by the comment from the block above:

```go
	// Staging cost the destination its own mode, so restore it before the rename.
	// Through the handle: a path-based chmod here could be pointed at another
	// file by anyone able to swap tempName for a symlink.
```

The umask is no longer mentioned, since it is not why the chmod exists.

utils/utils_test.go
jisung-02: Took the second option in 4223d6b — the stream now actually aborts, so the name is covered rather than lowered:

```go
	reader := &tempModeReader{reader: &failingReader{reader: strings.NewReader("replacement")}, dir: dir}
	_, err := SaveStreamAtomic(dest, reader, 0600)
	require.Error(t, err)
```

It also asserts the destination keeps its content and its `0644` afterwards, so the case pins the whole failure path rather than the mode alone. `TestSaveStreamAtomic_RetainsExistingFileOnReadError` stays as it is: it owns the "existing file survives" contract, and folding a mode assertion in would give it two reasons to fail.

On the `Glob` per `Read` from your review body: I left it. Sampling on every read is the point of the type — it is what would catch a widen that happens mid-stream rather than at the end — and every stream in this file is a single chunk. If a large-stream case ever lands here, sampling deserves a size cap at that point.

utils/file_save.go
jisung-02: Renamed to `finalPerm` in 3b74fbe, and the constraint you point at is now in the code:

```go
	// A replacement is staged narrow and widened once the content is complete, so
	// a download that dies mid-stream never leaves its partial content readable
	// to anyone else. A new file cannot be staged the same way: it would need a
	// chmod that bypasses the umask, and a caller passing 0666 is asking for the
	// umask to apply.
```

utils/file_save.go
jisung-02: Added in 3b74fbe:

```go
// stagingPerm is the mode a replacement is written under, never a final mode:
// it hides partial content from other local accounts until the write completes.
const stagingPerm = 0600
```

The declaration says it is a transient mode, so it cannot be mistaken for a default the callers fall back to.

utils/utils_test.go
jisung-02: Done in 4223d6b. `requireUnixModes(t)` covers the three mode cases; lines 292 and 320 keep their own skip, since they are about symlink privileges and would be wrong to file under a mode helper.

api/ftp/ftp.go
jisung-02: Agreed, and it is the gap that bothers me most about what ships here — the README sentence is thin cover for the exact case the issue named.

Out of scope for this PR, so I have drafted it as a follow-up issue: warn on stderr through `utils.CliWarning` when the destination already exists with a group or other read bit, leaving the behavior alone. It will be filed separately rather than in this branch.

README.md
jisung-02: Split in dd6d69c, and your accuracy note turned out to understate it.

I checked the archive against a live workspace rather than the extraction code alone. alpamon builds it with `zip.Writer.Create` (`pkg/utils/archive.go`), which writes no external attributes, so no entry carries a Unix mode at all:

```
$ unzip -Z -l <archive>
-rw----     2.0 fat        4 bl       10 defN 80-000-00 00:00 e2e-dir/a.txt
-rw----     2.0 fat       19 bl       25 defN 80-000-00 00:00 e2e-dir/b.sh
```

`fat` means MS-DOS attributes, and Go answers `0666` for those. Measured under `umask 022`, a remote `0600` key extracts at `0644` and a `0755` script arrives without its executable bit — so "the mode the archive records" was true of the code and misleading about the outcome. The sentence now states the `0666` and, per your point, the `0777` for directories from `os.ModePerm`.

The PR body carried the same wrong claim as its reason for leaving the archive path alone; that is corrected too, and the exposure is drafted as its own follow-up issue.

api/ftp/ftp.go
jisung-02: Filed as #346.

README.md
jisung-02: Filed as #345.

geunwoonoh (approved):
## Review (follow-up on `2acaced` -> `dd6d69c`)

All eight items from the previous review are applied. I checked the code and the standard library rather than taking the replies at their word.

| # | Location | Finding | Applied | Verified |
|---|---|---|---|---|
| W1 | `utils/file_save.go:96-113` | path-based `os.Chmod` -> fd | `3b74fbe` | `file.Chmod(finalPerm)` sits between `io.Copy` and `Close`, guarded by `copyErr == nil`. `tempName` is now only read by `replaceFile` |
| W2 | `utils/file_save.go:97-99` | comment misattributed the chmod to the umask | `3b74fbe` | umask is gone from the wording; the staging is named as the reason |
| S1 | `utils/utils_test.go:251` | test name promised an abort it never took | `4223d6b` | wraps `failingReader`, `require.Error`, and now also pins the destination's content and `0644` |
| S2 | `utils/file_save.go:61` | `perm` did not say its role | `3b74fbe` | `finalPerm`, plus a comment recording why the three-way split cannot collapse |
| S3 | `utils/file_save.go:15` | bare `0600` literal in the helper | `3b74fbe` | `const stagingPerm = 0600`, declaration order `const -> func` kept |
| S4 | `utils/utils_test.go:202` | five copies of the Windows skip | `4223d6b` | `requireUnixModes(t)` with `t.Helper()`; leaving the two symlink skips alone is the right call |
| S5 | `api/ftp/ftp.go:593` | existing-destination mode is kept | follow-up | #346 open |
| S6 | `README.md:161` | sentence length + directory modes | `dd6d69c` | split, and see below |

### Verified locally

- `go test -race -count=1 ./...` — 38 packages ok, no failures
- `go build ./...`, `go vet ./...`, `gofmt -l .` — all clean
- The claim that `file.Chmod` is not Unix-only holds. `$(go env GOROOT)/src/os/file_posix.go:5` carries the build tag `unix || (js && wasm) || wasip1 || windows`, and `f.chmod` calls `f.pfd.Fchmod`. `internal/poll/fd_windows.go:1145` does implement `Fchmod` against the read-only attribute. The cited file and line are both accurate
- #345 and #346 both exist and are open

### The archive finding turned out worse than what I raised

My README note said extracted directories use `os.ModePerm` rather than the archive's recorded mode. Checking the reply, files do not follow the archive either, and I confirmed it in the standard library.

`FileHeader.Mode()` (`archive/zip/struct.go:313`) switches on `CreatorVersion >> 8`. The agent's archive is `creatorFAT`, matching the `2.0 fat` column in the `unzip -Z -l` output, so it takes `msdosModeToFileMode` (`struct.go:349`), which answers `0666` for a file and `ModeDir | 0777` for a directory whenever the external attributes carry no Unix mode. A remote `0600` key therefore lands at `0644` under umask `022`, and the executable bit is lost.

The PR body's original reason for leaving that path alone — that it already reproduces a `0600` remote key as `0600` locally — was false, and `dd6d69c` corrects it with the measured values. The README now states the `0666` and `0777` outright.

Keeping that out of this PR is the right call: the cause is in the agent's zip writer, so fixing it is either a cross-repo change or an extraction-time mode override, and the override affects every extracted file including its executable bit. #345 is the right home for it.

### Critical issues (0) / Warnings (0) / Suggestions (0)

Nothing new. One thing I checked and dismissed: `chmodErr` is tested before `closeErr`, so a close error is dropped when both fail. Both paths remove the temp file and leave the destination untouched, so the outcome is identical. No action needed.

### Highlights

- W1 takes the proposed form and adds the `copyErr == nil` guard, so a failed write never widens and the temp file stays at `0600`. `TestSaveStreamAtomic_KeepsPartialReplacementUnreadable` now actually walks that path
- Checking the standard library's Windows implementation before answering on `file.Chmod`, with a file and line that match
- Recording in a comment why `createPerm` has to stay split three ways, which stops a later attempt to collapse it
- Pulling a real archive instead of only reading the extraction code, which found a wrong claim in this PR's own body and a larger exposure than the one I raised

---

**Verdict**: APPROVED

Good to merge. Both remaining exposures — #345 for the archive path, #346 for an existing destination — predate this branch, are now stated in the README, and are tracked. Nothing here makes either worse.

**Checklist**

- [x] Tests included and passing (38 packages, `-race`)
- [x] No security vulnerabilities — the chmod runs on the fd, with no second path resolution
- [x] Documentation updated

PR #344, opened 2026-08-13
fix(exec): treat an intent deviation as a pending approval

### LINKED ISSUE #288: fix(exec): sudo denial code list drifted from the server — --wait misses SUDO_INTENT_DEVIATION
## Background

The server ships the reason for a non-interactive sudo denial as a string code, and alpamon-pam's `alpacon_approval.c` prints it to the terminal as a single line.

```
Alpacon denied this sudo command (SUDO_RISK_DENIED).
```

The CLI parses that line to decide what to do next. The code list is hand-copied into `sudoDenialHints` at `cmd/exec/run.go:42-63`, and the comment at line 36 of the same file says so outright.

```go
// Codes are kept in sync with alpacon-server utils/error_codes.py.
```

Nothing enforces that sync, and it has drifted. The CLI knows four codes (`SUDO_NO_WORKSESSION_POLICY`, `SUDO_PRESENCE_REQUIRED`, `SUDO_APPROVAL_REQUIRED`, `SUDO_RISK_DENIED`); the server has added two since.

- `SUDO_INTENT_DEVIATION` — alpacon-server `b7f9a2485` (2026-06-16), `utils/error_codes.py:900`
- `SUDO_POLICY_MFA_REQUIRED` — alpacon-server `1404c4c91` (2026-07-24, #2608), `utils/error_codes.py:878`

## Problem 1: --wait does not wait for an intent-deviation approval request (bug)

The entry condition for `--wait` is `isApprovalDenial` (`cmd/exec/run.go:149-152`), which goes through `hasSudoApprovalDenial` (`run.go:100-102`) and looks at `SUDO_APPROVAL_REQUIRED` only.

But `sudo/services.py:1444-1460` on the server shows that an intent deviation takes the same HITL branch, creates an approval grant, and only swaps the code.

```python
hitl_reason = (
    ErrorCode.SUDO_INTENT_DEVIATION
    if result.intent_deviation
    else ErrorCode.SUDO_APPROVAL_REQUIRED
)
grant = _create_command_layer3_approval(
    iam_user, command_event, data, intent_deviation=result.intent_deviation
)
return _deny_exec_sudo(server, data, command_event, command, hitl_reason, pending_grant=grant)
```

`_deny_exec_sudo` (`sudo/services.py:882-922`) responds to alpamon with `approved=False` even when `pending_grant` is set (ADR 0016 §3: an interactive sudo cannot wait on an out-of-band approval). Only the live event is published as `approval_requested`. So what the CLI sees is the denial line alone, while the approval request really does exist in the background.

The result: even with `--wait`, `isApprovalDenial` is false, so the CLI never enters the wait loop and exits immediately. The approval is in flight but the user reads it as a failure. This is not a missing hint—the flag does not work.

For the same reason `HandlePendingApproval` (`run.go:253`) returns false, so the command exits 1 as a generic failure instead of `ExitCodePendingApproval` (4). CI and agents that branch on exit codes misread it as a non-retryable failure.

The interactive websh sudo path (`_handle_websh_branch`, `sudo/services.py:641-660`) has the same branch. Interactive websh in the CLI uses its own sudo MFA flow, so it is out of scope here, but it still needs to be checked.

## Problem 2: the SUDO_POLICY_MFA_REQUIRED hint is gone

Before the split, the "a policy matched but MFA is required" case also arrived as `SUDO_NO_WORKSESSION_POLICY`, and the CLI pointed the user at `work-session update --sudo`. After the split, `sudoDenialHint` matches nothing and the hint disappeared. The server-side comment states the split was made for the CLI.

```python
# Split from SUDO_NO_WORKSESSION_POLICY so the CLI can tell them apart (#2608).
```

The two situations call for different actions. `SUDO_NO_WORKSESSION_POLICY` means add a policy; `SUDO_POLICY_MFA_REQUIRED` means a policy already exists and an MFA step-up is needed.

## Proposal

1. Add both codes to `sudoDenialHints` (`run.go:42-63`).
   - `SUDO_POLICY_MFA_REQUIRED` — say that a policy matched and MFA is required. It must read differently from the "add a policy" guidance for `SUDO_NO_WORKSESSION_POLICY`.
   - `SUDO_INTENT_DEVIATION` — state that an approval request was created, and lead with the self-service path of re-declaring the session intent rather than waiting for a reviewer (ADR 0016 §4-5). In the CLI, `alpacon work-session update [SESSION_ID] --description` appears to be that mechanism, but the wording should be settled after confirming whether the server actually uses the session description for the intent-alignment check (the `description_aligned` argument).
2. Include `SUDO_INTENT_DEVIATION` in the set of codes recognized by `hasSudoApprovalDenial` / `isApprovalDenial`, so that `--wait` and `HandlePendingApproval` (exit 4) work. Replacing the single-code comparison with a lookup over a code set, and having the hint table and the approval code set read from one source, leaves less room for the next addition to drift again.
3. Consider whether `SUDO_POLICY_MFA_REQUIRED` belongs alongside `SUDO_PRESENCE_REQUIRED` as a code resolvable by a step-up. If it does, `RunExecWithPresenceStepUp` would prompt for MFA on an interactive terminal and retry once. This needs confirmation that a step-up alone is enough to pass when the server emits this code.

## Scope

- `cmd/exec/run.go` — `sudoDenialHints`, the `denialCodePresent` family, `isApprovalDenial`
- `cmd/exec/run_test.go` — hint matching for the two new codes and `--wait` entry tests

jisung-02:
## Background

The CLI's copy of the server's sudo denial codes had drifted. alpacon-server added `SUDO_INTENT_DEVIATION` and `SUDO_POLICY_MFA_REQUIRED`, and neither reached `sudoDenialHints`; `SUDO_COMMAND_NOT_AUTHORIZED` had never been there.

`SUDO_INTENT_DEVIATION` did more than cost a hint: it broke a flag. For a command the judge reads as off-purpose for the work session, `sudo/services.py` takes the same HITL branch, creates the approval grant, and only swaps the code, but `_deny_exec_sudo` still answers alpamon with `approved=False` (ADR 0016 §3). `hasSudoApprovalDenial` compared against `SUDO_APPROVAL_REQUIRED` alone, so:

- `--wait` never entered its re-attempt loop
- `HandlePendingApproval` returned false, and the command exited 1 as a generic failure instead of 4
- CI and AI agents that branch on exit 4 read a live approval request as non-retryable

## Changes

The hint table now carries a `pendingApproval` field, and both the hints and the pending-approval code set are read from it. Keeping them in one list rather than two is what stops them drifting apart again.

| Code | Hint | pendingApproval (exit 4, `--wait` applies) |
|---|---|---|
| `SUDO_INTENT_DEVIATION` | new | yes |
| `SUDO_POLICY_MFA_REQUIRED` | new | no |
| `SUDO_COMMAND_NOT_AUTHORIZED` | new | no |
| `WORKSPACE_SUDO_WITH_MFA_DISABLED` | new | no |
| `SUDO_SESSION_MISSING` | new | no |
| `WORK_SESSION_SCOPE_NOT_ALLOWED` | new | no |
| `SUDO_APPROVAL_REQUIRED` | existing | yes |
| `SUDO_PRESENCE_REQUIRED`, `SUDO_NO_WORKSESSION_POLICY`, `SUDO_RISK_DENIED` | existing | no |

Reaching exit 4 alone would still have hidden the new hint. `HandlePendingApproval` exits before `HandleCommandResult`, until now the only caller of `sudoDenialHint`, so the one way out of an intent deviation that does not wait on a reviewer never reached the user. That path prints the hint itself now, on stderr, leaving the `{"status":"pending_approval", ...}` envelope on stdout untouched.

```
alpacon work-session update [SESSION_ID] --title "<what you are doing>"
```

The command is title-only on purpose. `WorkSessionRiskContext` carries the session title and description and the judge reads both, but `work_sessions/services.py` `compute_modification_split` puts `title` in the immediate bucket and `description` in the queued one. On the approved or active session a denied sudo always runs under, a description edit therefore opens a modification approval of its own — the very wait this hint exists to skip. The hint says so rather than leaving the requester to discover it after ending up with two pending approvals and an unchanged description.

The same check applied to the rest of the table caught the two `--sudo` hints reading the same way. `sudo/services.py` `classify_sudo_policies_split` queues any change that expands privilege and a new policy always does, so "Add it and re-run" was describing something that waits on a reviewer first. Both hints now say the addition may require approval. `work-session update` warns about this on the table path, but under `--output json` the envelope is a plain `ok: true` success, so for the scripts and agents this branch is otherwise about, the hint is the only place it lands.

The pending path also picks its hint differently from `HandleCommandResult`. `sudoDenialHint` returns the first code in table order, while the pending classification looks at the `pendingApproval` codes alone; the two disagree as soon as one command line runs several sudo calls and each leaves a denial line (`sudo a; sudo b`). `pendingSudoDenial` scans the pending codes, so the hint comes from the code that drove the classification, and `hasSudoApprovalDenial` is now that scan's boolean rather than a second loop over the same table.

That scan takes two passes rather than returning on the first pending code it meets. `SUDO_APPROVAL_REQUIRED` sits above `SUDO_INTENT_DEVIATION` in the table, so a single-pass scan answered the same `sudo a; sudo b` output with the approval-required entry — which withholds its guidance on purpose — and dropped the intent-deviation hint entirely. A `selfService` entry now wins over any other pending code on the same output, whatever the table order. Both orders classify the command identically, so preferring the entry that carries actionable guidance costs nothing, and it keeps the table comment true: order among the non-gate codes means nothing again.

`SUDO_POLICY_MFA_REQUIRED` gets a hint but no MFA step-up. The server denies it because a matched policy does not bypass MFA and a Command cannot prompt for it at all, so a fresh presence would not change the verdict.

`SUDO_COMMAND_NOT_AUTHORIZED` is a Command with no `requested_by`, which is a service-token submission that only the token lane under `EXEC_SUDO_MODE=enforce` resolves. Self-hosted deployments on the default `monitor` are the ones that hit it, retrying with the same token cannot help, and the CLI cannot read the mode, so the hint names the principal to switch to instead of a setting the user may not control.

### Gates that run before the command is judged

Review of the above found the drift was not closed. Three codes reach the exec lane from *before* the risk verdict, and none had a hint:

| Code | Where | Why it needs its own wording |
|---|---|---|
| `WORKSPACE_SUDO_WITH_MFA_DISABLED` | `sudo/services.py:298`, before the websh/command branch split | No work session or policy can lift it. Only `alpacon workspace access-control update` can, and only a workspace admin |
| `SUDO_SESSION_MISSING` | `sudo/services.py:393` | The server could not match the sudo call to its Command. Nothing about the command line is wrong, so it reads as retryable, not as a policy problem |
| `WORK_SESSION_SCOPE_NOT_ALLOWED` | `sudo/services.py:2007`, the ADR 0014 scope ceiling | A session created with `--scope command` and no `--sudo` submits the command fine and is denied at the sudo call — the likeliest first-run miss of the three |

The scope hint follows `validateSessionForSudoUpdate` in `cmd/worksession`, which answers the same gap on the update path, including the caveat that `--scope` replaces the whole list. Without it the guidance would have the user silently drop every other scope.

The CLI already knows `work_session_scope_not_allowed` as a lowercase constant in `utils/error.go`, but `HandleWorkSessionError` reads it out of an HTTP error envelope. A denial embedded in the remote command's own output never reaches that handler, so the constant was no help here.

### Unknown codes no longer fail silently

`sudoDenialHint` returning `""` for an unrecognized code is what made each of those invisible — the user saw the bare plugin line and no next step. It now falls back to a hint naming the raw code. The server adds codes on its own release train and nothing enforces this table's sync with it, so a build that cannot explain a code should at least say which code it cannot explain.

That fallback is the one place remote command output reaches a hint, so it accepts only the `[A-Z0-9_]` shape `alpacon_approval.c`'s sanitizer emits, capped at the 63 characters its buffer holds. A command that prints a look-alike line of its own cannot smuggle escapes or newlines through it, and cannot suppress the hint for a real denial that follows it either — the scan continues past a malformed slot rather than giving up on the output.

### One instruction, printed once

`HandlePendingApproval` printed the pending code's hint and then its own message. For `SUDO_APPROVAL_REQUIRED` that said "an approval request was created, re-run after approval" twice in a row. The hint exists for the intent-deviation case, which carries a way past the wait; a code whose only way out *is* the wait has nothing to add to the message below it. `selfService` marks the difference, and `pendingSudoDenial` returns guidance only for entries that have one while still classifying on every pending code.

The rest is wording this change made stale: the exit-4 comment in `utils/error.go`, the inline comment in `cmd/exec/exec.go`, and row 4 of the exit-code table in `README.md`.

### Scanning the line, and saying what the code does

Three closing commits, none of them behavioral.

`firstDenialCode` (was `unknownDenialCode`) walks the output with `strings.Cut` instead of `strings.Index` plus `i+len(prefix)+len(open)` arithmetic, and `sudoDenialLinePrefix` now carries the opening paren so `denialCodePresent`, the scanner, and the fixture in `run_test.go` stop restating `" ("` in three places. The old name promised a table check the function never made: it answers with the first well-formed code, known or not, and only `sudoDenialHint` cares about the difference. The credential-inline helpers, which sat between `sudoDenialHint` and the sudo detectors reading the same table, moved below them so that family reads as one block.

A comment-width pass over what this branch wrote then found five assertions that were not the width of the code beside them:

| Where | Asserted | Code does |
|---|---|---|
| `sudoDenialHints` | codes "are kept in sync with alpacon-server" | nothing keeps them in sync—this PR exists because they drifted |
| `denialCodePresent` | "every detector routes through here" | `firstDenialCode` does not; it scans for the same prefix instead |
| `Invocation` | a defined type stops "a stray value" being assigned | it stops an arbitrary string variable, not a literal |
| `HandlePendingApproval` | handles a sudo denial | also handles a job parked at `awaiting_approval` (`PendingApprovalError`), which is not a denial |
| `ExitCodePendingApproval` | the action "was not waited on with `--wait`" | `cmd/event/wait.go:135,140` exits 4 *after* a wait that timed out or was interrupted |

The rest of that pass is compression: three entry comments restated the guidance string directly below them, and `denialHintLine` claimed to render "one table entry's guidance" when the unknown-code fallback goes through it too.

Rewriting those two comments in place left their line breaks where the old wording had put them, so `Invocation` and `HandlePendingApproval` each ended up with a short line mid sentence. gofmt does not reflow comment prose, so nothing would have caught it. A closing commit redraws the breaks; the wording is unchanged and the diff touches comment lines only.

## Testing

```
go test -race ./...     # no failures
go vet ./...            # no output
gofmt -l .              # no output
golangci-lint run ./... # 0 issues.
mcr docs --verify       # OK	no executable code changed
```

The last one covers the comment-width pass only, and was run at that pass's tip: it fingerprints the code before the edits and re-checks after, so a comment commit that moved an executable line would fail it. The two test-restructuring commits that follow touch no non-test file, so it does not apply to them.

The branch was rebased onto `origin/main` at `abcf8ef` (#348), which drops the earlier merge commit and leaves the diff unchanged. The four commands above were re-run on the rebased tip and again on the final tip, all passing, and `go test ./cmd/exec/` was run on each added commit at its own tip so none of them depends on a later one.

New tests:

- `TestExecIntentDeviationPrintsSelfServiceHint` drives the real exec path in a subprocess and asserts exit 4, the title-only command on stderr, and a still-parseable JSON signal on stdout
- `TestExecIntentDeviationPrintsSelfServiceHintAfterWaitTimeout` covers the other way into `HandlePendingApproval`: a `--wait` whose window closed returns the last denial to the same handler, so the hint has to survive the wait and not only reach the no-wait path. It runs with `--wait-approval 1ms`, shorter than the poll interval, so the window closes on the timer without a re-attempt. Both tests share `runIntentDeviationHelper` for the subprocess setup
- `TestRunExecWithApprovalWait_EntersLoopOnIntentDeviation` asserts `--wait` re-attempts on this code too
- `TestPendingSudoDenial` pins the divergence directly: for output carrying both `SUDO_NO_WORKSESSION_POLICY` and `SUDO_INTENT_DEVIATION`, it answers with the intent-deviation hint while `sudoDenialHint` still answers with the earlier `--sudo` one. It also pins that `SUDO_APPROVAL_REQUIRED` still classifies as pending while withholding its hint, and that on output carrying both pending codes the self-service entry wins over the one that sits earlier in the table. A closing commit adds the unknown-code row: the drift this branch exists to prevent reaches this scan first, and only the hint path had that case covered — not the classification path, which is the one that picks the exit code. An unrecognized code stays terminal, since claiming pending would tell a script to retry what the server may have refused outright
- `TestSudoDenialHintFallsBackToTheRawCode` covers the unknown-code hint, four forged shapes the sanitizer would never emit (ANSI escape, lowercase, embedded space, empty), a look-alike line that must not suppress the real denial after it, and both sides of the 63-character cap — the only rule stopping a longer code, which has the sanitizer's exact shape, from being echoed
- `TestSudoDenialHint`, `TestHasSudoApprovalDenial`, and `TestIsApprovalDenial` gain cases for the new codes, including a `NotContains` guard so a queued `--description` edit cannot reappear inside the reviewer-free command, and assertions that both `--sudo` hints keep the approval caveat
- `TestExecPendingApprovalExits4WithJSONSignal` gains a `NotContains(stderr, "Hint:")` guard for the duplicate-instruction regression

### Restructuring what the above added

Two closing commits, neither of them changing what is asserted.

`cmd/exec/sudo_hint_test.go` had grown to eleven subtests in one function and a nine-line block of `assert.False` calls in another, which is past where subtests still read well and past the repo's own convention (CLAUDE.md, "Test patterns"). Five tests are tables now, keyed on the denial code or the raw output. Two gains beyond the reading: a failure in `TestHasSudoApprovalDenial` names which code broke instead of pointing at a block covering nine of them, and the check that a hint did not come from the unknown-code fallback — previously on the risk-denied case alone — runs for every row, so a table entry dropped by accident cannot leave the other cases passing against fallback text.

Three checks stayed subtests because a row cannot hold them: the two policy hints are compared against each other, the 63/64 character cap asserts both sides of one boundary, and `TestReRunHint` verifies a different shape per case. No case was dropped; the empty-output and forged-token cases moved from `TestSudoDenialHint` to `TestSudoDenialHintFallsBackToTheRawCode`, where they test the anchor rather than a code's guidance.

The three test files each spelled the plugin's denial line their own way, and `stubApprovalWaitSeams` assembled it from `sudoDenialLinePrefix` — a test building its input from the constant under test, so a prefix changed to something the plugin never emits would have left `TestRunExecWithApprovalWait_*` passing against a line no real denial carries. All three call `denialLine` now, which holds the literal.

Each new assertion was confirmed to fail against the code without its guard, so none of them is a test that would pass either way:

| Reverted | Failing test |
|---|---|
| the three hint-printing lines | `TestExecIntentDeviationPrintsSelfServiceHint` — no `--title` line |
| `pendingSudoDenial` back to a single-pass scan | `TestPendingSudoDenial/prefers the self-service code…` — `""` does not contain `--title` |
| the `maxCodeLen` check | `TestSudoDenialHintFallsBackToTheRawCode/a code longer than…` — the 64-character code is echoed |
| the self-service print in `HandlePendingApproval` | `…AfterWaitTimeout` — stderr carries the timeout warning alone |

Each of the earlier review-fix commits was built, vetted, and run against `go test -race ./cmd/exec/` at its own tip before the next one was written, so none of them depends on a later commit to pass.

## Follow-up

Nothing verifies this table against alpacon-server. The contract is a three-repo mirror (server → alpamon-pam → CLI) and this repo has no precedent for a build-time dependency on a sibling repo's source, so the unknown-code fallback is the guard instead of a generated file. If a stronger check is wanted, it belongs in CI, not here.

`SUDO_APPROVAL_REQUIRED`'s guidance string is now unreachable for printing — `HandlePendingApproval` always intercepts that code before `HandleCommandResult`. It is kept as the code's description in the table; deleting it is a one-line follow-up if that reads as dead weight.

A transient error on a single poll tick ends `--wait-approval` and reports the still-open approval as exit 1. `RunExecWithApprovalWait` returns any error that is neither a `PendingApprovalError` nor an approval denial, so one 502 or dropped connection discards the rest of the window. Pre-existing loop behavior, unchanged here — but this PR widens which denials enter the loop, so intent deviations now inherit it. Filed as #350.

Codes confirmed **out of scope** for this table, since interactive websh keeps its own sudo flow and does not route through it: `SUDO_ACCOUNT_BOUND_TO_APPLICATION`, `SUDO_NO_ACCOUNTABLE_ACTOR`, `SUDO_MFA_FAILED`, `SUDO_SYSTEM_USER_NOT_FOUND`, `SUDO_NO_AUTHORITY`.

`alpacon work-session update --output json` reports a queued sudo policy or description edit as a plain `ok: true` success, with nothing marking it as awaiting approval. That belongs to `cmd/worksession`, not to this table, so it is left alone here.

Closes #288


cmd/exec/sudo_hint_test.go
```diff
@@ -3,147 +3,368 @@ package exec
...
+			wantHintContains: "--title",
+		},
+		{
+			// HandlePendingApproval's own message already says "re-run after
+			// approval", so returning this entry's guidance would print it twice.
+			name:        "withholds the hint for a code whose only way out is the wait",
+			output:      denialLine("SUDO_APPROVAL_REQUIRED"),
+			wantPending: true,
+		},
+		{"a terminal code is not pending", denialLine("SUDO_RISK_DENIED"), false, ""},
```
junho226, line 245: Suggestion (non-blocking): add the unknown-code case next to this one.

```go
{"an unknown code is not pending", denialLine("SUDO_BRAND_NEW_CODE"), false, ""},
```

The behavior is already correct and safe — `pendingSudoDenial` only matches table entries carrying `pendingApproval`, so a code the CLI does not know exits 1 with the `firstDenialCode` fallback hint rather than claiming a pending approval. That is the right fail direction: guessing "pending" on an unrecognized code would tell a script to retry something the server may have refused outright.

Worth pinning anyway, because the drift this PR exists to prevent is exactly the drift that reaches this scan. `TestSudoDenialHintFallsBackToTheRawCode` covers the unknown code for the *hint* path; nothing covers it for the *classification* path, which is the one that decides the exit code.

junho226 (dismissed):
## Summary

The fix is correct, and I verified the premise it rests on against the server rather than taking the PR description's word for it.

`sudo/services.py:1840-1856` picks `SUDO_INTENT_DEVIATION` and `SUDO_APPROVAL_REQUIRED` in the *same* `Verdict.REQUIRE_APPROVAL` branch — one ternary on `result.intent_deviation` — then calls `_create_command_layer3_approval` and passes the grant to `_deny_exec_sudo(..., pending_grant=grant)`. The websh branch does the same at `sudo/services.py:826-864` via `_create_layer3_approval`. `_deny_exec_sudo` (`sudo/services.py:1113-1152`) answers alpamon `approved=False` either way but publishes `approval_requested` rather than `denied` when a grant is attached, and its own docstring says the command "was routed to human approval, not hard-denied". `utils/error_codes.py:926` says the same: "routes to human approval rather than a terminal deny; the requester may instead re-declare the work session's intent (ADR 0016 §4-5)".

So the two codes are the same state with different labels, and the CLI calling one pending while calling the other a generic failure was the defect. Treating both as pending is what the server means.

I also checked exhaustiveness in the other direction. `pending_grant=` has exactly one call site server-side (`sudo/services.py:1855`), and `_create_layer3_approval` / `_create_command_layer3_approval` have three (`826`, `1078`, `1851`) — all inside HITL branches that emit one of these two codes. The token-exec lane (`_resolve_token_exec_sudo`) denies with `SUDO_RISK_DENIED` and no grant. No third code creates an approval request today, so the two-entry `pendingApproval` set is complete against current `main` of alpacon-server.

The other four codes the table gains are correctly *not* pending. `SUDO_POLICY_MFA_REQUIRED` ("Policy matched but needs MFA, which a non-interactive Command can't do", `error_codes.py:904`) and `SUDO_COMMAND_NOT_AUTHORIZED` (`error_codes.py:891`, service-token submission under the default `EXEC_SUDO_MODE=monitor`) both reach `_deny_exec_sudo`/`_send_sudo_failure_response` with no grant.

## Critical issues

None.

## Warnings

None.

## Suggestions

1. **The unknown-code path is safe by construction but not pinned by a test.** `pendingSudoDenial` only ever matches table entries carrying `pendingApproval`, so a code the CLI does not know is classified terminal: exit 1, and `HandleCommandResult` prints the `firstDenialCode` fallback hint naming the code. That is the right fail direction — a client that guessed "pending" on an unrecognized code would tell a script to retry something the server refused outright. `TestPendingSudoDenial` pins "a terminal code is not pending" and "empty output is not pending" but not "an unknown code is not pending", which is the case most likely to change silently when someone edits the scan. One line closes it (inline).

2. **Mixed output classifies as pending even when a terminal denial is also present.** For `sudo a; sudo b` where `a` is denied `SUDO_RISK_DENIED` and `b` lands `SUDO_INTENT_DEVIATION`, `pendingSudoDenial` returns pending and the process exits 4 — so a caller that re-runs after approval still hits the terminal denial on `a`. `TestPendingSudoDenial` pins this deliberately ("answers with the pending code, not an earlier terminal one"), and the alternative — exiting 1 and hiding a live approval request — is worse. Recording it as a known edge, not a change request.

3. **A transient error mid-`--wait` ends the wait and reports exit 1.** In `RunExecWithApprovalWait` the ticker arm returns any error that is neither a `PendingApprovalError` nor an approval denial, so a 502 or a dropped connection on a poll tick aborts the wait; `HandlePendingApproval` then declines it and the still-open approval surfaces as a generic failure. This is pre-existing loop behavior, unchanged by this PR — but the PR widens which denials enter the loop, so intent deviations now inherit it too. Out of scope here; worth a follow-up issue rather than a change to this branch.

## Highlights

- **The one-table design is the actual fix.** The bug was two lists drifting; deriving both the hints and the pending set from one table with a `pendingApproval` field means the next code added lands in both or neither. The comment block on `sudoDenialHints` says exactly that, and says why the gate codes lead the table while the rest carry no ordering meaning.
- **`firstDenialCode` closes the drift the table cannot.** An unknown code now yields a hint naming it and pointing at the console, instead of a bare denial line — so the *next* server-side addition is visible on first contact rather than after a bug report. The `[A-Z0-9_]` sanitizer check plus the 63-char cap (mirroring `alpacon_approval.c`'s `char[64]`) keeps a command's own output from smuggling escapes into that hint, and both bounds are tested at the boundary (63 echoes, 64 does not).
- **`selfService` is a genuine distinction, not decoration.** `HandlePendingApproval` exits before `HandleCommandResult`, so before this PR the one way out of an intent deviation that does not wait on a reviewer was unreachable. Printing only `selfService` entries avoids restating "re-run after approval", and `TestExecCommandPendingApproval...` pins the negative case: `SUDO_APPROVAL_REQUIRED` must produce no `Hint:` on stderr.
- **The hint goes to stderr, the envelope stays on stdout.** Verified end-to-end rather than by inspection: the subprocess test parses stdout as JSON and asserts `status == pending_approval` while asserting hint text on stderr. `cmd/exec/exec.go:196-214` buffers streamed output in JSON mode and discards it on the pending path, so the envelope is not corrupted by command output.
- **Every command and flag the new hints name exists and behaves as described.** I ran the built binary: `alpacon workspace access-control update` exists and opens an editor (matching the hint's "interactive terminal required" caveat); `work-session update` takes an optional `SESSION_ID`, `--title`, `--scope`, and `--sudo`, and its own long help already says "The additions may require approval before they take effect" — consistent with the two `--sudo` hints this PR corrects.
- **`sudoDenialLinePrefix` refactor is behavior-preserving.** Moving `" ("` from the `denialCodePresent` concatenation into the constant makes the same string; the test helper `denialLine` restates the wire format literally instead of reusing the constant, so an accidental edit to the constant fails the tests rather than passing vacuously.
- **Tests are end-to-end where it matters.** The two intent-deviation tests spawn a real subprocess against an `httptest` server and assert the actual process exit code, not a return value.

## Verification

Run in a clean worktree at `d164d4ab5e331320625f20c75eefa6a478b7b3ef` (`gh pr checkout 344`). CI is `.github/workflows/build-and-test.yaml` (`go build -v .`, `go test -race -v ./...`) and `.github/workflows/lint.yaml` (`golangci-lint` pinned to `v2.9.0`, plus `govulncheck`).

| Command | Result |
|---|---|
| `go build ./...` | pass (go1.25.13 darwin/amd64) |
| `go vet ./...` | pass, no findings |
| `gofmt -l .` | pass, no files listed |
| `go test -race -count=1 ./...` | pass, exit 0, 39 packages ok, 0 FAIL (`cmd/exec` ok 7.258s) |
| `golangci-lint run ./...` | pass, `0 issues` |

`golangci-lint` was not on `PATH`; I used `/Users/junho/go/bin/golangci-lint`, which reports `version 2.9.0 built with go1.25.12`. That matches the version the workflow pins, though CI resolves its own build of that release and my local Go toolchain is 1.25.13 vs the binary's 1.25.12.

**Base-revert check** — two probes, worktree restored and confirmed clean (`git status --short` empty) after each:

- *Full revert.* Restoring `cmd/exec/run.go` to the merge base with the branch's tests in place fails to compile: `undefined: sudoPresenceRequiredCode`, `undefined: pendingSudoDenial`. So the new tests cannot run against base sources unmodified, as expected.
- *Targeted mutation.* Keeping the whole branch and deleting only `pendingApproval: true` from the `SUDO_INTENT_DEVIATION` entry produced exactly the pre-fix failure: `TestExecIntentDeviationPrintsSelfServiceHint` and `TestExecIntentDeviationPrintsSelfServiceHintAfterWaitTimeout` both fail with `expected: 4, actual: 1`. That is the bug this PR describes — an intent deviation exiting 1 as a generic failure — reproduced and pinned.

**Not run:** `govulncheck ./...` (CI installs it at `@latest`; skipped as unrelated to this diff — CI reports it passing on this head, along with Build and Test, golangci-lint, CodeQL, and both Analyze jobs).

I did not exercise a live alpacon-server or alpamon; the server-side claims above are read from `sudo/services.py`, `utils/error_codes.py`, and `execution_gate/stages.py` on the current `main` of alpacon-server, not from a running deployment.

## Verdict

**Approve.** The client's new reading of an intent deviation matches what the server means by it, verified at the source. The pending set is exhaustive against current server code, the four non-pending additions are correctly non-pending, the wait is bounded, exit codes match the README contract, and the failure path is tested end-to-end including the negative and forged-input cases. The three suggestions are all non-blocking.

## Checklist

- [x] Server-side semantics of `SUDO_INTENT_DEVIATION` verified in `sudo/services.py` and `utils/error_codes.py` — same HITL branch, approval grant created
- [x] `pendingApproval` set checked for exhaustiveness against every `pending_grant` / `_create_*_layer3_approval` call site
- [x] Non-pending additions (`SUDO_POLICY_MFA_REQUIRED`, `SUDO_COMMAND_NOT_AUTHORIZED`, `WORKSPACE_SUDO_WITH_MFA_DISABLED`, `SUDO_SESSION_MISSING`, `WORK_SESSION_SCOPE_NOT_ALLOWED`) confirmed to create no grant
- [x] Unknown-code path traced: not pending, exit 1, fallback hint still reaches the user
- [x] Exit codes match the README table; `--wait` bounded by `utils.DefaultApprovalWaitTimeout` and `ParsePositiveDuration`
- [x] JSON stdout stays parseable with the hint on stderr — asserted end-to-end
- [x] Every command and flag named in the new hints verified against the built binary
- [x] Go specifics: `%w` wrapping intact, `errors.As` throughout, no unchecked type assertions, no swallowed errors
- [x] Build, vet, gofmt, race tests, golangci-lint v2.9.0 all run and passing
- [x] Base-revert / mutation probe confirms the tests fail without the fix
- [x] Existing review threads read — Copilot's review generated no comments; nothing re-litigated

cmd/exec/sudo_hint_test.go
jisung-02: Applied in 7b818a7, as suggested:

```go
// Claiming pending on a code the table does not carry would tell a script to
// retry what the server may have refused outright, so drift must fail terminal.
{"an unknown code is not pending", denialLine("SUDO_BRAND_NEW_CODE"), false, ""},
```

Your reasoning is the comment: the row alone reads as an arbitrary negative case, so it says why terminal is the right fail direction rather than an accident of the scan.

The split you name is the part I had missed. `TestSudoDenialHintFallsBackToTheRawCode` covers the unknown code for the hint path, and I had read that as covering the code being unknown at all — but the two paths scan the same table by different rules, and only this one picks the exit code. Confirmed the row fails without the behavior: adding `pendingApproval: true` to a fabricated entry for `SUDO_BRAND_NEW_CODE` turns it red, so it is not a test that passes either way.

`go test -race -count=1 ./...` passes, 39 packages, no failures; `gofmt -l .` and `golangci-lint run ./...` clean.

cmd/exec/run.go
```diff
@@ -43,65 +49,251 @@ var (
...
+// alpacon-server utils/error_codes.py, so returning "" for one would leave the
+// next drift invisible until someone reported a bare denial line.
 func sudoDenialHint(output string) string {
 	for _, h := range sudoDenialHints {
 		if denialCodePresent(output, h.code) {
-			return fmt.Sprintf("%s %s", utils.Yellow("Hint:"), h.guidance)
+			return denialHintLine(h.guidance)
 		}
 	}
+	if code := firstDenialCode(output); code != "" {
```
junho226, line 241: Suggestion (non-blocking): the codeless denial line is the one case in this family that still fails silently.

`alpacon_approval.c:450` emits the same line without a code — `"Alpacon denied this sudo command.\n"` — on the branch taken when the server sent no `error_code`, or one the sanitizer rejected whole (any character outside `[A-Z0-9_]`, or over 63 characters). `pam_alpamon.c:644` does the same. Now that `sudoDenialLinePrefix` ends in `(`, neither the table scan above nor `firstDenialCode` matches that line, so the user gets the bare denial and no next step.

That is not a regression — the old table scan missed it too — and the new `"prefix without the code slot"` row pins it as deliberate. But it is the sibling of the case this fallback exists to close, and after this PR it is the only remaining path where a real denial reaches the terminal with nothing actionable beside it. A generic branch on that exact literal would close it:

```go
if strings.Contains(output, "Alpacon denied this sudo command.") {
	return denialHintLine("sudo was denied, and the server named no reason. Read the denial in the Alpacon console (web).\n")
}
```

Worth noting that a rejected-by-sanitizer code lands here rather than on the fallback above, so a server that ships a malformed code hits the silent path instead of the one this PR added. Fine to leave for a follow-up — with no category from the server there is genuinely less to say — but it is the case I would expect to come back.

junho226 (approved):
## Summary

The fix is correct. `SUDO_INTENT_DEVIATION` reaches the exec lane from the same `Verdict.REQUIRE_APPROVAL` branch as `SUDO_APPROVAL_REQUIRED`, with the grant already created, so classifying it as pending is what the server means by it. Folding the pending set into the hint table rather than keeping a second list is the right shape for a contract that has already drifted once.

This is a re-review of the tip after the previously dismissed round. The delta is three lines — the `"an unknown code is not pending"` row in `TestPendingSudoDenial` plus its comment — applied as suggested and placed next to the terminal-code row it belongs beside. Nothing else moved, so the earlier round's verification still holds; I re-ran it against this head rather than assuming so.

## Toolchain, run on this head in a clean worktree

| Command | Result |
|---|---|
| `go build ./...` | pass, no output |
| `go vet ./...` | pass, no findings |
| `gofmt -l .` | pass, no files listed |
| `go test -race ./...` | pass, 39 packages `ok`, 0 `FAIL` (`cmd/exec` ok 7.16s) |

`golangci-lint` is not installed on this machine, so I did not run it — the repo's own `.golangci.yml` enables `errcheck`, `govet`, `ineffassign`, `staticcheck`, `testifylint`, `unused` plus `gofmt`/`goimports`. `go vet` and `gofmt` cover two of those; CI's `golangci-lint` job reports pass on this head, as do Build and Test, CodeQL, govulncheck, and both Analyze jobs. I did not exercise a live alpacon-server or alpamon; every server-side and agent-side claim below is read from source.

## What I checked beyond the diff

The table is a hand-maintained mirror of two sibling repos, so the parts worth verifying are the ones no test in this repo can reach.

**The pending set is exhaustive.** `sudo/services.py` has exactly one `pending_grant=` call site (the exec lane, in the `Verdict.REQUIRE_APPROVAL` branch that picks the code by a ternary on `result.intent_deviation`). The other approval-creating branch is the interactive websh one, which the PR correctly puts out of scope. So `SUDO_APPROVAL_REQUIRED` and `SUDO_INTENT_DEVIATION` are the complete pending set against current server code, and the four non-pending additions create no grant.

**The codes arrive in the shape the table expects.** The `ErrorCode` values are lowercase, and the plugin's sanitizer drops lowercase — so this only works because `send_sudo_approval_response` puts `denial_reason.upper()` on the wire as `error_code`. All six new codes have a value equal to their lowercased member name, so `.upper()` reproduces the table key exactly. `_send_sudo_failure_response` delegates to the same function, so the workspace-MFA gate is covered too.

**The three pre-verdict gates exist and the table's order matches the server's.** `WORKSPACE_SUDO_WITH_MFA_DISABLED` (before the branch split), `SUDO_SESSION_MISSING` (Branch B command resolution), `WORK_SESSION_SCOPE_NOT_ALLOWED` (the ADR 0014 ceiling), then `SUDO_COMMAND_NOT_AUTHORIZED` — the same order the table leads with, so the "answers with the gate the server checks earliest" comment is true rather than aspirational.

**The 63-character cap is exact, not approximate.** `alpacon_approval.c` rejects the whole code on any character outside `[A-Z0-9_]`, and rejects rather than truncates once `j >= sizeof(error_code_buf) - 1` on a `char[64]`. Maximum accepted length is therefore 63, which is what `maxCodeLen` and the boundary subtest assert. Because the accepted set excludes `\n`, `\r`, and ESC, the code `firstDenialCode` hands back cannot carry a newline or an escape even though it is printed verbatim — the scanner's `strings.Cut(rest, ").")` can span lines, but a span that does is rejected by `isSanitizedDenialCode`. The loop terminates: `rest` strictly shrinks on every iteration.

**Every command and flag the new hints name is real.** `work-session update --scope` accepts `sudo` and replaces the whole list; `alpacon workspace access-control update` exists; `work-session create --sudo` does add the `sudo` scope implicitly (`worksession_create.go:151-154`), so that sentence in the scope hint is accurate.

**The approval caveats are accurate.** `compute_modification_split` puts `title` in the immediate bucket and `description` in the queued one unconditionally, which is exactly the distinction the intent-deviation hint draws. `scopes` is queued iff the new set adds an element, and `sudo_policies` iff an entry expands privilege — so "may require approval before it takes effect" is the right hedge on both `--sudo` hints and on the scope hint, and the `NotContains(--description ")` guard pins the part that would mislead.

**The two stale comments are now correct.** `cmd/event/wait.go:131-140` does exit 4 after a timeout or an interrupt, so the `ExitCodePendingApproval` rewording matches. Websh command mode routes through `execCmd.RunRemoteExec` (`cmd/websh/websh.go:252`), the sole `HandlePendingApproval` call site, so the README's exit-4 claim for websh holds for the new code too.

One behavioral note on the fallback: `sudoDenialHint` is only reachable from a `RemoteCommandError`, so the unknown-code hint cannot fire on a command that succeeded while printing a look-alike line.

## Suggestion (non-blocking)

One silent path is left in this family — see the inline comment on the fallback.

## Verdict

**Approve.** The client's reading of an intent deviation matches the server's, verified at the source rather than from the description; the pending set is exhaustive; the four non-pending additions are correctly non-pending; the unknown-code path fails terminal, which is the safe direction and is now pinned on both the hint path and the classification path; and the guard-by-guard revert table in the description is the kind of evidence that makes the new assertions worth having. The one suggestion is a pre-existing gap, not a blocker.

## Checklist

- [x] `SUDO_INTENT_DEVIATION` semantics verified in `sudo/services.py` and `utils/error_codes.py` — same HITL branch, grant created, `approved=False`
- [x] Pending set checked for exhaustiveness against every `pending_grant` / `_create_*_layer3_approval` call site
- [x] Wire shape verified: `error_code = reason.upper()`, and all six new values equal their lowercased member name
- [x] Gate order in the table matches the server's check order
- [x] 63-char cap and `[A-Z0-9_]` shape verified against `alpacon_approval.c`; scanner cannot emit a newline or escape, and cannot loop
- [x] Every command, flag, and approval caveat named in the new hints verified against the CLI and the server
- [x] Exit codes match the README table; hint on stderr keeps the JSON envelope parseable (asserted end-to-end)
- [x] Build, vet, gofmt, and `-race` tests run on this head and passing
- [x] Prior review threads read — the single suggestion was applied as requested; nothing re-litigated
- [ ] `golangci-lint` not run locally (not installed); CI's job passes on this head

cmd/exec/run.go
jisung-02: Filed as #353, not changed here.

Confirmed both branches you name: `alpacon_approval.c:449-451` prints the codeless line when `error_code_buf` is empty, `pam_alpamon.c:643-645` the same when `code[0] == '\0'`. And the sanitizer (`alpacon_approval.c:380-404`) rejects the code *whole* on a disallowed character or on overflow past `sizeof(error_code_buf) - 1`, with the comment saying why—truncating would let padding spoof a legitimate code. So your last point is the sharp one: a malformed or over-63-character code from the server lands on this silent path rather than on the fallback this PR added for exactly that drift.

What kept it out of the branch is the anchor. Every detector here matches on the code slot so a command's own output cannot forge a denial signal; the codeless line has no slot, so the branch you sketch matches a bare literal and a command printing that sentence would produce the hint. `sudoDenialHint` is only reachable from a `RemoteCommandError`, which bounds it but does not remove it. That is a call on how much forgeability to accept for a hint that, with no category from the server, is thin by construction—and it changes what `TestSudoDenialHint/prefix without the code slot` should assert, which this PR added as a deliberate negative. The issue records both.

PR #347, opened 2026-08-14
build(go): raise the go directive to 1.25.13

jisung-02:
## Summary

The `govulncheck` job on `main` fails ([run #727](https://github.com/alpacax/alpacon-cli/actions/workflows/lint.yaml)) with four standard-library vulnerabilities. All four are fixed in go1.25.13, so this raises the `go` directive in `go.mod` from `1.25.12` to `1.25.13`—and then fixes the workflows, which turned out to be pinning the toolchain in a way that could not follow that bump.

| Advisory | Package | Reachable from |
| --- | --- | --- |
| [GO-2026-6218](https://pkg.go.dev/vuln/GO-2026-6218) | `net/url` — quadratic complexity in `resolvePath` | `client/client.go:366` |
| [GO-2026-6090](https://pkg.go.dev/vuln/GO-2026-6090) | `crypto/tls` — unbounded post-handshake messages | `pkg/tunnel/runtime/runtime.go:98`, `pkg/tunnel/pool.go:38`, `cmd/exec/run.go:379`, `client/client.go:366` |
| [GO-2026-5972](https://pkg.go.dev/vuln/GO-2026-5972) | `encoding/asn1` — unbounded recursion depth | `pkg/cert/cert.go:63` |
| [GO-2026-5026](https://pkg.go.dev/vuln/GO-2026-5026) | `net/http` (vendored `x/net/idna`) — ASCII-only Punycode labels accepted | `client/client.go:366`, `api/ftp/ftp.go:569` |

## Why the workflows had to change too

Raising the directive alone made the `govulncheck` job fail earlier and louder:

```
go: go.mod requires go >= 1.25.13 (running go 1.25.12; GOTOOLCHAIN=local)
```

Three things stack up to produce that:

1. `golang/govulncheck-action@v1` runs **its own `setup-go`** inside the composite action. `setup-go` sets `GOTOOLCHAIN=local` when it is handed an explicit version, which switches off toolchain auto-download.
2. The `GOTOOLCHAIN=auto` step we exported *before* the action is therefore overwritten by the action's own `setup-go`. It never had any effect on that job.
3. `go-version-input: '1.25.x'` resolved to **1.25.12**, the 1.25 build already sitting in the runner's tool cache. Against a go.mod that requires 1.25.13, the scan dies before it starts.

Pointing the action at go.mod is not straightforward: its `go-version-input` defaults to `'stable'` rather than an empty string, and `setup-go` resolves `go-version` ahead of `go-version-file`. Passing `go-version-input: ''` plausibly suppresses the composite default and lets the file input win, but that behavior is undocumented and would silently break on an upstream change. So this PR drops the action and inlines the same four steps it wraps—checkout, `setup-go`, `go install golang.org/x/vuln/cmd/govulncheck@latest`, `govulncheck ./...`—where our `GOTOOLCHAIN=auto` step actually applies.

While in there, every `setup-go` step across the three workflows now reads `go-version-file: go.mod`. Each one previously carried its own `1.25.x` wildcard, so the toolchain was whatever 1.25 build the runner cache happened to hold and drifted job by job. go.mod is now the single source, and the next patch bump needs one line changed instead of four.

## Release notes review

[go1.25.13](https://go.dev/doc/devel/release#go1.25.13) (2026-08-13) is a patch release:

> includes security fixes to the `go` command, and the `crypto/tls`, `encoding/asn1`, `encoding/xml`, `html/template`, `net/http`, and `net/url` packages, as well as bug fixes to the compiler, the runtime, and the `crypto/tls` and `os` packages.

I walked the [Go1.25.13 milestone](https://github.com/golang/go/issues?q=milestone%3AGo1.25.13+label%3ACherryPickApproved) (24 backports) and **found no functional change that this CLI has to adapt to**. Nothing in the release alters an API, a signature, or a documented behavior we depend on. The backports fall into three buckets:

- **Security hardening** — the four advisories above, plus `encoding/xml` recursion depth, `html/template` XSS, and three CVE fixes in the `go` command. All tighten limits on malformed input; none change the shape of a valid result.
- **Compiler and runtime miscompilation fixes** — `cmd/compile` prove/regalloc bugs and arch-specific miscompiles (riscv64, mips64le, arm64), runtime pointer bugs. These correct wrong codegen; they do not change language or library semantics.
- **Platform-specific `os` and `net/http` fixes** — `os.Root.MkdirAll` with trailing slashes, a `Server.UnencryptedHTTP2` header-timeout regression, and Plan 9 / NetBSD test fixes. None of these paths are reachable from this codebase.

So the Go side of this PR is a toolchain floor bump only. No source change accompanies it, and none is required.

## Review follow-ups

Three commits added after the first review round:

| Commit | Change | Review item |
| --- | --- | --- |
| `2f954d4` | `ci: blank GOOS/GOARCH when installing govulncheck` | Warning 2 — restores the one guard `golang/govulncheck-action` carried that did not survive the inlining |
| `aee9c66` | `docs(claude): point the Go version at the go directive` | Suggestion 3 — `CLAUDE.md:189` was the last version literal left in the repo |
| `f4e81b6` | `build(goreleaser): drop the redundant before hooks` | carried over from the same review on [alpamon#416](https://github.com/alpacax/alpamon/pull/416) |

On `.goreleaser.yaml`: `before.hooks` ran `go mod tidy` on every release build, but `build-and-test.yaml` already runs it followed by `git diff --exit-code go.mod go.sum`, and `release.yaml` gates the `goreleaser` job on that workflow. The second hook, `go generate ./...`, was a no-op—the repo carries no `go:generate` directive.

Deliberately not changed: `go install golang.org/x/vuln/cmd/govulncheck@latest` stays unpinned (Warning 1). govulncheck fetches its vulnerability database from `https://vuln.go.dev` at run time, so a pin would still catch every new advisory; what `@latest` buys is the scanner engine itself—new reachability analysis and symbol matching—which is the one thing a pin cannot get. The drift risk that argues for pinning is already closed here by the `GOTOOLCHAIN=auto` step above the install.

## Test plan

Ran locally against go1.25.13 (`GOTOOLCHAIN=auto`):

- `go build ./...` — clean
- `go test -race ./...` — every package `ok`, no failures
- `govulncheck ./...` — `No vulnerabilities found.`

CI on this branch: `govulncheck` and `golangci-lint` both pass on the rewritten workflow.

One vulnerability remains in the "imported but not called" bucket: [GO-2026-5942](https://pkg.go.dev/vuln/GO-2026-5942) (`x/net/dns/dnsmessage` panic on an invalid SVCB/HTTPS RR), fixed only in go1.26.6 with no 1.25 backport yet. Our code never calls it, so `govulncheck` still exits `0` and CI passes.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -52,8 +52,13 @@ jobs:
       - name: Allow Go toolchain auto-download
         run: echo "GOTOOLCHAIN=auto" >> "$GITHUB_ENV"
 
+      # Blank GOOS/GOARCH so the tool is always built for the runner, even if
+      # this job ever gains a cross-compilation matrix.
       - name: Install govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
+        env:
+          GOOS: ""
+          GOARCH: ""
 
       - name: Run govulncheck
         run: govulncheck ./...
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.25.x'
+          go-version-file: go.mod
 
       - name: Allow Go toolchain auto-download
         run: echo "GOTOOLCHAIN=auto" >> "$GITHUB_ENV"
@@ -39,10 +39,23 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      # golang/govulncheck-action pins its own setup-go with GOTOOLCHAIN=local
+      # and a `go-version-input` that overrides `go-version-file`, so a runner
+      # whose tool cache lags behind go.mod fails before the scan starts. Run
+      # the same steps directly and let go.mod pick the toolchain.
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
       - name: Allow Go toolchain auto-download
         run: echo "GOTOOLCHAIN=auto" >> "$GITHUB_ENV"
 
+      - name: Install govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
+
       - name: Run govulncheck
-        uses: golang/govulncheck-action@v1
-        with:
-          go-version-input: '1.25.x'
+        run: govulncheck ./...
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -42,10 +42,8 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
 
-      # golang/govulncheck-action pins its own setup-go with GOTOOLCHAIN=local
-      # and a `go-version-input` that overrides `go-version-file`, so a runner
-      # whose tool cache lags behind go.mod fails before the scan starts. Run
-      # the same steps directly and let go.mod pick the toolchain.
+      # Inlined instead of golang/govulncheck-action: it forces its own Go
+      # version with GOTOOLCHAIN=local, so a tool cache behind go.mod fails.
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,7 +186,7 @@ _ = json.NewEncoder(w).Encode(resp)
 
 ## Important notes
 
-- **Go version**: 1.25.13 (specified in go.mod)
+- **Go version**: see the `go` directive in `go.mod`
 - **Linter**: golangci-lint v2 with errcheck, govet, ineffassign, staticcheck, unused (see `.golangci.yml`)
 - **Config file**: `~/.alpacon/config.json` (dir `0700`, file `0600`)
 - **Alias**: `alpacon` can also be invoked as `ac`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,7 +186,7 @@ _ = json.NewEncoder(w).Encode(resp)
 
 ## Important notes
 
-- **Go version**: 1.25.12 (specified in go.mod)
+- **Go version**: 1.25.13 (specified in go.mod)
 - **Linter**: golangci-lint v2 with errcheck, govet, ineffassign, staticcheck, unused (see `.golangci.yml`)
 - **Config file**: `~/.alpacon/config.json` (dir `0700`, file `0600`)
 - **Alias**: `alpacon` can also be invoked as `ac`
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.25.x'
+          go-version-file: go.mod
 
       - name: Allow Go toolchain auto-download
         run: echo "GOTOOLCHAIN=auto" >> "$GITHUB_ENV"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -57,7 +57,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: 1.25.x
+          go-version-file: go.mod
 
       - name: Allow Go toolchain auto-download
         run: echo "GOTOOLCHAIN=auto" >> "$GITHUB_ENV"
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,10 +1,5 @@
 project_name: alpacon
 
-before:
-  hooks:
-    - go mod tidy
-    - go generate ./...
-
 builds:
   - env:
       - CGO_ENABLED=0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/alpacax/alpacon-cli
 
-go 1.25.12
+go 1.25.13
 
 require (
 	github.com/gorilla/websocket v1.5.3


.github/workflows/lint.yaml
```diff
@@ -39,10 +39,21 @@ jobs:
...
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
       - name: Allow Go toolchain auto-download
         run: echo "GOTOOLCHAIN=auto" >> "$GITHUB_ENV"
 
+      - name: Install govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
```
geunwoonoh, line 58: **Warning:** two points on this step, neither one a regression.

**1. `@latest` is unpinned.** The upstream action ran the identical `go install ...@latest` internally, so this matches prior behavior. But lines 33-35 of this file pin golangci-lint to an exact `v2.9.0`, and `x/vuln` v1.7.0 shipped on 2026-08-13, so CI is currently picking up a one-day-old release. A govulncheck release that changes exit-code behavior or adds a false positive breaks CI with no commit on our side.

```suggestion
      - name: Install govulncheck
        run: go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
```

**2. `GOOS`/`GOARCH` are no longer blanked.** The action deliberately set both to `''` on this step. Nothing sets them in lint.yaml today, so there is no live impact—but add a cross-compilation matrix to this job later and `go install` produces a binary the runner cannot execute. This is the one guard the action carried that did not survive the inlining. Optional:

```yaml
        env:
          GOOS: ''
          GOARCH: ''
```

CLAUDE.md
```diff
@@ -186,7 +186,7 @@ _ = json.NewEncoder(w).Encode(resp)
 
 ## Important notes
 
-- **Go version**: 1.25.12 (specified in go.mod)
+- **Go version**: 1.25.13 (specified in go.mod)
```
geunwoonoh: **Suggestion:** this PR removed the version literal from four workflows but left it here. The README badge already reads go.mod through shields.io and never drifts, so applying the same rule takes the next patch bump down to a single line in go.mod.

```suggestion
- **Go version**: see the `go` directive in `go.mod`
```

geunwoonoh (dismissed):
Reviewed the full diff, verified the test-plan claims locally on go1.25.13, and read the upstream `golang/govulncheck-action` definition to confirm the inlining is faithful.

## Summary

This raises the `go` directive to 1.25.13 to clear four stdlib advisories, and along the way fixes the workflow toolchain pinning that the bump exposed. No Go source changes.

Two threads run through it: every `setup-go` now reads `go-version-file: go.mod` instead of carrying its own `1.25.x` wildcard, and `golang/govulncheck-action` is replaced by the same four steps inlined.

## Critical issues (0)

None.

## Warnings (2)

### 1. `.github/workflows/lint.yaml:56`—`go install ...@latest` is unpinned

To be precise about the baseline: this is not a regression. I read the upstream `action.yml` and it runs the identical `go install golang.org/x/vuln/cmd/govulncheck@latest` internally, so behavior is unchanged from before this PR.

Two things still argue for pinning now that the call is explicit and in-repo:

- Lines 33-35 of the same file pin golangci-lint to an exact `v2.9.0`. One file, two policies.
- `x/vuln` v1.7.0 shipped on 2026-08-13, so CI is currently picking up a one-day-old release. A govulncheck release that changes exit-code behavior or introduces a false positive breaks CI with no commit on our side.

```yaml
- name: Install govulncheck
  run: go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
```

A `tool` directive in go.mod (Go 1.24+) with `go tool govulncheck ./...` would add go.sum checksums and let Dependabot bump it, but the one-line pin is enough on its own.

### 2. `.github/workflows/lint.yaml:55-56`—the inlined install drops `GOOS`/`GOARCH`

The upstream action deliberately blanks both on that step:

```yaml
- name: Install govulncheck
  run: go install golang.org/x/vuln/cmd/govulncheck@latest
  env:
    GOOS: ''
    GOARCH: ''
```

Nothing sets GOOS/GOARCH in lint.yaml today, so there is no live impact. It is a latent trap: add a cross-compilation matrix to this job later and `go install` produces a binary the runner cannot execute. This is the one guard the action carried that did not survive the inlining.

## Suggestions (5)

### 3. `CLAUDE.md:189`—the Go version is still hardcoded

This PR removed the version literal from four workflows but left it in the docs. The README badge already reads go.mod through shields.io and never drifts. Applying the same rule here takes the next patch bump down to a single line in go.mod.

### 4. govulncheck only runs on pull requests and pushes

A newly disclosed CVE goes unnoticed until someone happens to commit. This PR is what that gap produces. A `schedule:` cron would surface it during quiet periods.

### 5. No Dependabot config

Now that go.mod is the single source for the toolchain, an automation that raises that one line would prevent a repeat of this manual scramble. The `gomod` and `github-actions` ecosystems cover both halves.

### 6. The `setup-go` + `GOTOOLCHAIN=auto` pair is duplicated four times

One in build-and-test, two in lint, one in release. A composite action would collapse four six-line copies into one; whether that is worth it at this size is a team call.

Worth recording either way: now that go.mod supplies an exact version, `GOTOOLCHAIN=auto` is effectively a no-op for the main module. It still matters in the govulncheck job, where `go install` has to satisfy `x/vuln`'s own go requirement. The other three are a safety net. Not arguing for removal—just worth knowing why each one is there.

### 7. Action SHA pinning (repo-wide, not this PR)

The upstream action pinned `actions/checkout` and `actions/setup-go` to full SHAs internally. Inlining moved those to the mutable `@v6` tags. The repo convention is tags everywhere, so this is not this PR's doing, but it is worth a separate look given what this CLI is.

## A note on the PR description

"so the file input is always ignored" claims a little more than the evidence supports. The action does expose a `go-version-file` input, and passing `go-version-input: ''` explicitly skips the composite default, so a workaround plausibly exists. The conclusion—inline it—still holds. Since this sentence becomes the permanent record of why the action was dropped, something closer to "the workaround is undocumented and fragile" would age better.

## Go conventions, dead code, performance

**Go conventions**—zero Go source files changed, so there is no surface to review. Verified locally on go1.25.13:

```
go build ./...       → clean
go vet ./...         → clean (exit 0)
go test -race ./...  → every package ok, no failures
```

This matches the description's claim that the 1.25.13 backports touch no API, signature, or documented behavior we depend on.

**Duplication and dead code**—nothing at the Go level. At the workflow level, only item 6; this PR reduces duplication by removing four `1.25.x` literals.

**Performance**—the govulncheck job runs in 16s end to end. `go install` finished in 0.66s because setup-go's module cache hit (`Cache hit occurred on the primary key ...` in the job log). Job composition is the same as with the action, so no regression.

One small new cost: pinning to an exact version makes setup-go bypass the runner tool cache and download directly.

```
Acquiring go1.25.13 from https://go.dev/dl/go1.25.13.linux-amd64.tar.gz
```

The actions/go-versions manifest lags upstream Go releases by a day or two. It costs roughly 6s per job, a fine price for determinism, and it disappears once the manifest catches up.

## Highlights

- The root-cause analysis holds up. I checked `action.yml` directly: `go-version-input` does default to `'stable'`, and setup-go does resolve `go-version` ahead of `go-version-file`.
- The four inlined steps are behaviorally identical to the action under default inputs—`-C .` and `-format text` are govulncheck's own defaults.
- `GOTOOLCHAIN=auto` is correctly ordered after `setup-go` in every job.
- Good commit split. Lint genuinely fails on the go.mod-only commit (`4ec0f1d`) and passes on the next one, so the reason the workflows had to change is preserved as evidence in the history.
- Walking all 24 backports and recording why none require a source change is the part I would most want other version bumps to copy.

---

**Verdict: approved**

Nothing blocks the merge. Both warnings are non-regressions; only the `@latest` pin is worth a one-line follow-up.

- [x] Follows stack conventions (no Go source changes; workflows keep repo idiom)
- [x] Tests pass (local `-race` all ok, all CI checks green)
- [x] No security vulnerabilities (`govulncheck`: No vulnerabilities found)
- [x] Docs updated (CLAUDE.md reflects the bump; see item 3)

.github/workflows/lint.yaml
jisung-02: Thanks — split the two points, since they landed differently.

**1. `@latest` stays unpinned.** Deliberate, and I should have said so in the description rather than leaving it to look like an oversight.

govulncheck does not carry its vulnerability database in the binary; it fetches from `https://vuln.go.dev` on every run (the `-db` default). So a pin loses nothing on advisory freshness — `alpacon-event-server` runs `@v1.6.0` and still sees every new CVE. What `@latest` actually buys is the scanner engine: new reachability analysis, new symbol matching. That is the one thing a pin cannot get, and for a security scan I would rather have it.

The asymmetry with golangci-lint `v2.9.0` is intentional too: a linter bump changes what fails the build on unchanged code, so it should be a commit. A scanner that starts detecting more is the outcome we want.

On the drift half of the concern — that is already closed in this file. `lint.yaml:53-54` exports `GOTOOLCHAIN=auto` before the install, so if `x/vuln` ever raises its own `go` directive above the runner's toolchain, Go downloads the newer one instead of failing. That is the failure mode this PR set out to remove, and the install step is inside it, not outside.

Accepting the risk you name: a govulncheck release that changes exit-code behavior or adds a false positive can break CI with no commit on our side. If that happens I would pin at that moment rather than pre-emptively.

**2. `GOOS`/`GOARCH` — fixed in `2f954d4`.** You are right that this is the one guard the action carried that did not survive the inlining, and "no live impact today" is exactly the kind of thing that stops being true quietly.

```yaml
      # Blank GOOS/GOARCH so the tool is always built for the runner, even if
      # this job ever gains a cross-compilation matrix.
      - name: Install govulncheck
        run: go install golang.org/x/vuln/cmd/govulncheck@latest
        env:
          GOOS: ""
          GOARCH: ""
```

CLAUDE.md
jisung-02: Applied as suggested in `aee9c66`.

```
- **Go version**: see the `go` directive in `go.mod`
```

You are right that this was the same rule left half-applied: the PR removed four `1.25.x` literals from the workflows and then left one in the docs. Checked the rest of the repo for others and there are none — `README.md:3` is the shields.io badge reading go.mod, and the `Dockerfile` is `FROM alpine:latest` with the binary copied in, so no toolchain literal there either. `CLAUDE.md:189` was the last one.

geunwoonoh (approved):
Follow-up review at `f4e81b6`. Re-verified the two fixes, worked through the reasoning on the one point left unchanged, and independently checked the new `.goreleaser.yaml` commit that was not part of the first round.

## Where the previous review landed

| # | Item | Outcome |
| --- | --- | --- |
| W1 | `go install ...@latest` unpinned | Deliberately unchanged, answered and documented |
| W2 | `GOOS`/`GOARCH` not blanked | Fixed in `2f954d4` |
| S3 | Go version hardcoded in `CLAUDE.md` | Fixed in `aee9c66` |
| S4-S7 | cron, Dependabot, composite action, action SHA pinning | Left as follow-ups |
| — | "the file input is always ignored" in the description | Reworded |

## Verifying the fixes

**W2 (`2f954d4`)** lands at lint.yaml:55-61, and the job log confirms both variables actually reach the step empty:

```
Install govulncheck  GOOS: 
Install govulncheck  GOARCH: 
```

`""` instead of the action's `''` is the same empty scalar in YAML, and Go treats an empty `GOOS`/`GOARCH` as unset. The comment explaining why the step needs them is a good addition—it is the part a future reader would otherwise delete as noise.

**S3 (`aee9c66`)** took the suggestion verbatim. Checking the rest of the repo for surviving literals independently gives the same answer: `README.md:3` is the shields.io badge reading go.mod, and the Dockerfile copies a prebuilt binary onto alpine, so neither carries a toolchain version.

**The description edit** now says the workaround "plausibly suppresses the composite default" but is undocumented and would break silently on an upstream change. That is a more accurate statement than the one I asked to soften, and it gives the next maintainer the real reason rather than a flat assertion.

## W1: accepting the decision, with one correction to my own framing

One of the arguments here corrects a premise I had wrong. govulncheck does not carry its vulnerability database in the binary; it fetches from `https://vuln.go.dev` on every run. So pinning costs nothing in advisory freshness, and my original note carried an implication that a pin means staler detection. That implication was wrong. What `@latest` actually buys is the scanner engine—reachability analysis and symbol matching—and that genuinely cannot be had from a pin.

The residual risk is unchanged: a govulncheck release that adds a false positive or changes exit-code behavior breaks CI with no commit on our side. Accepting it explicitly and pinning reactively is a reasonable call on a judgment question, and it is now written down where the next person will find it. Not a blocker.

Two things I would still leave on the record.

The asymmetry argument against golangci-lint `v2.9.0` is softer than it reads. A govulncheck bump also has a "build fails on unchanged code" path—that is the risk accepted two paragraphs later. What actually separates them is the value judgment, not the mechanism: a scanner finding more is the outcome we want, a linter finding more is noise. That distinction holds; the mechanism-based framing does not quite.

And `alpacon-event-server` pinning `@v1.6.0` means the org now runs two policies on the same tool. Nothing for this PR to solve, but worth reconciling once somewhere.

## New in this round: dropping the goreleaser before hooks (`f4e81b6`)

Checked independently since it was not in the first review. It is safe, and it is slightly more than cleanup.

- `go generate ./...` was a genuine no-op—a full-repo search finds zero `go:generate` directives.
- `go mod tidy` is already gated upstream: the `goreleaser` job declares `needs: [build-and-test]`, and build-and-test runs `go mod tidy` followed by `git diff --exit-code go.mod go.sum`. Re-running it locally produces no diff.
- The hook could also mutate go.mod/go.sum inside the release working tree. So its only possible outcomes were: no-op, a binary built from a module graph different from the tested one, or a confusing dirty-tree failure mid-release. There was no case where it helped.

One caveat worth stating plainly: **the green checks on this PR do not cover this commit.** release.yaml only triggers on tag pushes, so nothing in PR CI exercises `.goreleaser.yaml`. I could not close that gap locally either—my `goreleaser` binary is v2 while this config is v1-style and CI pins `version: "~> v1"`. What I did confirm is that the file still parses with all nine remaining top-level keys intact, and that `before` was an optional key to begin with. Deleting an optional key is about as low-risk as a change to this file gets, but it rests on reasoning rather than on a passing job.

On scope: this sits outside "raise the go directive". It is small and the commit table documents it, so it is fine here. Since it touches the release path, though, that is the kind of change I would rather see land on its own.

## Verification at `f4e81b6`

```
go build ./...                       → clean
go vet ./...                         → clean
go mod tidy && git diff --exit-code  → no diff
```

CI is green on this exact SHA: Build and Test (which runs `go test -race -v ./...`), golangci-lint, govulncheck, and CodeQL. govulncheck reports `No vulnerabilities found.`

## Remaining follow-ups (out of scope here)

S4 (scheduled govulncheck run), S5 (Dependabot config), S6 (composite action for the `setup-go` + `GOTOOLCHAIN=auto` pair), S7 (action SHA pinning). None block this PR; worth filing so they do not evaporate.

---

**Verdict: approved**

Both code-level points were addressed, and the one left alone came back with reasoning that corrected a premise of mine rather than deflecting it. The new goreleaser commit checks out independently.

- [x] Follows stack conventions (no Go source changes; workflows keep repo idiom)
- [x] Tests pass (CI green on `f4e81b6`; build, vet, and tidy verified locally)
- [x] No security vulnerabilities (`govulncheck`: No vulnerabilities found)
- [x] Docs updated (`CLAUDE.md` now points at the `go` directive; description reworded)

PR #348, opened 2026-08-14
refactor(ftp): drop the unreachable colon checks in validatePaths

### LINKED ISSUE #340: validatePaths carries an unreachable missing-colon branch
## Background

Flagged during review of #337 ([comment](https://github.com/alpacax/alpacon-cli/pull/337#pullrequestreview-4923797679), Suggestion 4). Pre-existing code, outside that PR's scope.

## Problem

In `cmd/ftp/cp.go`, inside the `if isRemotePath(path)` block of `validatePaths`, the `if !strings.Contains(path, ":")` branch cannot run. `isRemotePath` delegates to `utils.IsRemoteTarget`, which strips a `user@` prefix before checking for a colon (`utils/ssh_parser.go`)—and `splitUserPrefix` only strips that prefix when the `@` sits before any colon, so whenever the stripped target contains a colon, the full string does too.

The `len(parts) != 2` check a few lines below is dead for the same reason: `strings.SplitN(path, ":", 2)` on a string that contains a colon always yields two parts.

## Suggested fix

Now that `utils.SplitPath` owns the missing-colon case (#337), `validatePaths` can keep just the `parts[0] == "" || parts[1] == ""` check and drop the dead branch along with its error text.

jisung-02:
## Background

`validatePaths` (`cmd/ftp/cp.go:216`) rejected a remote path with no colon from inside a branch that only runs when the path already has one.

`isRemotePath` calls `utils.IsRemoteTarget` (`utils/ssh_parser.go:39`), which passes the target through `splitUserPrefix` and then looks for a colon in what is left. `splitUserPrefix` (`utils/ssh_parser.go:44`) removes a leading `user@` and never inserts characters, so a colon surviving the strip was in the original path. `isRemotePath(path)` being true therefore implies `strings.Contains(path, ":")`.

Two checks were unreachable as a result:

- the `!strings.Contains(path, ":")` guard and the error message it returned
- the `len(parts) != 2` clause after `strings.SplitN(path, ":", 2)` — splitting a string that contains the separator always yields exactly two parts

## Changes

Both are gone. The check that does fire — server name and remote path both non-empty — now splits through `utils.SplitPath` instead of a second hand-rolled `SplitN`:

```go
if _, remotePath, err := utils.SplitPath(path); err != nil || remotePath == "" {
```

`utils.SplitPath` (`utils/utils.go:596`) is the helper `uploadObject` (`cmd/ftp/cp.go:259`) and `downloadObject` (`cmd/ftp/cp.go:302`) already call a few lines below. A missing colon and an empty server name come back as its error; the empty remote path stays a local condition, because `SplitPath` documents an empty path as accepted and left to the caller — a contract those two callers depend on.

No behavior changes. For any path that reaches the branch:

| Input | Before | After |
|---|---|---|
| `myserver:/tmp/f` | accepted | accepted |
| `:/tmp/f` | rejected, missing server name | rejected, `SplitPath` error |
| `myserver:` | rejected, missing path | rejected, `remotePath == ""` |

Removing the guard opens no validation gap for a path with no colon at all: `isRemotePath` classifies it as local, and `utils.SplitPath` rejects it at the `uploadObject`/`downloadObject` call sites (added in #337).

## Testing

`validatePaths` had no test, so `TestValidatePaths` now pins what survives — both accepted directions, the mixed local and remote source rejection, the empty server name, and the empty remote path. It follows `cmd/login_test.go:609` for the `wantErrSub` substring shape.

- `go test -race -count=1 ./...` — 39 packages ok, no failures
- `gofmt -l .` and `go vet ./...` — no output
- Every commit builds and passes on its own
- Rebased onto `origin/main` (`19afae6`); all of the above re-run after the rebase

## Review follow-up

Three suggestions from review, all applied in the two commits on top:

- `TestValidatePaths` gains a case for `admin@myserver:/tmp/file.txt`. `isRemotePath` strips a leading `user@`, `SplitPath` does not, so a path reaching `validatePaths` with the prefix intact is accepted with `admin@myserver` as the server name. `Run` (`cmd/ftp/cp.go:78`) rewrites the argument first, so the input is unreachable in practice and the behavior is the same before and after this PR — but that asymmetry is what the unreachability argument above rests on, and the case now fails if `IsRemoteTarget` changes.
- The comment above the guard now covers the `err != nil` half as well. `SplitPath`'s error is discarded on purpose: `cp` prints a returned error verbatim through `CliErrorWithExit`, and the guidance message names both failure shapes.
- The narration comment on `TestCpCommandSSHParsing` is restored. It was removed as a redundant restatement of the function name, but it belongs to a test this branch does not otherwise touch.

Closes #340

### DIFF
--- a/cmd/ftp/cp.go
+++ b/cmd/ftp/cp.go
@@ -240,16 +240,8 @@ func validatePaths(sources []string, dest string) error {
 	allPaths = append(allPaths, dest)
 	for _, path := range allPaths {
 		if isRemotePath(path) {
-			if !strings.Contains(path, ":") {
-				return fmt.Errorf("invalid remote path format: '%s'\n\n"+
-					"Remote paths must be in format 'servername:/path'\n"+
-					"Examples:\n"+
-					"  • myserver:/home/user/file.txt\n"+
-					"  • web-server:/var/www/index.html", path)
-			}
-
-			parts := strings.SplitN(path, ":", 2)
-			if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+			// SplitPath accepts an empty remote path; cp requires one.
+			if _, remotePath, err := utils.SplitPath(path); err != nil || remotePath == "" {
 				return fmt.Errorf("invalid remote path format: '%s'\n\n"+
 					"Remote paths must include both server name and path:\n"+
 					"  • Correct: myserver:/path/to/file\n"+
--- a/cmd/ftp/cp.go
+++ b/cmd/ftp/cp.go
@@ -240,7 +240,8 @@ func validatePaths(sources []string, dest string) error {
 	allPaths = append(allPaths, dest)
 	for _, path := range allPaths {
 		if isRemotePath(path) {
-			// SplitPath accepts an empty remote path; cp requires one.
+			// SplitPath accepts an empty remote path; cp requires one. Its error
+			// text is replaced by the guidance message below.
 			if _, remotePath, err := utils.SplitPath(path); err != nil || remotePath == "" {
 				return fmt.Errorf("invalid remote path format: '%s'\n\n"+
 					"Remote paths must include both server name and path:\n"+
--- a/cmd/ftp/cp_test.go
+++ b/cmd/ftp/cp_test.go
@@ -202,6 +202,7 @@ func TestValidatePaths(t *testing.T) {
 	}
 }
 
+// Test the SSH parsing logic used in the cp command
 func TestCpCommandSSHParsing(t *testing.T) {
 	tests := []struct {
 		name         string
--- a/cmd/ftp/cp_test.go
+++ b/cmd/ftp/cp_test.go
@@ -164,6 +164,13 @@ func TestValidatePaths(t *testing.T) {
 			dest:    "./",
 		},
 		{
+			// Run strips the user prefix before validatePaths sees it; this pins
+			// that SplitPath itself does not, so the server name keeps "admin@".
+			name:    "Remote source with a user prefix",
+			sources: []string{"admin@myserver:/tmp/file.txt"},
+			dest:    "./",
+		},
+		{
 			name:       "Mixed local and remote sources",
 			sources:    []string{"./test.txt", "myserver:/tmp/file.txt"},
 			dest:       "/tmp/",
--- a/cmd/ftp/cp_test.go
+++ b/cmd/ftp/cp_test.go
@@ -146,7 +146,55 @@ func TestIsLocalPaths(t *testing.T) {
 	}
 }
 
-// Test the SSH parsing logic used in the cp command
+func TestValidatePaths(t *testing.T) {
+	tests := []struct {
+		name       string
+		sources    []string
+		dest       string
+		wantErrSub string
+	}{
+		{
+			name:    "Local sources with remote destination",
+			sources: []string{"./test.txt", "/home/user/file.txt"},
+			dest:    "myserver:/tmp/",
+		},
+		{
+			name:    "Remote source with local destination",
+			sources: []string{"myserver:/tmp/file.txt"},
+			dest:    "./",
+		},
+		{
+			name:       "Mixed local and remote sources",
+			sources:    []string{"./test.txt", "myserver:/tmp/file.txt"},
+			dest:       "/tmp/",
+			wantErrSub: "cannot mix local and remote source paths",
+		},
+		{
+			name:       "Remote source without server name",
+			sources:    []string{":/tmp/file.txt"},
+			dest:       "./",
+			wantErrSub: "invalid remote path format",
+		},
+		{
+			name:       "Remote destination without path",
+			sources:    []string{"./test.txt"},
+			dest:       "myserver:",
+			wantErrSub: "invalid remote path format",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validatePaths(tt.sources, tt.dest)
+			if tt.wantErrSub == "" {
+				assert.NoError(t, err)
+				return
+			}
+			assert.ErrorContains(t, err, tt.wantErrSub)
+		})
+	}
+}
+
 func TestCpCommandSSHParsing(t *testing.T) {
 	tests := []struct {
 		name         string


cmd/ftp/cp.go
```diff
@@ -240,16 +240,8 @@ func validatePaths(sources []string, dest string) error {
...
-				return fmt.Errorf("invalid remote path format: '%s'\n\n"+
-					"Remote paths must be in format 'servername:/path'\n"+
-					"Examples:\n"+
-					"  • myserver:/home/user/file.txt\n"+
-					"  • web-server:/var/www/index.html", path)
-			}
-
-			parts := strings.SplitN(path, ":", 2)
-			if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+			// SplitPath accepts an empty remote path; cp requires one.
```
geunwoonoh: Suggestion: the comment explains the `remotePath == ""` half but not the `err != nil` half.

Discarding `SplitPath`'s error here is the right call — the CLI prints the returned error verbatim through `CliErrorWithExit("%s", err.Error())`, and the guidance message below is friendlier than `invalid remote path \"...\": missing server name`. Saying so in the comment makes the whole line self-explaining.

```suggestion
			// SplitPath accepts an empty remote path; cp requires one. Its error
			// text is replaced by the guidance message below.
```

cmd/ftp/cp_test.go
```diff
@@ -146,7 +146,55 @@ func TestIsLocalPaths(t *testing.T) {
...
+			sources:    []string{":/tmp/file.txt"},
+			dest:       "./",
+			wantErrSub: "invalid remote path format",
+		},
+		{
+			name:       "Remote destination without path",
+			sources:    []string{"./test.txt"},
+			dest:       "myserver:",
+			wantErrSub: "invalid remote path format",
+		},
```
geunwoonoh, line 190: Suggestion: consider a case for `admin@myserver:/tmp/file.txt`.

`isRemotePath` strips a leading `user@` via `splitUserPrefix` before deciding; `SplitPath` does not. An input reaching `validatePaths` with the prefix intact would therefore split into a server name of `admin@myserver` and still be accepted.

This is unreachable in production — the argument-rewriting loop in `Run` (cmd/ftp/cp.go:78) rewrites `admin@myserver:/tmp` to `myserver:/tmp` before `validatePaths` runs — and the behavior is identical before and after this PR. But that asymmetry between the two helpers is exactly what the unreachability argument rests on, so pinning it leaves a signal for whoever changes `IsRemoteTarget` next.

```suggestion
		},
		{
			// Run strips the user prefix before validatePaths sees it; this pins
			// that SplitPath itself does not, so the server name keeps "admin@".
			name:    "Remote source with a user prefix",
			sources: []string{"admin@myserver:/tmp/file.txt"},
			dest:    "./",
		},
```

cmd/ftp/cp_test.go
```diff
@@ -146,7 +146,55 @@ func TestIsLocalPaths(t *testing.T) {
...
+			if tt.wantErrSub == "" {
+				assert.NoError(t, err)
+				return
+			}
+			assert.ErrorContains(t, err, tt.wantErrSub)
+		})
+	}
+}
+
 func TestCpCommandSSHParsing(t *testing.T) {
```
geunwoonoh, line 206: Suggestion: this test's doc comment (`// Test the SSH parsing logic used in the cp command`) was removed in this diff, and the PR body does not mention it.

The removal is defensible on its own — the comment only restated the function name. But it belongs to a test this PR otherwise does not touch, so a reviewer has to stop and work out whether it was deliberate. Either restore it or add a line to the PR body.

geunwoonoh (dismissed):
## Review summary

This PR drops two branches in `validatePaths` that could never execute, and routes the surviving check through `utils.SplitPath` — the same helper `uploadObject` and `downloadObject` already call a few lines below. `validatePaths` had no test, so `TestValidatePaths` lands alongside it.

The unreachability argument holds. `IsRemoteTarget` strips a leading `user@` through `splitUserPrefix` and then looks for a colon in what is left; `splitUserPrefix` only removes a prefix and never inserts characters, so a colon surviving the strip was in the original string. `isRemotePath(path)` being true therefore implies `strings.Contains(path, ":")`, and `strings.SplitN` on a string that contains the separator always returns exactly two parts.

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (3, all nits)

1. **[cmd/ftp/cp_test.go:198]** The `// Test the SSH parsing logic used in the cp command` comment above `TestCpCommandSSHParsing` was removed in this diff with no mention in the PR body. The removal itself is defensible — the comment only restated the test name — but a comment quietly disappearing from an otherwise untouched test makes a reviewer stop and check. Either restore it or add a line to the PR body.

2. **[cmd/ftp/cp.go:243]** The error `SplitPath` produced is discarded in favour of the local guidance message. That is the right call — the CLI prints it verbatim via `CliErrorWithExit("%s", err.Error())` and the guidance is friendlier — and it matches the previous behavior. The comment currently explains only the `remotePath == ""` half, though; folding in the discarded error would make the whole line self-explaining.

3. **[cmd/ftp/cp_test.go:183]** No case covers `admin@myserver:/tmp/file.txt`. `isRemotePath` strips the `user@` prefix before deciding; `SplitPath` does not, so that input reaching `validatePaths` directly would yield a server name of `admin@myserver`. It is unreachable in practice because the argument-rewriting loop in `Run` (`cmd/ftp/cp.go:78`) strips the prefix first, and the before/after behavior is identical either way. But this asymmetry is exactly what the refactor's correctness argument leans on, so pinning it gives a signal to whoever touches `IsRemoteTarget` next.

### Go conventions

| Item | Result |
|---|---|
| errcheck (all error returns handled) | Pass — `err` is checked directly in the condition |
| Declaration order `const → var → type → func` | Unchanged, preserved |
| Error strings lowercase, no trailing punctuation (ST1005) | Pass (`invalid remote path format:`) |
| Table-driven tests with testify | Pass |
| `assert.ErrorContains` | Pass — avoids wrapping `strings.Contains` in `assert.True`, per the CLAUDE.md assertion rules |
| No message restating the assertion | Pass — the helpers print operands themselves |
| `wantErrSub` field name | Matches the `cmd/login_test.go:609` shape |
| Comments in English | Pass |
| Sentence-case test names | Pass |
| `gofmt -l`, `go vet` | No output |

### Duplicate and unused code

The two branches this PR deletes are genuinely dead. What remains is pre-existing and out of scope here:

- `isRemotePath` (`cp.go:198`) and `isLocalPath` (`cp.go:203`) are one-line pass-throughs to `utils.IsRemoteTarget` / `utils.IsLocalTarget`, and `isLocalPath(x)` is by definition `!isRemotePath(x)`. An indirection layer that earns nothing, but it predates this PR.
- `validatePaths` walks `sources` twice — once for the mixed check, once for the format check. With CLI-argument counts there is no reason to merge them, and the current split reads better.

### Performance

Nothing regresses, and it gets marginally faster:

- Before, per remote path: `strings.Contains` (one `Index`) plus `strings.SplitN` (one heap allocation for a two-element `[]string`)
- After: `strings.Cut` (one `Index`, no allocation)

One fewer heap allocation per path. `N` is the CLI argument count, so this is immaterial in practice, but the direction is right. `allPaths` is also already pre-sized with `make([]string, 0, len(sources)+1)`, so there is no regrowth.

### What I verified locally

- `go test -race -count=1 ./...` — all packages pass
- `go vet ./cmd/ftp/...` and `gofmt -l cmd/ftp/` — no output
- **Differential equivalence test**: generated every combination of length 1–3 over a 13-atom alphabet (`""`, `:`, `@`, `a`, `/`, `~`, `server`, `/tmp/f`, `a@b`, `::`, space, `\`, `c:`), kept the **1269** inputs that `isRemotePath` classifies as remote, and compared the old branch body against the new one for accept/reject. **Zero mismatches** — the "No behavior changes" claim is confirmed empirically. The scratch test file was deleted after the run and is not part of the branch.
- `validatePaths` statement coverage: 0% → **100%**
- `golangci-lint` is not installed on this machine, so I could not re-verify the `0 issues.` claim independently. CI covers it.

### Highlights

- Deletes dead code and, in the same move, funnels the surviving check into a helper the file already used. Path parsing now converges on one place.
- The comment pins down that `SplitPath` accepts an empty remote path while `cp` does not. That is the kind of note that stops someone from deleting the `remotePath == ""` check later.
- A function that had no test now has one, aimed precisely at the branches that survive.
- The PR body carries the unreachability argument and a before/after table per input, which makes this cheap to review.

---

**Verdict**: **APPROVED**

**Checklist**:
- [x] Documentation updated if needed (no behavior change, so README/CLAUDE.md need nothing)

cmd/ftp/cp.go
jisung-02: Applied in d908c59, taking the suggestion verbatim. The discarded error was deliberate but nothing said so, which is exactly the kind of line that reads as an oversight later.

cmd/ftp/cp_test.go
jisung-02: Applied in 6a01bf3. I had dropped this case as redundant while writing the table — the argument-rewriting loop makes it unreachable — but you are right that the asymmetry is load-bearing for the unreachability argument, so it belongs in the table as a tripwire on `IsRemoteTarget`. Placed with the other accepted cases rather than at the end, so the table stays grouped accepted-then-rejected.

cmd/ftp/cp_test.go
jisung-02: Restored in d908c59. Removing it was intentional and noted in the test commit body, but a trailer is not where a reviewer looks. The PR body now has a "Review follow-up" section covering all three points as well.

geunwoonoh (approved):
## Follow-up review (head `d908c597`)

All three suggestions from the previous round are applied, with a reply on each thread and a `Review follow-up` section in the PR body. Re-approving on the new head, since the earlier approval was dismissed when the commits landed.

| Suggestion | Commit | State |
|---|---|---|
| Cover the `err != nil` half in the guard comment | `d908c59` | Taken verbatim |
| Add an `admin@myserver:` case to `TestValidatePaths` | `6a01bf3` | Applied, placed with the other accepted cases so the table stays grouped accepted-then-rejected |
| Restore the narration comment on `TestCpCommandSSHParsing` | `d908c59` | Restored |

### Verified locally at `d908c597`

- `go build ./...` — passes
- `go test -race -count=1 ./...` — every package passes
- `go vet ./...` and `gofmt -l .` — no output
- `validatePaths` statement coverage — still 100%
- Differential equivalence sweep re-run: every combination of length 1–3 over a 13-atom alphabet, restricted to the **1269** inputs `isRemotePath` classifies as remote, old branch body vs new — **zero mismatches**. The logic is unchanged from the previous head; only the comment moved.

### One nit — and the wording is mine

**[cmd/ftp/cp_test.go:166-171, and the matching paragraph in the PR body]** The new case claims more than it pins.

The comment says it pins that `SplitPath` keeps the prefix "so the server name keeps `admin@`", and the PR body says "the case now fails if `IsRemoteTarget` changes". Neither is observable from the assertion: `validatePaths` discards the server name (`_`), and the case only asserts `NoError`. Measured:

```
prefix remote   (admin@myserver:/tmp/file.txt) : <nil>
stripped remote (myserver:/tmp/file.txt)       : <nil>
treated local   (./test.txt)                   : <nil>
```

Whether `SplitPath` starts stripping the prefix, or `IsRemoteTarget` starts classifying this input as local, the case stays green. What it actually guards is narrower but still real: `validatePaths` must not start rejecting a user-prefixed remote path.

I wrote that comment in the previous round and it was taken verbatim, so the overclaim originated on my side — flagging it rather than leaving it to mislead the next reader.

Either fix works, and one is enough:

1. Lower the claim to what holds — the case pins that a user-prefixed remote path is accepted. The PR body paragraph needs the same trim.
2. Put the real tripwire in `TestSplitPath` (`utils/utils_test.go:432`), whose table already asserts `wantServer` directly and currently has no `user@` case:

```go
{"user prefix is kept", "admin@myserver:/tmp/f", "admin@myserver", "/tmp/f", ""},
```

For reference, the input with discriminating power inside `validatePaths` is `admin@:/tmp/f`: accepted today because the server name is `admin@`, and it would flip to rejected if `SplitPath` ever stripped the prefix.

### Verdict

**APPROVED.** The refactor itself is unchanged from the round I already checked, and equivalence is re-confirmed at this head. The nit above is comment and prose only — no effect on shipped behavior — so it does not block the merge; take it or leave it.

**Checklist**:
- [x] Documentation updated if needed

Issue #330, opened 2026-08-13
fix(note): note list silently truncates to a single page


jisung-02:
## Background

Every list function in `api/` paginates through `api.FetchAllPages`—except `GetNoteList` (`api/note/note.go:17-34`), which issues a single `SendGetRequest` with an optional `page_size` param and returns whatever one page holds. When notes exceed one page, the output is silently truncated: no indicator, no way to reach the rest.

There is a second, related inconsistency at the cmd layer: `cmd/note/note_list.go:39` binds the `--tail`/`-t` flag ("Number of log entries to show from the end") to a variable named `pageSize`, which is then sent as `page_size`. The flag's documented semantic (last N entries) and its actual effect (first page of size N) disagree.

## Proposal

Convert `GetNoteList` to `api.FetchAllPages` like every other domain, and make `--tail` mean what it says—fetch all (or enough) and keep the last N, matching the tail semantics used elsewhere in the CLI. Decide whether a plain `note ls` with no flag should fetch everything (consistent with other list commands) or keep a default cap that is at least visible in the output.

## References

- `api/note/note.go:17-34`—single-request list, no pagination loop
- `cmd/note/note_list.go:39`—`--tail` flag bound to `pageSize`
- `api/pagination.go`—`FetchAllPages`, the standard used by every other domain

PR #351, opened 2026-08-14
fix: honor the --tail contract in websh ls and exec ls

### LINKED ISSUE #338: fix(cli): websh ls and exec ls break the --tail contract
## Background

`#330` settles what `--tail N` means across the CLI: **the newest N entries**. `log`, `audit`, and `webftp` already behave that way—each passes `tail` as the limit to `api.FetchCursorPages`, and the Elasticsearch-backed endpoints sort `-date` descending, so the first page holds the newest rows.

Two commands break that contract in two different ways.

## 1. `websh ls --tail` returns the oldest sessions

`cmd/websh/websh_list.go:29-31` fetches every page and then slices the *physical* end of the list:

```go
if tail > 0 && tail < len(sessionList) {
    sessionList = sessionList[len(sessionList)-tail:]
}
```

The sessions endpoint sorts newest first (`SessionViewSet.ordering = ['-added_at']` in alpacon-server `websh/api/views.py:130`), so the end of that list is the *oldest* sessions. `alpacon websh ls --tail 5` reports the five oldest sessions while the flag help says "Number of sessions to show".

Fix: slice the head (`sessionList[:tail]`), since the list is already ordered newest first.

## 2. `exec ls --tail` cannot exceed one page

`api/event/event.go:213-225` sends `page_size=tail` in a single request and never looks at `ListResponse.Next`:

```go
params := map[string]string{}
if pageSize > 0 {
    params["page_size"] = fmt.Sprintf("%d", pageSize)
}
responseBody, err := ac.SendGetRequest(utils.BuildURL(getEventURL, relativePath, params))
```

The server caps a page at 100 (`max_page_size` in alpacon-server `api/pagination.py`), so `alpacon exec ls --tail 500` silently returns 100 entries. This is the same defect shape as `#330`.

Fix: reuse the limit-bounded helper `#330` adds to `api/pagination.go`, so no second page loop is written:

```go
endpoint := path.Join(getEventURL, relativePath)
events, err := api.FetchPagesUpTo[EventDetails](ac, endpoint, nil, tail)
```

`api/event/chunks.go:38` already assembles an endpoint this way before calling `api.FetchAllPages`, so the shape matches existing code.

## Order

Part 2 depends on `#330` landing first—it calls the helper that PR introduces. Part 1 is independent and can go first.

## References

- `cmd/websh/websh_list.go:29-31`—tail slice on a descending list
- `api/event/event.go:213-225`—single request, `Next` ignored
- `api/pagination.go`—`FetchAllPages`, `FetchCursorPages`
- `#330`—`note ls` pagination and the `--tail` contract

jisung-02:
## Background

Issue #330 settled `--tail N` as the newest N entries, not the physical end of a descending list. `websh ls` and `exec ls` both broke that contract, each in its own way.

`websh ls` pulled every page with `FetchAllPages` and then sliced `sessionList[len-tail:]`. The endpoint sorts newest first (`SessionViewSet.ordering = ['-added_at']`), so that slice handed back the oldest N sessions.

`exec ls` read the right end of the list (`CommandViewSet.ordering = ['-scheduled_at']`) but could not reach past one page: it sent `page_size=tail` in a single request and the server caps `page_size` at 100, so `--tail 250` quietly returned 100 entries.

## Changes

Both commands now go through `api.FetchPagesUpTo`, the helper #330 added: it takes tail as its limit and stops once it holds that many items. `note ls` already reads its tail this way.

| Command | Before | After |
|---|---|---|
| `websh ls` | fetch every page, slice from the end | stop once tail items are held |
| `exec ls` | single `page_size=tail` request | walk pages up to tail |

`FetchPagesUpTo` returns nothing for a limit of 0, where the old `websh ls` code read 0 as "no limit". `utils.RequirePositiveInt` keeps that from turning into a silently empty table, matching every other `--tail` command.

Help text is aligned in the same pass. `exec ls` said "most recent last" in its `Long` and "from the end" in its `--tail` help, describing neither the endpoint's order nor the flag's contract. The deprecated `alpacon event` shares that code path, and `log`, `audit` and `webftp-log` already returned newest-first while their help said otherwise. `websh ls` said only "Number of sessions to show", which named no end at all. All of them now use the wording `note ls` already had: "Number of X to show, newest first".

`websh ls` also picks up the `-t` shorthand that `note`, `audit`, `log`, `webftp` and `exec` all register, and the two list projections in `api/event` and `api/websh` are preallocated now that the final length is known before the loop.

## Testing

`TestGetEventList_NoExtraPagination` pinned the old single-request behavior and is replaced by `TestGetEventList_PassesTailAsTheLimit`, which runs `--tail 250` against a fake server serving five full pages and asserts exactly 250 commands come back—unreachable unless tail arrives as the helper's limit. The fake server stops at five pages rather than paginating forever, so a regression to an unbounded walk fails on the length immediately instead of hanging until the test timeout. `api/websh` gets the same test for `GetSessionList`. The page walk itself stays pinned in `api.TestFetchPagesUpTo_*`.

`TestGetEventList_PutsTheResolvedFilterIDsInThePath` covers the other half of the `exec ls` rewrite. `--server` and `--user` are resolved to IDs and joined into the request path, not passed as query params, so a broken join does not fail the request—it returns every command as if no filter had been given. It is table-driven over all three cases: both filters, server only, and user only. The last one is recorded as current behavior rather than as the desired contract—see the Residual section.

`cmd/websh` gains what `cmd/exec` already had. `TestWebshLsRoutesAndParsesFlags` pins that `ls` still routes and parses flags under the parent's `DisableFlagParsing`, and `TestWebshLsTailZeroExitsWithUsageErrorCode` pins the exit code 2 rejection of `--tail 0` through a helper process, since `RequirePositiveInt` exits the process.

`go build ./...`, `go vet ./...` and `go test -race ./...` all pass.

## Residual

Server-side ordering has no tiebreaker: `CommandViewSet` and `SessionViewSet` both sort on a single key, so equal sort values can repeat one row and skip another at a page boundary under offset pagination. Not fixable from the CLI—filed as alpacax/alpacon-server#3044.

`exec ls --user NAME` with no `--server` builds a URL that collides with the command detail route, because `path.Join` drops the empty server segment. Pre-existing, not introduced here—the old code produced the same URL. Which URL the server expects for a user-only filter cannot be settled from the CLI, so it is filed as #363 and the current behavior is pinned by a test in the meantime.

Closes #338

### DIFF
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -217,7 +217,7 @@ func GetEventList(ac *client.AlpaconClient, tail int, serverName string, userNam
 		return nil, err
 	}
 
-	var eventList []EventAttributes
+	eventList := make([]EventAttributes, 0, len(events))
 	for _, event := range events {
 		eventList = append(eventList, EventAttributes{
 			Server:      event.Server.Name,
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -193,7 +193,9 @@ func gapFillInterval(noProgress int) time.Duration {
 	return d
 }
 
-func GetEventList(ac *client.AlpaconClient, pageSize int, serverName string, userName string) ([]EventAttributes, error) {
+// GetEventList returns the newest tail commands. The endpoint sorts -scheduled_at,
+// so the newest ones arrive first and the walk stops as soon as tail is reached.
+func GetEventList(ac *client.AlpaconClient, tail int, serverName string, userName string) ([]EventAttributes, error) {
 	var serverID, userID string
 	var err error
 	if serverName != "" {
@@ -209,23 +211,14 @@ func GetEventList(ac *client.AlpaconClient, pageSize int, serverName string, use
 		}
 	}
 
-	relativePath := path.Join(serverID, userID)
-	params := map[string]string{}
-	if pageSize > 0 {
-		params["page_size"] = fmt.Sprintf("%d", pageSize)
-	}
-	responseBody, err := ac.SendGetRequest(utils.BuildURL(getEventURL, relativePath, params))
+	endpoint := path.Join(getEventURL, serverID, userID)
+	events, err := api.FetchPagesUpTo[EventDetails](ac, endpoint, nil, tail)
 	if err != nil {
 		return nil, err
 	}
 
-	var response api.ListResponse[EventDetails]
-	if err = json.Unmarshal(responseBody, &response); err != nil {
-		return nil, err
-	}
-
 	var eventList []EventAttributes
-	for _, event := range response.Results {
+	for _, event := range events {
 		eventList = append(eventList, EventAttributes{
 			Server:      event.Server.Name,
 			Shell:       event.Shell,
--- a/api/event/event_test.go
+++ b/api/event/event_test.go
@@ -18,66 +18,81 @@ import (
 	"time"
 
 	"github.com/alpacax/alpacon-cli/api"
-	"github.com/alpacax/alpacon-cli/api/types"
 	"github.com/alpacax/alpacon-cli/client"
 	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestGetEventList_NoExtraPagination(t *testing.T) {
-	var eventRequestCount atomic.Int32
+// The page walk itself is pinned in api.TestFetchPagesUpTo_*; what is specific to
+// GetEventList is that tail reaches the helper as its limit, so a tail larger than the
+// server's 100-item page cap still yields exactly tail commands.
+func TestGetEventList_PassesTailAsTheLimit(t *testing.T) {
+	// Twice the pages the tail needs, so a walk that ignored the limit comes back with 500
+	// and fails the length assertion instead of running until the test timeout.
+	const lastPage = 5
 
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 
-		count := eventRequestCount.Add(1)
-		if count > 1 {
-			t.Errorf("extra request detected: request #%d to %s (should be single request)", count, r.URL.String())
+		size, sizeErr := strconv.Atoi(r.URL.Query().Get("page_size"))
+		page, pageErr := strconv.Atoi(r.URL.Query().Get("page"))
+		if sizeErr != nil || pageErr != nil {
+			t.Errorf("page and page_size must be integers, got page=%q page_size=%q",
+				r.URL.Query().Get("page"), r.URL.Query().Get("page_size"))
+			http.Error(w, "bad pagination query", http.StatusBadRequest)
 			return
 		}
 
-		pageSize := r.URL.Query().Get("page_size")
-		if pageSize != "25" {
-			t.Errorf("expected page_size=25, got %s", pageSize)
+		next := page + 1
+		if page >= lastPage {
+			next = 0
 		}
-
-		var results []EventDetails
-		for i := range 25 {
-			results = append(results, EventDetails{
-				ID:          fmt.Sprintf("evt-%d", i),
-				Server:      types.ServerSummary{Name: "test-server"},
-				Shell:       "bash",
-				Line:        fmt.Sprintf("cmd-%d", i),
-				RequestedBy: types.UserSummary{Name: "admin"},
-			})
-		}
-
-		resp := api.ListResponse[EventDetails]{
-			Count:   200, // more items exist on server
-			Results: results,
-		}
-		_ = json.NewEncoder(w).Encode(resp)
+		results := make([]EventDetails, size)
+		_ = json.NewEncoder(w).Encode(api.ListResponse[EventDetails]{Next: next, Results: results})
 	}))
 	defer ts.Close()
 
-	ac := &client.AlpaconClient{
-		HTTPClient: ts.Client(),
-		BaseURL:    ts.URL,
-	}
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
 
-	events, err := GetEventList(ac, 25, "", "")
-	if err != nil {
-		t.Fatalf("GetEventList error: %v", err)
-	}
+	events, err := GetEventList(ac, 250, "", "")
 
-	totalRequests := int(eventRequestCount.Load())
-	if totalRequests != 1 {
-		t.Errorf("expected 1 request, got %d", totalRequests)
-	}
-	if len(events) != 25 {
-		t.Errorf("expected 25 events, got %d", len(events))
-	}
+	require.NoError(t, err)
+	assert.Len(t, events, 250)
+}
+
+// The --server and --user filters are path segments, not query params, so a broken join
+// does not fail the request—it returns every command as if no filter had been given.
+func TestGetEventList_PutsTheResolvedFilterIDsInThePath(t *testing.T) {
+	// Mirrors the lookup endpoints of api/server and api/iam, whose consts are unexported.
+	const (
+		serverLookupURL = "/api/servers/servers/"
+		userLookupURL   = "/api/iam/users/"
+	)
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+
+		switch {
+		case strings.HasPrefix(r.URL.Path, serverLookupURL):
+			_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"srv-uuid"}]}`))
+		case strings.HasPrefix(r.URL.Path, userLookupURL):
+			_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"usr-uuid"}]}`))
+		case strings.HasPrefix(r.URL.Path, getEventURL):
+			assert.Equal(t, getEventURL+"srv-uuid/usr-uuid/", r.URL.Path)
+			_, _ = w.Write([]byte(`{"count":0,"results":[]}`))
+		default:
+			t.Errorf("unexpected request path %q", r.URL.Path)
+			http.Error(w, "unexpected path", http.StatusNotFound)
+		}
+	}))
+	defer ts.Close()
+
+	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+
+	_, err := GetEventList(ac, 25, "my-server", "admin")
+
+	require.NoError(t, err)
 }
 
 func TestPollCommandExecution(t *testing.T) {
--- a/api/event/event_test.go
+++ b/api/event/event_test.go
@@ -70,29 +70,60 @@ func TestGetEventList_PutsTheResolvedFilterIDsInThePath(t *testing.T) {
 		userLookupURL   = "/api/iam/users/"
 	)
 
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Content-Type", "application/json")
+	tests := []struct {
+		name         string
+		serverName   string
+		userName     string
+		expectedPath string
+	}{
+		{
+			name:         "both filters",
+			serverName:   "my-server",
+			userName:     "admin",
+			expectedPath: getEventURL + "srv-uuid/usr-uuid/",
+		},
+		{
+			name:         "server only",
+			serverName:   "my-server",
+			expectedPath: getEventURL + "srv-uuid/",
+		},
+		{
+			// path.Join drops the empty server segment, so a lone user ID lands on the address
+			// the command detail route also answers on. Recorded as current behavior, not as
+			// the desired one—confirming it needs the server-side route (alpacax/alpacon-cli#363).
+			name:         "user only",
+			userName:     "admin",
+			expectedPath: getEventURL + "usr-uuid/",
+		},
+	}
 
-		switch {
-		case strings.HasPrefix(r.URL.Path, serverLookupURL):
-			_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"srv-uuid"}]}`))
-		case strings.HasPrefix(r.URL.Path, userLookupURL):
-			_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"usr-uuid"}]}`))
-		case strings.HasPrefix(r.URL.Path, getEventURL):
-			assert.Equal(t, getEventURL+"srv-uuid/usr-uuid/", r.URL.Path)
-			_, _ = w.Write([]byte(`{"count":0,"results":[]}`))
-		default:
-			t.Errorf("unexpected request path %q", r.URL.Path)
-			http.Error(w, "unexpected path", http.StatusNotFound)
-		}
-	}))
-	defer ts.Close()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
 
-	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+				switch {
+				case strings.HasPrefix(r.URL.Path, serverLookupURL):
+					_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"srv-uuid"}]}`))
+				case strings.HasPrefix(r.URL.Path, userLookupURL):
+					_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"usr-uuid"}]}`))
+				case strings.HasPrefix(r.URL.Path, getEventURL):
+					assert.Equal(t, tt.expectedPath, r.URL.Path)
+					_, _ = w.Write([]byte(`{"count":0,"results":[]}`))
+				default:
+					t.Errorf("unexpected request path %q", r.URL.Path)
+					http.Error(w, "unexpected path", http.StatusNotFound)
+				}
+			}))
+			defer ts.Close()
 
-	_, err := GetEventList(ac, 25, "my-server", "admin")
+			ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
 
-	require.NoError(t, err)
+			_, err := GetEventList(ac, 25, tt.serverName, tt.userName)
+
+			require.NoError(t, err)
+		})
+	}
 }
 
 func TestPollCommandExecution(t *testing.T) {
--- a/api/websh/websh_test.go
+++ b/api/websh/websh_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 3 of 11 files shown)


api/event/event_test.go
```diff
@@ -18,66 +18,40 @@ import (
...
-		}
-
-		resp := api.ListResponse[EventDetails]{
-			Count:   200, // more items exist on server
-			Results: results,
-		}
-		_ = json.NewEncoder(w).Encode(resp)
+		// Never runs out: a walk that ignored the limit would go on forever.
+		results := make([]EventDetails, size)
+		_ = json.NewEncoder(w).Encode(api.ListResponse[EventDetails]{Count: 1000, Next: page + 1, Results: results})
```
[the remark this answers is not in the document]
jisung-02: Fixed in 62ec164. The fake server now stops at five pages instead of paginating forever:

```go
next := page + 1
if page >= lastPage {
    next = 0
}
```

Five pages is twice what the tail needs, so the assertion stays decisive—an unbounded walk comes back with 500 items and fails `assert.Len(t, events, 250)` right away. Verified by swapping `FetchPagesUpTo` for `FetchAllPages`: the test now fails in 0.00s where it used to hang to the test timeout.

api/websh/websh_test.go
```diff
@@ -29,6 +30,36 @@ const (
...
+		if sizeErr != nil || pageErr != nil {
+			t.Errorf("page and page_size must be integers, got page=%q page_size=%q",
+				r.URL.Query().Get("page"), r.URL.Query().Get("page_size"))
+			http.Error(w, "bad pagination query", http.StatusBadRequest)
+			return
+		}
+
+		// Never runs out: a walk that ignored the limit would go on forever.
+		results := make([]SessionDetailResponse, size)
+		_ = json.NewEncoder(w).Encode(api.ListResponse[SessionDetailResponse]{Count: 1000, Next: page + 1, Results: results})
```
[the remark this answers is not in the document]
jisung-02: Fixed in 1c0adf9, same shape as the `api/event` test:

```go
next := page + 1
if page >= lastPage {
    next = 0
}
```

Five pages is twice what the tail needs, so `assert.Len(t, list, 250)` still cannot pass unless tail arrives as the helper limit—an unbounded walk returns 500 and fails immediately instead of hanging.

api/event/event.go
```diff
@@ -209,23 +211,14 @@ func GetEventList(ac *client.AlpaconClient, pageSize int, serverName string, use
 		}
 	}
 
-	relativePath := path.Join(serverID, userID)
-	params := map[string]string{}
-	if pageSize > 0 {
-		params["page_size"] = fmt.Sprintf("%d", pageSize)
-	}
-	responseBody, err := ac.SendGetRequest(utils.BuildURL(getEventURL, relativePath, params))
+	endpoint := path.Join(getEventURL, serverID, userID)
```
geunwoonoh, line 214: Warning: `--user` without `--server` builds a URL that lands on the detail endpoint (pre-existing, but this line moved here).

`path.Join` drops empty elements, so with `--user admin` and no `--server`:

```
path.Join("/api/events/commands/", "", "usr-uuid") -> "/api/events/commands/usr-uuid"
```

A single UUID segment under that prefix is the **command detail** address, used at `event.go:265` and `event.go:364` as `BuildURL(getEventURL, cmdID, nil)`. So `alpacon exec ls --user admin` sends a user UUID to the command detail route.

The old code (`relativePath := path.Join(serverID, userID)`) produced the same URL, so this is not a regression from this PR. For contrast, `api/webftp/webftp.go:24,31` sends the same filters as query params instead of path segments.

Confirming it needs the server-side route, which the CLI cannot tell us — I'd file a separate issue rather than widen this PR.

api/event/event_test.go
```diff
@@ -18,66 +18,81 @@ import (
...
-		resp := api.ListResponse[EventDetails]{
-			Count:   200, // more items exist on server
-			Results: results,
+	require.NoError(t, err)
+	assert.Len(t, events, 250)
+}
+
+// The --server and --user filters are path segments, not query params, so a broken join
+// does not fail the request—it returns every command as if no filter had been given.
+func TestGetEventList_PutsTheResolvedFilterIDsInThePath(t *testing.T) {
```
geunwoonoh, line 66: Warning: this pins only the case where **both** filters are set, while the case that actually breaks — one set, one empty — is left uncovered. The comment right above names the hazard ("a broken join does not fail the request"), which is exactly what the one-sided case runs into (see my comment on `api/event/event.go:214`).

Adding the one-filter case to record current behavior would be cheap.

cmd/websh/websh_list.go
```diff
@@ -15,25 +15,22 @@ var webshListCmd = &cobra.Command{
   alpacon websh ls --tail 50`,
 	Run: func(cmd *cobra.Command, args []string) {
 		tail, _ := cmd.Flags().GetInt("tail")
+		utils.RequirePositiveInt("tail", tail)
```
geunwoonoh: Suggestion: nothing pins this new validation.

`WebshCmd` sets `DisableFlagParsing: true` (`cmd/websh/websh.go:212`), and `cmd/exec/list_test.go` has `TestExecLsRoutesAndParsesFlags` aimed at exactly that situation — its comment reads "DisableFlagParsing only covers the parent; this pins that a child still routes and parses flags normally."

`websh ls` sits under the same parent and now also gains `RequirePositiveInt`, yet neither the routing nor the `--tail 0` rejection (exit code 2) is pinned. Porting `cmd/exec/list_test.go` over would be enough.

cmd/websh/websh_list.go
```diff
@@ -15,25 +15,22 @@ var webshListCmd = &cobra.Command{
...
-			sessionList = sessionList[len(sessionList)-tail:]
-		}
-
 		utils.PrintTable(sessionList)
 	},
 }
 
 func init() {
-	webshListCmd.Flags().Int("tail", 25, "Number of sessions to show")
+	webshListCmd.Flags().Int("tail", 25, "Number of sessions to show, newest first")
```
geunwoonoh: Suggestion: `-t` is the only shorthand missing across the `--tail` commands.

`note` / `audit` / `log` / `webftp` / `exec` all register `IntP(..., "t", ...)`; `websh ls` is the lone exception. This PR already edits the line. The parent's hand-rolled `-s` (share) and `-u` (username) do not collide, since routing to the subcommand happens first.

```suggestion
	webshListCmd.Flags().IntP("tail", "t", 25, "Number of sessions to show, newest first")
```

api/event/event.go
```diff
@@ -209,23 +211,14 @@ func GetEventList(ac *client.AlpaconClient, pageSize int, serverName string, use
...
 	if err != nil {
 		return nil, err
 	}
 
-	var response api.ListResponse[EventDetails]
-	if err = json.Unmarshal(responseBody, &response); err != nil {
-		return nil, err
-	}
-
 	var eventList []EventAttributes
```
geunwoonoh: Suggestion: preallocate the projection slice.

It grows from nil by append; at tail 250 that is eight-plus reallocations and copies, and `len(events)` is already known here. `utils.PrintTable` prints `[]` for both nil and empty slices (`utils/output.go:130`), so `--output json` is unaffected.

```suggestion
	eventList := make([]EventAttributes, 0, len(events))
```

`api/websh/websh.go:51` (`var list []SessionListItem`) has the same shape and would benefit the same way — it falls outside the diff so I could not anchor a comment there.

api/websh/websh_test.go
```diff
@@ -29,6 +30,43 @@ const (
 	receivedBufferSize = 8
 )
 
+// The page walk itself is pinned in api.TestFetchPagesUpTo_*; what is specific to
+// GetSessionList is that tail reaches the helper as its limit, so a tail larger than the
+// server's 100-item page cap still yields exactly tail sessions.
+func TestGetSessionList_PassesTailAsTheLimit(t *testing.T) {
```
geunwoonoh, line 36: Suggestion (non-blocking): this and `api/event/event_test.go:30` differ only in the type parameter and the function under test — about 30 lines are identical. Sharing them across packages would need a test helper package such as `api/apitest`, and with only two call sites the duplication is fine as is. Worth extracting once a third one shows up.

geunwoonoh (dismissed):
## Summary

#330 settled `--tail N` as "the newest N entries", and two commands were breaking that contract in different ways. This PR routes both through `api.FetchPagesUpTo`.

- `websh ls` fetched every page and then sliced `sessionList[len-tail:]`. The endpoint sorts `-added_at`, so that slice returned the **oldest** N.
- `exec ls` had the sort direction right but sent a single `page_size=tail` request, so the server's 100 cap silently reduced `--tail 250` to 100 entries.

The help text of `log` / `audit` / `webftp-log` / `event` is aligned with what those endpoints actually return (newest first) in the same pass.

## Verification (run locally)

```
go build ./...                      → exit 0
go vet ./...                        → exit 0
go test -race ./api/... ./cmd/...   → all ok
gofmt -l .                          → no output
```

`golangci-lint` is not installed on my machine, so I could not run it. CI has to confirm that one.

---

## Critical issues (0)

None.

---

## Warnings (1)

### 1. `api/event/event.go:214` — `--user` without `--server` builds a URL that points at the detail endpoint (pre-existing, but this PR touches the line)

```go
endpoint := path.Join(getEventURL, serverID, userID)
```

`path.Join` drops empty elements. With `--user admin` and no `--server`, `serverID` is `""`, so:

```
path.Join("/api/events/commands/", "", "usr-uuid") → "/api/events/commands/usr-uuid"
```

A single UUID segment under that prefix is the **command detail** address, used twice in the same file (`event.go:265`, `event.go:364`, both `BuildURL(getEventURL, cmdID, nil)`). So `alpacon exec ls --user admin` sends a user UUID to the command detail route.

The old code (`relativePath := path.Join(serverID, userID)`) produced the same URL, so this is not a regression introduced here. What makes it worth raising is that the new `TestGetEventList_PutsTheResolvedFilterIDsInThePath` (`event_test.go:66`) pins only the **both set** case, while the case that actually breaks — one filter set, one empty — is left uncovered. The test comment even names the hazard: "a broken join does not fail the request".

For contrast, the sibling package `api/webftp/webftp.go:24,31` sends the same filters as query params (`params["server"]`, `params["user"]`) rather than path segments. `api/event` is the only one using path segments.

Confirming this needs the server-side route, which the CLI cannot tell us. I'd keep it out of this PR's scope and file a separate issue, but adding the one-filter case to the test to record current behavior would be cheap.

---

## Suggestions (4)

### 1. `cmd/websh/websh_list.go` — `websh ls` has no flag-parsing test

`WebshCmd` sets `DisableFlagParsing: true` (`cmd/websh/websh.go:212`). `cmd/exec/list_test.go` has a test aimed at exactly that situation, `TestExecLsRoutesAndParsesFlags`, whose comment explains why: "DisableFlagParsing only covers the parent; this pins that a child still routes and parses flags normally."

`websh ls` sits under the same parent and now also gains `utils.RequirePositiveInt("tail", tail)`, yet nothing pins either the routing or the `--tail 0` rejection (exit code 2). Porting `cmd/exec/list_test.go` over would be enough.

### 2. `cmd/websh/websh_list.go:35` — the `-t` shorthand is the only one missing

```go
webshListCmd.Flags().Int("tail", 25, "Number of sessions to show, newest first")
```

`note` / `audit` / `log` / `webftp` / `exec` all register `IntP(..., "t", ...)`. `websh ls` is the lone exception. This PR already edits the line, so switching to `IntP` removes the outlier. The parent's hand-rolled `-s` (share) and `-u` (username) do not collide, since routing to the subcommand happens first.

### 3. `api/event/event.go:220`, `api/websh/websh.go:51` — preallocate the projection slices

```go
var eventList []EventAttributes   // event.go:220
var list []SessionListItem        // websh.go:51
```

Both grow from nil by append. At tail 250 that is eight-plus reallocations and copies, and the length is already known at that point:

```go
eventList := make([]EventAttributes, 0, len(events))
```

`utils.PrintTable` prints `[]` for both nil and empty slices (`utils/output.go:130`), so `--output json` is unaffected.

### 4. The two new tests share a near-identical fake server

`TestGetEventList_PassesTailAsTheLimit` (`api/event/event_test.go:30`) and `TestGetSessionList_PassesTailAsTheLimit` (`api/websh/websh_test.go:36`) differ only in the type parameter and the function under test; about 30 lines are the same. Sharing them across packages would need a test helper package such as `api/apitest`, and with only two call sites the duplication is fine as is. Worth extracting once a third one shows up.

---

## Duplicate / dead code check

Checking the specific things that were asked about:

- **Unused imports/symbols**: none. `go build` and `go vet` pass, and `fmt` (18 uses) and `json` (3 uses) in `event.go`, plus `atomic` (19 uses) in `event_test.go`, are all still live elsewhere. `api/types` was correctly dropped from `event_test.go`.
- **Dead logic**: the `if tail > 0 && tail < len(...)` slice in `websh_list.go` is the core deletion of this diff, and nothing is left behind.
- **Stale help text**: a search for `from the end` / `most recent last` / `last N` returns nothing. All six `--tail` flags now read `newest first`.
- **`FetchAllPages` call sites**: swept all of them. No command with a `--tail` flag still uses it. `note` / `websh ls` / `exec ls` use `FetchPagesUpTo`; `log` / `audit` / `webftp-log` use `FetchCursorPages`. The remaining `FetchAllPages` callers (`iam`, `server`, `cert`, `packages`, …) list things that have no tail concept, so leaving them alone is right.

## Performance

- **`websh ls` is a clear win.** Walking the entire session list becomes `ceil(tail/100)` requests. At the default `--tail 25` that is one 25-item request instead of N pages of 100. The busier the workspace, the larger the gap.
- **`exec ls` does not regress.** For `tail <= 100` it is still a single request. Beyond that the request count grows, but that is the cost of returning the data it should have returned all along.
- **Over-fetch stays under one page.** At `--tail 250`, `page_size` is pinned to 100, so 300 items are fetched and truncated to 250. The comment in `pagination.go` explains why `page_size` has to stay fixed for the whole walk (offset math), so this reads as a deliberate trade-off.

## Highlights

- The fix closes **the whole bug class**, not just the two reported sites. I verified that all six `--tail` commands now go through a bounded helper.
- **A test that was pinning the wrong behavior got removed.** `TestGetEventList_NoExtraPagination` was guarding the "only one request" bug. Deleting it and replacing it with a test whose failure mode is inverted is the right call.
- **The new tests fail fast by design.** The fake server stops at five pages, so a regression to an unbounded walk breaks on the length assertion instead of hanging until the test timeout — and the comment says why.
- Comments explain WHY rather than WHAT: the endpoint ordering rationale and the choice of `lastPage = 5` are both things the code cannot say on its own.
- testify conventions from CLAUDE.md are respected: `assert.Len` over a wrapped predicate, `require` only for preconditions, `assert` and `t.Errorf` inside httptest handlers.
- The Residual section of the PR body records the missing server-side tiebreaker (alpacax/alpacon-server#3044) as a limit the CLI cannot fix.

---

**Verdict**: APPROVED

No critical issues, and the single warning is a pre-existing bug rather than something this PR introduced. All four suggestions can be deferred.

**Checklist**:
- [x] Code follows stack conventions (gofmt clean, `const → var → type → func` order kept, errcheck patterns respected)
- [x] Tests included and passing (`go test -race` passes)
- [x] No security vulnerabilities (no hardcoded secrets or injection paths; auth header handling untouched)
- [x] Documentation updated (six help strings aligned; the README exit-code-2 entry already covers the positive-value check behind `--tail`)

api/event/event.go
jisung-02: Agreed on both counts—out of this PR's scope, and worth recording. Filed as #363 with the route comparison against `api/webftp` and the note that the server-side definition is what settles it.

The cheap half is done in 1310ce8: `TestGetEventList_PutsTheResolvedFilterIDsInThePath` is now table-driven over all three cases (both filters, server only, user only), so the colliding URL is pinned as current behavior rather than left uncovered. The comment on the user-only case says explicitly that it records what happens today, not what should, and points at #363.

api/event/event_test.go
jisung-02: Fixed in 1310ce8. The test is table-driven now and the one-sided cases are in:

```go
{
    name:         "server only",
    serverName:   "my-server",
    expectedPath: getEventURL + "srv-uuid/",
},
{
    // path.Join drops the empty server segment, so a lone user ID lands on the address
    // the command detail route also answers on. Recorded as current behavior, not as
    // the desired one—confirming it needs the server-side route (alpacax/alpacon-cli#363).
    name:         "user only",
    userName:     "admin",
    expectedPath: getEventURL + "usr-uuid/",
},
```

The user-only case asserts the colliding URL on purpose—it is a description of today's behavior, so if the route is corrected under #363 this test fails and forces the update.

cmd/websh/websh_list.go
jisung-02: Fixed in ef70e0d. `TestWebshLsRoutesAndParsesFlags` is ported from `cmd/exec/list_test.go`, carrying the same comment about `DisableFlagParsing` covering the parent only, and covers defaults / `--tail` / `-t` / the `list` alias.

The `--tail 0` rejection needed one more thing: `RequirePositiveInt` calls `os.Exit`, so it is pinned through a helper process, the same shape `cmd/note/note_list_test.go` uses:

```go
func TestWebshLsTailZeroExitsWithUsageErrorCode(t *testing.T) {
	helper := osexec.Command(os.Args[0], "-test.run=^TestWebshLsTailZeroHelperProcess$")
	helper.Env = append(os.Environ(), "GO_WANT_WEBSH_LS_TAIL_HELPER=1")

	err := helper.Run()

	var exitErr *osexec.ExitError
	require.ErrorAs(t, err, &exitErr)
	assert.Equal(t, utils.ExitCodeUsageError, exitErr.ExitCode())
}
```

That required lifting the `Run` body into `runWebshList(tail int)` (73c9d78) so the helper can reach the validation without a client—`cmd/note` and `cmd/exec` are already shaped that way, so this removes an inconsistency rather than adding one.

cmd/websh/websh_list.go
jisung-02: Fixed in 73c9d78—applied as suggested, so `websh ls` is no longer the outlier:

```go
webshListCmd.Flags().IntP("tail", "t", 25, "Number of sessions to show, newest first")
```

Confirmed the parent's `-s` / `-u` do not collide: the `short flag` case in the new `TestWebshLsRoutesAndParsesFlags` resolves `ls -t 5` through `WebshCmd.Find` and reads back 5.

api/event/event.go
jisung-02: Fixed in b1a30b6, both sites—`api/event/event.go:220` as suggested, and `api/websh/websh.go:51` which you flagged as out of anchor range:

```go
eventList := make([]EventAttributes, 0, len(events))
list := make([]SessionListItem, 0, len(sessions))
```

`utils.PrintTable` printing `[]` for nil and empty alike is the part that made this safe, so the commit body records it—the switch from a nil slice to an empty one does not move `--output json`.

api/websh/websh_test.go
jisung-02: Agreed, leaving it. Two call sites do not pay for a new `api/apitest` package, and the duplicated part is the fake server rather than the assertion, so the two tests still read on their own.

Noting the trigger you set: a third one is when it gets extracted.

geunwoonoh (approved):
## Follow-up review — all previous findings addressed

Re-reviewed at `1310ce8` (four commits since my approval at `786c5f9`, which GitHub dismissed as stale). All six comments from that review got a reply; five are applied in code and one is deferred with reasoning I agree with.

| Previous finding | Commit | Status |
|---|---|---|
| Warning — `--user` alone collides with the detail route | `1310ce8` + #363 | Addressed (scoped out, behavior pinned) |
| Warning — test covered only the both-filters case | `1310ce8` | Fixed |
| Suggestion 1 — no flag/validation test for `websh ls` | `ef70e0d` | Fixed |
| Suggestion 2 — `-t` shorthand | `73c9d78` | Fixed |
| Suggestion 3 — preallocate the projection slices | `b1a30b6` | Fixed (both sites) |
| Suggestion 4 — duplicated fake server | — | Deferred with reasoning (agreed) |

### On each

**Warning (`1310ce8`)** — Splitting the fix out was the right call. `TestGetEventList_PutsTheResolvedFilterIDsInThePath` is table-driven now and carries all three cases (`both filters` / `server only` / `user only`). The comment on the user-only case states plainly that it records today's behavior rather than the desired one and points at #363, so a server-side correction breaks the test and forces the update. That removes what I was actually worried about: the broken path staying symptomless. #363 is open and its body carries the `api/webftp` comparison.

**Suggestion 1 (`ef70e0d`)** — `runWebshList` extracted to package level, and `cmd/websh/websh_list_test.go` added. `TestWebshLsRoutesAndParsesFlags` mirrors `cmd/exec/list_test.go`, including the `DisableFlagParsing` comment. The `--tail 0` rejection is pinned through a helper process because `RequirePositiveInt` calls `os.Exit` — same shape as `cmd/note/note_list_test.go:75`, down to the `osexec` alias. It follows a pattern the repo already has.

**Suggestion 3 (`b1a30b6`)** — Both sites, including `api/websh/websh.go:51`, which fell outside the diff so I could not anchor a comment on it. I checked the nil-to-empty change separately: each function has exactly one caller (`cmd/exec/list.go:57`, `cmd/websh/websh_list.go:35`), there is no `== nil` comparison anywhere in the repo, and `utils.PrintTable` prints `[]` for nil and empty alike. Safe.

**Suggestion 4** — Deferring is right. Two call sites do not pay for a new `api/apitest` package, and the duplicated part is the fake server rather than the assertions, so both tests still read on their own.

### Verification

```
go build ./...        → exit 0
go vet ./...          → exit 0
gofmt -l .            → no output
go test -race ./...   → all packages ok
```

CI is green across the board. **`golangci-lint` — which I could not run locally last time — reports SUCCESS**, as do `CodeQL`, `govulncheck` and `Build and Test`.

### Mutation checks on the new tests

I confirmed the two new tests actually bite rather than passing vacuously (both mutations reverted afterwards):

- Changing the `user only` expected path to `MUTANT/` fails that subtest alone, so the assertion really executes.
- Removing the `RequirePositiveInt` call from `runWebshList` fails `TestWebshLsTailZeroExitsWithUsageErrorCode` immediately, so the exit-code-2 contract is genuinely pinned.

### New findings in the follow-up commits

None. Two trivia not worth a comment: `runWebshList` stutters slightly against the package name but is consistent with the existing `webshListCmd` naming here, and the unused `args` in the `Run` signature predates this PR and no enabled linter flags it.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Code follows stack conventions (gofmt clean, `const → var → type → func` order kept, `golangci-lint` green in CI)
- [x] Tests included and passing (`go test -race` passes; both new tests verified to fail under mutation)
- [x] No security vulnerabilities (`CodeQL` and `govulncheck` green; no secrets or injection paths introduced)
- [x] Documentation updated (six help strings aligned; the pre-existing routing bug recorded in #363)

PR #359, opened 2026-08-18
fix(utils): sanitize server-controlled text in the Cli* helpers

### LINKED ISSUE #324: fix(utils): plain-text error path prints server-controlled text unsanitized
Found while surveying residual gaps of the same harm class as #294, after the review of #323.

## Problem

`utils.CliError`, `CliErrorWithExit`, `CliErrorWithExitCode`, and `CliWarning` print their formatted message to stderr with no sanitize pass:

```go
// utils/log.go:17-21
func CliError(msg string, args ...any) {
	errorMessage := fmt.Sprintf(msg, args...)
	fmt.Fprintf(os.Stderr, "%s: %s\n", Red("Error"), errorMessage)
	reportCLIError()
}
```

`client.parseAPIError` (client/client.go:476) puts the server's `detail` field into that message — or the truncated response body verbatim when the response is not JSON — and nearly every command renders API failures through `%s` on these helpers. A bidi override or an escape sequence inside a server error message therefore lands raw on the operator's terminal: the same Trojan Source class #294 closed on the audit-review surface, reachable through any command's error output.

## Why it is reachable

- Self-hosted deployments talk to a server the workspace operator does not necessarily control end to end, and a compromised or hostile server writes `detail` freely.
- Reflected input can reach `detail` too: a request that echoes an attacker-chosen name or path back in the error message carries the attacker's bytes into the reviewer's terminal.
- The asymmetry is already visible in the code: the `--output json` error envelope goes through `PrintJSONValue` → `escapeJSONControls` and is safe; only the default text form is exposed.

Not verified: whether the server sanitizes `detail` before serving it. As with #294, treat the CLI side as defence in depth.

## Proposal

Wrap the message in `utils.SanitizeTerminalText` at the `CliError*`/`CliWarning`/`CliInfo` level (utils/log.go), so every caller is covered at the single choke point rather than at each of the ~200 call sites. stderr is a human-facing terminal surface, so stripping — the treatment `PrintTable` cells already get — is the right default there, not escaping.

One decision to make deliberately: the helpers also print caller-authored format strings containing `\n` for multi-line guidance; `SanitizeTerminalText` drops `\n`. Either sanitize only the interpolated arguments, or keep the newline out of the control strip for this path.

jisung-02:
## Background

`utils.CliError`, `CliErrorWithExit`, `CliErrorWithExitCode` and `CliWarning` printed their rendered message to stderr with no sanitize pass.

`client.parseAPIError` (`client/client.go:476`) puts the server's `detail` field into that message, or the truncated response body verbatim when the response is not JSON. Nearly every command renders API failures through `%s` on these helpers, so a bidi override or an escape sequence chosen by the server lands raw on the operator's terminal. Same class as the Trojan Source issue #294 closed on the audit-review surface, this time reachable through any command's error output.

A self-hosted workspace talks to a server the operator does not necessarily control end to end, and reflected input reaches the same path whenever an error message echoes an attacker-chosen name or path back. The `--output json` error envelope already goes through `escapeJSONControls` and is safe, so only the default text form was exposed.

Whether the server sanitizes `detail` before serving it is not verified. As with #294, the CLI side is treated as defence in depth.

## Changes

`cliMessage` in `utils/log.go` is the single choke point every `Cli*` helper now passes through, rather than the roughly 200 call sites.

It sanitizes per line. `SanitizeTerminalText` drops `\n` and callers pass multi-line guidance, so splitting first is what keeps their formatting alive.

The label colors (`Red("Error")` and friends) sit outside `cliMessage` in the `Fprintf` call and survive unchanged.

The token key highlight in `cmd/server` did not. It was passed as `utils.Green(key)` into a `%s` argument, so the sanitize pass stripped it. `warnTokenKey` (`cmd/server/token_create.go:113`) now prints both one-time keys, colorizing after sanitizing. The key is shown once and never again, so the highlight is what tells the user which part of the line to copy.

That contract—colorize after sanitizing, never hand a `Color()`-wrapped value to a `Cli*` helper—is stated on `cliMessage` and pinned by a test. `cmd/login.go` dropped its own `SanitizeTerminalText` call, which the choke point had made redundant; nothing hands a pre-sanitized or pre-colored value to a `Cli*` helper any more.

The decision the issue left open went to sanitizing the rendered message per line, not the interpolated arguments alone.

## Safety

Splitting after the format is rendered means a `\n` from the server survives too, so a `detail` can open a line that looks like the CLI's own prefix.

That is pinned by a test rather than fixed. `\r` is still dropped, so injected text can only appear below what is already on screen and never overwrites it.

`CliErrorWithExit`, `CliErrorWithExitCode` and `CliInfoWithExit` are absent from the table test because `os.Exit` makes their stderr uncapturable without a subprocess. All three go through the same `cliMessage`.

Registration guide strings take a different path—`cmd/server/server_create.go` prints them raw with `fmt.Fprintln`—and are out of scope here. Filed as #360, since those are the strings the operator is told to paste into a shell.

## Testing

Five emitters (`CliError`, `CliWarning`, `CliInfo`, `CliSuccess`, `CliDebug`) are crossed with four inputs: ANSI escape, bidi override, carriage return, 8-bit CSI. Caller newline preservation and server newline pass-through are each pinned separately, and one more test pins that `cliMessage` strips its arguments too, so a `Color()`-wrapped value comes out plain.

`colorEnabled` is fixed per test rather than left to the environment. It is decided once at package init by `term.IsTerminal`, so the assertion that stderr holds no `\x1b` at all was resting on `go test` happening to pipe the test binary's output—build the binary and run it under a pty and it fails on the label's own escape.

```
$ gofmt -l .                    (no output)
$ go vet ./...                  (no output, exit 0)
$ golangci-lint run ./...       0 issues.
$ go test -race -count=1 ./...  39 packages ok, 0 failures
```

`config.TestGetOrCreateDeviceID_ConcurrentCreation` fails intermittently in this repo (about half of `-count=20` iterations). It predates this branch and is untouched by it—filed as #361.

Closes #324

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -177,7 +177,7 @@ with the saved target as the default. Non-interactive login requires a HOST or
 		switch classifyWhoamiVerification(whoErr) {
 		case whoamiVerified:
 			if who.PrincipalType == auth.PrincipalTypeApplication && who.Application != nil {
-				utils.CliInfo("Authenticated as application '%s'.", utils.SanitizeTerminalText(who.Application.Name))
+				utils.CliInfo("Authenticated as application '%s'.", who.Application.Name)
 			}
 		case whoamiFallback:
 			// Old server without whoami: fall back to legacy prefix-based verification.
--- a/cmd/server/server_create.go
+++ b/cmd/server/server_create.go
@@ -266,8 +266,7 @@ func createTokenAndWarn(ac *client.AlpaconClient, name string) string {
 	if err != nil {
 		utils.CliErrorWithExit("Failed to create the registration token: %s.", err)
 	}
-	fmt.Fprintf(os.Stderr, "%s: New token created. Save the key now—it will not be shown again: %s\n",
-		utils.Yellow("Warning"), utils.Green(utils.SanitizeTerminalText(response.Key)))
+	warnTokenKey("New token created. Save the key now—it will not be shown again", response.Key)
 	return response.ID
 }
 
--- a/cmd/server/server_create.go
+++ b/cmd/server/server_create.go
@@ -266,7 +266,8 @@ func createTokenAndWarn(ac *client.AlpaconClient, name string) string {
 	if err != nil {
 		utils.CliErrorWithExit("Failed to create the registration token: %s.", err)
 	}
-	utils.CliWarning("New token created. Save the key now—it will not be shown again: %s", utils.Green(response.Key))
+	fmt.Fprintf(os.Stderr, "%s: New token created. Save the key now—it will not be shown again: %s\n",
+		utils.Yellow("Warning"), utils.Green(utils.SanitizeTerminalText(response.Key)))
 	return response.ID
 }
 
--- a/cmd/server/token_create.go
+++ b/cmd/server/token_create.go
@@ -72,8 +72,7 @@ The token key is displayed once at creation time and cannot be retrieved again.`
 		}
 
 		utils.CliSuccess("Registration token created: %s", resp.Name)
-		fmt.Fprintf(os.Stderr, "%s: Save this key now—it will not be shown again: %s\n",
-			utils.Yellow("Warning"), utils.Green(utils.SanitizeTerminalText(resp.Key)))
+		warnTokenKey("Save this key now—it will not be shown again", resp.Key)
 		if resp.ExpiresAt != nil {
 			utils.CliInfo("Token expires at: %s", *resp.ExpiresAt)
 		}
@@ -107,4 +106,11 @@ func resolveGroupIDs(ac *client.AlpaconClient, entries []string) ([]string, erro
 		}
 	}
 	return ids, nil
+}
+
+// warnTokenKey prints the one-time key like CliWarning, except it colors the
+// key after sanitizing—CliWarning would strip the highlight with the escapes.
+func warnTokenKey(prefix, key string) {
+	fmt.Fprintf(os.Stderr, "%s: %s: %s\n",
+		utils.Yellow("Warning"), prefix, utils.Green(utils.SanitizeTerminalText(key)))
 }
--- a/cmd/server/token_create.go
+++ b/cmd/server/token_create.go
@@ -3,6 +3,7 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
 	iamAPI "github.com/alpacax/alpacon-cli/api/iam"
@@ -71,7 +72,8 @@ The token key is displayed once at creation time and cannot be retrieved again.`
 		}
 
 		utils.CliSuccess("Registration token created: %s", resp.Name)
-		utils.CliWarning("Save this key now—it will not be shown again: %s", utils.Green(resp.Key))
+		fmt.Fprintf(os.Stderr, "%s: Save this key now—it will not be shown again: %s\n",
+			utils.Yellow("Warning"), utils.Green(utils.SanitizeTerminalText(resp.Key)))
 		if resp.ExpiresAt != nil {
 			utils.CliInfo("Token expires at: %s", *resp.ExpiresAt)
 		}
--- a/utils/log.go
+++ b/utils/log.go
@@ -19,17 +19,33 @@ func reportCLIError() {
 	fmt.Fprintln(os.Stderr, "For issues, check the latest version or report on", gitIssueURL)
 }
 
+// cliMessage sanitizes a rendered message before it reaches the terminal:
+// client.parseAPIError puts the server's error detail into every one of these
+// helpers through %s, and the --output json envelope is the only path escaping
+// it today (escapeJSONControls).
+//
+// Per line, because SanitizeTerminalText drops \n and callers pass multi-line
+// guidance. Splitting after the format is rendered means a \n from the server
+// survives too, so the detail can add a line that looks like our own prefix.
+// Accepted: \r is still dropped, so it can only append below, never overwrite
+// what is already on screen.
+func cliMessage(msg string, args ...any) string {
+	lines := strings.Split(fmt.Sprintf(msg, args...), "\n")
+	for i, line := range lines {
+		lines[i] = SanitizeTerminalText(line)
+	}
+	return strings.Join(lines, "\n")
+}
+
 // CliError handles all error messages in the CLI.
 func CliError(msg string, args ...any) {
-	errorMessage := fmt.Sprintf(msg, args...)
-	fmt.Fprintf(os.Stderr, "%s: %s\n", Red("Error"), errorMessage)
+	fmt.Fprintf(os.Stderr, "%s: %s\n", Red("Error"), cliMessage(msg, args...))
 	reportCLIError()
 }
 
 // CliErrorWithExit handles all error messages in the CLI.
 func CliErrorWithExit(msg string, args ...any) {
-	errorMessage := fmt.Sprintf(msg, args...)
-	fmt.Fprintf(os.Stderr, "%s: %s\n", Red("Error"), errorMessage)
+	fmt.Fprintf(os.Stderr, "%s: %s\n", Red("Error"), cliMessage(msg, args...))
 	reportCLIError()
 	os.Exit(1)
 }
@@ -38,8 +54,7 @@ func CliErrorWithExit(msg string, args ...any) {
 // Unlike CliErrorWithExit it omits the "report an issue" footer—use it for expected,
 // machine-distinguishable refusals (e.g. a busy server) that scripts branch on by exit code.
 func CliErrorWithExitCode(code int, msg string, args ...any) {
-	errorMessage := fmt.Sprintf(msg, args...)
-	fmt.Fprintf(os.Stderr, "%s: %s\n", Red("Error"), errorMessage)
+	fmt.Fprintf(os.Stderr, "%s: %s\n", Red("Error"), cliMessage(msg, args...))
 	os.Exit(code)
 }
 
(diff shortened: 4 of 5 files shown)


geunwoonoh (dismissed):
## Review

Verified locally:

```
$ go build ./...                                        OK
$ go vet ./...                                          OK (no output)
$ gofmt -l .                                            (no output)
$ go test -race -count=1 ./utils/... ./cmd/server/...   ok
$ go test ./cmd/...                                     all packages ok
```

### Summary

Every `Cli*` emitter now passes its rendered message through one choke point, `cliMessage` (`utils/log.go:32`), before it reaches the terminal. The path `client.parseAPIError` opens—the server's `detail` flowing through `%s`—is spread over roughly 200 call sites, and closing it at one point instead of at all of them is the right call. The reason for splitting into lines before sanitizing (`SanitizeTerminalText` drops `\n`) is in the comment, so the next person has little room to undo it by accident.

**No critical issues.** Everything below is non-blocking.

### On the design decision

Sanitizing the arguments alone—which would strip the server's `\n` too and kill line spoofing outright—looks like the better option until you check the call sites. In `cmd/worksession/worksession_output.go:144`, `utils.CliWarning("Approver adjusted your request:\n%s", block)` passes a **genuinely multi-line argument**: `block` is built with `strings.Join(lines, "\n")` on line 138 of the same file. Collapsing newlines per argument would flatten that block. Nothing can tell a caller's newline from the server's, so per-line sanitization of the rendered string is a forced choice, and the trade-off write-up in the PR body is accurate.

In exchange for accepting the `\n` spoof, `\r` is still dropped, so injected text can never overwrite what is already on screen. On a TTY there is a second signal: only the real label is bold-colored by `Red("Error")`, while a spoofed `Success:` prints plain. That distinction disappears when the output is piped, but in that case the machine-readable contract is the exit code and the `--output json` envelope, so the residual risk is low.

### Warnings (2)

**1. `utils/log_test.go:98`—an assertion that breaks the moment color is on**

```go
stderr := captureStderr(t, func() { emit("%s", tt.detail) })
assert.NotContains(t, stderr, tt.notWant, name)
```

`stderr` holds the `Red("Error")` label as well as the message. Since `notWant` is `"\x1b"` for the ANSI escape case, this assertion rests on `colorEnabled` happening to be false. The PR body says as much, but the test fails as soon as the test binary's stderr is a TTY, because of the label's own `\033[1;31m`. That is the kind of failure that takes a while to diagnose.

`colorEnabled` is a package-level `var` in the same package, so the test can pin it:

```go
original := colorEnabled
colorEnabled = false
t.Cleanup(func() { colorEnabled = original })
```

**2. The two changed lines in `cmd/server` are untested**

The regression this PR discovered itself—the choke point stripping the ANSI out of `utils.Green(key)` and taking the highlight with it—is not pinned by a test. `cmd/server/token_create_test.go` and `server_create_test.go` already exist but do not touch the warning output.

No call site anywhere passes a colored argument into a `Cli*` helper today:

```
$ grep -rnE 'Cli(Error|Warning|Info|Success|Debug)[A-Za-z]*\(.*(Green|Red|Yellow|Blue|Bold)\(' --include="*.go" .
(no matches)
```

But the next `utils.CliInfo("%s", utils.Bold(x))` someone writes will lose its color silently. Two cheap guards:

- State the contract in the `cliMessage` comment: colorize after sanitizing, and never hand a `Color()`-wrapped value to a `Cli*` helper.
- Pin the fact that `cliMessage` strips ANSI out of its arguments in a `utils` unit test, so the test doubles as the contract.

### Suggestions (4)

**3. Duplication between `cmd/server/server_create.go:269` and `cmd/server/token_create.go:75`**

Both files are `package server`, so one four-line unexported helper removes the copy:

```go
// warnTokenKey prints the one-time key with the highlight applied after
// sanitizing—CliWarning would strip the color along with the escapes.
func warnTokenKey(prefix, key string) {
	fmt.Fprintf(os.Stderr, "%s: %s: %s\n",
		utils.Yellow("Warning"), prefix, utils.Green(utils.SanitizeTerminalText(key)))
}
```

"Two call sites is below the bar for a helper" is usually right. What got duplicated here is not a value though—it is `CliWarning`'s **output format** (`"%s: %s\n"` plus `Yellow("Warning")`), and the original of that format lives in `utils`. If the label format ever changes, these two lines drift without anything saying so.

**4. `cmd/login.go:180` now sanitizes twice**

```go
utils.CliInfo("Authenticated as application '%s'.", utils.SanitizeTerminalText(who.Application.Name))
```

`CliInfo` goes through the choke point, so the inner call is redundant now. It is the only line in the repo that hands a `SanitizeTerminalText` result to a `Cli*` helper. The other call sites (`worksession_*`, `websh_watch`, `whoami`) either write through `fmt.Fprintf` directly or use the stripped result for an empty-string fallback check, so those should stay.

**5. `utils/log_test.go:13 captureStderr` duplicates `pkg/testutil.CaptureOutput`**

`output_test.go` and `pending_approval_test.go` in this same `utils` package already use `pkg/testutil`. The shared version is the sturdier one: its stop func is idempotent and registered with `t.Cleanup`, so a panic in `fn` leaves no pipe installed, whereas `captureStderr` never reaches `writer.Close()` on that path and leaks the reader goroutine. Pre-existing, so not this PR's debt—but this PR adds seven more call sites to it. Worth a separate cleanup.

**6. `utils/log_test.go:77`—a slice instead of a map for `emitters`**

Map iteration order changes per run, and the emitter name currently survives only in `msgAndArgs`, not in the test path. Switching to `[]struct{ name string; emit func(string, ...any) }` with one more `t.Run` layer puts the failing emitter in the test name. Taste-level.

### Adjacent gap (not this PR—suggest a separate issue)

`cmd/server/server_create.go` writes server-supplied strings to stderr **raw**, in the same file:

```
303: fmt.Fprintln(os.Stderr, installCmd)            // guide.InstallCommands
309: fmt.Fprintln(os.Stderr, guide.RegisterCommand)
320: fmt.Fprintln(os.Stderr, guide.CollectionInstall)
327: fmt.Fprintln(os.Stderr, guide.InventorySnippet)
```

All of them are JSON fields from `api/server/types.go:81-90`. These are also the strings the operator is told to **copy and paste into a shell**, which pays an attacker better than #324 does: one bidi override is enough to split what the screen shows from what gets pasted. `cmd/login.go:134-135` prints `VerificationURIComplete` and `UserCode` the same way, at lower risk.

That is outside this PR's scope (error output), so a follow-up issue fits better than a fix here.

### Performance

Measured on an Apple M4 Pro, 95-character single-line error message:

| Target | ns/op | allocs/op |
|---|---|---|
| `cliMessage`, single line | 617 | 5 |
| `cliMessage`, 4 lines | 842 | 14 |
| `SanitizeTerminalText` alone | 533 | 3 |
| `StripANSIEscapes` (regexp) alone | 64 | 3 |
| `StripFormatChars` alone | 300 | 0 |

Three things to read out of that:

- The cost is dominated by `StripFormatChars`'s per-rune `unicode.Is(unicode.Cf, r)`, not by the regexp—the regexp is 12% of it.
- There is no hot path. `exec`'s streaming output goes straight out through `os.Stdout.Write` at `cmd/exec/exec.go:214` and never touches `Cli*`. A CLI run emits a handful of these messages, so the added cost is microseconds.
- A single-line fast path that skips `strings.Split`/`Join` moves 617ns to 602ns and saves one alloc. **Not worth the branch. Leave the code as it is.**

If it ever does matter, the lever is `utils/utils.go:510`: no Cf rune exists below U+00AD, so a single `if r < 0xad { return r }` takes that function from 300ns to 128ns (-57%). Outside this diff, and unnecessary today.

### Go conventions

No violations against CLAUDE.md:

- Declaration order `const → var → type → func` holds, and placing `cliMessage` right below `reportCLIError` reads in the right order too.
- The new `fmt.Fprintf(os.Stderr, ...)` calls are correct without `_, _ =`, since `.golangci.yml` lists `fmt.Fprintf` under `errcheck.exclude-functions`. They also match the dominant style in `server_create.go` (44 bare calls).
- testify conventions hold. The dedicated `NotContains`/`Contains` helpers are used, and the `name` passed as `msgAndArgs` carries something the assertion cannot say on its own (which emitter failed), so it is not the restatement the convention forbids.
- No dead code or unused symbols; dropping the local variables (`errorMessage` and friends) shrank the diff.

### Highlights

- One choke point instead of 200 call sites. Best decision in this PR.
- The accepted risk (`\n` spoofing) is **pinned explicitly** by `TestCliHelpers_KeepServerNewlines` rather than left implicit. If that behavior ever changes, the test says so first.
- The 8-bit CSI case (U+009B) is covered. The ESC-led regexp never sees that path, so it is easy to miss.
- Splitting the work across two commits (`0ad18e6` for the choke point, `6b3345d` for the highlight) leaves both the regression and its fix in the history.

---

**Verdict: APPROVED.** The two warnings are worth folding in before merge, but neither blocks.

- [x] Follows Go and repo conventions
- [x] Tests included, `-race` clean
- [x] No new security vulnerabilities (the adjacent gap is split out into its own issue)
- [x] No docs update needed (no user-facing behavior change)

geunwoonoh (approved):
## Re-review (`6b3345d` → `be74d42`)

Six new commits, one per item from the previous review.

### Review follow-up: 6 of 6 addressed

| # | Item | Commit | Status |
|---|---|---|---|
| W1 | Assertion that breaks when color is on | `8a232c7` | Done |
| W2 | Colored-argument contract not pinned | `5350f2e` | Done |
| S3 | Duplication across the two `cmd/server` sites | `a89110b` | Done |
| S4 | Double sanitize at `cmd/login.go:180` | `6285aed` | Done |
| S5 | `captureStderr` duplicating `testutil` | `48f1f35` | Done |
| S6 | `emitters` map → slice | `be74d42` | Done |
| Adjacent gap | Raw registration guide strings | — | Split out as #360 |

### Verification

**Build, lint, tests**—run against the PR head (`be74d42`) checked out into an isolated worktree:

```
$ go build ./...                  OK
$ go vet ./...                    OK (no output)
$ gofmt -l .                      (no output)
$ go test -race -count=1 ./...    all packages ok
```

**Mutation testing**—to confirm the new tests actually bite.

Flipping `pinColor(t, false)` to `true` breaks at exactly the predicted spot:

```
--- FAIL: TestCliHelpers_SanitizeServerControlledText/ANSI_escape/CliError
    Error: "\x1b[1m\x1b[31mError\x1b[0m: deniedapproved\n..." should not contain "\x1b"
```

The label's own escape is right there in the failure message. The fragility was real, and `pinColor` closes exactly that. S6 pays off here too: the failure path now reads `/ANSI_escape/CliError`, so the failing emitter is visible without opening the test.

Ripping the sanitize out of `cliMessage` fails 29 assertions, `TestCliMessage_StripsColorFromArguments` among them. The contract test is not idling.

**Checking the #361 claim**—the PR body attributes `config.TestGetOrCreateDeviceID_ConcurrentCreation` to a pre-existing defect, so I checked that the branch cannot be the cause:

```
PR head (be74d42):  go test -race -count=20 -run TestGetOrCreateDeviceID_ConcurrentCreation ./config/   FAIL
origin/main:        same command                                                                        FAIL
```

The three-dot diff against `main` shows `config/` untouched by this branch (all five changed files are under `cmd/` and `utils/`). Pre-existing, and splitting it into #361 is right.

### The new code

**`warnTokenKey` (`cmd/server/token_create.go:113`)**—the shape suggested, and the comment records *why* `CliWarning` is not used, which is what stops the next person from folding it back into the choke point. `prefix` is not sanitized, but both call sites pass compile-time constants, so that is fine.

**The `cliMessage` comment (`utils/log.go:32`)**—states the contract next to the code. Doing both the comment and the test goes past what was asked.

**`TestCliMessage_StripsColorFromArguments` pinning color to `true`**—the most careful detail in this round. With color off, `Green("secret")` returns plain `"secret"`, so the test would pass while verifying nothing. Forcing color on is what gives it teeth. The ask was "pin it"; the answer picked the right direction per test.

**Re-checking that the contract actually holds** at the PR head:

```
$ grep -rnE 'Cli(Error|Warning|Info|Success|Debug)[A-Za-z]*\(.*(Green|Red|Yellow|Blue|Bold)\(' --include="*.go" .   → 0 hits
$ Cli* calls passing a SanitizeTerminalText result                                                                  → 0 hits
```

Neither a pre-colored value nor a double sanitize is left in the repo.

**The PR body** was rewritten to match the code. The "two call sites is below the bar for a helper" paragraph is now the `warnTokenKey` description, and the reason for pinning `colorEnabled` is stated better than in my own comment ("build the binary and run it under a pty and it fails on the label's own escape"). Nothing in the description disagrees with the code.

### Remaining notes (non-blocking, nothing to change here)

1. **`warnTokenKey` has 0% coverage** (per `go tool cover`). Both call sites sit on network/interactive paths that are awkward to unit test, and the contract it depends on is pinned in `utils`, so this is fine as is.
2. **Helper placement**—defined in `token_create.go`, used from `server_create.go`. Same `package server`, so nothing breaks, and token creation is a reasonable home for it. Taste-level.
3. **Performance**—the previous conclusion stands (microseconds, measured; leave it alone). This round does not touch the runtime cost.

---

**Verdict: APPROVED.** No critical issues, both prior warnings resolved, all four suggestions applied, the adjacent gap and the pre-existing flake split into #360 and #361. Full suite including `-race` verified directly at the PR head.

- [x] Follows Go and repo conventions
- [x] Tests included, and mutation-checked for effectiveness
- [x] No new security vulnerabilities
- [x] PR body matches the code

PR #366, opened 2026-08-19
feat(utils): warn when a kept file mode is wider than a new download

### LINKED ISSUE #346: cp keeps an existing destination's mode without telling the user
## Background

`utils.SaveStreamAtomic` keeps the destination's own mode when the file already exists (#326, PR #342). That is deliberate: clamping every download to `0600` would break `alpacon cp my-server:/etc/nginx/nginx.conf /etc/nginx/nginx.conf`, and re-permissioning a file the user set up is not the CLI's call.

The cost is that this passes silently:

```bash
umask 022
touch ./id_rsa          # 0644
alpacon cp my-server:/root/.ssh/id_rsa ./id_rsa
ls -l id_rsa            # -rw-r--r--
```

A remote private key lands world-readable. The README and PR #342 both document the rule, but a line of documentation is thin cover for the exact case the original issue was about, and the user gets no signal at the moment it happens.

## Proposal

Warn on stderr when the destination already exists and carries a group or other read bit. Behavior stays as it is; the user is only told what just happened.

```
Warning: ./id_rsa already existed at 0644 and kept that mode; a new file would have been 0600.
```

`utils.CliWarning` already writes to stderr, so stdout stays clean and `--output json` consumers are unaffected.

Worth considering as part of it: whether the warning should name the alternative (`rm` the file first, or download to a fresh path) rather than only stating the mode.

## References

- PR #342 review comment on `api/ftp/ftp.go:593`
- #326

jisung-02:
## Background

`utils.SaveStreamAtomic` keeps the destination's own mode when the file already exists. That rule comes from #326 and PR #342 and it stays: clamping every download to `0600` would break `alpacon cp my-server:/etc/nginx/nginx.conf /etc/nginx/nginx.conf`, and re-permissioning a file the user set up is not the CLI's call.

The cost is that the dangerous case passes silently. A remote private key downloaded over an existing `0644` file lands world-readable, and nothing in the output says so:

```bash
umask 022
touch ./id_rsa          # 0644
alpacon cp my-server:/root/.ssh/id_rsa ./id_rsa
ls -l id_rsa            # -rw-r--r--
```

The README documented the kept-mode rule, but documentation is thin cover for exactly the case the original issue was about.

### The warning

`warnKeptWiderMode` in `utils/file_save.go` emits one stderr warning when the kept mode leaves the content readable beyond what a fresh download would be:

```
Warning: ./id_rsa already existed at 0644 and kept that mode; a new file would have been at most 0600. Remove it first for owner-only permissions.
```

The decision itself lives in `keptModeIsWider`, which fires only when the kept mode carries a group or other read bit and the mode a new file would have been given does not. `warnKeptWiderMode` adds the third condition, that the write is replacing a file that already exists. The read-bit condition on `newFilePerm` is what keeps the warning quiet for downloads that were never going to be private anyway — a `0666` package artifact over an existing `0644` file is not a leak, so it says nothing. Comparing full modes instead of the read bits would have been wrong for the same reason: an execute or write bit on the destination says nothing about who can read the secret.

`keptModeIsWider` returns false on Windows before comparing anything. Windows has no Unix modes; Go synthesizes them in `os.Stat` from `FILE_ATTRIBUTE_READONLY` alone, so every writable file reports `0666` and the comparison would fire on every single-file overwrite download. The remedy the message names does not hold there either, since removing a file changes no ACL.

"at most" in the message is deliberate: `newFilePerm` is the mode requested before the umask, so a new file lands at that mode or narrower, never wider. Naming the bare mode would be wrong under a umask that clears more. The same asymmetry means a restrictive umask can only silence a warning that was due, never produce one that was not.

Following the issue's open question, the message names the remedy rather than only stating the mode. Removing the file first is the one action that gets the caller back to an owner-only download. It names the path the user typed rather than the symlink-resolved target, because that is the path removing actually applies to.

### Getting the warning onto a line the user can read

The call sits before any writing rather than after the rename, so a user who sees the warning can abort a long download instead of learning about it once the bytes have already landed. That places it inside the download spinner's window, and a spinner leaves the cursor mid-line, so the warning arrived glued to a spinner frame.

`utils.CliWarning` now leads with `spinnerLineBreak()`, which returns a newline while a spinner is animating stderr and `""` otherwise. The break and the text go out in a single `Fprintf` — as two writes a 100ms repaint can land between them and reproduce the collision, just less often.

Two decisions in that helper are worth naming, because the obvious shapes are both wrong:

- A newline, not `\r\033[K`. stdout and stderr share a cursor on a terminal, so erasing the current line takes any streamed output not yet ended by a newline with it. `alpacon exec` in table mode writes chunk content to stdout and can call `CliWarning` on the next statement, with chunks that are not line-aligned; an erase there deletes a line of real remote output and the warning that deleted it reads `output may be incomplete`. The last spinner frame stays on screen in trade, and the spinner redraws on the line below.
- Gated on a spinner running, not on stderr being a terminal. Those are different questions, and `CliWarning` has 77 call sites — most of them nowhere near a spinner. `activeSpinners` (`utils/spinner.go`) is an `atomic.Int32` incremented inside the branch of `Start` that actually animates and decremented in `Stop` after the goroutine has returned, so it is already the TTY test as well: `s.enabled` is `term.IsTerminal(os.Stderr)`. Off a TTY the static progress line already ends in a newline, so a second one would only add a blank line.

`CliDebug` takes the same break. The refresh-token degrade at `api/auth0/auth0.go:286` logs from inside the `Refreshing access token...` spinner that `client.NewAlpaconAPIClient` starts (`client/client.go:81`), and that line exists to tell a degrade which always succeeds apart from the path it replaced — which it cannot do from mid-frame.

Those two are the only helpers that take it, and the helper's comment now says why rather than only listing them. They are the two printed from inside a call a spinner wraps: `SaveStreamAtomic` warns during a download, and the refresh degrade logs during `Refreshing access token...`. `CliInfo`, `CliSuccess`, and `CliError` print from the cmd layer or an interactive prompt, and every one of the eleven `HandleCommonErrors` call sites is in `cmd/`, so the api call whose `defer spinner.Stop()` owned the animation has already returned by then. One of them needing the break later would mean it has gained a call site under a spinner, which is the thing worth looking at.

### Streaming out from behind a spinner

The line break covers a message. It does not cover a caller that streams: a spinner repaints over streamed output and `Stop` then erases the whole line it drew on.

`Spinner.StopWriter(io.Writer) io.Writer` wraps a writer so the spinner retires on the first non-empty write. Callers that wait on something which may start producing output hand their writer through it instead of having to know when the first byte arrives. An empty write is not output, so it leaves the spinner alone.

Two waits are wired through it:

- `utils.HandleCommonErrors` (`utils/error_handler.go`) — the new MFA flow now stops its spinner as soon as the poll reports completion, before `RetryOperation` can stream a command to stdout, rather than at each of the exits below it. The legacy flow gets a static `CliInfo` line instead of a spinner: every tick of that loop retries the operation, so there is no stretch of it where an animation would be safe. The label both flows print is now one `mfaWaitMessage` constant.
- `cmd/exec/run.go` — the `--wait` approval loop. Every tick re-submits the command, so a tick whose request is still pending writes its denial line while the spinner animates, and that first pending tick spends the one-shot handover meant for the approved output. `Spinner.Stop` is terminal — it closes channels `Start` does not reopen — so a retired spinner cannot be revived. Each stretch of waiting therefore gets a spinner and a writer of its own, via a `restartSpinner` closure called at the two points where the poll re-arms. A wait that reused one writer would animate through its first poll gap and then sit silent for the rest of what `--wait` exists to cover.

### README

The kept-mode paragraph gains one clause: the archive path taken by recursive and multi-source downloads leaves an existing local file's mode alone too, and no warning covers that path. `extractFile` (`utils/utils.go:428`) opens with `os.OpenFile(..., O_CREATE, f.Mode())`, and `open(2)` ignores the mode argument for a file that already exists.

## Safety

The download behavior is unchanged. Nothing is chmod'd, no write path is altered, and no error case is added — the change is a warning and a README clause.

`utils.CliWarning` and `utils.CliDebug` write to stderr, so stdout stays clean and `--output json` consumers are unaffected. Neither emits an ANSI escape at all now; the only added byte is a leading newline, and only while a spinner is animating.

Both `newFilePerm` values in the tree were checked. `api/ftp/ftp.go:599` passes `0600`, which is where the warning is meant to fire. `api/packages/packages.go:172` passes `0666` for public artifacts, so the `newFilePerm` condition suppresses it there.

The archive path also reaches `SaveStreamAtomic`, through `fetchFromURLToFile` at `api/ftp/ftp.go:646` and `:763`, with the same `0600`. It stays quiet for a different reason: the destination is a temp path reserved by `reserveDownloadArchiveTempPath` with `os.CreateTemp` (`:624`), which the standard library creates at `0600`, so the kept-mode condition filters it out. That is an implicit invariant rather than an enforced one — widening that temp file's mode later would surface a warning naming a `.alpacon-download-*.zip` path the user never created. The README clause is scoped to the single-file download to match.

Downloads run sequentially. `collectConcurrentFailures` in `api/ftp/ftp.go` is used only by the bulk upload and poll paths, so there is no concurrent writer to interleave warnings.

On the spinner side, `restartSpinner` reassigns `spinner` inside the poll loop, so every exit was read to confirm none can leave a spinner unstopped and leak the `activeSpinners` count: the timeout case, the pending-error branch, the give-up-after-`MaxConsecutivePollFailures` branch, and the `default` case all call `spinner.Stop()` on the current spinner.

## Testing

`TestKeptModeIsWider` in `utils/utils_test.go` pins the predicate directly, one case per branch: `0644` kept over a `0600` download warns, a destination already at `0600` does not, and `0644` kept under a `0666` download does not. Calling it directly rather than through the filesystem is what makes the third case meaningful — `os.WriteFile` only requests a mode, so under `umask 077` a `0644` setup file lands at `0600` and the case exits at the kept-mode condition without ever evaluating the one it is named after.

`TestSaveStreamAtomic_WarnsWhenKeptModeIsWiderThanANewFile` is the end-to-end regression proof: `0600` written over an existing `0644` file, asserting both modes appear in the captured stderr. Its pre-existing mode is pinned with `os.Chmod` for the same umask reason. `TestSaveStreamAtomic_NoWarningForANewFile` covers the remaining branch, a destination that does not exist yet, which needs a real write.

For the presentation half:

| Test | Pins |
|---|---|
| `TestCliWarning_BreaksTheLineOnlyWhileASpinnerRuns` | a break with a spinner counted, none without, and no `\033[K` either way |
| `TestCliDebug_BreaksTheLineOnlyWhileASpinnerRuns` | the same for the debug line |
| `TestSpinner_DoesNotArmTheLineBreakOffATerminal` | a spinner started off a TTY leaves the count at zero |
| `TestSpinner_AnimatingBranchPairsTheActiveCount` | the increment and decrement pair, forcing `enabled` to reach the branch `go test` never runs |
| `TestSpinner_StopWriterRetiresTheSpinnerBeforeTheFirstByte` | the spinner is stopped by the first non-empty write, and an empty write leaves it alone |
| `TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnSpinner` | each tick of the wait receives a distinct writer, which is one spinner per tick |

Each new test was seen fail against the change it covers before being trusted. Deleting `activeSpinners.Add(1)` fails the pairing test with `expected: 0, actual: -1`; emptying `restartSpinner` fails the per-tick test with both writers pointing at the same spent `sync.Once`; dropping `spinnerLineBreak()` from `CliDebug` fails with `the break must lead the write: "Debug: fell back to the device scope\n"`.

```
$ gofmt -l .                                     (no output)
$ go vet ./...                                   (no output)
$ golangci-lint run ./...                        0 issues.
$ go test -race ./...                            39 ok, 0 FAIL
$ GOOS=windows go build ./...                    (no output)
$ umask 077 && go test ./utils/ -run 'TestKeptModeIsWider|TestSaveStreamAtomic'   ok
```

Closes #346


utils/utils_test.go
```diff
@@ -226,6 +226,54 @@ func TestSaveStreamAtomic_PreservesExistingFileMode(t *testing.T) {
 	assert.Equal(t, os.FileMode(0600), info.Mode().Perm())
 }
 
+func TestSaveStreamAtomic_WarnsWhenKeptModeIsWiderThanANewFile(t *testing.T) {
+	requireUnixModes(t)
+
+	dest := filepath.Join(t.TempDir(), "id_rsa")
+	require.NoError(t, os.WriteFile(dest, []byte("existing"), 0644))
+
```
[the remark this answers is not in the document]
jisung-02: Confirmed and fixed in 9ee6fb7. Reproduced it: `umask 077 && go test ./utils/ -run TestSaveStreamAtomic_WarnsWhenKeptModeIsWiderThanANewFile` failed with `"" does not contain "already existed at 0644 and kept that mode"`, because the destination landed at 0600 and `warnKeptWiderMode` saw no group/other bits.

Added `require.NoError(t, os.Chmod(dest, 0644))` after the write. `TestSaveStreamAtomic_KeepsPartialReplacementUnreadable` predates this branch and asserts `0644` the same way, so it got the same line — it was failing under umask 077 too (`expected: 0x1a4, actual: 0x180`).

The other two new cases are umask-safe: the `0600` setup has no bits a umask could clear, and the `0666` case returns early on `newFilePerm` regardless of the file's actual mode.

Verified: `umask 077 && go test ./utils/ -run TestSaveStreamAtomic` -> ok, `go test -race ./...` -> 39 ok / 0 FAIL, `golangci-lint run ./...` -> 0 issues.

utils/file_save.go
```diff
@@ -120,6 +121,19 @@ func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (in
 	return written, nil
 }
 
+// warnKeptWiderMode signals when the kept mode leaves the content readable
+// beyond what a fresh download would be (a private key over a 0644 file). A
+// warning, not a chmod—re-permissioning a file the user set up is not the
+// CLI's call (#326).
+func warnKeptWiderMode(fileName string, replacing bool, keptPerm, newFilePerm os.FileMode) {
+	const groupOtherRead = os.FileMode(0o044)
+	if !replacing || keptPerm&groupOtherRead == 0 || newFilePerm&groupOtherRead != 0 {
```
geunwoonoh: **Critical:** this fires on every overwrite download on Windows.

Windows has no Unix modes — Go's `os.Stat` synthesizes them from `FILE_ATTRIBUTE_READONLY` alone (`$GOROOT/src/os/types_windows.go:177-182`):

```go
if fs.FileAttributes&syscall.FILE_ATTRIBUTE_READONLY != 0 {
    m |= 0444
} else {
    m |= 0666   // every writable file
}
```

So on the Windows `cp` path both mode checks always pass:

| Condition | Value | Result |
|---|---|---|
| `keptPerm` | `0666` (any writable file) | `0666 & 0044 != 0` → passes |
| `newFilePerm` | `0600` (`api/ftp/ftp.go:599`) | `0600 & 0044 == 0` → passes |

Unless the destination carries the read-only attribute, every single-file overwrite download on Windows warns. The message does not hold there either: deleting the file changes no ACL, so "Remove it first for owner-only permissions" is not a remedy.

Windows is a first-class release target (`.goreleaser.yaml` builds amd64/arm/arm64) and this package already carries `utils/file_replace_windows.go`. `requireUnixModes` skips Windows, so CI cannot catch it.

Extracting the predicate gives the guard a home and also fixes the umask-dependent subtest at `utils/utils_test.go:271`:

```go
func warnKeptWiderMode(fileName string, replacing bool, keptPerm, newFilePerm os.FileMode) {
	if !replacing || !keptModeIsWider(keptPerm, newFilePerm) {
		return
	}
	CliWarning("%s already existed at %04o and kept that mode; a new file would have been at most %04o. Remove it first for owner-only permissions.",
		fileName, keptPerm, newFilePerm)
}

// keptModeIsWider reports whether keptPerm exposes the content to readers
// newFilePerm would not. Windows has no Unix modes—os.Stat synthesizes 0666 for
// every writable file—so the comparison carries no meaning there.
func keptModeIsWider(keptPerm, newFilePerm os.FileMode) bool {
	if runtime.GOOS == "windows" {
		return false
	}
	const groupOtherRead = os.FileMode(0o044)
	return keptPerm&groupOtherRead != 0 && newFilePerm&groupOtherRead == 0
}
```

`config/device.go:218` already uses `runtime.GOOS == "windows"`, so this is in keeping with the package. Build-tagged files the way `replaceFile` does would also work, but are heavy for a one-liner.

utils/file_save.go
```diff
@@ -69,6 +69,7 @@ func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (in
 	} else if !os.IsNotExist(err) {
 		return 0, fmt.Errorf("failed to access file: %w", err)
 	}
+	warnKeptWiderMode(fileName, replacing, finalPerm, newFilePerm)
```
geunwoonoh, line 73: **Warning:** this runs inside the download spinner's window, so on a TTY the warning arrives glued to a spinner frame.

`downloadSingleFileWithResult` starts the spinner at `api/ftp/ftp.go:677-678` and calls `saveDownloadedURL` at `:703`. The spinner goroutine repaints stderr every 100ms with `fmt.Fprintf(os.Stderr, "\r%s %s", ...)` (`utils/spinner.go:84`), and `CliWarning` does not clear the line first:

```
⠹ Downloading id_rsa...Warning: ./id_rsa already existed at 0644 and kept that mode; ...
⠸ Downloading id_rsa...
```

The text survives, but it lands on the one line the user has to read. Non-TTY output is fine — the spinner prints a single static message there — so this is interactive users only.

Either clear the line before `CliWarning`, or move the warning out of `SaveStreamAtomic`: carry a flag on `DownloadedFile` and warn from the caller once the spinner has stopped. The second changes signatures but takes CLI presentation out of the utility layer. Worth noting the current placement is not against the grain here — `utils/prompt.go` and `utils/pending_approval.go` already call `CliWarning` from this layer.

utils/utils_test.go
```diff
@@ -226,6 +226,57 @@ func TestSaveStreamAtomic_PreservesExistingFileMode(t *testing.T) {
...
+		stderr := captureStderr(t, func() {
+			_, err := SaveStreamAtomic(dest, strings.NewReader("y"), 0600)
+			require.NoError(t, err)
+		})
+		assert.Empty(t, stderr)
+	})
+
+	t.Run("a new file would be just as readable", func(t *testing.T) {
+		dest := filepath.Join(dir, "package")
+		require.NoError(t, os.WriteFile(dest, []byte("x"), 0644))
```
geunwoonoh: **Warning:** this subtest is umask-dependent and never reaches the branch it is named after.

Same defect `9ee6fb7` fixed in the other two tests. Measured:

```
$ umask 077 → WriteFile(0644) → 0600
```

Under `umask 077`, `keptPerm` is `0600` and the guard exits at the **second** condition (`keptPerm & 0044 == 0`). The **third** condition (`newFilePerm & 0044 != 0`) — the one this case exists to cover — is never evaluated. Deleting that third condition outright would leave this test green.

An `os.Chmod` would patch it, but extracting `keptModeIsWider` (see the comment on `utils/file_save.go:130`) lets all three branches be pinned without touching the filesystem or the umask at all:

```go
func TestKeptModeIsWider(t *testing.T) {
	requireUnixModes(t)
	tests := []struct {
		name              string
		kept, newFilePerm os.FileMode
		want              bool
	}{
		{"private key over a group-readable file", 0644, 0600, true},
		{"destination already owner-only", 0600, 0600, false},
		{"a new file would be just as readable", 0644, 0666, false},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			assert.Equal(t, tt.want, keptModeIsWider(tt.kept, tt.newFilePerm))
		})
	}
}
```

`TestSaveStreamAtomic_WarnsWhenKeptModeIsWiderThanANewFile` alone then suffices at the `SaveStreamAtomic` level as proof the wiring is in place.

utils/file_save.go
```diff
@@ -120,6 +121,19 @@ func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (in
 	return written, nil
 }
 
+// warnKeptWiderMode signals when the kept mode leaves the content readable
+// beyond what a fresh download would be (a private key over a 0644 file). A
+// warning, not a chmod—re-permissioning a file the user set up is not the
+// CLI's call (#326).
+func warnKeptWiderMode(fileName string, replacing bool, keptPerm, newFilePerm os.FileMode) {
```
geunwoonoh, line 135: **Suggestion:** `replacing` is dead at the only call site.

`finalPerm` is overwritten with `info.Mode().Perm()` only when `replacing` is true (lines 61-71). So when `!replacing`, `keptPerm == newFilePerm`, and the warning condition reduces to `X & 0044 != 0 && X & 0044 == 0` — never satisfiable. `!replacing ||` is behaviorally a no-op today.

Not necessarily an argument for removing it — the point is that the guard's safety currently rests on a call-site invariant rather than on anything local. If it stays, one line of comment saying why spares the next reader the derivation. Extracting `keptModeIsWider` also concentrates the judgement in one place.

utils/utils_test.go
```diff
@@ -226,6 +226,57 @@ func TestSaveStreamAtomic_PreservesExistingFileMode(t *testing.T) {
...
+	stderr := captureStderr(t, func() {
+		_, err := SaveStreamAtomic(dest, strings.NewReader("key"), 0600)
+		require.NoError(t, err)
+	})
+
+	assert.Contains(t, stderr, "already existed at 0644 and kept that mode")
+	assert.Contains(t, stderr, "at most 0600")
+}
+
+func TestSaveStreamAtomic_NoWarningWhenNothingIsWider(t *testing.T) {
```
geunwoonoh: **Suggestion:** three subtests with effectively the same body.

Both `CLAUDE.md` and the Go stack rules ask for table-driven tests. Folding these into the predicate table proposed on line 271 handles both this and the umask dependency in one move.

utils/file_save.go
```diff
@@ -120,6 +121,19 @@ func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (in
...
+	const groupOtherRead = os.FileMode(0o044)
```
geunwoonoh: **Suggestion:** octal literal style is mixed within this file.

`0o044` is the only `0o`-prefixed literal here; `0755` (lines 24, 57) and `0600` (line 15) use the traditional form. `gofmt` does not care, but one file should pick one.

```suggestion
	const groupOtherRead = os.FileMode(0044)
```

utils/file_save.go
```diff
@@ -120,6 +121,19 @@ func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (in
...
+	if !replacing || keptPerm&groupOtherRead == 0 || newFilePerm&groupOtherRead != 0 {
+		return
+	}
+	CliWarning("%s already existed at %04o and kept that mode; a new file would have been at most %04o. Remove it first for owner-only permissions.",
```
geunwoonoh, line 139: **Suggestion:** the path named here differs from the one the sibling error names.

The warning names `fileName` (the user-supplied path) while `destination is a directory` at line 65 names `targetName` (the symlink-resolved path). The mode is read from `targetName` and attributed to `fileName`.

For the remedy, `fileName` is the right choice — removing the symlink does leave a fresh `0600` file at that path. If that is deliberate, a short comment records it so the inconsistency does not read as an oversight.

geunwoonoh (changes requested):
## Summary

The kept-mode policy from #326 / #342 stays untouched and this adds one stderr warning for the case it was silently covering. Behavior is unchanged, and the three-condition AND guard correctly stays quiet for downloads that were never going to be private anyway. Direction and scope are both right, and the reasoning in the PR body for comparing read bits rather than full modes is sound.

Local verification:

```
gofmt -l .            no output
go vet ./...          no output
go test -race ./...   all packages pass
golangci-lint         not installed locally, not run
```

---

## Critical issues (1)

### 1. `utils/file_save.go:128-135` — the warning fires on every overwrite on Windows

Windows has no Unix modes. Go's `os.Stat` synthesizes them from `FILE_ATTRIBUTE_READONLY` alone (`$GOROOT/src/os/types_windows.go:177-182`):

```go
if fs.FileAttributes&syscall.FILE_ATTRIBUTE_READONLY != 0 {
    m |= 0444
} else {
    m |= 0666   // every writable file
}
```

So on the Windows `cp` path the guard always passes:

| Condition | Value | Result |
|---|---|---|
| `keptPerm` | `0666` (any writable file) | `0666 & 0044 != 0` → passes |
| `newFilePerm` | `0600` (`api/ftp/ftp.go:599`) | `0600 & 0044 == 0` → passes |

Unless the destination carries the read-only attribute, **every single-file overwrite download on Windows emits the warning**. The message is also meaningless there: deleting the file does not change any ACL, so "Remove it first for owner-only permissions" is not a remedy that holds.

Windows is a first-class release target — `.goreleaser.yaml` builds amd64/arm/arm64, and this package already carries `utils/file_replace_windows.go`. The tests skip Windows through `requireUnixModes`, so CI cannot catch this.

Suggested fix — extracting the predicate gives the Windows guard a home and also resolves warning 3 below:

```go
func warnKeptWiderMode(fileName string, replacing bool, keptPerm, newFilePerm os.FileMode) {
	if !replacing || !keptModeIsWider(keptPerm, newFilePerm) {
		return
	}
	CliWarning("%s already existed at %04o and kept that mode; a new file would have been at most %04o. Remove it first for owner-only permissions.",
		fileName, keptPerm, newFilePerm)
}

// keptModeIsWider reports whether keptPerm exposes the content to readers
// newFilePerm would not. Windows has no Unix modes—os.Stat synthesizes 0666 for
// every writable file—so the comparison carries no meaning there.
func keptModeIsWider(keptPerm, newFilePerm os.FileMode) bool {
	if runtime.GOOS == "windows" {
		return false
	}
	const groupOtherRead = os.FileMode(0o044)
	return keptPerm&groupOtherRead != 0 && newFilePerm&groupOtherRead == 0
}
```

`config/device.go:218` already uses `runtime.GOOS == "windows"`, so that is in keeping with the package. Splitting into build-tagged files the way `replaceFile` does would also work, but is heavy for a one-liner.

---

## Warnings (2)

### 2. `utils/file_save.go:72` — the warning is emitted while the spinner is animating

The call site sits inside the spinner window of `downloadSingleFileWithResult` (`api/ftp/ftp.go:677-678` starts it, `:703` calls `saveDownloadedURL`). The spinner goroutine repaints stderr every 100ms with `fmt.Fprintf(os.Stderr, "\r%s %s", ...)` (`utils/spinner.go:84`), and `CliWarning` does not clear the line first. On a TTY that reads as:

```
⠹ Downloading id_rsa...Warning: ./id_rsa already existed at 0644 and kept that mode; ...
⠸ Downloading id_rsa...
```

The text survives, but it arrives glued to a spinner frame — on the one line the user is meant to read. Non-TTY output is unaffected, since the spinner prints a single static message there, so this hits interactive users only.

Two ways out: clear the line before `CliWarning`, or move the warning out of `SaveStreamAtomic` — carry a flag on `DownloadedFile` and warn from the caller once the spinner has stopped. The second changes signatures but takes CLI presentation out of the utility layer. That said, the current placement is not against the grain of this repo: `utils/prompt.go` and `utils/pending_approval.go` already call `CliWarning` from the same layer.

### 3. `utils/utils_test.go:269-277` — the third subtest is umask-dependent and never reaches its branch

```go
t.Run("a new file would be just as readable", func(t *testing.T) {
    dest := filepath.Join(dir, "package")
    require.NoError(t, os.WriteFile(dest, []byte("x"), 0644))   // no chmod
```

This is the same defect `9ee6fb7` fixed in the other two tests. Measured:

```
$ umask 077 → WriteFile(0644) → 0600
```

Under `umask 077`, `keptPerm` is `0600` and the guard exits at the **second** condition (`keptPerm & 0044 == 0`). The **third** condition (`newFilePerm & 0044 != 0`) — the one this subtest is named after — is never evaluated. The test passes for the wrong reason: deleting the third condition outright would leave it green.

Extracting `keptModeIsWider` as in issue 1 lets all three branches be pinned by a table test that touches neither the filesystem nor the umask:

```go
func TestKeptModeIsWider(t *testing.T) {
	requireUnixModes(t)
	tests := []struct {
		name              string
		kept, newFilePerm os.FileMode
		want              bool
	}{
		{"private key over a group-readable file", 0644, 0600, true},
		{"destination already owner-only", 0600, 0600, false},
		{"a new file would be just as readable", 0644, 0666, false},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			assert.Equal(t, tt.want, keptModeIsWider(tt.kept, tt.newFilePerm))
		})
	}
}
```

`TestSaveStreamAtomic_WarnsWhenKeptModeIsWiderThanANewFile` alone then suffices at the `SaveStreamAtomic` level as proof the wiring is in place.

---

## Suggestions (5)

### 4. `utils/file_save.go:130` — `replacing` is dead at the only call site

`finalPerm` is overwritten with `info.Mode().Perm()` only when `replacing` is true (`file_save.go:61-71`). So when `!replacing`, `keptPerm == newFilePerm`, and the warning condition reduces to `X & 0044 != 0 && X & 0044 == 0` — never satisfiable. `!replacing ||` is behaviorally a no-op today.

Not necessarily an argument for deleting it; the point is that the guard's safety currently rests on a call-site invariant. If it stays, a one-line comment saying why spares the next reader the derivation. Extracting `keptModeIsWider` also concentrates the judgement in one place.

### 5. `utils/utils_test.go:247-278` — three subtests with effectively the same body

Both `CLAUDE.md` and the Go stack rules ask for table-driven tests. Folding these into the predicate table from suggestion 3 handles it.

### 6. `utils/file_save.go:130` — a restrictive umask produces a false negative, not a false positive

`newFilePerm` is the pre-umask mode, so passing `0666` under `umask 077` still lands at `0600`. If the existing file is `0644`, the kept mode genuinely *is* wider and the guard stays silent. The only call site passing `0666` is `api/packages/packages.go:172`, and packages are public artifacts, so there is no practical exposure.

Reading the umask via `syscall.Umask(0)` and restoring is racy and absent on Windows, so the current design is the right call. A line in the `keptModeIsWider` comment recording the limitation would save rediscovering it if more `newFilePerm` values appear later.

### 7. `utils/file_save.go:129` — octal literal style is mixed within one file

`0o044` is the only `0o`-prefixed literal; `0755` (lines 24, 57) and `0600` (line 15) in the same file use the traditional form. `gofmt` does not care, but one file should pick one.

### 8. `utils/file_save.go:133` — the path named in the message differs from the sibling error's

The warning names `fileName` (the user-supplied path) while the `destination is a directory` error at line 65 names `targetName` (the symlink-resolved path). The mode is read from `targetName` and attributed to `fileName`.

For the remedy, `fileName` is the correct choice — removing the symlink does leave a fresh `0600` file at that path. If that is deliberate, a short comment records it.

### Additionally — the PR body's archive-path claim is inaccurate

The body states the archive path "does not go through `SaveStreamAtomic` at all". It does: both `saveDownloadedURL` (`api/ftp/ftp.go:646`) and `downloadBulk` (`:763`) reach it through `fetchFromURLToFile` with `0600`.

The conclusion (no warning) holds, but for a different reason: `reserveDownloadArchiveTempPath` uses `os.CreateTemp` (`:624`), which the standard library creates at `0600`, so the second condition filters it out. Harmless today, but it is an implicit invariant nothing enforces — widening that temp file's mode later would produce a warning naming a `.alpacon-download-*.zip` path the user never created. Correcting the body before merge is enough.

---

## Performance

No cost on the warning path. `finalPerm` reuses the `os.Stat` already performed, so no syscall is added; the guard is two bit masks with no allocation; `fmt.Fprintf` runs only when the warning actually fires. Placing the call before `io.Copy` also keeps the streaming loop untouched.

---

## Highlights

- Reuses the existing `os.Stat` result instead of adding a second one
- Warns before any bytes land, so the user can still abort
- The three-condition AND guard filters the package download case precisely; the rationale for comparing read bits rather than full modes is well argued in the body
- stdout is untouched, so `--output json` consumers are unaffected
- The umask point from the Copilot round was resolved with `at most`, which is the bound that actually holds — and the regression test asserting `at most 0600` keeps the wording from silently drifting back
- The `os.Chmod` added to `TestSaveStreamAtomic_KeepsPartialReplacementUnreadable` fixes a latent umask-dependent failure along the way

---

**Verdict**: CHANGES REQUESTED

The Windows false positive (issue 1) puts a wrong warning on every overwrite download on a shipped platform, so it needs handling before merge. The test gap (issue 3) falls out of the same fix. Everything else is optional.

**Checklist**:
- [x] Follows Go conventions (mixed octal literals are a minor blemish)
- [ ] Tests actually exercise the branches they name (issue 3)
- [x] No security vulnerabilities (the change is itself a security improvement)
- [ ] Documentation accurate (PR body's archive-path claim)
- [ ] Cross-platform behavior (issue 1)

utils/file_save.go
jisung-02: Fixed in 5573295, taking the suggested shape: the predicate moved into `keptModeIsWider` with a `runtime.GOOS == "windows"` early return, matching the existing use in `config/device.go:218`. Agreed on skipping build-tagged files for a one-liner.

Verified with `GOOS=windows go build ./...` and `GOOS=windows go vet ./utils/`, both clean. `requireUnixModes` still skips the mode tests there, so the guard itself is what carries the platform behavior.

utils/file_save.go
jisung-02: Fixed in f3f1c12, taking the first option: `CliWarning` now clears the current line before printing.

The clear and the text go out in one `Fprintf` rather than two. As two writes the 100ms repaint can land between them and reproduce the same collision, just less often. The escape is emitted only when `term.IsTerminal(os.Stderr.Fd())`, so redirected stderr and CI logs are unchanged.

I left the warning in `SaveStreamAtomic` rather than carrying a flag out to the caller. Warning after `spinner.Stop()` means warning after the download finished, and the point of the current placement is that it lands before any bytes are written so a long transfer can still be aborted. Thanks for confirming the layering is not against the grain here.

utils/utils_test.go
jisung-02: Fixed in 32b4d94 with the table test as suggested. Confirmed it now pins the branch it names — deleting the third condition from `keptModeIsWider` fails it:

```
Error: Not equal: expected: false, actual: true
Test:  TestKeptModeIsWider/a_new_file_would_be_just_as_readable
FAIL
```

The old subtest stayed green under the same mutation.

`TestSaveStreamAtomic_WarnsWhenKeptModeIsWiderThanANewFile` is left as the wiring proof, plus `TestSaveStreamAtomic_NoWarningForANewFile` for the one branch that still needs a real write.

utils/file_save.go
jisung-02: Kept, with the derivation recorded in 5664dd9:

```go
// !replacing is redundant at today's only call site (keptPerm equals
// newFilePerm there, which no comparison satisfies) and stays so the guard does
// not rest on that invariant.
```

Your reading is right — with `keptPerm == newFilePerm` the condition is unsatisfiable. I would rather the guard not depend on a caller keeping `finalPerm` and `newFilePerm` in sync. The judgement is now concentrated in `keptModeIsWider` as you noted.

utils/utils_test.go
jisung-02: Folded into the predicate table in 32b4d94. The three filesystem subtests are gone; only `TestSaveStreamAtomic_NoWarningForANewFile` remains outside the table, because a destination that does not exist yet cannot be expressed as a `keptPerm`/`newFilePerm` pair.

utils/file_save.go
jisung-02: Fixed in 5664dd9: `0o044` is now `0044`, matching `0600` and `0755` in the same file.

utils/file_save.go
jisung-02: Deliberate, and now recorded in 5664dd9:

```go
// The message names fileName rather than the
// symlink-resolved targetName: removing the name the user typed is what gets
// them a fresh owner-only file.
```

Same conclusion you reached — the remedy applies to the path the user typed.

utils/log.go
```diff
@@ -91,8 +93,23 @@ func CliInfo(msg string, args ...any) {
...
+
+// clearStderrLine returns the escape that wipes whatever is on the current
+// line, or "" when stderr is not a terminal—there is no spinner animation to
+// clear, and the escape would only be noise in a log.
+func clearStderrLine() string {
+	if !term.IsTerminal(int(os.Stderr.Fd())) {
+		return ""
+	}
+	return "\r\033[K"
 }
```
geunwoonoh: **Critical:** this clear applies to every `CliWarning` in the codebase, and outside the spinner case it erases real output.

My finding was about one call site, `SaveStreamAtomic`. The fix landed in the helper, which has 70+ call sites. Each now emits `\r\033[K` whenever stderr is a TTY.

`\r\033[K` erases the terminal's **current line**, not "the line the spinner drew". stdout and stderr share a cursor on a terminal, so unterminated stdout content is what gets wiped.

`alpacon exec` streaming is exactly that shape:

- `cmd/exec/exec.go:198` — table mode sets `out = os.Stdout`
- `api/event/event.go:112-118` (`giveUpGap`) — writes chunk content to `out`, then calls `CliWarning` on the very next statement
- `api/event/event.go:578-584` (`applyChunk`) — same on a failed gap fetch
- Chunks are not line-aligned: the package's own tests use `{Seq: 3, Content: "hi"}` and `{Seq: 0, Content: "first"}` (`api/event/commandoutput_listener_test.go:26,211`)

Verified with a binary linked against the real `utils.CliWarning`, run under a pty:

```
raw: b'2026-08-20 12:00:01 INFO important remote output\r\x1b[K\x1b[1m\x1b[33mWarning\x1b[0m: chunk seq(s) 5..7 not arrived...\r\n'
```

Rendered:

```
=== AFTER  (this PR) ===
   |Warning: chunk seq(s) 5..7 not arrived; output may be incomplete|

=== BEFORE (main) ===
   |2026-08-20 12:00:01 INFO important remote outputWarning: chunk seq(s) 5..7 not arrived; output may be incomplete|
```

A line of real remote output is gone, and the warning that deleted it reads `output may be incomplete` — the user is told output is missing by the thing that removed it, with no way to tell what. Before, the same case was ugly but lossless. `api/event/event.go` alone has 13 warnings that can fire mid-stream (`:117`, `:141`, `:145`, `:148`, `:164`, `:178`, `:181`, `:373`, `:582`, `:634`, `:662`, `:664`, `:666`).

The root cause is ownership: the spinner drew the line, but the erase decision lives here and keys off "is stderr a TTY", which is not the same question as "is a spinner running". Gating on the latter restores the fix's intended scope.

The suggestion below is **one of three parts** — on its own it leaves `x/term` unused and will not build:

1. add the counter in `utils/spinner.go`
2. drop the now-unused `golang.org/x/term` import from this file
3. apply the suggestion

```go
// utils/spinner.go
// activeSpinners counts spinners animating stderr. A warning printed while one
// is running lands mid-frame, so CliWarning clears the line first—but only
// then, since the escape erases whatever else is on that line.
var activeSpinners atomic.Int32
```

`activeSpinners.Add(1)` after the `if !s.enabled` block in `Start()`, `Add(-1)` at the matching point in `Stop()`; the `s.running` early returns keep the pairing symmetric and nesting counts correctly. `s.enabled` is already `term.IsTerminal(os.Stderr)`, so the TTY test is subsumed.

```suggestion
func clearStderrLine() string {
	if activeSpinners.Load() == 0 {
		return ""
	}
	return "\r\033[K"
}
```

The download warning still prints clean (its spinner is active), exec streaming is untouched, and the clear still ships in one write.

Not covered by this: `cmd/exec/run.go:425-428` enters `streamApprovedCommand` with the approval spinner running, so `--wait` streaming would still clear. That path already repaints frames over streamed output every 100ms independently of this PR — a follow-up issue rather than a change here.

utils/log.go
```diff
@@ -91,8 +93,23 @@ func CliInfo(msg string, args ...any) {
...
 
 // CliWarning handles all warning messages in the CLI.
+//
+// A warning can land while a Spinner is animating stderr (SaveStreamAtomic
+// warns from inside a download), and the spinner leaves the cursor mid-line.
+// The clear and the text go out as one write so a spinner frame cannot wedge
+// itself between them.
 func CliWarning(msg string, args ...any) {
-	fmt.Fprintf(os.Stderr, "%s: %s\n", Yellow("Warning"), cliMessage(msg, args...))
+	fmt.Fprintf(os.Stderr, "%s%s: %s\n", clearStderrLine(), Yellow("Warning"), cliMessage(msg, args...))
```
geunwoonoh: **Suggestion:** the new branch has no test coverage.

`captureStderr` swaps stderr for a pipe, so `term.IsTerminal` is false and the escape path never runs in the suite. That is why the existing `CliWarning` assertions still pass — and why the regression above cleared CI.

Keying on `activeSpinners` instead makes it injectable: warn with no spinner running (no escape), warn with one started (escape). Two cases cover it.

By contrast the Windows early return in `keptModeIsWider` genuinely resists testing — CI is ubuntu-only and `runtime.GOOS` cannot be faked. `GOOS=windows go build ./...` plus inspection is the right level of assurance there; no change wanted.

utils/file_save.go
```diff
@@ -120,6 +122,41 @@ func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (in
...
+// keptModeIsWider reports whether keptPerm leaves the content readable to
+// accounts newFilePerm would shut out. Windows has no Unix modes—os.Stat
+// synthesizes 0666 for every file without the read-only attribute—so the
+// comparison would fire on every overwrite there, and removing the file is no
+// remedy for an ACL anyway.
+//
+// newFilePerm is the mode a new file is requested at, before the umask narrows
+// it, so a restrictive umask can only silence a warning that was due—never
+// invent one.
+func keptModeIsWider(keptPerm, newFilePerm os.FileMode) bool {
```
geunwoonoh, line 152: Resolved, and pulling the predicate out is what let the test problem in the other thread close at the same time.

The umask note here — "a restrictive umask can only silence a warning that was due—never invent one" — states the bound more precisely than my original comment did.

utils/utils_test.go
```diff
@@ -226,6 +226,58 @@ func TestSaveStreamAtomic_PreservesExistingFileMode(t *testing.T) {
...
+	})
+
+	assert.Contains(t, stderr, "already existed at 0644 and kept that mode")
+	assert.Contains(t, stderr, "at most 0600")
+}
+
+// TestKeptModeIsWider pins the guard directly: driving the same branches
+// through the filesystem makes them umask-dependent, and a umask that narrows
+// the setup mode retires the branch instead of testing it.
+func TestKeptModeIsWider(t *testing.T) {
```
geunwoonoh, line 250: Resolved. Confirming it by deleting the third condition and watching this fail is the check that separates a test which pins a branch from one that merely passes — thanks for running that.

geunwoonoh (changes requested):
## Re-review at `5664dd9`

Seven of the eight findings are resolved, and the ones you chose to document rather than change (`replacing`, the umask bound, `fileName` vs `targetName`) all carry their reasoning in the code now, so nobody has to re-derive them. Confirming the umask table test by deleting the third condition and watching it fail is the right way to prove a test pins what it names, and the `keptModeIsWider` note — "a restrictive umask can only silence a warning that was due—never invent one" — states the bound more precisely than my original comment did.

| # | Finding | Commit | Status |
|---|---|---|---|
| 1 | Windows false positive (critical) | `5573295` | Resolved |
| 2 | Spinner collision (warning) | `f3f1c12` | **Regression introduced** |
| 3 | umask-dependent test (warning) | `32b4d94` | Resolved |
| 4 | Dead `replacing` condition | `5664dd9` | Reasoning recorded |
| 5 | Table-driven tests | `32b4d94` | Resolved |
| 6 | umask false negative | `5664dd9` | Limitation documented |
| 7 | Mixed octal literals | `5664dd9` | Resolved |
| 8 | `fileName` vs `targetName` | `5664dd9` | Intent recorded |
| — | PR body's archive-path claim | — | Corrected |

Local verification on `5664dd9`: `gofmt -l .` and `go vet ./...` clean, `go test -race ./...` passes, and CI is 6/6 including golangci-lint and govulncheck.

The one blocker is that fix 2 was applied to the `CliWarning` helper rather than to the call site the finding was about.

---

## Critical issues (1)

### 1. `utils/log.go:101-113` — the spinner fix widened to every warning and now erases real output

My finding was about one call site, `SaveStreamAtomic`. The fix landed in `CliWarning` itself, which has 70+ call sites across the codebase. Every one of them now emits `\r\033[K` whenever stderr is a TTY.

`\r\033[K` does not erase "the line the spinner drew" — it erases **the terminal's current line**. stdout and stderr share a cursor on a terminal, so any stdout content not yet terminated by a newline is what gets wiped.

The `alpacon exec` streaming path has exactly that shape:

- `cmd/exec/exec.go:198` — table mode sets `out = os.Stdout`
- `api/event/event.go:112-118` (`giveUpGap`) — writes chunk content to `out`, then calls `CliWarning` on the very next statement
- `api/event/event.go:578-584` (`applyChunk`) — same shape on a failed gap fetch
- Chunks are not line-aligned. The package's own tests use `{Seq: 3, Content: "hi"}` and `{Seq: 0, Content: "first"}` (`api/event/commandoutput_listener_test.go:26,211`)

Verified by running a binary linked against the real `utils.CliWarning` under a pty:

```
raw: b'2026-08-20 12:00:01 INFO important remote output\r\x1b[K\x1b[1m\x1b[33mWarning\x1b[0m: chunk seq(s) 5..7 not arrived...\r\n'
```

Rendered:

```
=== AFTER  (this PR) ===
   |Warning: chunk seq(s) 5..7 not arrived; output may be incomplete|

=== BEFORE (main) ===
   |2026-08-20 12:00:01 INFO important remote outputWarning: chunk seq(s) 5..7 not arrived; output may be incomplete|
```

A line of real remote command output is gone. The warning riding along with it reads `output may be incomplete`, so the user is told output is missing by a message that just deleted some — with no way to tell what. Before this change the same case was ugly but lossless.

`api/event/event.go` alone holds 13 warnings that can fire mid-stream: `:117`, `:141`, `:145`, `:148`, `:164`, `:178`, `:181`, `:373`, `:582`, `:634`, `:662`, `:664`, `:666`.

**The root cause is ownership.** The spinner drew the line, but the decision to erase it sits in `CliWarning` and keys off "is stderr a TTY" — which is not the same question as "is a spinner running". Narrowing the clear to the case where one actually is restores the fix's original scope:

```go
// utils/spinner.go
// activeSpinners counts spinners animating stderr. A warning printed while one
// is running lands mid-frame, so CliWarning clears the line first—but only
// then, since the escape erases whatever else is on that line.
var activeSpinners atomic.Int32
```

`activeSpinners.Add(1)` after the `if !s.enabled` block in `Start()`, `Add(-1)` at the matching point in `Stop()` — the early returns on `s.running` keep the pairing symmetric, and nesting counts correctly. Since `s.enabled` is already `term.IsTerminal(os.Stderr)`, the TTY test is subsumed and `utils/log.go` no longer needs `x/term`.

```go
// utils/log.go
func clearStderrLine() string {
	if activeSpinners.Load() == 0 {
		return ""
	}
	return "\r\033[K"
}
```

The download warning still prints clean because its spinner is active; exec streaming is left untouched; and the clear still ships in the same single write.

One case this does not cover: `cmd/exec/run.go:425-428` enters `streamApprovedCommand` with the approval spinner still running, so `--wait` streaming would still clear. That path is already repainting frames over streamed output every 100ms independently of this PR, so it is a separate problem — worth a follow-up issue rather than a change here.

---

## Suggestions (1)

### 2. `utils/log.go:108` — the new branch has no test

`captureStderr` swaps stderr for a pipe, so `term.IsTerminal` is false and the TTY branch never executes in the suite. That is why the existing `CliWarning` assertions still pass, and also why this regression cleared CI.

Keying the decision on `activeSpinners` instead makes it injectable: warn with no spinner running (expect no escape), warn with one started (expect the escape). Two cases cover it.

For contrast, the Windows early return in `keptModeIsWider` genuinely resists testing — CI is ubuntu-only and `runtime.GOOS` cannot be faked. `GOOS=windows go build ./...` plus inspection is the right level of assurance there; no change wanted.

---

## Highlights

- Issue 1 was taken in the suggested shape, and extracting `keptModeIsWider` resolved the test gap in issue 3 at the same time
- Proving the table test by deleting the third condition and watching it fail — that is the check that distinguishes a test that pins a branch from one that merely passes
- Choosing documentation over change for 4, 6, and 8, with the derivations in the code rather than only in review replies
- The PR body's archive-path correction, including the point that `os.CreateTemp` at `0600` is an implicit invariant nothing enforces
- `f3f1c12`'s single-`Fprintf` construction is right on its own terms: a spinner repaint cannot wedge between the clear and the text. Only the scope needs narrowing

---

**Verdict**: CHANGES REQUESTED

`utils/file_save.go` and the tests are in approvable shape. The remaining work is roughly ten lines across `utils/log.go` and `utils/spinner.go`.

**Checklist**:
- [x] Seven of eight prior findings resolved
- [x] Follows Go conventions
- [x] Tests pass (`go test -race ./...`, gofmt, vet, CI 6/6)
- [ ] No new regressions — `CliWarning` scope (critical 1)
- [x] Documentation accurate

utils/log.go
jisung-02: Confirmed, and fixed across a5c88b0 and 5e6be75. Your reproduction held: the erase was reaching all 77 `CliWarning` sites, and on `alpacon exec` streaming it was deleting a line of real remote output.

The counter landed as you specified. `activeSpinners atomic.Int32` sits in `utils/spinner.go`, `Add(1)` after the `if !s.enabled` block in `Start`, `Add(-1)` in `Stop` after `<-s.doneCh` so the window where the last frame is still on screen keeps its break. `golang.org/x/term` is gone from `utils/log.go`. Your point about ownership is the one that decided the shape: "is stderr a TTY" is not "is a spinner running", and `s.enabled` already subsumes the first.

One departure from the suggestion, and it comes from your own closing note. `cmd/exec/run.go` enters `streamApprovedCommand` with the approval spinner still running, so on that path `activeSpinners > 0` and stdout is streaming at the same time. Gating the erase on the counter narrows it but does not make it lossless there — the `--wait` case would still lose a line. So `spinnerLineBreak` returns `"\n"` rather than `"\r\033[K"`:

```go
func spinnerLineBreak() string {
	if activeSpinners.Load() == 0 {
		return ""
	}
	return "\n"
}
```

The trade is that the last spinner frame stays on screen and the spinner redraws on the line below. That costs one line of cosmetics and it cannot destroy output on any path, which the erase cannot promise while a spinner and a stream share a line.

I also fixed the follow-up you scoped out rather than deferring it, because it is what made the erase unsafe under the counter. `Spinner.StopWriter(io.Writer) io.Writer` retires a spinner on the first non-empty write, so a caller that waits on something which may start producing output no longer has to know when the first byte arrives (a939c20). The `--wait` loop needed more than that: every tick re-submits, so a tick whose request is still pending writes its denial line and spends the one-shot handover meant for the approved output, and `Spinner.Stop` is terminal — it closes channels `Start` does not reopen. Each stretch of waiting therefore gets a spinner and a writer of its own (3ab8148). Without that the wait animated through its first poll gap and then sat silent for the rest of what `--wait` exists to cover. The MFA wait in `utils/error_handler.go` got the same treatment: the new flow stops its spinner as soon as the poll reports completion, before `RetryOperation` can stream, and the legacy flow prints a static line instead, since every tick of that loop retries the operation and no stretch of it is safe to animate.

One call site beyond `CliWarning` needed the break too. `CliDebug` (702b972): the refresh-token degrade at `api/auth0/auth0.go:286` logs from inside the `Refreshing access token...` spinner that `client.NewAlpaconAPIClient` starts (`client/client.go:81`), and a line whose whole job is to tell a silent degrade apart from the path it replaced cannot do that from mid-frame.

Verified: `go test -race ./...` 39 ok / 0 FAIL, `golangci-lint run ./...` 0 issues, `gofmt -l .` and `go vet ./...` clean, `GOOS=windows go build ./...` clean.

utils/log.go
jisung-02: Fixed in 8c301ce and the tests alongside it. Your diagnosis of why the regression cleared CI was the useful part: `captureStderr` swaps stderr for a pipe, so `term.IsTerminal` is false and the escape path never ran in the suite.

Keying on `activeSpinners` made it injectable as you said, so the two cases you named exist:

```go
func TestCliWarning_BreaksTheLineOnlyWhileASpinnerRuns(t *testing.T) {
	t.Run("nothing is animating stderr", func(t *testing.T) { ... })

	t.Run("a spinner is animating stderr", func(t *testing.T) {
		activeSpinners.Add(1)
		t.Cleanup(func() { activeSpinners.Add(-1) })

		stderr := captureStderr(t, func() { CliWarning("output may be incomplete") })

		assert.True(t, strings.HasPrefix(stderr, "\n"), "the break must lead the write: %q", stderr)
		assert.NotContains(t, stderr, "\033[K")
	})
}
```

`TestCliDebug_BreaksTheLineOnlyWhileASpinnerRuns` mirrors it for the debug line.

Faking the counter covers `CliWarning`, but it also means nothing checks that a real spinner ever sets it — a dropped `Add(1)` would leave the whole suite green, because the branch that counts is the one `go test` never enters. So `TestSpinner_AnimatingBranchPairsTheActiveCount` forces `s.enabled` and a 1ms interval to reach it, and pins the pairing:

```go
s := NewSpinner("Waiting for approval...")
s.enabled = true
s.interval = time.Millisecond

require.Equal(t, int32(0), activeSpinners.Load())
s.Start()
assert.Equal(t, int32(1), activeSpinners.Load(), "an animating spinner arms the line break")
s.Stop()
assert.Equal(t, int32(0), activeSpinners.Load(), "and Stop disarms it")
```

Being in the same package is what makes that possible without cutting a seam into production code. `TestSpinner_DoesNotArmTheLineBreakOffATerminal` holds the other side, that the non-TTY path leaves the count at zero.

Each one was seen fail against the change it covers. Deleting `activeSpinners.Add(1)`: `expected: 0, actual: -1`. Dropping `spinnerLineBreak()` from `CliDebug`: `the break must lead the write: "Debug: fell back to the device scope\n"`. Emptying the `restartSpinner` body in `cmd/exec/run.go`: both writers point at the same spent `sync.Once`, which is `TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnSpinner`.

Agreed on `keptModeIsWider`'s Windows early return — `GOOS=windows go build ./...` plus inspection stays the level of assurance, no change made.

utils/log.go
```diff
@@ -91,8 +96,32 @@ func CliInfo(msg string, args ...any) {
...
+// A warning can land while a Spinner is animating stderr (SaveStreamAtomic
+// warns from inside a download), and the spinner leaves the cursor mid-line.
+// The break and the text go out as one write so a spinner frame cannot wedge
+// itself between them.
 func CliWarning(msg string, args ...any) {
-	fmt.Fprintf(os.Stderr, "%s: %s\n", Yellow("Warning"), cliMessage(msg, args...))
+	fmt.Fprintf(os.Stderr, "%s%s: %s\n", spinnerLineBreak(), Yellow("Warning"), cliMessage(msg, args...))
+}
+
+// spinnerLineBreak returns the newline that moves a message off the line a
```
geunwoonoh, line 108: This is a better answer than the one I suggested, and worth saying so.

Gating `\r\033[K` on a running spinner — my proposal — rests on "erasing is safe while a spinner is animating", and that premise does not hold: a partial stdout line longer than the spinner's frame keeps a tail to the right of it, and `\033[K` still takes that tail. A newline cannot erase anything, whatever is on the line.

Confirmed under a pty with the real binary: with no spinner running, `activeSpinners == 0` emits nothing at all, so exec streaming is byte-identical to `main`; with the download spinner running, the warning gets its own line and the frame above survives.

Recording the reasoning ("stdout and stderr share a cursor on a terminal") rather than only the behavior is what makes this hold up under the next edit.

utils/spinner.go
```diff
@@ -113,4 +129,25 @@ func (s *Spinner) Stop() {
 
 	close(s.stopCh)
 	<-s.doneCh
+	// After the goroutine has returned, so the window where the last frame is
+	// still on screen keeps its line break.
+	activeSpinners.Add(-1)
```
geunwoonoh, line 134: **Suggestion (pre-existing, non-blocking):** `Start()` on a stopped spinner panics on a TTY.

`Stop()` leaves `stopCh` closed, so a subsequent `Start()` launches a goroutine that falls straight through its `select` to the deferred `close(s.doneCh)` — on a channel the previous goroutine already closed. Reproduced under a pty:

```
panic: close of closed channel
	utils.(*Spinner).Start.func1() utils/spinner.go:85
```

This predates the PR, but `StopWriter` changes its character: stopping is now implicit, so `RunExecWithApprovalWait`'s correctness rests on never restarting a spinner, and the only thing enforcing that is the comment at `cmd/exec/run.go:459-463`.

What makes it worth hardening is the blind spot: off a TTY `s.enabled` is false and `Start` returns before the goroutine, and **`go test` is never a TTY** — so a future regression here passes the entire suite and panics on a user's terminal. Re-initializing `stopCh`/`doneCh` in `Start()`, or making `Start` a no-op once stopped, closes it cheaply.

No change needed for this PR; the current call sites all construct a fresh spinner.

utils/log.go
```diff
@@ -91,8 +96,32 @@ func CliInfo(msg string, args ...any) {
...
+// streamed output not yet ended by a newline with it. Spinner.StopWriter is what
+// keeps a spinner off a line a caller is about to stream to, but it only covers
+// the call sites that opt in, and this helper backs every CliWarning and
+// CliDebug site—far too many to each be audited. The last spinner frame stays on
+// screen in trade; the spinner redraws on the line below.
+//
+// Gated on a spinner rather than on stderr being a terminal: off a TTY the
+// static progress line already ends in a newline, so a second one would only
+// add a blank line.
+func spinnerLineBreak() string {
```
geunwoonoh, line 128: **Suggestion:** worth a line on why `CliInfo` / `CliSuccess` / `CliError` are exempt.

The current split is correct — every spinner-adjacent call site among them runs after `spinner.Stop()` (the MFA timeout's `CliErrorWithExit`, the completion `CliSuccess`). But this comment says the helper backs "every CliWarning and CliDebug site" without saying why the rest need nothing, so the next person adding a `CliInfo` under a spinner has to re-derive it.

utils/spinner.go
```diff
@@ -113,4 +129,25 @@ func (s *Spinner) Stop() {
...
+// to and Stop erases that line, so output streamed while one is running gets
+// drawn over and then wiped. Callers that wait on something which may start
+// producing output—an approved command resuming, an operation retried after
+// MFA—hand their writer through here instead of having to know when the first
+// byte arrives. An empty write is not output, so it leaves the spinner alone.
+func (s *Spinner) StopWriter(w io.Writer) io.Writer {
+	return &spinnerStopWriter{spinner: s, w: w}
+}
+
+func (w *spinnerStopWriter) Write(p []byte) (int, error) {
```
geunwoonoh, line 148: Nice shape. Callers hand over a writer instead of having to know when the first byte arrives, which is the part that would otherwise get re-derived wrong at every site.

The `len(p) > 0` guard matters more than it looks — an empty write from a streaming layer would otherwise retire the spinner before there is anything to show — and it is pinned by a test.

I walked every exit of `RunExecWithApprovalWait` with this in mind: timeout, status-hold switch, poll-failure give-up, and answered all pass through `spinner.Stop()`, `restartSpinner` stops before constructing the next one, and a tick that already wrote makes both a no-op. No path leaks a spinner or pushes the count past one.

utils/spinner_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	assert.Equal(t, "first and second", out.String())
+}
+
+// Only the animating branch counts itself, and off a TTY—which is every go test
+// run—Start returns before reaching it. Forcing enabled is what lets a test see
+// that branch at all, and the pairing it pins is what CliWarning's line break
+// reads: a dropped increment, or a decrement moved ahead of the goroutine's
+// exit, would otherwise leave the whole suite green.
+func TestSpinner_AnimatingBranchPairsTheActiveCount(t *testing.T) {
```
geunwoonoh, line 68: This is the test that makes the rest trustworthy.

Off a TTY `Start` returns before the counting branch, and `go test` is never a TTY — so without forcing `enabled` the increment/decrement pair has no coverage at all, and the comment says exactly that. A dropped increment, or a decrement moved ahead of the goroutine's exit, would otherwise leave the whole suite green while `CliWarning`'s break silently stopped working.

geunwoonoh (dismissed):
## Re-review at `702b972` — approving

The critical issue is resolved, and resolved better than what I proposed. My suggestion was to keep `\r\033[K` and gate it on a running spinner. Breaking to a new line instead is the stronger answer: a newline cannot erase anything, whatever happens to be on that line.

```go
func spinnerLineBreak() string {
	if activeSpinners.Load() == 0 {
		return ""
	}
	return "\n"
}
```

My version rested on "erasing is safe while a spinner is animating", and that premise does not hold: a partial stdout line longer than the spinner's frame keeps a tail to the right of it, and `\033[K` would still take that tail. The newline covers that case too, and the comment records exactly why ("stdout and stderr share a cursor on a terminal").

Verified both paths with the real binary under a pty:

```
--- exec streaming (no spinner) ---
   |2026-08-20 12:00:01 INFO important remote outputWarning: chunk seq(s) 5..7 not arrived...|

--- cp download (spinner running) ---
   |⠹ Downloading id_rsa.  |
   |Warning: .../id_rsa already existed at 0644 and kept that mode; a new file would have been at most 0600...|
```

exec streaming is byte-identical to `main` — `activeSpinners == 0`, so nothing is emitted at all — and the download warning gets its own line. `gofmt`, `go vet`, and `go test -race ./...` are clean locally; CI is 6/6.

### The deferred work landed too

`Spinner.StopWriter` covers the two overlaps I had said to split into a follow-up:

- `cmd/exec/run.go:445` — `--wait` approval streaming, spinner retired at the first byte
- `utils/error_handler.go:59-80` — the new MFA flow stops before `RetryOperation`; the legacy flow drops the animation entirely for a static `CliInfo`, since every tick there retries an operation that can stream

I walked every exit of `RunExecWithApprovalWait`: timeout, status-hold switch, poll-failure give-up, and answered all pass through `spinner.Stop()`, and `restartSpinner` stops before constructing the next one. No path leaks a spinner or pushes the counter past one.

Also confirmed the two supporting claims: the `CliDebug` break is genuinely needed (`client/client.go:82` wraps `auth0.RefreshAccessToken` in a spinner and `api/auth0/auth0.go:286` logs from inside it), and the new README clause is accurate (`utils/utils.go:428` passes a mode to `O_CREATE`, which an existing file ignores).

---

## Suggestions (3) — none blocking

### 1. `utils/spinner.go:85` — `Start()` after `Stop()` panics on a TTY (pre-existing)

Reproduced under a pty:

```
panic: close of closed channel
	utils.(*Spinner).Start.func1() utils/spinner.go:85
```

`Stop()` leaves `stopCh` closed, so a new goroutine falls straight through to its deferred `close(s.doneCh)` on an already-closed channel. Not introduced here — but `StopWriter` changes its character, because stopping is now implicit. `RunExecWithApprovalWait`'s correctness depends on never restarting a spinner, and the only thing enforcing that is the comment at `run.go:459-463`.

Off a TTY, `s.enabled` is false and `Start` returns before the goroutine, so **`go test` can never reach this** — a future regression would pass the whole suite and panic on a user's terminal. Re-initializing the channels in `Start()`, or making `Start` a no-op on a stopped spinner, closes it cheaply.

### 2. Scope grew from 3 files / +67 to 10 files / +362

`utils/spinner.go` and `utils/log.go` are load-bearing for the regression fix, but `utils/error_handler.go` and `cmd/exec/run.go` fix bugs this PR did not introduce. A change titled "warn when a kept file mode is wider than a new download" that also reworks the MFA spinner is hard to revert as a unit.

The commits are cleanly separated (`fix(exec): give every stretch of an approval wait its own spinner` says what it does), so the history stays honest and this is not worth another round. For next time: landing `spinnerLineBreak` here and the exec/MFA overlaps as a follow-up would have kept both reviewable.

### 3. `utils/log.go` — `CliInfo` / `CliSuccess` / `CliError` do not take the break

Correct as it stands: every spinner-adjacent call site among them (the MFA timeout's `CliErrorWithExit`, the completion `CliSuccess`) runs after `spinner.Stop()`. But `spinnerLineBreak`'s comment only says it backs "every CliWarning and CliDebug site", so the reason the others are exempt is not written down. One line would stop someone adding a `CliInfo` under a spinner later.

---

## Highlights

- The newline over the erase — safer than what I suggested, with the reasoning recorded rather than just the behavior
- `StopWriter` — callers hand over a writer instead of having to know when the first byte arrives, and an empty write correctly leaves the spinner alone
- `TestSpinner_AnimatingBranchPairsTheActiveCount` forcing `s.enabled = true` to reach a branch `go test` structurally cannot, with a comment naming exactly what would otherwise stay green
- Calling the surviving last frame an accepted trade in the comment rather than leaving it as an unexplained artifact — losslessness is the right thing to prioritize there
- The README gaining the archive path's kept-mode behavior and the fact that no warning covers it

---

**Verdict**: APPROVED

The critical issue is fixed, the two items I said could wait were handled, and the three remaining points are non-blocking — the first of them pre-existing.

**Checklist**:
- [x] Prior review addressed, critical included
- [x] Follows Go conventions
- [x] Tests pass (`go test -race ./...`, gofmt, vet, CI 6/6), with coverage for the new branches
- [x] No regressions (verified under a pty)
- [x] Documentation accurate

utils/log.go
jisung-02: Written down in f6afce2. The rule was in my head and not in the file, which is exactly the gap you named.

```go
// Only CliWarning and CliDebug take the break, because they are the two that
// print from inside a call a spinner wraps: SaveStreamAtomic warns during a
// download, and the refresh degrade logs during "Refreshing access token...".
// CliInfo, CliSuccess, and CliError print from the cmd layer or an interactive
// prompt, and by then the api call that owned the spinner has returned. One of
// them needing the break means it has gained a call site under a spinner—look
// at that site rather than only adding the break here.
```

I checked the claim before writing it rather than restating your reading of the call sites. Every spinner window in the tree is `api/ftp/ftp.go` (6), `api/mfa/mfa.go:117`, `api/auth0/auth0.go:180`, `client/client.go:81`, `utils/error_handler.go:58`, and `cmd/exec/run.go:440,452`. The only `CliInfo`/`CliSuccess`/`CliError` calls below the cmd layer are `api/iam/iam.go:365,366,373,386,398` and `pkg/tunnel/runtime/runtime.go:289,306`, and neither runs under a spinner: `HandleUsernameRequired` is reached through `HandleCommonErrors`, and all eleven of its call sites are in `cmd/` (`cmd/ftp/cp.go:107,136`, `cmd/websh/websh.go:281`, `cmd/edit/edit.go:143,154`, `cmd/exec/run.go:628`, `cmd/exec/exec.go:158`, `cmd/server/server_action.go:32`, `cmd/tunnel/tunnel.go:183`, `cmd/workspace/*_update.go`), so the api call whose `defer spinner.Stop()` owned the animation has already returned by then. The tunnel runtime has no spinner anywhere in its path.

So the line the comment now states is the actual invariant, not a description of today's call list: the two that need the break are the two printed from inside library code a spinner wraps.

Comment-only change—`gofmt -l .` and `go vet ./...` clean, `go test ./utils/` ok.

utils/spinner.go
jisung-02: Reproduced and filed as #373, not fixed here.

Your trace held exactly, including the line. Forcing `enabled` in a same-package test:

```
panic: close of closed channel

goroutine 38 [running]:
github.com/alpacax/alpacon-cli/utils.(*Spinner).Start.func1()
	utils/spinner.go:85 +0x404
created by github.com/alpacax/alpacon-cli/utils.(*Spinner).Start in goroutine 36
	utils/spinner.go:75 +0x1c4
```

`NewSpinner` at `utils/spinner.go:52-53` is the only place `stopCh` and `doneCh` are made, so a second `Start` runs a goroutine that falls through the closed `stopCh` case straight into its deferred `close(s.doneCh)`.

Keeping it out of this PR is the same argument as your point 2: the scope already grew from 3 files to 10, and this is a pre-existing panic in code no current call site can reach. The issue carries the reproduction, both fixes you suggested (re-initialize the channels in `Start`, or make `Start` inert once stopped), the note that any fix needs `s.enabled` forced by hand since `go test` cannot reach the branch otherwise, and the `activeSpinners` pairing a reusable spinner must not drift.

Point 2 taken for the next one. Landing `spinnerLineBreak` alone and the exec/MFA overlaps as a follow-up would have left both reviewable, and I should have split when the second bug turned up rather than after.

geunwoonoh (approved):
## Re-approving at `f6afce2`

Re-approving because the push dismissed the previous approval — the repo dismisses stale approvals on new commits, and `702b972` → `f6afce2` was enough to trip it even though the commit only adds comment lines.

`f6afce2` is documentation only: eight `//` lines inside `spinnerLineBreak`'s doc block, no executable change. The code is identical to what I approved at `702b972`. `gofmt`, `go vet`, and `go test -race ./...` are clean locally; CI is 6/6.

### The three suggestions from last round

| # | Item | Outcome |
|---|---|---|
| 1 | `Spinner.Start()` after `Stop()` panics on a TTY | Filed as #373, not fixed here — as recommended |
| 2 | PR scope growth | A note for next time; nothing to do |
| 3 | Why `CliInfo` / `CliSuccess` / `CliError` are exempt | Written down in `f6afce2` |

The new comment holds up. Both premises check out — `SaveStreamAtomic` warns from inside a download, and the refresh degrade logs from inside "Refreshing access token..." (`client/client.go:82` → `api/auth0/auth0.go:286`). The closing sentence is the part that will earn its keep:

> One of them needing the break means it has gained a call site under a spinner—look at that site rather than only adding the break here.

That points the next person at the cause rather than at the symptom, which is the failure mode a helper like this invites.

Splitting #373 out rather than folding it in was the right call: it is pre-existing, unreachable from current call sites, and its fix touches channel lifecycle that has nothing to do with a file-mode warning.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Prior review addressed
- [x] Follows Go conventions
- [x] Tests pass (`go test -race ./...`, gofmt, vet, CI 6/6)
- [x] No regressions — code unchanged since the last approval
- [x] Documentation accurate

PR #367, opened 2026-08-19
fix(ftp): keep the trailing colon on a user-prefixed remote target

### LINKED ISSUE #349: cp drops the trailing colon from `user@server:`, so the wrong error fires
## Summary

`alpacon cp file.txt alice@myserver:` reports an invalid source/destination combination instead of the missing remote path. The same destination without the user prefix—`myserver:`—reports the accurate error.

## Cause

`cmd/ftp/cp.go:77` reassembles every argument holding both `@` and `:` after parsing it:

```go
if sshTarget.Path != "" {
    args[i] = sshTarget.Host + ":" + sshTarget.Path
} else {
    args[i] = sshTarget.Host
}
```

For `alice@myserver:`, `ParseSSHTarget` returns `Host: "myserver"`, `Path: ""` (`utils/ssh_parser.go:19`), so the else branch drops the colon along with the user. `validatePaths` receives a bare `myserver`, which `utils.IsRemoteTarget` classifies as local; both operands local, it falls to the final else. `myserver:` never enters the loop—the condition needs an `@`—so it survives intact and hits the remote-path check.

## Reproduction

```
$ alpacon cp file.txt alice@myserver:
Error: Invalid combination of source and destination paths.
...

$ alpacon cp file.txt myserver:
Error: invalid remote path format: 'myserver:'

Remote paths must include both server name and path:
  • Correct: myserver:/path/to/file
  • Incorrect: :myfile (missing server name)
  • Incorrect: myserver: (missing path)
```

Both exit `1`; only the message differs.

## Impact

Non-blocking—no transfer runs either way and the exit code is unchanged, so this misleads the reader rather than losing data. An empty path is the only case where the reassembly discards something the user typed.

## Suggested fix

Keep the colon when the parsed path is empty but the original argument ended with one:

```go
} else if strings.HasSuffix(arg, ":") {
    args[i] = sshTarget.Host + ":"
} else {
    args[i] = sshTarget.Host
}
```

The destination then stays remote and reaches the existing missing-path check. `alice@myserver` keeps falling to the bare-host branch.

Found while reviewing #348.

jisung-02:
## Background

`alpacon cp file.txt alice@myserver:` reported an invalid source/destination combination instead of the missing remote path. The same destination without the user prefix — `myserver:` — reported the accurate error.

`cmd/ftp/cp.go` reassembled every argument holding both `@` and `:` after parsing it:

```go
if sshTarget.Path != "" {
    args[i] = sshTarget.Host + ":" + sshTarget.Path
} else {
    args[i] = sshTarget.Host
}
```

For `alice@myserver:`, `utils.ParseSSHTarget` returns `Host: "myserver"`, `Path: ""`, so the else branch dropped the colon along with the user. `validatePaths` then received a bare `myserver`, which `utils.IsRemoteTarget` classifies as local; with both operands local it fell through to the mismatch error. `myserver:` never entered the loop — the condition needs an `@` — so it survived intact and hit the remote-path check.

An empty path is the only case where the reassembly discarded something the user typed. No transfer ran either way and the exit code was already `1`, so this misled the reader rather than losing data.

## Changes

Five commits, ordered so review reads top to bottom.

| Commit | What |
|---|---|
| `fix(ftp)` | the reassembly moves into `stripUserPrefix`, where a trailing colon in the original argument keeps the target remote |
| `refactor(ftp)` | drop the branch that helper's own guard made dead |
| `test(ftp)` | pin the empty-host target, drop the case that proved nothing |
| `refactor(ftp)` | the normalization loop moves into `normalizeArgs` so `TestCpCommandSSHParsing` calls it instead of inlining a copy |
| `test(ftp)` | pin the username precedence that copy left unreachable |

### Why the dead branch went

`stripUserPrefix` returns early unless the argument holds both `@` and `:`, so by the time `ParseSSHTarget` runs the colon is always present. `splitUserPrefix` (`utils/ssh_parser.go:44`) only shifts the colon, never removes it, so `Path` always comes from the `SplitN` and is empty exactly when the argument ends with a colon. That made the first cut of the fix — `if sshTarget.Path != "" || strings.HasSuffix(arg, ":")` — a tautology, with an unreachable return below it:

```go
sshTarget := utils.ParseSSHTarget(arg)
return sshTarget.Host + ":" + sshTarget.Path, sshTarget.User
```

Keeping the branch bought nothing. It dropped the colon and the path, which is the defect being fixed here, so it was never a usable fallback if `ParseSSHTarget` changed. `TestStripUserPrefix` holds the observable behavior instead.

## Testing

`cmd/ftp/cp_test.go` — `TestStripUserPrefix`, six cases over the helper:

| Argument | Rewritten | User |
|---|---|---|
| `alice@myserver:/tmp/f` | `myserver:/tmp/f` | `alice` |
| `alice@myserver:` | `myserver:` | `alice` |
| `alice@:` | `:` | `alice` |
| `myserver:/tmp/f` | unchanged | — |
| `./mail@example.txt` | unchanged | — |
| `alice@myserver:/tmp/mail@example.txt` | `myserver:/tmp/mail@example.txt` | `alice` |

`alice@myserver:` is the regression proof — the exact input from the report. `alice@:` is adjacent: it reached the old code as an empty string, a local argument nobody typed, and now rewrites to `:` so `validatePaths` reports the missing server name.

`TestCpCommandSSHParsing` used to inline the normalization loop rather than call `cp`, and that copy still carried the branch the `fix` commit deletes — it pinned `alice@myserver:` rewriting to `myserver` as correct. The loop now lives in `normalizeArgs` and the eight cases run it, so deleting `stripUserPrefix` no longer leaves the file green. The copy was the only use of the `strings` and `utils` imports in the test file, so both go with it.

`TestNormalizeArgsUsername` pins the username rule the copy left unreachable: `-u` outranks an inline user, and the first inline user outranks later ones. The old `if username == "" && sshTarget.User != ""` became `if username == "" { username = user }`; when `user` is empty that assigns `""` onto a variable already `""`, so the precedence is unchanged.

`utils.ParseSSHTarget` is untouched, so `websh`, `edit`, and `exec` are unaffected.

```
$ gofmt -l .                 (no output)
$ go vet ./...               (no output)
$ golangci-lint run ./...    0 issues.
$ go test -race ./...        39 ok, 0 FAIL
```

Closes #349

### DIFF
--- a/cmd/ftp/cp.go
+++ b/cmd/ftp/cp.go
@@ -74,13 +74,7 @@ Requires an active WorkSession when using Browser login (Auth0); Token auth (API
 			return
 		}
 
-		for i, arg := range args {
-			rewritten, user := stripUserPrefix(arg)
-			if username == "" {
-				username = user
-			}
-			args[i] = rewritten
-		}
+		username = normalizeArgs(args, username)
 
 		sources := args[:len(args)-1]
 		dest := args[len(args)-1]
@@ -203,6 +197,19 @@ func isLocalPaths(paths []string) bool {
 		}
 	}
 	return true
+}
+
+// normalizeArgs strips inline user prefixes in place and resolves the username:
+// the -u flag wins, then the first inline user.
+func normalizeArgs(args []string, username string) string {
+	for i, arg := range args {
+		rewritten, user := stripUserPrefix(arg)
+		if username == "" {
+			username = user
+		}
+		args[i] = rewritten
+	}
+	return username
 }
 
 // stripUserPrefix rewrites user@host:path as host:path, handing back the user.
--- a/cmd/ftp/cp.go
+++ b/cmd/ftp/cp.go
@@ -214,10 +214,7 @@ func stripUserPrefix(arg string) (rewritten, user string) {
 		return arg, ""
 	}
 	sshTarget := utils.ParseSSHTarget(arg)
-	if sshTarget.Path != "" || strings.HasSuffix(arg, ":") {
-		return sshTarget.Host + ":" + sshTarget.Path, sshTarget.User
-	}
-	return sshTarget.Host, sshTarget.User
+	return sshTarget.Host + ":" + sshTarget.Path, sshTarget.User
 }
 
 func validatePaths(sources []string, dest string) error {
--- a/cmd/ftp/cp.go
+++ b/cmd/ftp/cp.go
@@ -75,19 +75,11 @@ Requires an active WorkSession when using Browser login (Auth0); Token auth (API
 		}
 
 		for i, arg := range args {
-			if strings.Contains(arg, "@") && strings.Contains(arg, ":") {
-				// Parse SSH-like target: user@host:path (requires : to distinguish from local files with @)
-				sshTarget := utils.ParseSSHTarget(arg)
-				if username == "" && sshTarget.User != "" {
-					username = sshTarget.User
-				}
-				// Reconstruct the argument without the user part
-				if sshTarget.Path != "" {
-					args[i] = sshTarget.Host + ":" + sshTarget.Path
-				} else {
-					args[i] = sshTarget.Host
-				}
+			rewritten, user := stripUserPrefix(arg)
+			if username == "" {
+				username = user
 			}
+			args[i] = rewritten
 		}
 
 		sources := args[:len(args)-1]
@@ -211,6 +203,21 @@ func isLocalPaths(paths []string) bool {
 		}
 	}
 	return true
+}
+
+// stripUserPrefix rewrites user@host:path as host:path, handing back the user.
+// The : is required to tell the syntax from a local file name holding an @. A
+// trailing colon survives ("alice@myserver:" stays remote), so validatePaths
+// reports the missing remote path rather than a source/destination mismatch.
+func stripUserPrefix(arg string) (rewritten, user string) {
+	if !strings.Contains(arg, "@") || !strings.Contains(arg, ":") {
+		return arg, ""
+	}
+	sshTarget := utils.ParseSSHTarget(arg)
+	if sshTarget.Path != "" || strings.HasSuffix(arg, ":") {
+		return sshTarget.Host + ":" + sshTarget.Path, sshTarget.User
+	}
+	return sshTarget.Host, sshTarget.User
 }
 
 func validatePaths(sources []string, dest string) error {
--- a/cmd/ftp/cp_test.go
+++ b/cmd/ftp/cp_test.go
@@ -1,10 +1,8 @@
 package ftp
 
 import (
-	"strings"
 	"testing"
 
-	"github.com/alpacax/alpacon-cli/utils"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -271,28 +269,13 @@ func TestCpCommandSSHParsing(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			// Simulate the parsing logic from the cp command
 			args := make([]string, len(tt.args))
 			copy(args, tt.args)
-			username := ""
 
-			// Apply the same parsing logic as in cp.go
-			for i, arg := range args {
-				if strings.Contains(arg, "@") && strings.Contains(arg, ":") {
-					sshTarget := utils.ParseSSHTarget(arg)
-					if username == "" && sshTarget.User != "" {
-						username = sshTarget.User
-					}
-					if sshTarget.Path != "" {
-						args[i] = sshTarget.Host + ":" + sshTarget.Path
-					} else {
-						args[i] = sshTarget.Host
-					}
-				}
-			}
+			username := normalizeArgs(args, "")
 
-			assert.Equal(t, tt.expectedArgs, args, "Arguments should match expected after parsing")
-			assert.Equal(t, tt.expectedUser, username, "Username should match expected")
+			assert.Equal(t, tt.expectedArgs, args)
+			assert.Equal(t, tt.expectedUser, username)
 		})
 	}
 }
--- a/cmd/ftp/cp_test.go
+++ b/cmd/ftp/cp_test.go
@@ -368,3 +368,29 @@ func TestStripUserPrefix(t *testing.T) {
 		})
 	}
 }
+
+func TestNormalizeArgsUsername(t *testing.T) {
+	tests := []struct {
+		name     string
+		args     []string
+		flagUser string
+		wantUser string
+	}{
+		{
+			name:     "the -u flag wins over an inline user",
+			args:     []string{"test.txt", "alice@myserver:/tmp/"},
+			flagUser: "carol",
+			wantUser: "carol",
+		},
+		{
+			name:     "the first inline user wins",
+			args:     []string{"alice@myserver:/tmp/f", "bob@myserver:/tmp/g"},
+			wantUser: "alice",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.wantUser, normalizeArgs(tt.args, tt.flagUser))
+		})
+	}
+}
--- a/cmd/ftp/cp_test.go
+++ b/cmd/ftp/cp_test.go
@@ -357,3 +357,31 @@ func TestRequiredCpPatterns(t *testing.T) {
 		})
 	}
 }
+
+func TestStripUserPrefix(t *testing.T) {
+	tests := []struct {
+		name          string
+		arg           string
+		wantRewritten string
+		wantUser      string
+	}{
+		{name: "user, host and path", arg: "alice@myserver:/tmp/f", wantRewritten: "myserver:/tmp/f", wantUser: "alice"},
+		{
+			name:          "user and host with a trailing colon",
+			arg:           "alice@myserver:",
+			wantRewritten: "myserver:",
+			wantUser:      "alice",
+		},
+		{name: "no user is untouched", arg: "myserver:/tmp/f", wantRewritten: "myserver:/tmp/f"},
+		{name: "trailing colon without a user is untouched", arg: "myserver:", wantRewritten: "myserver:"},
+		{name: "local path with an @ but no colon is untouched", arg: "./mail@example.txt", wantRewritten: "./mail@example.txt"},
+		{name: "path may itself contain an @", arg: "alice@myserver:/tmp/mail@example.txt", wantRewritten: "myserver:/tmp/mail@example.txt", wantUser: "alice"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rewritten, user := stripUserPrefix(tt.arg)
+			assert.Equal(t, tt.wantRewritten, rewritten)
+			assert.Equal(t, tt.wantUser, user)
+		})
+	}
+}
--- a/cmd/ftp/cp_test.go
+++ b/cmd/ftp/cp_test.go
@@ -372,8 +372,8 @@ func TestStripUserPrefix(t *testing.T) {
 			wantRewritten: "myserver:",
 			wantUser:      "alice",
 		},
+		{name: "user and an empty host", arg: "alice@:", wantRewritten: ":", wantUser: "alice"},
 		{name: "no user is untouched", arg: "myserver:/tmp/f", wantRewritten: "myserver:/tmp/f"},
-		{name: "trailing colon without a user is untouched", arg: "myserver:", wantRewritten: "myserver:"},
 		{name: "local path with an @ but no colon is untouched", arg: "./mail@example.txt", wantRewritten: "./mail@example.txt"},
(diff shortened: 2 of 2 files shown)


geunwoonoh (commented):
## PR Review: #367 — fix(ftp): keep the trailing colon on a user-prefixed remote target

**Author**: @jisung-02
**Branch**: `349-cp-trailing-colon` → `main`
**Stack**: golang (no `alpacax` frontmatter in `CLAUDE.md` — inferred from file extensions)
**Changes**: +44 −12 in 2 files (`cmd/ftp/cp.go`, `cmd/ftp/cp_test.go`)

---

### Summary

This fixes `alpacon cp file.txt alice@myserver:` reporting "invalid source/destination combination" instead of the missing remote path. The argument reassembly moves into a `stripUserPrefix` helper, and the `else` branch that dropped the colon on an empty path is gone.

**The fix itself is correct.** What I verified:

- Once an argument passes the guard, `ParseSSHTarget` always reaches `SplitN(target, ":", 2)`, so `Host + ":" + Path` reproduces the original exactly. The "dead branch" argument in the PR body holds.
- Tracing the regression: `alice@myserver:` → `myserver:` → `isRemotePath` true → `SplitPath` returns `remotePath == ""` → the precise guidance message fires.
- One more case improved as a side effect. `alice@:` used to collapse to an empty string and get classified as a local path; it now survives as `:` and gets the "missing server name" message.
- I ran `gofmt -l .` (no output), `go vet ./...` (no output), and `go test -race ./cmd/ftp/... ./utils/...` (pass). golangci-lint is not installed in my environment, so I could not run it.

### Critical Issues (0)

None.

### Warnings (1)

**1. `cmd/ftp/cp_test.go:275-289` — a verbatim copy of the old buggy loop is still in the tree**

`TestCpCommandSSHParsing` does not call production code. It inlines the whole loop into the test body, and that copy still carries the exact branch this PR deletes:

```go
if sshTarget.Path != "" {
    args[i] = sshTarget.Host + ":" + sshTarget.Path
} else {
    args[i] = sshTarget.Host   // ← the bug this PR removed from cp.go
}
```

I confirmed the two implementations have actually diverged, with a throwaway test:

```
arg="alice@myserver:"       copy="myserver"        production="myserver:"       diverges=true
arg="alice@:"               copy=""                production=":"               diverges=true
arg="alice@myserver:/tmp/f" copy="myserver:/tmp/f" production="myserver:/tmp/f" diverges=false
```

Three problems follow:

- The test never calls `stripUserPrefix`. Delete the helper outright and this test still goes green. That leaves `TestStripUserPrefix` carrying the regression guard alone.
- The repo now fixes the bug in one file and pins the buggy behavior as correct in another. The next reader has nothing to tell them which one is the intent.
- This copy is the only thing keeping the `strings` and `utils` imports alive in `cp_test.go` (grep confirms lines 281 and 282 are the sole uses). Removing it cleans up both imports.

This PR was rigorous enough to spend a separate commit deleting a branch its own guard had made dead. Applying the same standard to the test side would be consistent.

### Suggestions (5)

**1. Lift the whole loop into a helper so the copied test can be folded in (the fix for the warning above)**

Because `stripUserPrefix` handles one argument at a time, the "multiple arguments, first inline user wins" rule still lives untested inside the `Run` closure — `TestStripUserPrefix` cannot reach it either. Extracting the normalization loop lets the eight cases in `TestCpCommandSSHParsing` point at production code with no copy in between:

```go
// normalizeArgs strips inline user prefixes in place and resolves the username:
// the -u flag wins, then the first inline user.
func normalizeArgs(args []string, username string) string {
	for i, arg := range args {
		rewritten, user := stripUserPrefix(arg)
		if username == "" {
			username = user
		}
		args[i] = rewritten
	}
	return username
}
```

**2. `cmd/ftp/cp.go:213` — the `@` check does not change any result**

When a colon is present, `Host + ":" + Path` always reproduces the original (minus the user prefix), so the colon check is the only guard that is load-bearing. I compared both versions over a 20-case corpus — Windows drive letters, arguments with several `@`, arguments with several `:`, the empty string — and found no input where they diverge.

```go
func stripUserPrefix(arg string) (rewritten, user string) {
	if !strings.Contains(arg, ":") {
		return arg, ""
	}
	sshTarget := utils.ParseSSHTarget(arg)
	return sshTarget.Host + ":" + sshTarget.Path, sshTarget.User
}
```

The trade-off: arguments without an `@` currently exit at the guard, whereas after the change they allocate an identical string through `Host + ":" + Path` (an allocation the old code also paid). Keeping the check as an intent marker is defensible, so treat this as taste. One thing that is not taste, though: the comment "The `:` is required to tell the syntax from a local file name holding an @" describes only the colon check, which puts it slightly out of step with the code as written.

**3. `cmd/ftp/cp_test.go:368-378` — single-line and multi-line table entries are mixed arbitrarily**

The entry that was expanded across lines would be 113 characters on one line, while its neighbours that stayed inline are 123 (line 377) and 154 (line 378). Only the shorter one got expanded, so there is no rule a reader can pick up. `gofmt` accepts either, so I would collapse them all to one line each.

**4. `cmd/ftp/cp_test.go` — case-name capitalization departs from the rest of the file**

Existing tests in this file start with a capital: `"Local file path"`, `"Simple local to remote copy"`. The new cases are lowercase — `"user, host and path"`. The sentence-case rule in `CLAUDE.md` targets document headings, so this is not a violation, but two conventions now coexist in one file.

**5. The `myserver:` control case from the issue report is missing from the table**

The PR body leans on "`myserver:` without a user prefix already worked" as the axis of its argument, yet that input is not in the table. The `@` guard covers it today, but suggestion 2 changes which path it takes, so a control case would start earning its keep:

```go
{name: "no user with a trailing colon is untouched", arg: "myserver:", wantRewritten: "myserver:"},
```

### Duplicate / unused code

| Location | Finding |
|---|---|
| `cmd/ftp/cp_test.go:275-289` | Copy of the production loop, including the deleted buggy branch (see the warning) |
| `cmd/ftp/cp.go:213` | The `@` check has no effect on the result (verified over a corpus) |
| `cmd/websh/websh.go:239-245`, `cmd/exec/parse.go:153-159` | Two more near-identical blocks of the pattern extracted here. They are not directly mergeable — those use `!strings.Contains(":")`, so they accept only a path-less `user@host`. Candidates to lift into `utils/ssh_parser.go` together, but out of scope for this PR |
| `cmd/ftp/cp_test.go:305` | The `description` field on `TestCpCommandSSHParsing` is declared but referenced by no assertion and no `t.Run` name (pre-existing) |
| `cmd/ftp/cp_test.go:295-296` | Messages like `"Arguments should match expected after parsing"` fall foul of the `CLAUDE.md` rule against messages that restate the assertion (pre-existing). Could be cleaned up alongside warning 1 |

No unused functions or dead imports in this diff. `stripUserPrefix` has one call site, and the `strings` / `utils` imports in `cp.go` are all live.

### Performance

No measurable impact. I would leave it alone.

- The loop runs once per process, over as many arguments as were typed — usually two or three.
- Two `strings.Contains`, two `strings.Index` inside `ParseSSHTarget`, and two `SplitN` scan a short string six times. Cutting that down with `strings.Cut` buys nanoseconds.
- The old code allocated a fresh string at the same `Host + ":" + Path` spot, so the allocation count is unchanged.

No performance regression in this PR.

### Security

Nothing to flag. No credentials, no injection surface, no log exposure. If anything it tightens things slightly: `alice@:` can no longer slip through to an API request with an empty server name, because `SplitPath` validation now catches it.

### Highlights

- Pulling the reassembly into a helper, and using the name plus the comment to record *why* the colon has to survive
- Spotting that the first cut — `if sshTarget.Path != "" || strings.HasSuffix(arg, ":")` — was a tautology, and spending a commit on deleting the branch. Unreachable fallbacks are meant to go, per Go convention
- Named return values `(rewritten, user string)` used purely for documentation, with no naked return
- Three commits stacked `fix` → `refactor` → `test` so review reads top to bottom
- Pinning both the reported input (`alice@myserver:`) and the adjacent boundaries (`alice@:`, an `@` inside the path)

---

**Verdict**: COMMENTED

Nothing here blocks the merge. I would fold warning 1 into this PR, though — otherwise a commit that fixes a bug and a test that pins that same bug as correct end up living in the tree together.

**Checklist**
- [x] Follows stack conventions (verified `gofmt` / `go vet` clean)
- [x] Tests included and passing (`go test -race ./cmd/ftp/... ./utils/...`)
- [x] No documentation change needed (`[USER@]SERVER:/path` in `Long` / `Example` stays accurate)
- [ ] Tests point at production code — warning 1

geunwoonoh (approved):
## PR Review (follow-up): #367 — fix(ftp): keep the trailing colon on a user-prefixed remote target

**Author**: @jisung-02
**Branch**: `349-cp-trailing-colon` → `main`
**Head**: `757f785` (previous review looked at `9b8010c`)
**Changes**: +83 −35 in 2 files — two new commits (`05de6cb`, `757f785`)

---

### Status of the previous review

| Item | Status |
|---|---|
| **Warning 1** — copy of the old buggy loop | **Applied** — `TestCpCommandSSHParsing` calls `normalizeArgs` directly, the copy is gone |
| **Suggestion 1** — lift the loop into a helper | **Applied** — `normalizeArgs`, with the signature as proposed |
| Dead `strings` / `utils` imports | **Applied** — removed along with the copy |
| Two assertion-restating messages | **Applied** — removed |
| The untested "first inline user wins" rule | **Applied** — `TestNormalizeArgsUsername` pins both halves, `-u` precedence and first-inline-user precedence |
| **Suggestion 2** — redundant `@` guard + comment mismatch | **Deferred, moved to #369** |
| **Suggestions 3, 4** — table wrapping, case-name capitalization | **Deferred, moved to #369** |
| **Suggestion 5** — the `myserver:` control case | **Deferred, moved to #369** (reasoning holds) |
| Unused `description` field, websh/exec commonization | **Deferred, moved to #369** |

Every deferred item is recorded in #369. That issue also picks up a fourth call site I had missed (`cmd/edit/edit.go:182`) and records the judgement that the four guards encode different notions of a valid target, so they are not mechanically mergeable. That reads as moved, not dropped.

The reasoning for not reinstating suggestion 5 is right, too. `myserver:` carries no `@`, so it exits on the first half of the guard and takes exactly the same path as `myserver:/tmp/f`, which is already in the table. Adding it now would put back a case that proves nothing.

### Verification

What I ran myself:

```
gofmt -l .          no output
go vet ./...        no output
go build            ok
go test -race ./... 39 ok, 0 FAIL
```

golangci-lint is not installed in my environment, so I could not run it. The author reports `0 issues`, and I confirmed `go vet` and the build.

**I mutation-tested whether the copied-loop problem is genuinely gone.** The point of the previous review was that deleting the helper left the file green, so the question is whether the suite now dies.

```
Mutation 1: restore the original bug (return Host alone when Path == "")
  → FAIL TestStripUserPrefix/user_and_host_with_a_trailing_colon
  → FAIL TestStripUserPrefix/user_and_an_empty_host

Mutation 2: break username precedence (drop the `if username == ""` guard)
  → FAIL TestCpCommandSSHParsing/Remote_with_user_to_local
  → FAIL TestNormalizeArgsUsername/the_-u_flag_wins_over_an_inline_user
  → FAIL TestNormalizeArgsUsername/the_first_inline_user_wins
```

Both mutants die. Mutation 2 failing inside `TestCpCommandSSHParsing` is the direct evidence that the copy is gone — the old copy carried the logic itself, so no amount of damage to production code could turn it red.

### Production behaviour

Unchanged. The loop body inside `normalizeArgs` is byte-for-byte what the previous review looked at, and moving the call site to `username = normalizeArgs(args, username)` preserves the `-u`-wins rule. It runs after the `len(args) < 2` guard, so an empty slice never reaches it.

### Critical Issues (0) · Warnings (0)

### Suggestions (1, non-blocking)

**`cmd/ftp/cp_test.go:392` — `TestNormalizeArgsUsername` passes `tt.args` without copying**

`normalizeArgs` rewrites the slice in place, so this call mutates the table entry. It is harmless today: each case is used once and nothing reads `tt.args` afterwards. But `TestCpCommandSSHParsing` right above it calls the same function through a `copy`, so two conventions now sit next to each other, and it goes quietly wrong the day someone adds an assertion on `tt.args`. Not worth a commit of its own — a line on the #369 cleanup list is about the right weight.

### Highlights

- Rather than taking the suggestion as written, folding the two changes into one commit because `unused` rejects an extraction with no caller. Letting the linter set the commit boundary is the right call
- Spotting that the `TestCpCommandSSHParsing` table cannot express `-u` precedence at all — every case starts from an empty username — and standing up a separate test for it
- Not leaving the deferrals as prose in a comment thread but moving them into #369 with their reasoning. "Whichever guard the shared helper settles on, the comment has to be rewritten against it, so fixing it in place first would be wasted work" is a sharper statement of the problem than my original note
- Finding `cmd/edit/edit.go` as a fourth call site, which my review had missed
- Updating the commit table and the regression write-up in the PR body to match the new commits

---

**Verdict**: APPROVED

The one blocking item is resolved, and I confirmed it with mutation tests rather than by reading. Everything left over is what I had classified as taste or pre-existing, and it is filed in #369 with reasoning attached. The remaining suggestion is test hygiene and has no business holding up a merge.

**Checklist**
- [x] Follows stack conventions (`gofmt` / `go vet` / build verified)
- [x] Tests included and passing (`go test -race ./...` — 39 ok, 0 FAIL)
- [x] Tests point at production code (verified by mutation testing)
- [x] No documentation change needed
- [x] Deferred items tracked (#369)

PR #374, opened 2026-08-21
fix(utils): let a stopped spinner start again, and hand its goroutine the channels it waits on

### LINKED ISSUE #373: Spinner.Start() after Stop() panics on a TTY, and go test structurally cannot catch it
## Problem

`Spinner.Stop()` closes `stopCh` and `doneCh` and never reopens them, but `Start()` does not re-initialize either. Calling `Start()` on a stopped spinner therefore launches a goroutine whose `select` falls straight through the already-closed `stopCh` case, returns, and runs its deferred `close(s.doneCh)` on a channel the previous goroutine already closed.

Reproduced by forcing `enabled` in a same-package test (`go test` is never a TTY, so this is the only way to reach the branch):

```
panic: close of closed channel

goroutine 38 [running]:
github.com/alpacax/alpacon-cli/utils.(*Spinner).Start.func1()
	utils/spinner.go:85 +0x404
created by github.com/alpacax/alpacon-cli/utils.(*Spinner).Start in goroutine 36
	utils/spinner.go:75 +0x1c4
```

`NewSpinner` (`utils/spinner.go:52-53`) is the only place the two channels are made.

## Why the blind spot is the real hazard

Off a TTY `s.enabled` is false and `Start` returns at `utils/spinner.go:68-72`, before the goroutine exists. `go test` is never a TTY, so no test in the suite can reach the panicking path without setting `s.enabled` by hand. A regression that reintroduces a restart passes the entire suite and panics on a user's terminal.

## Why it is worth hardening now rather than leaving as-is

The bug predates #366, and every call site today constructs a fresh spinner, so nothing is broken in practice. What changed in #366 is the character of the rule: `Spinner.StopWriter` makes stopping implicit—the spinner retires on the first non-empty write, with no `Stop()` call at the site—so "never restart a spinner" is now a constraint a caller can violate without writing anything that looks like a restart. In `cmd/exec/run.go` the only thing enforcing it is the comment at `cmd/exec/run.go:459-463`, which is why `restartSpinner` builds a new `Spinner` per stretch of waiting instead of reusing one.

## Proposed fix

Either is cheap; the first keeps `Spinner` reusable, the second makes the current contract explicit.

1. Re-initialize `stopCh` and `doneCh` in `Start()` when the spinner is not running, so a stopped spinner can be started again.
2. Make `Start()` a no-op once the spinner has been stopped, so a restart is inert rather than fatal.

Whichever lands, it needs a test that forces `s.enabled = true` and a short `interval`—the pattern `TestSpinner_AnimatingBranchPairsTheActiveCount` in `utils/spinner_test.go` already uses—since no test that leaves `enabled` alone can cover this.

Option 1 also has to keep the `activeSpinners` accounting paired: `Start` increments after the `!s.enabled` early return and `Stop` decrements after `<-s.doneCh`, and a reusable spinner must not drift that count.

## Reference

Raised by @geunwoonoh in review of #366: https://github.com/alpacax/alpacon-cli/pull/366#discussion_r3826886299

jisung-02:
## Background

`Spinner.Stop()` closes `stopCh` and `doneCh`, and `NewSpinner` was the only place either was made. A second `Start()` therefore launched a goroutine whose `select` fell straight through the already-closed `stopCh`, returned, and ran its deferred `close(s.doneCh)` on a channel the first goroutine had closed: `panic: close of closed channel`.

Nothing was broken in practice, because no caller restarted a spinner. `cmd/exec/run.go` built a fresh `Spinner` for every stretch of approval waiting, and a comment was the only thing enforcing that. What makes the panic worth closing anyway is that `go test` is never a TTY: `Start` returns at the `!s.enabled` branch before the goroutine exists, so a regression here passes the whole suite and panics on a user's terminal. Reaching the path at all requires a same-package test that sets `enabled` by hand, which is what `TestSpinner_AnimatingBranchPairsTheActiveCount` already does.

### A stopped spinner can start again

`Start()` now makes a fresh `stopCh`/`doneCh` pair inside the same critical section that flips `running`, and `Stop()` copies both out under that lock before closing them.

The `activeSpinners` accounting is untouched: `Start` still increments after the `!s.enabled` early return and `Stop` still decrements after the goroutine has exited, so a reusable spinner does not drift the count that `CliWarning` reads to decide whether to break off the animated line.

### The goroutine holds the channels it was handed

Making the pair per run introduced a second problem: the goroutine read both back off the struct, `doneCh` once at spawn and `stopCh` on every turn of the `select`. A restart that lands between `Stop` flipping `running` and the old goroutine noticing its close leaves that goroutine selecting on the new `stopCh`, which nobody closes. It animates forever while the `Stop` that meant to end it blocks on a `doneCh` that never closes.

`Start` now captures both channels in locals and the goroutine closes over those. No caller can hit this today—every `Start`/`Stop` pair in the tree runs on one goroutine, and the writes `StopWriter` retires a spinner from come from that same goroutine—but `Spinner` is exported and `Stop`'s own comment already promised the guard.

### exec keeps one spinner

`restartSpinner` in `cmd/exec/run.go` no longer rebuilds the `Spinner`; it stops it, starts it again, and takes a fresh `StopWriter` per tick. The writer's retirement is still one-shot, which is what the existing test observes, so `TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnSpinner` is renamed to `...EachTickWaitsBehindItsOwnWriter` to say what it actually asserts.

## Testing

`go test -race ./...` passes. Both new tests were seen red against the unfixed code:

- `TestSpinner_RestartsAfterStop` forces `enabled` and a 1ms interval, then runs `Start`/`Stop` twice. Without the per-run channels it fails with `panic: close of closed channel` at `utils/spinner.go:89`.
- `TestSpinner_RestartAndStopRaceOnTheSameSpinner` drives `Start`/`Stop` from two goroutines. With the goroutine reading the struct fields again, `-race` reports a read at `utils/spinner.go:80` in `Start.func1` against a write at `utils/spinner.go:69` in `Start`.

Closes #373

## Review follow-up

Three commits address the review's two warnings and three suggestions.

- **Heartbeat made explicit** (warning 2): the `!s.enabled` branch keeps printing the static line on every restart on purpose—during a long approval wait that repeated line is the only sign of life a redirected log gets—and now says so in a comment.
- **Restart regression is observable** (warning 1): `TestRunExecWithApprovalWait_EachRestartPrintsTheHeartbeatLine` asserts the message-line count on captured stderr, which off a TTY is exactly the number of `Start` calls. Deleting `spinner.Start()` from `restartSpinner` now fails it (expected 3, got 1).
- **The goroutine runs on locals** (suggestion 1): `Start()` copies `message`, `frames`, `dots`, and `interval` in the same critical section that hands over the channels, so the goroutine touches no struct field and the per-frame lock is gone.
- **`Stop()` returns immediately** (suggestion 2): the `default:` + `time.Sleep` loop is now a two-way select on a `time.Ticker`, so a close of `stopCh` is seen without waiting out a frame.
- **Deferred backstop** (suggestion 3): the poll-loop spinner gets `defer spinner.Stop()` for return paths added later; the explicit `Stop()` calls that order against `CliWarning` output stay.

## Second review follow-up

The re-review left one optional suggestion and two nits. All three are now closed.

- **The prompt-`Stop` property has a test** (suggestion 1): `TestSpinner_StopDoesNotWaitOutAFrame` gives the spinner a 2s frame, waits for the goroutine to reach the select, and asserts `Stop` returns inside 500ms. Reverting the select to `default:` plus `time.Sleep(interval)` fails it at 1.9s.
- **A non-positive interval no longer reaches the ticker** (nit 2): `Start` falls back to `defaultSpinnerInterval` when the field is not positive. `time.NewTicker` panics where the `time.Sleep` it replaced took a zero happily, and it panics on the spinner's own goroutine, so the process dies. Unreachable today, but a stderr animation should not be able to take the CLI down.
- **The draw-before-check ordering is on record** (nit 3): intended. A `Start` answered immediately paints one frame and erases it, and `Stop` blocks on `doneCh`, so both writes land before the caller's output and nothing is left on screen. The comment now says so.

`go test -race ./...`, `go vet ./...`, and `golangci-lint run` are all clean. Both new tests were seen red: the ticker revert fails the timing test at 1.9s against a 500ms budget, and dropping the interval fallback takes the test binary down with `panic: non-positive interval for NewTicker`.

### DIFF
--- a/cmd/exec/run.go
+++ b/cmd/exec/run.go
@@ -456,14 +456,12 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
 	// command's own output once approved. StopWriter retires the spinner at the
 	// first byte of it.
 	spinnerOut := spinner.StopWriter(out)
-	// That retirement is permanent (StopWriter fires once, and Stop closes
-	// channels Start does not reopen) and the first still-pending tick spends it,
-	// so every further stretch of waiting gets a spinner and a writer of its own.
-	// Stop first: a tick that returned without writing left the previous spinner
-	// animating, and two of them would draw over each other.
+	// A writer retires the spinner once and the first still-pending tick spends
+	// it, so every further stretch of waiting restarts the spinner and takes a
+	// fresh writer. Stop first: a tick that returned without writing left the
+	// spinner animating, and Start on a running spinner is a no-op.
 	restartSpinner := func() {
 		spinner.Stop()
-		spinner = utils.NewSpinner(approvalWaitMessage)
 		spinner.Start()
 		spinnerOut = spinner.StopWriter(out)
 	}
--- a/cmd/exec/run.go
+++ b/cmd/exec/run.go
@@ -451,6 +451,10 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
 
 	spinner := utils.NewSpinner(approvalWaitMessage)
 	spinner.Start()
+	// Every return path below stops the spinner before it writes, which is
+	// load-bearing for output ordering; this backstops paths added later. Stop
+	// on a stopped spinner is a no-op.
+	defer spinner.Stop()
 	// Every tick re-submits, so a tick's output starts while the spinner is
 	// animating—the denial line of a request still pending as much as the
 	// command's own output once approved. StopWriter retires the spinner at the
--- a/cmd/exec/run_test.go
+++ b/cmd/exec/run_test.go
@@ -321,6 +321,39 @@ func TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnWriter(t *testing.T) {
 	assert.NotSame(t, writers[2], writers[3], "and so does the third")
 }
 
+// Off a TTY every Start prints the wait message once, so the number of message
+// lines on stderr is exactly the number of Start calls. That pins the heartbeat
+// contract—a redirected log gets one line per stretch of waiting—and it is the
+// only test that notices restartSpinner forgetting to restart: drop its Start
+// and every other test stays green while the user sees no spinner after the
+// first tick.
+func TestRunExecWithApprovalWait_EachRestartPrintsTheHeartbeatLine(t *testing.T) {
+	denial := stubApprovalWaitSeams(t, 10*time.Millisecond, "SUDO_APPROVAL_REQUIRED")
+
+	calls := 0
+	runPresenceStepUp = func(_ *client.AlpaconClient, _, _, _, _ string, _ map[string]string, _ string, w io.Writer) error {
+		calls++
+		_, _ = w.Write([]byte(denialLine("SUDO_APPROVAL_REQUIRED")))
+		if calls < 4 {
+			return denial
+		}
+		return nil
+	}
+	streamApprovedCommand = func(*client.AlpaconClient, string, io.Writer, time.Duration) error {
+		t.Fatal("a denial-code wait re-attempts the command; it never resumes a held job")
+		return nil
+	}
+
+	var err error
+	stderr := captureStderr(t, func() {
+		err = RunExecWithApprovalWait(nil, "srv", "whoami", "", "", nil, "", time.Second, io.Discard)
+	})
+
+	assert.NoError(t, err)
+	// One line for the wait's opening Start, one per still-pending tick's restart.
+	assert.Equal(t, 3, strings.Count(stderr, approvalWaitMessage))
+}
+
 // Reporting the last poll failure instead would exit 1 on a request still open.
 func TestRunExecWithApprovalWait_TimeoutReportsThePendingDenial(t *testing.T) {
 	const waitTimeout = 80 * time.Millisecond
--- a/cmd/exec/run_test.go
+++ b/cmd/exec/run_test.go
@@ -294,8 +294,8 @@ func TestRunExecWithApprovalWait_TransientPollFailureKeepsWaiting(t *testing.T) 
 // that reused a single writer would therefore animate through its first poll gap
 // and sit silent for the rest of what --wait is there to cover. Writer identity
 // is all a test can see of the spinner from here, and one writer per tick means
-// one spinner per tick.
-func TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnSpinner(t *testing.T) {
+// one restart per tick.
+func TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnWriter(t *testing.T) {
 	denial := stubApprovalWaitSeams(t, 10*time.Millisecond, "SUDO_APPROVAL_REQUIRED")
 
 	var writers []io.Writer
@@ -317,7 +317,7 @@ func TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnSpinner(t *testing.T) 
 	assert.NoError(t, err)
 	// writers[0] is the first attempt, which runs before any spinner exists.
 	require.Len(t, writers, 4)
-	assert.NotSame(t, writers[1], writers[2], "the second wait needs a spinner of its own")
+	assert.NotSame(t, writers[1], writers[2], "the second wait needs a writer of its own")
 	assert.NotSame(t, writers[2], writers[3], "and so does the third")
 }
 
--- a/utils/spinner.go
+++ b/utils/spinner.go
@@ -67,51 +67,57 @@ func (s *Spinner) Start() {
 	stopCh := make(chan struct{})
 	doneCh := make(chan struct{})
 	s.stopCh, s.doneCh = stopCh, doneCh
+	// The rest of what the goroutine needs is copied in the same critical
+	// section, so it never reaches back into the struct during its run.
+	msg := s.message
+	frames, dots := s.frames, s.dots
+	interval := s.interval
 	s.mu.Unlock()
 
 	if !s.enabled {
-		// Print a static message so users still see progress in non-TTY output
-		fmt.Fprintln(os.Stderr, s.message)
+		// Print a static message so users still see progress in non-TTY output.
+		// Deliberately printed again on every restart: a long approval wait
+		// restarts the spinner each poll tick, and the repeated line is the
+		// heartbeat a redirected log gets from an otherwise silent wait.
+		fmt.Fprintln(os.Stderr, msg)
 		return
 	}
 	activeSpinners.Add(1)
 
 	go func() {
 		defer close(doneCh)
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
 		frameIdx := 0
 		dotIdx := 0
 		dotCounter := 0
 		for {
+			frame := frames[frameIdx%len(frames)]
+
+			// Animate dots if message ends with "..."
+			displayMsg := msg
+			if baseMsg, ok := strings.CutSuffix(msg, "..."); ok {
+				displayMsg = baseMsg + dots[dotIdx%len(dots)]
+			}
+
+			fmt.Fprintf(os.Stderr, "\r%s %s", Yellow(frame), displayMsg)
+			frameIdx++
+
+			// Update dots every 3 frames (300ms)
+			dotCounter++
+			if dotCounter >= 3 {
+				dotIdx++
+				dotCounter = 0
+			}
+
+			// Blocking here instead of sleeping lets Stop return as soon as it
+			// closes stopCh rather than up to a frame later.
 			select {
 			case <-stopCh:
 				// Clear the spinner line
 				fmt.Fprint(os.Stderr, "\r\033[K")
 				return
-			default:
-				s.mu.Lock()
-				msg := s.message
-				s.mu.Unlock()
-
-				frame := s.frames[frameIdx%len(s.frames)]
-
-				// Animate dots if message ends with "..."
-				displayMsg := msg
-				if strings.HasSuffix(msg, "...") {
-					baseMsg := strings.TrimSuffix(msg, "...")
-					displayMsg = baseMsg + s.dots[dotIdx%len(s.dots)]
-				}
-
-				fmt.Fprintf(os.Stderr, "\r%s %s", Yellow(frame), displayMsg)
-				frameIdx++
-
-				// Update dots every 3 frames (300ms)
-				dotCounter++
-				if dotCounter >= 3 {
-					dotIdx++
-					dotCounter = 0
-				}
-
-				time.Sleep(s.interval)
+			case <-ticker.C:
 			}
 		}
 	}()
--- a/utils/spinner.go
+++ b/utils/spinner.go
@@ -91,6 +91,9 @@ func (s *Spinner) Start() {
 		frameIdx := 0
 		dotIdx := 0
 		dotCounter := 0
+		// Drawing before the first stopCh check is deliberate: a Start answered
+		// at once paints one frame and erases it, and Stop waits on doneCh, so
+		// both land before the caller's own output.
 		for {
 			frame := frames[frameIdx%len(frames)]
 
--- a/utils/spinner.go
+++ b/utils/spinner.go
@@ -61,9 +61,12 @@ func (s *Spinner) Start() {
 		return
 	}
 	s.running = true
-	// Stop closes both channels, so a restart needs a fresh pair.
-	s.stopCh = make(chan struct{})
-	s.doneCh = make(chan struct{})
+	// Stop closes both channels, so a restart needs a fresh pair—and the
+	// goroutine below takes its own copy, since the fields belong to whichever
+	// Start ran last rather than to the run it was spawned for.
+	stopCh := make(chan struct{})
+	doneCh := make(chan struct{})
+	s.stopCh, s.doneCh = stopCh, doneCh
 	s.mu.Unlock()
 
 	if !s.enabled {
@@ -74,13 +77,13 @@ func (s *Spinner) Start() {
 	activeSpinners.Add(1)
 
 	go func() {
-		defer close(s.doneCh)
+		defer close(doneCh)
 		frameIdx := 0
 		dotIdx := 0
 		dotCounter := 0
 		for {
 			select {
-			case <-s.stopCh:
+			case <-stopCh:
 				// Clear the spinner line
 				fmt.Fprint(os.Stderr, "\r\033[K")
 				return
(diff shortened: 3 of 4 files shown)


geunwoonoh (dismissed):
## Summary

`Spinner.Stop()` closes `stopCh` and `doneCh`, but both were only ever made in `NewSpinner`, so a second `Start()` launched a goroutine that fell straight through the already-closed `stopCh` and died in `close(s.doneCh)` with `panic: close of closed channel`. This PR makes a fresh channel pair per `Start()` so a spinner can restart, and closes the follow-on problem that introduced (the goroutine re-reading `s.stopCh` on every turn of the select) by handing the goroutine its own locals. `cmd/exec` stops rebuilding a `Spinner` per tick and restarts one instead.

The implementation is correct. Lock ordering, `activeSpinners` accounting, and channel lifetimes all hold up. Everything below is either pre-existing or a suggestion.

## Critical issues (0)

None.

## Warnings (2)

### 1. `cmd/exec/run.go:463`—deleting `spinner.Start()` from `restartSpinner` leaves the whole suite green

Verified directly: drop that one line, run `go test -race ./cmd/exec/... ./utils/...`, and everything still passes. That is a regression where the user sees no spinner at all after the first tick, and nothing catches it.

The gap predates this PR—the writer-identity test never observed `Start` in the first place—and the rename is the right call, since it stops the test from claiming more than it checks. What is missing is something to replace the guarantee that was dropped. There is a cheap replacement, but it is entangled with warning 2 below.

### 2. `utils/spinner.go:72-76`—each restart adds another copy of the same line off a TTY

The `!s.enabled` branch of `Start()` prints `fmt.Fprintln(os.Stderr, s.message)`, so now that a restart happens every poll tick, a redirected log collects one repeat per tick:

```
$ go test -run TestRunExecWithApprovalWait_EachTickWaitsBehindItsOwnWriter ./cmd/exec/ -v
Waiting for approval in the Alpacon console...
Waiting for approval in the Alpacon console...
Waiting for approval in the Alpacon console...
```

A 30-minute `--wait-approval` piped into a CI log turns that into dozens of lines. The line count is unchanged from before this PR, so it is not a regression, but it is a genuine fork in the road. Read as log noise, the fix is a `staticPrinted bool` on `Spinner` so a reused instance prints once; read as a liveness heartbeat for a long silent wait, the current behavior is already right. Every other caller builds a fresh spinner per operation, so either choice leaves them untouched.

Either way, please make the intent explicit in the code. If the heartbeat reading wins, asserting the number of message lines on captured stderr also closes warning 1, since that count is exactly the number of `Start` calls. If the noise reading wins, that observation point disappears and warning 1 stays an accepted limitation covered only by the `utils` tests.

## Suggestions (3)

### 1. `utils/spinner.go:91-93`—the per-frame lock guards nothing

`s.message` has no writer after `NewSpinner`. Yet the goroutine takes `s.mu` every 100ms to read it, while `frames`, `dots`, and `interval` are read unlocked right beside it. Lifting `msg` in the same critical section that now hands over `stopCh`/`doneCh` matches the rule this PR just established:

```go
	stopCh := make(chan struct{})
	doneCh := make(chan struct{})
	s.stopCh, s.doneCh = stopCh, doneCh
	msg := s.message
	s.mu.Unlock()
```

The goroutine then touches no struct field at all, and the per-frame lock/unlock goes away.

### 2. `utils/spinner.go:84-116`—`default:` plus `time.Sleep` makes `Stop()` block for up to a frame

A sleeping goroutine cannot see the closed `stopCh`, so `Stop()`'s `<-doneCh` waits up to `interval` (100ms in production). Pre-existing, but this PR moves `Stop` from once per wait to once per poll tick, which multiplies the cost. Against a 5s poll interval that is roughly 1% and nothing a user would notice. A `time.Ticker` with a two-way select would let `Stop` return immediately and drop the busy `default` branch. Fine to leave out of scope here.

### 3. `cmd/exec/run.go:452`—`defer spinner.Stop()` is now available

Before, `spinner` was reassigned, so `defer spinner.Stop()` would have pinned its receiver to the first spinner. With one instance for the whole function it is simply correct. All five return paths call `Stop()` explicitly today, so this is a net for paths added later rather than a fix. The `Stop()` calls that precede `CliWarning` are load-bearing for output ordering and should stay.

## Go conventions, dead code, performance

- **Conventions**: declaration order (`var` → `type` → `func`), naming, testify usage, the `require`/`assert` split, and the assertion-message rule all match `CLAUDE.md`. `for range 2` is fine on `go 1.25.13`. `go vet ./...` is clean.
- **Dead or duplicated code**: nothing new is dead. The `stopCh`/`doneCh` fields are still needed, since `Stop()` has to reach the current run's channels. The only surplus is the per-frame lock in suggestion 1, which predates this PR. The five `spinner.Stop()` calls in `RunExecWithApprovalWait` each order against different output, so I did not count them as duplication.
- **Performance**: net positive. `term.IsTerminal` drops from one syscall per tick to one per wait, and `Spinner` allocation does the same. Against that, every `Start()` now allocates two channels, but the old code allocated the same number via a per-tick `NewSpinner`, so it is a wash (off a TTY they are allocated and never used, which is negligible). Suggestion 2 is the only item left.
- **Escape check**: once `spinnerOut` is replaced by a restart, a write through the stale writer would now stop a live spinner, where before it hit a dead object and did nothing. Traced down to `streamSubscribed`: `out` is only written from the select loop and `finish`, and never escapes into a goroutine that outlives the call, so this is unreachable today.

## Highlights

- **The commit split is exact.** The first commit opens up restarting; the second closes the field swap that opening created. The PR body narrates that causality faithfully.
- **Both new tests actually go red.** Reverted `utils/spinner.go` to `0c4dcde` and ran them: `-race` reports the read at `spinner.go:80` against the write at `spinner.go:69`, and the run dies with `panic: close of closed channel`. Reproduced across `-count=3`.
- **`Stop()` has its lock ordering right.** It copies the channels under the lock, then closes and waits outside it. Holding the lock across `<-doneCh` would have deadlocked against a goroutine that takes the same lock every frame.
- **`activeSpinners` stays balanced across a restart.** `Add(1)` precedes the goroutine spawn and `Add(-1)` follows the `doneCh` receive, so each pair lands in order. `running` also forces `Start` and `Stop` critical sections to strictly alternate, so `Stop` can never grab another run's channels.
- **The test comments explain why the tests exist**, in particular that `go test` is never a TTY and an ordinary test cannot reach this path at all.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Follows stack conventions (`go vet` clean, matches the `CLAUDE.md` test rules)
- [x] Tests included and passing (`go test -race ./...` green; the new tests confirmed red against the unfixed code)
- [x] No security concerns (no credential or input handling; stderr output only)
- [ ] Documentation—not applicable (internal behavior change, no user-facing docs)

geunwoonoh (dismissed):
## Summary

Re-review at `19cf97d`. Three commits landed against the two warnings and three suggestions from my previous pass:

| Commit | Addresses |
|---|---|
| `01a10c8` refactor(utils): drive the spinner goroutine from locals and a ticker | Suggestions 1 and 2 |
| `8365156` refactor(exec): defer the approval-wait spinner's Stop | Suggestion 3 |
| `19cf97d` test(exec): count heartbeat lines to pin spinner restarts | Warnings 1 and 2 |

All of it applied correctly. Where a call was mine to raise but yours to make, the reasoning is now in the code rather than just satisfied in form.

## What I verified

**Warning 1 (deleting `spinner.Start()` left the suite green)—closed.** `TestRunExecWithApprovalWait_EachRestartPrintsTheHeartbeatLine` counts message lines on stderr. Reran the same mutation:

```
--- FAIL: TestRunExecWithApprovalWait_EachRestartPrintsTheHeartbeatLine (0.03s)
        Error:   Not equal:
                 expected: 3
                 actual  : 1
```

The expected 3 checks out too: one line for the wait's opening `Start`, one per still-pending tick's restart, and no restart on the tick that answers.

**Warning 2 (repeated line off a TTY)—settled on the heartbeat reading and said so.** `utils/spinner.go:78-81` records the intent, and the new test pins the contract. Picking one of the two readings, writing down which, and tying it to a test is exactly the resolution I was asking for.

**Suggestion 1 (per-frame lock)—closed.** `msg`, `frames`, `dots`, and `interval` are copied in the same critical section as the channel pair. The goroutine body no longer has a single `s.` access.

**Suggestion 2 (`time.Sleep` → ticker)—applied.** `time.NewTicker` plus a two-way select, so `Stop()` returns as soon as it closes `stopCh`, and the busy `default` branch is gone.

**Suggestion 3 (`defer spinner.Stop()`)—applied.** The five explicit `Stop()` calls stay, the backstop is added, and the comment explains that the explicit ones are load-bearing for output ordering. `Stop` guards on `running`, so the double call is harmless.

## Critical issues (0) / Warnings (0)

None.

## Suggestions (1) and nits (2)

### 1. (optional) The ticker's prompt-`Stop` property is the one change no test guards

I reverted the select to `default:` plus `time.Sleep(interval)` and both `./utils/...` and `./cmd/exec/...` stayed green. Raising it only for symmetry, since the other two suggestions did get pinned. A large `interval` (say 2s) with an assertion that `Start` followed by `Stop` returns well inside it would catch a revert. It is a timing test, so it is not a must, and deferring it past this PR is fine.

### 2. (nit) `time.NewTicker(interval)` panics on `interval <= 0`

`NewSpinner` supplies 100ms, `interval` is unexported, and the only other writers are three same-package test lines using 1ms, so this is unreachable today. Just noting that `time.Sleep(0)` used to be harmless where the ticker is not. No real risk.

### 3. (nit, intent check) The frame is now drawn before the first `stopCh` check

In the `Start`-then-immediate-`Stop` race the old loop returned without drawing; the new one draws a frame and then clears it. `Stop()` blocks on `<-doneCh`, so both the draw and the clear are guaranteed to finish before the caller's output, and the clear follows immediately, so nothing is visible on screen. Not a problem—just confirm the ordering is what you intended.

## Highlights

- **One improvement arrived unasked**: `strings.CutSuffix` collapses the `HasSuffix` plus `TrimSuffix` double scan into one.
- **The `(300ms)` comment is now true.** A ticker fires on exactly `interval`; `time.Sleep` added the draw time on top.
- **The ticker's backpressure is safe.** Its buffer of one drops ticks when the consumer lags, so frames never arrive in a burst.
- **The goroutine no longer reads the struct at all.** "A spinner can restart" is now expressed by handing values over rather than by a rule about which fields may be touched, which is much harder to break later.
- **The comments say only what the code cannot.** The heartbeat intent, why the `defer` is a backstop, and why the select blocks—each records a why and none restates a what.

## Verification

```
go test -race ./...        all pass
go vet ./...               clean
mutation A: drop Start() from restartSpinner  → new test fails 3 vs 1 (was green before)
mutation B: revert ticker to sleep            → still green (basis for suggestion 1)
```

---

**Verdict**: APPROVED

What is left is one optional suggestion and two nits. Nothing here should hold up the merge.

**Checklist**:
- [x] Follows stack conventions (`go vet` clean; declaration order and testify rules match)
- [x] Tests included and passing (two mutations run to confirm the new test actually bites)
- [x] No security concerns (stderr output only; no change in surface)
- [ ] Documentation—not applicable (internal behavior change)

geunwoonoh (approved):
## Summary

Re-review at `1e2c822`. One commit each for the optional suggestion and two nits left over from the last pass:

| Commit | Addresses |
|---|---|
| `1e2c822` test(utils): pin that Stop does not wait out a frame | Suggestion 1 |
| `e2a83c3` fix(utils): fall back when the spinner's frame interval is not positive | Nit 2 |
| `793d5ef` docs(utils): say why the spinner draws before its first stop check | Nit 3 |

**Nothing is outstanding.** Every item raised across three rounds is closed.

## What I verified

**Suggestion 1 (no test guarded the prompt `Stop`)—closed.** `TestSpinner_StopDoesNotWaitOutAFrame` sets a 2s interval and checks that `Stop()` returns inside 500ms. Reverted the ticker to `default:` plus `time.Sleep` to confirm it bites:

```
--- FAIL: TestSpinner_StopDoesNotWaitOutAFrame (2.00s)
        Error: "1.899643917s" is not less than "500ms"
```

The design is the part worth calling out. A `Stop` issued right after `Start` returns immediately however the loop is written, because the goroutine has not been scheduled yet—so the test sleeps 100ms first to get a frame genuinely underway before it starts the clock. That also makes the failure mode one-sided: on a machine too slow for 100ms to be enough, the test goes vacuous rather than flaky-failing.

**Nit 2 (`time.NewTicker` panics on a non-positive period)—closed.** `defaultSpinnerInterval` gives `NewSpinner` and the fallback one source, and `Start` clamps `interval <= 0` back to it. Dropped the fallback to check, and the result matches the test comment exactly:

```
panic: non-positive interval for NewTicker
```

The comment's claim that the test "does not fail, it takes the test binary down" holds, including that a panic on the spinner's own goroutine is not something a caller could recover from. Applying the fallback to the local and leaving `s.interval` alone is also the right call—`Start` does not quietly rewrite configuration.

**Nit 3 (drawing before the first `stopCh` check)—documented.** `utils/spinner.go:100-102` records both the intent and why it is safe, and it keeps the part that matters: `Stop` waits on `doneCh`, so the draw and the erase both land before the caller's own output.

## Critical issues (0) / Warnings (0) / Suggestions (0)

None.

## Notes (no action needed)

- **An empty `frames` or `dots` would divide by zero at `frames[frameIdx%len(frames)]`.** Same shape as the interval panic, different provenance. The interval one was *introduced* by this PR when `Sleep` became `Ticker`, which is what justified guarding it; the `frames` one predates the PR and has no writer outside `NewSpinner`. Extending the guard there would be over-defensive—stopping where you did is the right line.
- **If the timing test ever flakes in CI**, raise the 500ms budget rather than shortening the interval: the gap between the 2s frame and the budget *is* the test's signal-to-noise ratio. `-count=3` passed at 0.10s each, so there is plenty of margin today.

## Highlights

- **The constant removes a magic-number duplication** that would otherwise have sat in two places, and the declaration order (`const` → `var` → `type` → `func`) matches `CLAUDE.md`.
- **The test comments make checkable claims.** I reproduced "this does not fail, it takes the test binary down" verbatim. A comment written so it can be falsified is a rare thing.
- **`assert.Less` over a wrapped predicate.** Because it is not `assert.True(t, elapsed < ...)`, the failure prints the real measurement, `1.899643917s`. Exactly the assertion rule in `CLAUDE.md`.
- **Across three rounds nothing was closed in form only.** Each item came back with the reasoning written into the code or a comment, and pinned by a test.

## Verification

```
go test -race ./...        all pass
go vet ./...               clean
new tests -count=3         3/3 pass (0.10s / 0.00s)
mutation A: revert ticker to sleep      → fails, 1.899s vs 500ms
mutation B: drop the interval fallback  → panic: non-positive interval for NewTicker
```

---

**Verdict**: APPROVED

**Checklist**:
- [x] Follows stack conventions (`go vet` clean; declaration order and testify assertion rules match)
- [x] Tests included and passing (two mutations run to confirm the new tests actually bite)
- [x] No security concerns (stderr output only; no change in surface)
- [ ] Documentation—not applicable (internal behavior change)

PR #384, opened 2026-08-24
refactor(api): drop the approve/reject clients ADR 0015 stranded

### LINKED ISSUE #381: refactor(api): remove dead approve/reject functions left behind by ADR 0015
## Summary

ADR 0015 made the CLI an execution/request surface only: `alpacon approval approve/reject` and `alpacon work-session approve/reject` now print a guidance error and exit without calling the API (`cmd/approval/approval_approve.go`, `cmd/worksession/worksession_approve.go`). The API-layer functions they used to call were left behind and are now dead code—nothing outside their own tests reaches them.

## Dead code

`golang.org/x/tools/cmd/deadcode` (entry point `main.go`) reports all four as unreachable, and `grep -rn` confirms the only callers are their tests:

- `api/approval/approval.go:77`—`ApproveRequest`
- `api/approval/approval.go:82`—`RejectRequest`
- `api/worksession/worksession.go:112`—`ApproveWorkSession`
- `api/worksession/worksession.go:117`—`RejectWorkSession`

Tests that exist only to cover the dead functions:

- `api/approval/approval_test.go`—`TestApproveRequest_*` / `TestRejectRequest*` (5 functions)
- `api/worksession/worksession_test.go`—`TestApproveWorkSession_*` / `TestRejectWorkSession` (3 functions)

## Proposed change

Delete the four API functions and their eight test functions—roughly 110–170 lines. No user-visible behavior changes: the command layer already blocks these paths, so the deletion only removes code that can never run.

## Notes

- The server endpoints themselves are untouched; this only removes the CLI-side client functions that no longer have a caller.
- Does not conflict with open PRs #379/#380 (different files).

jisung-02:
## Background

ADR 0015 made the CLI an execution and request surface only. Approval and rejection happen out of band in the Alpacon console, and the server refuses both from the CLI credential channel, so `6d0c754` stopped the command layer from calling the API: `alpacon approval approve|reject` and `alpacon work-session approve|reject` print guidance and exit 1 without a request.

That left four API-layer functions with no caller outside their own tests—`ApproveRequest`, `RejectRequest`, `ApproveWorkSession`, `RejectWorkSession`. `deadcode` reports all four as unreachable from `main`.

`#221` kept them on the grounds that the 403-propagation tests document the server behavior. Reading the tests, they do not. `TestApproveRequest_403PropagatesError` stands up an `httptest` server that the test itself tells to return 403, so what it pins is that a `detail` body reaches the caller as an error—work that lives in `client.SendPostRequest`, not in the deleted function. `TestCancelRequest_403PropagatesError` asserts the same thing through `CancelRequest`, which is still reachable, and the reason the server refuses CLI approvals is already written down at `cmd/approval/approval_approve.go:8-13`.

### Dead clients and their tests are gone

The four functions, their payload types (`ApproveOptions`, `WorkSessionApproveRequest`—used by nothing else), and the eight tests that only covered them. The remaining six of those tests checked the shape of a request nobody builds any more: the POST method, an `/approve/` or `/reject/` URL suffix, and the two adjustment fields in the body.

The shared URL constants stay: `approvalURL` and `workSessionURL` are read by `ls`, `describe`, `cancel`, `extend`, and the rest.

### The two command groups now answer the same way

Separate from the deletion, `cmd/worksession/worksession.go:66` listed `alpacon work-session approve` and `alpacon work-session reject` among the subcommands to use, though both exit 1. `cmd/approval/approval.go:30` already leaves them out and points at the console instead, so that closing sentence now appears in both groups and they read alike.

Copilot flagged the sentence it had carried since `6d0c754`: "Approve and reject" put bare verbs in the subject slot. Both groups now say "Approval and rejection happen in the Alpacon console (web) or Slack."

The subcommands stay registered. `6d0c754` kept them so a script that calls one gets an actionable exit instead of "unknown command", and only the guidance text changes.

This drift dates to `6d0c754` rather than to the deletion, but it is the same ADR 0015 leftover and reads better reviewed together.

### Review follow-ups

Three commits were added after review.

`2d51ec6` finishes the noun-form fix. `915017e` had changed only the two group-level fallbacks, leaving `approveRejectExcludedMessage` in both packages still reading "approve and reject happen out of band". That constant is the more visible of the two: the fallback prints only for a bare group command, while the constant prints on every `approve` or `reject` invocation.

`409e5cd` names Slack alongside the web console. The new closing sentence said "in the Alpacon console (web)" while the `Long` text above it says "in the Alpacon console (web) or Slack", so the two read as different guidance.

`0ee95e3` adds `TestWorkSessionGuidanceMatchesRegisteredSubcommands` and `TestApprovalGuidanceMatchesRegisteredSubcommands`. Each walks the group's registered subcommands and asserts that `approve` and `reject` are absent from the guidance while every other command is named, so the drift this PR fixes cannot return unnoticed. Both directions were seen failing first: putting `approve` back into the list, and dropping `revoke` from it.

Two review points are deliberately not in this PR. The hand-maintained enumeration is the root cause of the drift, but it spans 23 group commands under `cmd/` and is tracked in #385. The duplicated `approveRejectExcludedMessage` constant is left in place under the rule of three—two copies do not yet justify a shared home, and a third group needing the same message is the trigger to extract it.

## Testing

No user-visible behavior changes from the deletion. `cmd/` carries guidance text changes and nothing else.

- `go build -v .`, `go vet ./...`: exit 0.
- `go test -race ./...`: no `FAIL`; `api/approval`, `api/worksession`, `cmd/approval`, and `cmd/worksession` all `ok`.
- `golangci-lint run ./...`: `0 issues.` The `unused` linter is enabled, so an orphaned symbol would have been caught.
- `deadcode` reports nothing on this branch. On the pre-deletion commit the same command reports exactly the four functions, so the empty output is the tool working rather than the tool failing.
- The built binary still exits 1 with guidance for `approval approve`, `approval reject`, `work-session approve`, and `work-session reject`, and each now prints "approval and rejection happen out of band". `alpacon work-session` and `alpacon approval` with no subcommand close with the console-or-Slack sentence.

Closes #381

### DIFF
--- a/cmd/approval/approval.go
+++ b/cmd/approval/approval.go
@@ -27,7 +27,7 @@ Subcommands for tracking:
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon approval ls', 'alpacon approval describe', or 'alpacon approval cancel'. Approval and rejection happen in the Alpacon console (web). Run 'alpacon approval --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon approval ls', 'alpacon approval describe', or 'alpacon approval cancel'. Approval and rejection happen in the Alpacon console (web) or Slack. Run 'alpacon approval --help' for more information")
 	},
 }
 
--- a/cmd/approval/approval.go
+++ b/cmd/approval/approval.go
@@ -27,7 +27,7 @@ Subcommands for tracking:
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon approval ls', 'alpacon approval describe', or 'alpacon approval cancel'. Approve and reject happen in the Alpacon console (web). Run 'alpacon approval --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon approval ls', 'alpacon approval describe', or 'alpacon approval cancel'. Approval and rejection happen in the Alpacon console (web). Run 'alpacon approval --help' for more information")
 	},
 }
 
--- a/cmd/worksession/worksession.go
+++ b/cmd/worksession/worksession.go
@@ -63,7 +63,7 @@ Run 'alpacon whoami' to check your WorkSession requirement and active session.`,
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session approve', 'alpacon work-session reject', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Run 'alpacon work-session --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Approve and reject happen in the Alpacon console (web). Run 'alpacon work-session --help' for more information")
 	},
 }
 
--- a/cmd/worksession/worksession.go
+++ b/cmd/worksession/worksession.go
@@ -63,7 +63,7 @@ Run 'alpacon whoami' to check your WorkSession requirement and active session.`,
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Approve and reject happen in the Alpacon console (web). Run 'alpacon work-session --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Approval and rejection happen in the Alpacon console (web). Run 'alpacon work-session --help' for more information")
 	},
 }
 
--- a/cmd/worksession/worksession.go
+++ b/cmd/worksession/worksession.go
@@ -63,7 +63,7 @@ Run 'alpacon whoami' to check your WorkSession requirement and active session.`,
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Approval and rejection happen in the Alpacon console (web). Run 'alpacon work-session --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Approval and rejection happen in the Alpacon console (web) or Slack. Run 'alpacon work-session --help' for more information")
 	},
 }
 
--- a/api/approval/approval.go
+++ b/api/approval/approval.go
@@ -74,16 +74,6 @@ func GetApprovalRequestRaw(ac *client.AlpaconClient, id string) ([]byte, error) 
 	return ac.SendGetRequest(utils.BuildURL(approvalURL, id, nil))
 }
 
-func ApproveRequest(ac *client.AlpaconClient, id string, opts ApproveOptions) error {
-	_, err := ac.SendPostRequest(utils.BuildURL(approvalURL, path.Join(id, "approve"), nil), opts)
-	return err
-}
-
-func RejectRequest(ac *client.AlpaconClient, id string) error {
-	_, err := ac.SendPostRequest(utils.BuildURL(approvalURL, path.Join(id, "reject"), nil), struct{}{})
-	return err
-}
-
 func CancelRequest(ac *client.AlpaconClient, id string) error {
 	_, err := ac.SendPostRequest(utils.BuildURL(approvalURL, path.Join(id, "cancel"), nil), struct{}{})
 	return err
--- a/api/approval/approval_test.go
+++ b/api/approval/approval_test.go
@@ -90,55 +90,6 @@ func TestGetApprovalRequest(t *testing.T) {
 	assert.Equal(t, "sudo", req.RequestType)
 }
 
-func TestApproveRequest_NoAdjustments(t *testing.T) {
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		assert.Equal(t, http.MethodPost, r.Method)
-		assert.True(t, strings.HasSuffix(r.URL.Path, "apr-abc/approve/"))
-		var body ApproveOptions
-		assert.NoError(t, json.NewDecoder(r.Body).Decode(&body))
-		assert.Nil(t, body.AdjustedScopes)
-		assert.Nil(t, body.AdjustedServers)
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
-	}))
-	defer ts.Close()
-
-	err := ApproveRequest(newTestClient(ts), "apr-abc", ApproveOptions{})
-	assert.NoError(t, err)
-}
-
-func TestApproveRequest_WithAdjustments(t *testing.T) {
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		assert.True(t, strings.HasSuffix(r.URL.Path, "apr-abc/approve/"))
-		var body ApproveOptions
-		assert.NoError(t, json.NewDecoder(r.Body).Decode(&body))
-		assert.Equal(t, []string{"command"}, body.AdjustedScopes)
-		assert.Equal(t, []string{"srv-uuid-1"}, body.AdjustedServers)
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
-	}))
-	defer ts.Close()
-
-	err := ApproveRequest(newTestClient(ts), "apr-abc", ApproveOptions{
-		AdjustedScopes:  []string{"command"},
-		AdjustedServers: []string{"srv-uuid-1"},
-	})
-	assert.NoError(t, err)
-}
-
-func TestRejectRequest(t *testing.T) {
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		assert.Equal(t, http.MethodPost, r.Method)
-		assert.True(t, strings.HasSuffix(r.URL.Path, "apr-abc/reject/"))
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
-	}))
-	defer ts.Close()
-
-	err := RejectRequest(newTestClient(ts), "apr-abc")
-	assert.NoError(t, err)
-}
-
 func TestCancelRequest(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		assert.Equal(t, http.MethodPost, r.Method)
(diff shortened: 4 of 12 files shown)


cmd/worksession/worksession.go
```diff
@@ -63,7 +63,7 @@ Run 'alpacon whoami' to check your WorkSession requirement and active session.`,
...
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session approve', 'alpacon work-session reject', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Run 'alpacon work-session --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon work-session ls', 'alpacon work-session create', 'alpacon work-session describe', 'alpacon work-session use', 'alpacon work-session current', 'alpacon work-session activate', 'alpacon work-session complete', 'alpacon work-session extend', 'alpacon work-session update', 'alpacon work-session revoke', 'alpacon work-session cancel', 'alpacon work-session timeline', or 'alpacon work-session recording'. Approval and rejection happen in the Alpacon console (web). Run 'alpacon work-session --help' for more information")
```
geunwoonoh: **Warning:** the noun-form fix from `915017e` is only half applied.

This line now reads "Approval and rejection happen…", but `approveRejectExcludedMessage` at `cmd/worksession/worksession_approve.go:15` still says "approve and reject happen out of band"—the same bare-imperative-as-subject construction Copilot flagged. That constant is the more visible of the two: this fallback prints only when `alpacon work-session` is run with no subcommand, while the constant prints on every `approve` or `reject` invocation. Confirmed against the built binary:

```
$ alpacon work-session approve ses-x
Error: Approvals must be done in the Alpacon console (web), not the CLI. The CLI is an
execution and request surface only; approve and reject happen out of band. Use
'alpacon work-session ls' to track a session's status.
```

Suggested fix in `cmd/worksession/worksession_approve.go` (and the same in `cmd/approval/approval_approve.go`):

```go
const approveRejectExcludedMessage = "Approvals must be done in the Alpacon console (web), not the CLI. " +
	"The CLI is an execution and request surface only; approval and rejection happen out of band. " +
	"Use 'alpacon work-session ls' to track a session's status."
```

---

**Suggestion:** nothing pins this string, so the drift can recur. A table test walking `WorkSessionCmd.Commands()` and asserting that every command except `approve` and `reject` appears in the error string would catch it. `cmd/worksession` already has several test files to host it.

**Suggestion:** the enumeration itself is what drifted. `cmd.Help()` on the line above already prints the full list, and Cobra prints usage again on the returned error, so with no arguments `alpacon work-session` emits 92 lines: the subcommand list three times (help, usage, this enumeration) and the long error sentence twice. Repo-wide convention across 20+ group commands, so better as a separate issue than a change here.

cmd/approval/approval.go
```diff
@@ -27,7 +27,7 @@ Subcommands for tracking:
 		if err := cmd.Help(); err != nil {
 			return err
 		}
-		return errors.New("a subcommand is required. Use 'alpacon approval ls', 'alpacon approval describe', or 'alpacon approval cancel'. Approve and reject happen in the Alpacon console (web). Run 'alpacon approval --help' for more information")
+		return errors.New("a subcommand is required. Use 'alpacon approval ls', 'alpacon approval describe', or 'alpacon approval cancel'. Approval and rejection happen in the Alpacon console (web). Run 'alpacon approval --help' for more information")
```
geunwoonoh: **Warning:** same gap as in `cmd/worksession/worksession.go`. `approveRejectExcludedMessage` at `cmd/approval/approval_approve.go:15` still reads "approve and reject happen out of band", so `alpacon approval approve` and `alpacon approval reject` keep printing the construction this PR set out to remove. Verified against the built binary.

**Suggestion:** the `Long` text above says approving and rejecting happen "in the Alpacon console (web) or Slack", while this sentence mentions only the web. Minor, but the two read as different guidance.

geunwoonoh (dismissed):
## PR review: #384 - refactor(api): drop the approve/reject clients ADR 0015 stranded

**Stack**: golang (no `alpacax` frontmatter in CLAUDE.md; inferred from `.go` files)
**Changes**: +2 -162 across 8 files (3 commits ahead of `origin/main`, 0 behind)

---

### Summary

ADR 0015 made the CLI an execution and request surface only, and `6d0c754` cut the command layer's approve/reject calls. This PR removes what that stranded: four API clients (`ApproveRequest`, `RejectRequest`, `ApproveWorkSession`, `RejectWorkSession`), two request types, and eight tests that only covered them. It also fixes the `work-session` group guidance, which still listed `approve` and `reject` among the usable subcommands, and applies Copilot's noun-form wording to both group commands.

The change is almost entirely deletion. Runtime behavior is unchanged; one line of user-facing guidance text differs.

### Critical issues (0)

None.

### Warnings (1)

**1. `cmd/worksession/worksession_approve.go:15`, `cmd/approval/approval_approve.go:15` — the noun-form fix is only half applied**

Commit `915017e` is titled "use noun forms in the console approval guidance", but it changed only the two group-level fallback strings. The `approveRejectExcludedMessage` constant—the message users actually hit—still carries the construction Copilot flagged.

From the built binary:

```
$ alpacon work-session approve ses-x
Error: Approvals must be done in the Alpacon console (web), not the CLI. The CLI is an
execution and request surface only; approve and reject happen out of band. Use
'alpacon work-session ls' to track a session's status.
```

`approve and reject happen` puts bare imperatives in the subject slot, exactly what the commit set out to remove. This string is also more visible than the one that was fixed: the group fallback prints only when someone runs `alpacon work-session` with no subcommand, while this constant prints on every `approve` or `reject` invocation. `cmd/approval/approval_approve.go:15` has the identical sentence.

Two lines, so it belongs in this PR:

```go
const approveRejectExcludedMessage = "Approvals must be done in the Alpacon console (web), not the CLI. " +
	"The CLI is an execution and request surface only; approval and rejection happen out of band. " +
	"Use 'alpacon work-session ls' to track a session's status."
```

### Suggestions (4)

**1. `cmd/worksession/worksession.go:66` — the hand-maintained subcommand list is what drifted**

The drift this PR fixes exists because the list is maintained by hand, so it can recur. The line right above already calls `cmd.Help()`, which prints the full list, and Cobra prints usage again when `RunE` returns an error. Measured with no arguments:

```
$ alpacon work-session          # 92 lines total
stdout: one "Available Commands" block (57 lines)
stderr: Error + usage with a second "Available Commands" block, then
        "Error: While executing the command: ..." repeating the same long sentence (35 lines)
```

The subcommand list reaches the user three times (help, usage, the enumeration inside the error string) and the long error sentence twice. This pattern is shared by 20+ group commands under `cmd/`, so it is repo-wide convention rather than something this PR introduced. Better as a separate issue.

**2. No regression test pins the guidance**

Nothing under `cmd/` asserts the guidance string or the subcommand list, so the exact drift this PR fixes can come back silently. A table test walking `WorkSessionCmd.Commands()` and asserting that every command except `approve` and `reject` appears in the error string would catch it. `cmd/worksession` already has several test files to host it.

**3. `approveRejectExcludedMessage` is duplicated**

Defined twice, in `cmd/approval` and `cmd/worksession`, with near-identical doc comments and text differing only in the trailing `ls` command. The four stub `Long` blocks are near-duplicates too. Pre-existing, and not worth extracting until a third group needs the same treatment. Noting it only for the record.

**4. The new sentence drops the Slack path**

Surrounding `Long` text says "in the web console or Slack", while the new `Approval and rejection happen in the Alpacon console (web).` mentions only the web. Minor, but the two read as different guidance.

### Verification

Only what was actually run locally:

| Check | Result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./api/... ./cmd/...` | exit 0 |
| `go test ./api/approval/... ./api/worksession/... ./cmd/approval/... ./cmd/worksession/...` | all four packages `ok` |
| `golangci-lint`, `deadcode` | not installed locally, so those claims were not reproduced. CI is the gate |
| Binary behavior | `work-session approve`, `approval reject`, and bare `work-session` all exit 1 with the guidance |

**Dead-code sweep.** Every remaining symbol was checked; the deletion left no orphans.

- All 5 exported functions in `api/approval` and all 12 in `api/worksession` have at least one non-test caller (`GetWorkSession` has 8)
- All 14 remaining types across both `types.go` files are referenced outside their own file
- Imports are correct: `path` is still used by `CancelRequest`, `RevokeWorkSession`, and others, and `strings` / `encoding/json` are still used by surviving assertions in both test files. This is the usual miss in a deletion PR and it is clean here

**Coverage loss.** The evidence is stronger than the PR body claims. `TestApproveRequest_403PropagatesError` and `TestRejectRequest_404PropagatesError` were exercising `client.SendPostRequest`'s `detail`-body propagation, and `client/client_test.go:70-106` already covers 404 while `:226-279` covers 403. That coverage lives in `client`, not only in `TestCancelRequest_403PropagatesError`, so nothing was lost.

**Performance.** Pure deletion; no runtime path is touched. Two POST wrappers and two types leave the binary marginally smaller. Nothing gained or lost beyond that.

### Highlights

- The deletion is complete: functions, their exclusive types, and their exclusive tests go together, while the shared URL constants (`approvalURL`, `workSessionURL`) stay
- Import hygiene is correct
- The PR body re-reads `#221`'s counter-argument instead of working around it, and is right about what those 403 tests actually pin
- Clean commit split: API deletion, drift fix, and wording fix are independent and separately revertable
- Keeping the subcommands registered is the right call—an existing script gets an intentional exit 1 with guidance instead of "unknown command"

---

**Verdict**: APPROVED

The one warning is not a merge blocker, but it is the sentence that commit was written to fix, so folding in those two lines closes it cleanly.

**Checklist**
- [x] Follows stack conventions (declaration order, errcheck, lowercase error strings with no trailing punctuation)
- [x] Tests for deleted code removed with it; no coverage lost
- [x] No security concerns (no credentials, injection, or log exposure in scope)
- [x] No docs to update (no stale references in README or docs; no CHANGELOG in this repo)
- [ ] `approveRejectExcludedMessage` wording aligned (warning 1)

cmd/worksession/worksession.go
jisung-02: Fixed, and both suggestions are handled.

**Warning (noun form).** `2d51ec6` changes `approveRejectExcludedMessage` in both `cmd/worksession/worksession_approve.go` and `cmd/approval/approval_approve.go`. You are right that this is the more visible of the two strings, and fixing only one package would have restored the drift `915017e` just closed. From the rebuilt binary:

```
$ alpacon work-session approve ses-x
Error: Approvals must be done in the Alpacon console (web), not the CLI. The CLI is an
execution and request surface only; approval and rejection happen out of band. Use
'alpacon work-session ls' to track a session's status.
exit=1
```

**Suggestion (nothing pins the string).** `0ee95e3` adds `TestWorkSessionGuidanceMatchesRegisteredSubcommands` in `cmd/worksession/worksession_guidance_test.go` and the matching `TestApprovalGuidanceMatchesRegisteredSubcommands` in `cmd/approval`. Each walks `Commands()` and asserts `approve` and `reject` are absent from the guidance while every other subcommand is named. Both directions were seen failing before the tests were kept:

```
# approve put back into the list
Error: "...'alpacon work-session update', 'alpacon work-session approve', ..." should not contain "'alpacon work-session approve'"

# revoke dropped from the list
Error: "...'alpacon work-session update', 'alpacon work-session cancel', ..." does not contain "'alpacon work-session revoke'"
```

**Suggestion (the enumeration itself).** Agreed, and filed as #385 rather than changed here. Measured on this branch, a bare `alpacon work-session` writes 57 lines to stdout and 92 to stderr, with an `Available Commands` block three times and the long sentence twice. 23 files under `cmd/` carry the pattern, so it is one sweep or none. The two tests above hold the groups ADR 0015 touched until that lands.

cmd/approval/approval.go
jisung-02: Both handled.

**Warning (same gap here).** `2d51ec6` changes `approveRejectExcludedMessage` in `cmd/approval/approval_approve.go` together with its twin in `cmd/worksession`. Verified on the rebuilt binary:

```
$ alpacon approval reject req-x
Error: Approvals must be done in the Alpacon console (web), not the CLI. The CLI is an
execution and request surface only; approval and rejection happen out of band. Use
'alpacon approval ls' to track a request's status.
exit=1
```

**Suggestion (the Slack path).** `409e5cd` makes the closing sentence match the `Long` text above it in both groups: "Approval and rejection happen in the Alpacon console (web) or Slack." Slack is a real approval path, so the shorter sentence read as different guidance rather than a summary of the same thing.

geunwoonoh (approved):
## Re-review: #384 - refactor(api): drop the approve/reject clients ADR 0015 stranded

**Head**: `0ee95e3` (3 commits since the previous review at `915017e`)
**Changes**: +82 -164 across 12 files, 0 behind `origin/main`, MERGEABLE

My earlier APPROVE was dismissed when the new commits landed, so this re-approves the current head.

---

### What landed

| Previous finding | Status | Commit |
|---|---|---|
| Warning 1—`approveRejectExcludedMessage` noun-form fix only half applied | Resolved | `2d51ec6` |
| Suggestion 2—nothing pins the guidance string | Resolved | `0ee95e3` |
| Suggestion 4—new sentence dropped the Slack path | Resolved | `409e5cd` |
| Suggestion 1—hand-maintained subcommand enumeration | Split out as #385 | Correctly out of scope |
| Suggestion 3—duplicated constant | Unchanged | Agreed; not worth extracting until a third group needs it |

Deferring the two out-of-scope items to #385 instead of touching the code here is the right call, and #385 is actionable as written: it carries the measurement, the root cause, and the blast radius (23 files under `cmd/`).

### Review of the new code

The only new code is 78 lines of tests: `cmd/worksession/worksession_guidance_test.go` and `cmd/approval/approval_guidance_test.go`.

**The tests do what they claim.** Verified independently by mutating the guidance string and watching them fail in both directions:

```
# Mutation A: put 'alpacon work-session approve' back in the list (the original drift)
FAIL  worksession_guidance_test.go:34
      ... should not contain "'alpacon work-session approve'"

# Mutation C: drop 'alpacon work-session extend' from the list
FAIL  ... does not contain "'alpacon work-session extend'"

# after reverting: ok  github.com/alpacax/alpacon-cli/cmd/worksession
```

Both directions are covered: `NotContains` catches a console-only subcommand being offered as runnable again, and `Contains` catches a newly registered subcommand that never made it into the guidance. Failure output prints both operands, so a future breakage is readable without opening the test.

**Conventions.** Matches the repo's test rules:

- Dedicated `Contains` / `NotContains` helpers, no predicate wrapped in `True`
- `require.Error` only where the next line depends on it (`err.Error()`), `assert` for the independent per-subcommand checks
- No message restating the assertion
- Declaration order `var → func`, every import used
- `SetOut`/`SetErr` capture `cmd.Help()` into a buffer and `t.Cleanup` restores it, so `-v` runs stay clean instead of dumping 57 lines of help

**Isolation.** The tests mutate the package-level `WorkSessionCmd` / `ApprovalCmd`, but neither package uses `t.Parallel()` and `t.Cleanup` restores the writers, so nothing leaks into the other tests in those packages. Confirmed under `-race`.

### Verification

| Check | Result |
|---|---|
| CI on `0ee95e3` | Build and Test, golangci-lint, CodeQL (go + actions), govulncheck all SUCCESS |
| `go build ./...` | exit 0 |
| `go vet ./cmd/... ./api/...` | exit 0 |
| `go test -race` on `api/{approval,worksession}` and `cmd/{approval,worksession}` | all four packages `ok` |
| Mutation test | failed in both directions, reverted, working tree clean |
| Binary | `work-session approve` and `approval reject` both print the noun-form wording and still exit 1 |
| Residual wording | grep finds zero remaining `approve and reject happen` constructions |

### Remaining items (0 blocking)

For completeness: the new tests pin the subcommand list but not the presence of the `Approval and rejection happen in the Alpacon console (web) or Slack.` sentence itself. The harmful direction—offering a console-only command as runnable—is held by `NotContains`, so there is no real exposure here. Not worth a change.

---

**Verdict**: APPROVED

All three actionable findings are in, and the evidence is measured rather than asserted. Running the same bidirectional mutation check independently and posting the output is the right way to close a "nothing pins this" finding. The two out-of-scope items are filed. Good to merge.

**Checklist**
- [x] Follows stack conventions (testify helpers, require/assert split, declaration order, errcheck)
- [x] Tests included and passing; regression coverage verified by mutation, CI green on head
- [x] No security concerns
- [x] No docs to update; out-of-scope follow-up captured in #385
- [x] Previous review findings resolved or explicitly deferred

PR #390, opened 2026-08-25
fix(exec): hint on a sudo denial that carries no code

### LINKED ISSUE #353: exec: a sudo denial with no error code produces no hint
## Summary

A sudo denial that carries no error code reaches the terminal with nothing actionable beside it. The plugin prints `Alpacon denied this sudo command.`—no code, no parenthesis—and neither the hint table nor the unknown-code fallback matches that line, so the user gets the bare denial and no next step.

## Where

`sudoDenialHint` (`cmd/exec/run.go:235-247`) has two paths, and both anchor on `sudoDenialLinePrefix`:

```go
const sudoDenialLinePrefix = "Alpacon denied this sudo command ("
```

- the table scan, via `denialCodePresent(output, h.code)` → `sudoDenialLinePrefix + code + ")."`
- the fallback, via `firstDenialCode(output)` → `strings.Cut(rest, sudoDenialLinePrefix)`

The prefix ends in `(`, so a line without the code slot matches neither and `sudoDenialHint` returns `""`.

## When the code is absent

Two branches on the agent side, both reached on a real denial:

- `alpacon_approval.c:449-451` — `g_plugin_printf(SUDO_CONV_ERROR_MSG, "Alpacon denied this sudo command.\n")` when `error_code_buf` is empty
- `pam_alpamon.c:643-645` — `pam_error(pamh, "Alpacon denied this sudo command.")` when `code[0] == '\0'`

`error_code_buf` ends up empty in two cases: the server sent no `error_code`, or the sanitizer rejected the one it sent. The sanitizer (`alpacon_approval.c:380-404`) rejects the code *whole* rather than truncating it—on any character outside `[A-Z0-9_]`, and on overflow past `sizeof(error_code_buf) - 1` on a `char[64]`, with the comment "a truncated code could be spoofed by prefixing a legitimate code with padding". So a server that ships a malformed or over-63-character code lands on this silent path, not on the unknown-code fallback that #344 added for exactly that drift.

## Not a regression

Pre-existing: the old table scan missed this line too. `TestSudoDenialHint/prefix without the code slot` (`cmd/exec/sudo_hint_test.go:293`) pins the current behavior as deliberate—the fallback must not fire on a line with no code to name.

Raised on #344 as a non-blocking suggestion and left out of scope. It is the sibling of the case that PR's fallback exists to close, and after it, the only remaining path where a real denial reaches the terminal with nothing actionable.

## Suggested direction

A generic branch on the codeless literal, after the table scan and the `firstDenialCode` fallback both miss:

```go
if strings.Contains(output, "Alpacon denied this sudo command.") {
	return denialHintLine("sudo was denied, and the server named no reason. Read the denial in the Alpacon console (web).\n")
}
```

With no category from the server there is genuinely less to say, so the guidance is thin by necessity—the point is that the user is pointed somewhere instead of nowhere.

Worth settling before writing it: this matches on a bare literal rather than on the plugin's exact line shape, so a command's own output containing that sentence would produce the hint. The existing detectors all anchor on the code slot precisely to keep command output from forging a denial signal; the codeless line has no slot to anchor on. `sudoDenialHint` is only reachable from a `RemoteCommandError`, which bounds the exposure but does not remove it.

## Acceptance

- [ ] A denial line with no code slot produces a hint pointing at the Alpacon console
- [ ] The existing table scan and unknown-code fallback still win when a code is present
- [ ] `TestSudoDenialHint/prefix without the code slot` updated to the new expectation, and `the code slot without the prefix` still yields no hint
- [ ] Forged-input case covered: decide and pin what a command printing the codeless sentence in its own output should do

jisung-02:
## Background

A sudo denial can reach the terminal with no error code in it. The agent prints `Alpacon denied this sudo command.`—no parentheses, no code—and until this PR the CLI had nothing to say about it.

`sudoDenialHint` had two paths and both anchored on `sudoDenialLinePrefix` (`cmd/exec/run.go:26`), which ends at the opening parenthesis of the code slot: the table scan through `denialCodePresent`, and the unknown-code fallback through `firstDenialCode`. A line with no slot matched neither, so the function returned `""` and the user got the bare denial and no next step.

Two agent branches reach that line on a real denial: `alpacon_approval.c` when `error_code_buf` came out empty, and `pam_alpamon.c` when `code[0] == '\0'`. `error_code_buf` ends up empty either because the server sent no `error_code`, or because the agent's sanitizer rejected the one it sent. That sanitizer rejects a bad code whole rather than truncating it—on any character outside `[A-Z0-9_]`, and on overflow past a `char[64]`—so a server shipping a malformed or over-63-character code landed on this silent path instead of the unknown-code fallback #344 added for exactly that drift.

This was the last path where a real denial arrived with nothing actionable beside it. It is not a regression; the table scan missed this line before #344 as well.

### A third branch, running last

`sudoDenialHint` gains a branch after the table scan and the `firstDenialCode` fallback have both missed, so a denial that does carry a code still gets the guidance written for it. With no category from the server there is genuinely less to say, so the hint only points at the Alpacon console—but that is a next step where there was none.

The guidance says the denial carried no code rather than that the server named no reason, which is how the issue sketched it. The code may well have been sent and then rejected by the agent, and telling the user the server was silent points them at the wrong place to look.

### The literal is spelled out, not derived

`sudoDenialCodelessLine` is its own constant rather than `sudoDenialLinePrefix` with an empty slot. Because the sanitizer drops a bad code whole, the parentheses go with it, so `Alpacon denied this sudo command ().` is not a shape the agent ever emits. The pre-existing test row `empty code slot` still expects no hint for that reason, and now carries a comment saying so instead of reading as a coverage gap.

## Safety

The match is a plain `strings.Contains`, so a failed command that printed the same sentence in its own output gets the hint too. That is the decision the issue asked to settle, and it is deliberate.

The code-carrying detectors anchor on the full line for a reason: a forged match there can trigger the MFA step-up (`hasSudoPresenceDenial`), wedge `--wait` into a re-run loop (`hasSudoApprovalDenial`), or echo attacker-chosen text back to the terminal. This branch reaches none of those. It returns a fixed advisory line, interpolates nothing, and does not touch the exit code—`HandleCommandResult` writes the hint to stderr and exits on the code it already computed. What a forged match buys is one extra advisory line on a command that has already failed, since `sudoDenialHint` is reachable only from a `RemoteCommandError`.

Anchoring harder buys nothing against that and risks a false negative on the silent denial this branch exists to close. The commit carries a `Directive` trailer saying as much.

## Testing

`TestSudoDenialHintCodelessDenial` pins four cases: the codeless line producing the hint, the forged-input decision above, and a table code and an unknown code each still winning on output that also carries the codeless line.

The pre-existing row `prefix without the code slot` pinned the old empty return as deliberate, so it had to go; it is replaced by the new test rather than dropped. The row `the code slot without the prefix` is untouched and still yields no hint (`cmd/exec/sudo_hint_test.go:293`).

Every case was seen red against a mutated target.

Removing the branch entirely—the original bug:

```
--- FAIL: TestSudoDenialHintCodelessDenial/the_codeless_line_points_at_the_console
    Error:      	"" does not contain "no code"
    Error:      	"" does not contain "Alpacon console"
--- FAIL: TestSudoDenialHintCodelessDenial/a_failed_command_that_printed_the_sentence_itself
    Error:      	"" does not contain "Alpacon console"
```

Anchoring the match on a whole line instead of a substring, so a command echoing the sentence no longer matches. Only the forged-input case fails, which is what makes that case the one pinning the decision:

```
--- FAIL: TestSudoDenialHintCodelessDenial/a_failed_command_that_printed_the_sentence_itself
    Error:      	"" does not contain "Alpacon console"
```

Hoisting the branch above the two code-carrying paths:

```
--- FAIL: TestSudoDenialHintCodelessDenial/a_code_the_table_carries_still_wins_on_the_same_output
    Error:      	"Hint: sudo was denied, and the denial carried no code to explain it.\nRead the denial in the Alpacon console (web).\n" does not contain "runtime risk assessment"
--- FAIL: TestSudoDenialHintCodelessDenial/a_code_the_table_does_not_carry_still_wins_on_the_same_output
    Error:      	"Hint: sudo was denied, and the denial carried no code to explain it.\nRead the denial in the Alpacon console (web).\n" does not contain "SUDO_BRAND_NEW_CODE"
```

Suite and linters on the final diff:

- `go test -race ./...`: exit 0, 39 `ok`, no `FAIL`.
- `go vet ./...`: no output.

README is unchanged on purpose. Its "When a command is denied" section documents the WorkSession gate and the inline-credential refusal, not the sudo hint table, and a codeless denial creates no approval request—so it stays exit code `1`, which the existing table row already covers.

Closes #353

### DIFF
--- a/cmd/exec/run.go
+++ b/cmd/exec/run.go
@@ -25,6 +25,13 @@ import (
 // command whose own output prints "(SUDO_RISK_DENIED)" from forging a hint.
 const sudoDenialLinePrefix = "Alpacon denied this sudo command ("
 
+// sudoDenialCodelessLine is the denial line the agent emits with no code to
+// name: alpacon_approval.c with an empty error_code_buf, pam_alpamon.c with
+// code[0] == '\0'. Its own literal rather than sudoDenialLinePrefix with an
+// empty slot—the sanitizer rejects a bad code whole, so the parentheses go with
+// it and no empty-slot form ever reaches the terminal.
+const sudoDenialCodelessLine = "Alpacon denied this sudo command."
+
 // sudoPresenceRequiredCode is the one denial code the CLI resolves in-flow (an
 // MFA step-up). The hint table and hasSudoPresenceDenial both name it from
 // here, so renaming it cannot leave one of them behind and silently stop the
@@ -242,6 +249,9 @@ func isSanitizedDenialCode(code string) bool {
 // on its own release train and nothing enforces this table's sync with
 // alpacon-server utils/error_codes.py, so returning "" for one would leave the
 // next drift invisible until someone reported a bare denial line.
+//
+// A denial carrying no code gets a last, thinner hint: without a category from
+// the server there is only the console to point at.
 func sudoDenialHint(output string) string {
 	for _, h := range sudoDenialHints {
 		if denialCodePresent(output, h.code) {
@@ -252,6 +262,15 @@ func sudoDenialHint(output string) string {
 		return denialHintLine(fmt.Sprintf(
 			"sudo was denied (%s). This build carries no guidance for that code—the server may be newer than the CLI.\n"+
 				"Read the denial in the Alpacon console (web), and update the CLI so a later run explains it.\n", code))
+	}
+	// No code slot to anchor on, so a command printing the same sentence in its own
+	// output gets the hint too. Accepted: this branch only prints a fixed line—a
+	// forged match reaches no step-up, no --wait loop, and echoes nothing the
+	// command chose. Anchoring harder re-opens the silent denial this closes.
+	if strings.Contains(output, sudoDenialCodelessLine) {
+		return denialHintLine(
+			"sudo was denied, and the denial carried no code to explain it.\n" +
+				"Read the denial in the Alpacon console (web).\n")
 	}
 	return ""
 }
--- a/cmd/exec/sudo_hint_test.go
+++ b/cmd/exec/sudo_hint_test.go
@@ -290,7 +290,6 @@ func TestSudoDenialHintFallsBackToTheRawCode(t *testing.T) {
 			output:       denialLine("not a code") + denialLine("SUDO_BRAND_NEW_CODE"),
 			wantContains: []string{"SUDO_BRAND_NEW_CODE"},
 		},
-		{"prefix without the code slot", "Alpacon denied this sudo command.\n", nil},
 		{"the code slot without the prefix", "build finished (SUDO_BRAND_NEW_CODE).\n", nil},
 		{"empty output", "", nil},
 		{
@@ -307,6 +306,8 @@ func TestSudoDenialHintFallsBackToTheRawCode(t *testing.T) {
 		{"escape sequence in the code slot", denialLine("\x1b[31mSUDO_FAKE"), nil},
 		{"lowercase code", denialLine("sudo_lowercase"), nil},
 		{"space in the code", denialLine("SUDO FAKE"), nil},
+		// nil, not a coverage gap: the sanitizer drops a bad code whole, parentheses
+		// included, so an empty slot is not a shape the agent ever emits.
 		{"empty code slot", denialLine(""), nil},
 	}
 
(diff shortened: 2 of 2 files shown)


junho226 (approved):
Approving. The branch is correct, it runs last so a coded denial still wins, and what it prints is a fixed literal—there is nothing a command's own output can steer.

## What I verified

**The agent really emits this line, verbatim.** Both branches the description names are in `alpamon-pam` at its current HEAD: `alpacon_approval.c:450` (`g_plugin_printf(SUDO_CONV_ERROR_MSG, "Alpacon denied this sudo command.\n")` when `error_code_buf[0] == '\0'`) and `pam_alpamon.c:644` (`pam_error(pamh, "Alpacon denied this sudo command.")` when `code[0] == '\0'`). `sudoDenialCodelessLine` matches both, and no other message in that repo carries this sentence as a substring, so the `strings.Contains` cannot pick up a different agent line by accident.

**The empty-slot shape genuinely does not exist.** Both sanitizers reject a bad code whole rather than truncating it: `alpacon_approval.c:380-405` sets `reject = 1` on any character outside `[A-Z0-9_]` and on overflow past `sizeof(error_code_buf) - 1`, and `pam_alpamon.c:204-223` does the same two rejections ("reject the whole code on any disallowed character", "reject overly-long codes rather than truncate"). The parentheses always go with the code, so deriving the literal from `sudoDenialLinePrefix` with an empty slot would have anchored on a shape the agent never emits. Spelling the constant out is the right call, and the new comment on the `empty code slot` row puts that reasoning where the next reader will look for it.

**Precedence holds by construction, not just by branch order.** `"Alpacon denied this sudo command."` is not a substring of `"Alpacon denied this sudo command (CODE)."`—the period sits exactly where the space and the paren do—so one line can never satisfy both literals, and the position after the table scan and `firstDenialCode` is what settles mixed output. Both directions are pinned by the two precedence cases. The over-long-code case still returns `""` too: `denialLine(strings.Repeat("A", 64))` matches neither literal, so the new branch does not quietly start answering for a code the old code deliberately dropped.

**Blast radius of a forged match.** `sudoDenialHint` has one caller, `HandleCommandResult` (`cmd/exec/run.go:688`), reached only inside `errors.As(err, &remoteErr)`, so a command that succeeded cannot reach it at all. The hint goes to stderr and the process exits on the code `remoteCommandOutcome` already computed. The branch returns a constant and interpolates nothing. The MFA step-up (`hasSudoPresenceDenial`) and the `--wait` re-run loop (`hasSudoApprovalDenial` via `pendingSudoDenial`) both route through `denialCodePresent`, which still demands the code slot, so this match reaches neither. One extra advisory line on a command that has already failed is the whole exposure. Agreed with the decision, and the `Directive` trailer is the right place to record it.

**No new edge.** A forged coded line still outranks a real codeless denial on the same output, but that was already true before this PR—the forged coded hint won over nothing. This branch neither opens that nor widens it.

Local run on `2e728c7`: `go build ./...` and `go vet ./...` silent, `gofmt -l cmd/exec/` empty, `go test ./...` all `ok`, `go test -race ./cmd/exec/...` `ok`. CI is green on every check. The README claim checks out as well—"When a command is denied" documents the WorkSession gate and the inline-credential refusal, and a codeless denial stays on exit code `1`, which the existing table row already covers.

## Non-blocking, and not for this PR

The troubleshooting page in `alpacax/docs` still opens with "When Alpacon blocks a `sudo` command, it returns a message with an error code" and lists only the coded table (`content/en/reference/troubleshooting/pam-issues.mdx:74`). The codeless form has no entry there. Nothing in this PR contradicts that page, since the new hint points at the console rather than the docs—but a user who searches that sentence lands on a page telling them the code is always present. Worth a line in that repo whenever someone is next in it.

PR #391, opened 2026-08-25
fix(client): renew a stale access token mid-flight instead of failing the request

jisung-02:
## Background

Item 2 of #312 asked `pollForApproval` to retry a transient error instead of ending the wait, and `cmd/worksession/worksession_create.go:447` now does that through `utils.IsTransientRequestError`. One failure still slips past it, and it is the one that gets more likely the longer the wait runs.

`NewAlpaconAPIClient` refreshes an expired access token once, when it builds the client. A command that then holds that client through a long wait never re-enters the refresh, so the token expires under it. From that point every poll gets a 401, and `IsFatalClientError` classifies a 401 as fatal, so `IsTransientRequestError` returns false and the loop takes the terminal branch. A half-hour `--wait` ends with `polling failed: Authentication credentials were not provided.` while the approval is still open.

It is deterministic rather than transient: the longer a human takes to approve, the more certain the abort. Running the same poll as separate invocations always worked, because each process refreshed at construction.

## Changes

Nine commits, each its own reviewable idea.

### Renew mid-flight instead of reporting the 401 (2180b1a)

`sendRequest` now renews the access token and replays the request once. alpacon-server rejects a stale token in its permission layer (`utils/api/permissions.py`), before the view runs, so the first attempt changed nothing and the replay is the same request rather than a second one.

The renewal is bounded on every side: once per request, bearer requests only, and only for a body `net/http` can rewind. Two requests sharing one expiry spend one refresh-token grant between them.

Two smaller corrections ride along, both needed for the above. `RefreshToken` rejects an empty refresh token before the Auth0 round trip rather than sending one that cannot work, and a download's 401/403 now carries its status the way every other API error does—without it, `utils.HTTPStatusCode` read 0 there and the renewal could not see the rejection.

The fix is in the client rather than in each poll loop on purpose. `exec --wait`, `work-session --wait` and the MFA polls all sit on the same client, and a refresh branch inside `pollForApproval` would fix the two reported loops while leaving every other long-lived client with the same expiry.

### Keep the grant off the token read lock (85eea7a)

Putting the token behind `tokenMu` created a second problem. `setHTTPHeader` reads it through `accessToken()`, and both refresh paths held that same lock across `auth0.RefreshAccessToken`—two HTTP round trips with no deadline on them, since nothing in `client/`, `api/auth0/` or `config/` sets an `http.Client.Timeout` or builds the request with a context. Every request on the client would block behind one slow Auth0 answer.

That is reachable today: `api/event/sudolistener.go` spawns a goroutine per MFA frame (`:83`) against the client the terminal WebSocket already shares (`:47`), and both of its branches call `ac.RefreshToken()` (`:117`, `:144`).

The locks are now split by job. `refreshMu` serializes the grant, which is what keeps concurrent requests on one expiry down to a single round trip; `tokenMu` guards only the field, in two critical sections that call nothing.

### Renew only on a 401 alpacon-server itself wrote (1767fbc)

The stale-credential test was an empty error-code slot, and a body that is not JSON produces the same empty slot—`parseAPIErrorPayload` gives up on an HTML page and returns no code. A proxy, a WAF or an mTLS gate answering 401 ahead of the server looked exactly like a stale credential, so the client would spend an Auth0 round trip and a `config.json` rewrite on a refusal no token changes.

The discriminator is now whether the body is a JSON object, which every alpacon-server error response is. It cannot be the presence of a `detail` or a `code` instead: `authStatusMessage` already carries a branch for a 401 with neither, so the server is expected to send one, and gating on either would strand a genuinely stale token there.

### Leave a trace when the renewal fails (3d69e07)

A failed renewal wrote nothing—the caller surfaced the server's own 401, which says to run `alpacon login` but not that the refresh token was rejected. The reason now goes to stderr under `ALPACON_DEBUG` through `utils.CliDebug`.

### State what the uncoded-401 test actually proves (2b7c4c8)

Comments only, from review. `isStaleCredential` claimed every deliberate 401 carries a code, and alpacon-server does not promise that: its Auth0 authenticator returns no user on every bearer rejection, so a workspace-claim mismatch, the authenticator-level MFA gate and an uninvited user all collapse into the same uncoded `NotAuthenticated` an expired token produces. Each costs one grant and one replay and then surfaces the same error, so the behavior is right and only the stated invariant was too strong. The comment now says what the code relies on—an uncoded JSON 401 is the only 401 a new token could plausibly move.

Two neighbours get the same treatment. `apiPayload` is a filter rather than a provenance flag: only "not JSON, not the server" holds, and the field name invites the converse reading. `renewAccessToken`'s `sent` check collapses nothing when the grant fails, since the token is unchanged and every other in-flight request runs its own grant—the bound is the in-flight count, now stated rather than incidental.

### Say how the authenticator reaches the uncoded 401 (6560b3f)

Comments only, from review. The previous commit said the authenticator's catch-all absorbs all three rejections it names, and only one of them gets there that way—`Auth0JWTInvitationAuthentication.get_user` raises inside the base `authenticate`'s try. The workspace-claim mismatch and the authenticator-level MFA gate are plain `return None` branches after that try, so they decline outright rather than having anything swallowed.

Nothing downstream tells the two apart, and the code slot is empty either way, which is the only property `isStaleCredential` reads. The comment now states the property that covers both: the authenticator returns no user on every bearer rejection, whether it absorbs an exception or declines outright.

### Replace the config file atomically (2bfd908)

`saveConfig` opened `config.json` with `O_TRUNC` and streamed the encoder into it. A second alpacon process saving at the same moment truncates the same path, and a reader landing in that window gets a config that does not parse—which `LoadConfig` reports as a broken file and the command answers with "run `alpacon login`", throwing away a session that was fine.

The window is not hypothetical on this branch. Every command that meets an expired access token already writes this file, since `auth0.RefreshAccessToken` saves the new token before returning, and `sendRequest` now renews mid-flight too. Two alpacon commands in two terminals, or two CI steps in parallel, is all it takes.

The encoder now writes a sibling temp file, chmod'd to `0600` before anything goes into it, which is then renamed over the target. Rename is atomic, so a concurrent reader sees one whole config or the other.

### Make the concurrency test able to fail (2d4f2ed)

From review. The two writers in `TestSaveConfig_ConcurrentWritesNeverLeaveAnUnreadableFile` discarded every error `SaveRefreshedAuth0Token` returned. A `saveConfig` that failed on every call therefore left the reader re-reading the file `CreateConfig` wrote during setup, which parses every time—so the test passed without a single replacement ever happening, and could not fail for the reason it exists.

Each writer now records its first error in its own slot and stops, and the reader asserts the joined result before it checks its own read count.

### Scope the atomic-write guarantee to concurrent readers (94fc654)

From review. The comment on `saveConfig` said a reader sees one whole config or the other without saying which reader. Nothing calls `Sync` before the rename, so the guarantee holds against another process and not against a crash—a power loss can still bring the config back empty or reverted, which is the same unparseable-config outcome the atomic write exists to prevent, reached the other way.

The comment now says so. No `Sync` call was added: the payload is a renewed access token, so losing it costs one refresh and nothing else, and the overstated comment was what needed correcting.

## Safety

The renewal never widens what a credential can do. It runs only for a bearer request, so a legacy API key or a service token—neither of which has a refresh token behind it—is never sent to Auth0. The gateway check above narrows it further: a 401 written by something in front of the server no longer reaches the grant at all.

The replay is the same request, not a second one. `replayableClone` refuses a body `net/http` cannot rewind, so a streamed upload that is not a file is reported rather than sent truncated.

A 401 that names its refusal with an error code—MFA required, IP not allowed, token ACL—is left alone, so none of them is retried behind the user's back or relabeled as an authentication failure. The converse does not hold: the server's Auth0 authenticator returns no user for every bearer rejection alike, so an uncoded 401 is not proof of expiry, only the absence of a stated refusal. The cost of that is bounded and visible—one grant and one replay, after which the same error surfaces unchanged. The exit-code contract is untouched: `3`, `4` and `6` arrive on a 403 or in a 200 body, which `isStaleCredential` never looks at.

The debug line goes through `cliMessage`, so a server-supplied string in the error chain cannot write escape sequences to the terminal.

The config file keeps its `0600` mode with no window at a wider one: `os.CreateTemp` creates the temp file private to the user, and the explicit `Chmod` runs before the encoder writes a byte. Two writers can still lose each other's access token, which costs the loser one refresh and nothing else, so that race is left alone rather than fixed with a per-platform file lock.

## Testing

`go test -race ./...`: exit 0, 39 packages `ok`, no `FAIL` and no `DATA RACE`. `go vet ./...`: no output. `golangci-lint run ./...`: `0 issues.`

Every new test was seen red first.

`TestRenewAccessToken_DoesNotBlockTokenReads` holds the grant open on a channel and asserts a concurrent token read still returns. Against the reverted lock split:

```
--- FAIL: TestRenewAccessToken_DoesNotBlockTokenReads (2.00s)
    client_test.go:1028: a token read blocked on an in-flight refresh
```

`TestSendRequest_GatewayUnauthorizedIsNotRenewed` pins an HTML body, a plain-text body and no body at all. Against the reverted guard, all three renewed once and sent the request twice:

```
--- FAIL: TestSendRequest_GatewayUnauthorizedIsNotRenewed/no_body
    client_test.go:847: expected: 0  actual: 1
        a 401 alpacon-server did not write is not a stale credential
    client_test.go:848: expected: 1  actual: 2
        a gateway refusal must not be replayed
```

The renewal itself is pinned by the tests 2180b1a added: a stale token renews and replays, a POST body survives the replay, a coded 401 and a legacy token are left alone, the retry happens at most once per request, and concurrent stale requests spend one grant.

`TestSaveConfig_ConcurrentWritesNeverLeaveAnUnreadableFile` runs two writers against one `config.json` while a reader parses whatever is on disk. Against the reverted `saveConfig`:

```
--- FAIL: TestSaveConfig_ConcurrentWritesNeverLeaveAnUnreadableFile (0.00s)
    config_test.go:341: Received unexpected error:
        	unexpected end of JSON input
        Messages:   	read a config that does not parse: ""
```

The writer assertion 2d4f2ed added was seen red the same way. With `saveConfig` returning an error only when the target already exists—so setup still succeeds and the reader still has a parseable file to read:

```
--- FAIL: TestSaveConfig_ConcurrentWritesNeverLeaveAnUnreadableFile (0.00s)
    config_test.go:340: Received unexpected error:
        	simulated replace failure
        	simulated replace failure
```

The debug line has no test. `CliDebug` writes to `os.Stderr` directly and the two capture helpers that exist are unexported copies in `utils` and `api/auth0`; a third copy to assert one diagnostic line costs more than the line is worth.

## Not in this PR

Item 3 of #312, the MFA step-up polls, is untouched. `api/mfa/mfa.go` and `api/event/sudolistener.go` still use a fixed-interval `time.NewTicker`. This PR does not close the issue, so that item stays open.

Refs #312


client/client_test.go
```diff
@@ -665,3 +667,367 @@ func TestLoadCurrentUser_ErrorIsCachedOnFailure(t *testing.T) {
...
+
+	var wg sync.WaitGroup
+	errs := make([]error, 2)
+	for i := range errs {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			_, errs[i] = ac.SendGetRequest("/api/test/")
+		}()
+	}
```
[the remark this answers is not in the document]
jisung-02: This does not apply on this module. `go.mod` declares `go 1.25.13`, and since Go 1.22 a `for` loop declares its variables per iteration, so each goroutine closes over its own `i`. The pre-1.22 capture bug is what the change gated on the `go` directive removed.

- `for i := range errs` gives each iteration a distinct `i`, so no goroutine can write another's index.
- The two goroutines write `errs[0]` and `errs[1]`, which are distinct memory locations, so the slice write is not a race either.
- `go test -race -run TestSendRequest_ConcurrentStaleRequestsRenewOnce -count=50 ./client/` reports `ok  github.com/alpacax/alpacon-cli/client  1.746s` with no `DATA RACE`.

client/client.go
```diff
@@ -246,7 +273,112 @@ func readJSONResponse(resp *http.Response) ([]byte, error) {
...
+	}
+	return true
+}
+
+// isStaleCredential reports whether err is the 401 a mid-flight token expiry
+// produces. alpacon-server's Auth0 authenticator swallows an expired token and
+// returns no user (auth0/auth.py), so the request falls through to
+// IsAuthenticatedOr401 and raises DRF's NotAuthenticated—which no branch of the
+// server's error_code_handler rewrites, leaving a 401 with a detail and no code.
+// Every deliberate 401 (MFA required, IP not allowed, token ACL) carries a code,
```
junho226: I checked this against the server, and the claim is a little stronger than what alpacon-server actually guarantees. `Auth0JWTAuthentication.authenticate` catches every exception out of `super().authenticate(request)` and returns `None` (`auth0/auth.py:138-142`), so *every* Auth0-bearer rejection collapses into the same uncoded `NotAuthenticated`, not just expiry. Reachable examples: the workspace-claim mismatch at `:149`, `_validate_mfa_methods` returning False at `:158` (the authenticator-level MFA gate, which is a different path from the coded `check_mfa` in `auth0/api/mixins.py` that the test models), and `Auth0JWTInvitationAuthentication.get_user`'s explicit `AuthenticationFailed(_('User is not invited'))` at `:554`, which that same catch-all swallows into an uncoded 401 too.

Nothing breaks. Each of those costs one grant plus one replay and then surfaces the same error, and in the MFA-claim case a refresh may legitimately fix it, since `refreshWithDeviceScopeFallback` restores the claim through the device scope. So this is about the comment rather than the code: `isStaleCredential` is really "any Auth0-bearer 401 the server produced", and a reader who takes "every deliberate 401 carries a code" as an invariant will build on something the server does not promise. Softening it to what the code actually relies on would hold up better — an uncoded JSON 401 is the only 401 a new token could plausibly move, so it is the only one worth spending one retry on.

client/client.go
```diff
@@ -246,7 +273,112 @@ func readJSONResponse(resp *http.Response) ([]byte, error) {
...
+// in hand. sent is the token the rejected request carried: a different one means
+// another request in flight already renewed it, and retrying with that one beats
+// spending a second refresh-token grant on the same expiry.
+func (ac *AlpaconClient) renewAccessToken(sent string) bool {
+	ac.refreshMu.Lock()
+	defer ac.refreshMu.Unlock()
+	if ac.accessToken() != sent {
+		return true
+	}
+	if err := refreshAccessToken(ac); err != nil {
```
junho226, line 334: A failed grant is not remembered, so the `sent` check cannot short-circuit the next one: the token is unchanged, so every other in-flight request that took the same 401 runs its own full grant — `FetchAuthEnv` plus a token POST, and twice over when `refreshWithDeviceScopeFallback` retries without the device scope.

Bounded rather than unbounded, which is why this is a note and not a request: it is capped by the in-flight count, each request still retries at most once, and `pollForApproval` ends on the first fatal 401 so it never accumulates across ticks. Recording it so the bound is a known one rather than an incidental one — if a caller ever fans requests out across one client, remembering the failure for the life of the client would collapse it back to a single grant.

client/client.go
```diff
@@ -130,7 +144,20 @@ func checkAuthStatus(statusCode int, body []byte) error {
...
+		code:       code,
+		source:     source,
+		apiPayload: isJSONObject(body),
+	}
+}
+
+// isJSONObject reports whether body is a JSON object. alpacon-server renders
+// every error as one, so anything else on a 401—an HTML page, a bare string,
+// nothing at all—came from a proxy, a WAF or an mTLS gate ahead of it.
+func isJSONObject(body []byte) bool {
```
junho226, line 160: The exclusion is sound and I would keep it, but the direction is worth spelling out for whoever touches this next. Only "not JSON ⇒ not the server" holds; the converse does not, since any intermediary that emits a JSON object clears this test. That is fine for the job it does here — it is a filter that removes gateway refusals, not evidence of who wrote the response.

Worth naming, because `apiPayload` reads like a provenance flag and the field name invites that reading. Nothing downstream treats it as authenticity today; the risk is a later change that does.

junho226 (dismissed):
## Verdict

Approving. I traced the lock discipline, the 401 discrimination and the replay path against the code rather than the commit messages, and checked the server-side premise in `alpacon-server` itself. The design holds. Three non-blocking notes below, all about how sharp the discrimination is *claimed* to be rather than about what the code does.

## Verification run

- `go test -race ./...` — exit 0, 39 packages `ok`, no `FAIL`, no `DATA RACE` (go1.25.13 darwin/amd64).
- `go test -race -count=10` over the six new client tests, and `-count=5` over `TestPollForApproval_SurvivesATokenExpiryMidWait` — both clean, so the concurrency tests are not passing by luck.
- `go vet ./...` — no output. `golangci-lint` is not installed on this machine, so I did not re-run it.

## Concurrency

The lock split does what 85eea7a claims. Traced by hand rather than from the message:

- `renewAccessToken` takes `refreshMu`, calls `accessToken()` (which takes and releases `tokenMu`), then runs the grant holding only `refreshMu`. `refreshLocked` reacquires `tokenMu` for the duration of `setAccessToken` and nothing else. So the grant genuinely runs off the read lock, and `setHTTPHeader` on another goroutine never waits on Auth0.
- Lock order is one-directional throughout: `refreshMu`→`tokenMu`, never the reverse, and no path takes `tokenMu` and then blocks on anything. No deadlock is reachable.
- No re-entrancy. `auth0.RefreshAccessToken` reaches `FetchAuthEnv` and `requestRefreshedToken`, and both call `httpClient.Do` directly (`api/auth0/auth0.go:57`, `:366`). Neither goes back through `sendRequest`, so `refreshMu` cannot be re-taken from underneath itself. The comment on `refreshLocked` is accurate.
- The thundering herd is collapsed by the `sent` comparison *inside* the critical section, which is the part that matters — the loser of the race compares against the token the winner already installed and skips its own grant. Publication is safe too: the new token is written under `tokenMu` and the replay reads it back through `setHTTPHeader`, so no goroutine sees a stale value.
- One interaction worth recording as deliberate rather than lucky: `api/event/sudolistener.go` and `api/mfa/mfa.go` call the public `RefreshToken()`, which now takes the same `refreshMu`. An explicit grant racing a mid-flight renewal therefore costs one round trip between them, not two.

## Security

**Can an upstream forge the marker?** Yes, trivially — and it buys nothing new. The test is "body parses as a JSON object" plus "no `code` field", and any intermediary that emits JSON satisfies both. What the code uses is only the safe direction of that implication: non-JSON ⇒ definitely not the server ⇒ exclude. That is a sound exclusion filter and it is exactly what the new tests pin. It is not an authenticity check, and nothing downstream treats it as one today.

**Can it be driven to replay the refresh token at an attacker-chosen endpoint?** No. The grant's destination is `https://` + `envInfo.Auth0.Domain`, and `envInfo` comes from `FetchAuthEnv(ac.BaseURL)` — the configured workspace, over the workspace's own TLS. Nothing derived from the 401 body reaches that URL. A captive portal or an off-path proxy cannot reach this code at all, because it cannot produce a valid TLS response for the workspace host. Someone who can (a TLS-terminating proxy, or a workspace logged into with `--insecure`) does gain the ability to trigger the grant on demand, which is new — but the destination still comes from the channel the CLI already trusts completely, and that same position already yields the bearer token straight off the request header. No escalation.

**Is the refresh token ever logged?** No. `CliDebug` formats only `err`, and every error on that path is token-free: the `no refresh token stored` sentinel, `FetchAuthEnv`'s `response status: %s`, `*url.Error` (which carries the URL, while the token travels in the POST body), `json.Unmarshal` failures, and `oauthError`, whose `Error()` renders code and description only (`api/auth0/auth0.go:44`). It is also gated behind `DebugEnabled()`.

**Does the bearer gate hold?** Yes, and it is load-bearing. `api/servicetoken/auth.py:56,59` raise uncoded 401s that would otherwise match `isStaleCredential` exactly. They are unreachable from here: `get_auth_token` parses `token="..."` (`api/apitoken/auth.py:38`), so a service token never arrives with `Bearer`, and `renewedRequest`'s `CutPrefix` rejects it before the grant.

## Correctness

- **Body replay.** Correct for every shape that exists. `SendPostRequest` and `SendPatchRequest` pass a `*bytes.Buffer`, so `http.NewRequest` installs `GetBody` itself; the one real `SendMultipartStreamRequest` caller (`api/packages/packages.go:123`) passes the `*os.File` that `SpoolToTempFile` returns, and the pre-existing `GetBody` reopens it by name for the replay. Anything else refuses to replay rather than sending a truncated body. Reading the `Authorization` header back off `req` after `Do` is safe — `net/http` forks the request rather than mutating the caller's headers — and the `[]string{"Bearer stale", "Bearer fresh"}` assertion pins it.
- **No retry loop.** The renewal is attempted once per `sendRequest`, and a second 401 propagates. A persistent 401 costs exactly two round trips and one grant. `pollForApproval` cannot turn that into a loop either: `IsFatalClientError` still classifies 401 as fatal, so a renewal that genuinely cannot help ends the wait on the first pass rather than re-grinding it every tick.
- **When the renewal itself 401s.** `requestRefreshedToken` never re-enters the client, so it returns an `oauthError` or a decode error, `renewAccessToken` returns false, and the caller surfaces the server's own 401. Pinned by `TestSendRequest_RenewalFailureSurfacesTheOriginal401`.
- **The server-side premise checks out.** I verified it in `alpacon-server` rather than taking the comment's word, because it is the one assumption the whole feature rests on. `Auth0JWTAuthentication.authenticate` wraps `super().authenticate(request)` in `except Exception: return None` (`auth0/auth.py:138-142`), so an expired token never surfaces as simplejwt's `InvalidToken` — which would have shipped a `token_not_valid` code and quietly defeated the empty-slot test. It falls through to `NotAuthenticated`, and `error_code_handler` rewrites only `PermissionDenied`/`AuthenticationFailed` (`utils/handlers.py:73`), never `NotAuthenticated`. The uncoded 401 is real.
- **Contexts and deadlines.** `replayableClone` propagates `req.Context()` correctly. Worth knowing there is nothing to propagate: `createRequest` uses `http.NewRequest`, and `NewAlpaconAPIClient` sets no `http.Client.Timeout`, so no request on this client carries a deadline and a renewal blocked on a hung Auth0 connection is bounded only by the TCP stack. Pre-existing and untouched here — I mention it because the lock split makes "one grant at a time" a property whose bound is now worth knowing.

## YAGNI

Shaped for the paths that exist, not past them. Two call sites need renewal and both get it through the same `renewedRequest`; the five helpers are each one idea and each reachable. `replayableClone`'s unrewindable branch is the only speculative-looking one, and it is five lines guarding a real footgun `SendMultipartStreamRequest` can hand it. Routing `SendMultipartStreamRequest` through `sendRequest` also deletes a duplicated round-trip block rather than adding one. Nothing here to cut.

## One note that has no line to sit on

The `refreshAccessToken` seam is used by `renewAccessToken` but not by the public `RefreshToken()`, which calls `refreshLocked` directly. That is the right call — the worksession test exercises the real grant end to end, and it should not be stubbable — but someone who stubs the seam and finds `RefreshToken()` unaffected will lose a few minutes to it. One clause in the seam's comment would pay for itself.

client/client.go
jisung-02: Fixed in 2b7c4c8. You are right that the invariant was the server's to make and it does not make it—the catch-all in `Auth0JWTAuthentication.authenticate` puts the workspace-claim mismatch, the authenticator-level MFA gate and the uninvited user in the same uncoded slot as expiry, and the comment invited a reader to rule those out.

The comment now states the weaker thing the code actually rests on: an uncoded JSON 401 is the only 401 a new token could plausibly move, so it is the only one worth one retry. It names the three cases you found and says what they cost (one grant, one replay, same error), and notes the MFA-claim case a refresh may genuinely fix through the device scope. The behavior is unchanged.

The same sentence appeared in two other places and both are updated in the same commit: the `CLAUDE.md` stale-token-renewal entry and the doc comment on `TestSendRequest_CodedUnauthorizedIsNotRenewed`.

client/client.go
jisung-02: Fixed in 2b7c4c8. Recorded on `renewAccessToken` itself, since the `sent` check is where the wrong expectation would form.

The comment now says the grant failure is not remembered, so the check collapses nothing on that path—every other in-flight request that took the same 401 runs its own `FetchAuthEnv` plus token POST, twice over when the device-scope fallback retries. It names the bound (the in-flight count, one retry per request) and the fix if a caller ever fans out widely enough for it to matter: remember the failure for the life of the client.

No code change, per your note that this is a known bound rather than a request.

client/client.go
jisung-02: Fixed in 2b7c4c8. The direction is now spelled out in both places a reader arrives from.

The `apiPayload` field comment says it is a filter and not a provenance flag—only the negative holds, since anything in front of the server can emit a JSON object too—and that nothing may read it as proof of who wrote the body. `isJSONObject` carries the same one-line statement so the function does not read stronger than the field.

Kept as is otherwise, since the filter does the job it is there for.

client/client.go
```diff
@@ -246,7 +275,122 @@ func readJSONResponse(resp *http.Response) ([]byte, error) {
...
+}
+
+// isStaleCredential reports whether err is a 401 a fresh access token could
+// plausibly move. alpacon-server's Auth0 authenticator catches every exception
+// out of the authentication it wraps and returns no user (auth0/auth.py), so the
+// request falls through to IsAuthenticatedOr401 and raises DRF's
+// NotAuthenticated—which no branch of the server's error_code_handler rewrites,
+// leaving a 401 with a detail and no code. An expired token lands there, and so
+// does every other Auth0-bearer rejection that same catch-all absorbs: a
+// workspace-claim mismatch, the authenticator-level MFA gate, an uninvited user.
```
junho226, line 352: Nit, and the only thing I found in the delta—the conclusion is right and I checked all three examples land on the same uncoded 401, but two of them do not get there through the catch-all.

`except Exception` covers only what `super().authenticate(request)` raises (`auth0/auth.py:139-142`). The workspace-claim mismatch (`:149-155`) and the authenticator-level MFA gate (`:158-159`) are plain `return None` branches *after* that try, so they produce `NotAuthenticated` by returning no user, not by having an exception absorbed. The uninvited user is the one example the catch-all really does absorb: `Auth0JWTInvitationAuthentication.get_user` raises at `:554` from inside `super().authenticate()`, and `Auth0JWTNonMFAAuthentication` overrides only `_validate_mfa_methods` and `get_user`, so the base `authenticate` and its try are what run.

Nothing downstream cares—all three return no user, the request falls through to `IsAuthenticatedOr401`, and the code slot is empty either way, which is the only property `isStaleCredential` reads. I am raising it because this comment now exists to be precise about exactly this, and a reader who follows it into `auth0/auth.py` looking for a swallowed exception on a claim mismatch will find a bare `return None` and wonder which of the two is wrong. Something like "returns no user on every bearer rejection, whether it absorbs an exception or declines outright" carries the same point without the misattribution. `CLAUDE.md:84` phrases it the same way and would move with it.

junho226 (dismissed):
## Verdict

Re-approving. `2b7c4c89` closes all three notes from my review at `3d69e07d`, and it closes the first one properly—by weakening the claim to what the server actually promises, not by rewording around it. One nit inline, non-blocking.

## Prior findings

- **`client.go:343`, "every deliberate 401 carries a code" (closed).** The comment no longer asserts an invariant the server does not hold. It now says the empty code slot "is not proof of expiry—it is the only 401 worth spending one retry on", and names the other rejections that land in the same slot. That is the right scope: the code never needed the strong claim, only the weak one, and the comment now states the weak one.
- **`client.go:327`, a failed grant is not remembered (closed as documented).** The doc comment on `renewAccessToken` now records that the `sent` check collapses nothing when the grant fails, that the in-flight count is the bound, and what to do if a caller ever fans out widely enough for the bound to hurt. That was a note about a known-vs-incidental bound, so writing the bound down settles it.
- **`client.go:158`, `apiPayload` reads like a provenance flag (closed).** Fixed in both places a reader lands: the field comment now says "It is a filter, not a provenance flag: only the negative holds" and adds "nothing may read it as proof of who wrote the body", and `isJSONObject` gains "Only that direction holds: a JSON object clears the test whoever wrote it." The second sentence is the one that matters—it forecloses the later change that treats the flag as authenticity.

## Is the new wording true?

I re-derived it against `alpacon-server` rather than accepting the edit, since precision about the server is the entire point of the commit.

- The catch-all is real and unchanged: `except Exception as exc: ... return None` (`auth0/auth.py:139-142`).
- The three named examples all reach an uncoded 401. Workspace-claim mismatch returns no user at `:149-155`; the authenticator-level MFA gate returns no user at `:158-159`; `Auth0JWTInvitationAuthentication.get_user` raises `AuthenticationFailed(_('User is not invited'))` at `:554`, and since `Auth0JWTNonMFAAuthentication` overrides only `_validate_mfa_methods` and `get_user`, that raise runs inside the base `authenticate`'s try and is swallowed.
- `NotAuthenticated` subclasses `APIException`, not `AuthenticationFailed`, so `error_code_handler` (`utils/handlers.py:73`) leaves it alone. The code slot really is empty.
- The new closing clause—"a coded refusal names what it wants and a new token is not it"—holds for the MFA case, which is the one that could plausibly have gone the other way. The coded 401 comes from `MFARequiredMixin.check_mfa`, whose payload is `{'code': AUTH_MFA_REQUIRED, 'source': ...}` (`auth0/api/mixins.py:101-109`), and that is the step-up path (`check_mfa_completed`, timestamp-windowed). A silent refresh-token grant performs no challenge, so it cannot move the completion timestamp. Declining to renew there is correct, and it is the conservative direction regardless.

So the claim is now scoped to what the test covers and no longer asserts anything `auth0/auth.py`'s catch-all makes false.

## Scope of the delta

Comment-only, verified mechanically rather than by eye: every added and removed line under `client/` in `3d69e07d..2b7c4c89` begins with `//`. The only other file is `CLAUDE.md:84`, whose bullet gets the same softening. No executable line changed, so I did not re-run the concurrency suite—`go test -race ./...` and the `-count=10` repeats from my previous review still describe this binary. `go vet ./...` is clean and `go test ./client/...` is `ok` at this head, which is enough to catch a stray edit inside a comment block.

One test comment moved with it, on `TestSendRequest_CodedUnauthorizedIsNotRenewed`. `staleTokenHandler`'s comment was left alone and correctly so: it describes the shape the fixture emits on expiry, and never claimed expiry was the only thing that emits it.

## Not re-derived

The lock split and its ordering, the thundering-herd collapse via the `sent` comparison inside the critical section, the refresh-token handling and replay path, and the bearer gate that keeps service-token 401s out of the grant were all verified against the code at `3d69e07d`. Nothing in this delta touches them, so I did not walk them again.

client/client.go
jisung-02: Fixed in 6560b3f, with your wording. You are right that only the invitation case goes through the catch-all—`Auth0JWTInvitationAuthentication.get_user` raises at `:554` inside the base `authenticate`'s try—and that the claim mismatch and the MFA gate are `return None` branches after it.

The comment now says the authenticator "returns no user on every bearer rejection, whether it absorbs an exception or declines outright", which covers both routes without asserting either for a case that does not take it. `CLAUDE.md:84` carried the same sentence and moved with it.

One thing your note prompted that is not in the delta you reviewed: 2bfd908 makes `saveConfig` atomic. It opened `config.json` with `O_TRUNC` and encoded straight into it, and this branch made that write happen on any request that meets an expired token rather than once at startup, so two alpacon processes could leave a config the next command cannot parse. It writes a `0600` temp file and renames now. The PR body no longer defers it under "Not in this PR".

config/config.go
```diff
@@ -66,19 +74,37 @@ func saveConfig(config *Config) error {
...
 		return fmt.Errorf("failed to encode config to JSON: %v", err)
 	}
 
+	if err = file.Close(); err != nil {
+		return fmt.Errorf("failed to write config file: %v", err)
+	}
+
+	configFile := filepath.Join(configDir, ConfigFileName)
+	if err = os.Rename(tempFile, configFile); err != nil {
+		return fmt.Errorf("failed to replace config file: %v", err)
```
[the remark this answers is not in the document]
jisung-02: Checked this and the premise does not hold: `os.Rename` already replaces an existing destination on Windows.

Go's `os.Rename` calls `internal/syscall/windows.Rename` (`os/file_windows.go:260`), which is `MoveFileEx(from, to, MOVEFILE_REPLACE_EXISTING)` (`internal/syscall/windows/syscall_windows.go:364`). The repo's `utils/file_replace_windows.go` wraps `golang.org/x/sys/windows.Rename`, and at v0.47.0 that function is the same `MoveFileEx` call with the same flag (`windows/syscall_windows.go:708-718`)—`os.Rename` additionally runs `fixLongPath` on both paths.

So switching would not change the syscall, and the atomic-replace guarantee this commit is after already holds on Windows. Leaving `os.Rename` in place; `config/device.go:236` publishes its temp file the same way.

config/config_test.go
```diff
@@ -294,3 +296,49 @@ func TestGetAuthMethod(t *testing.T) {
...
+	}()
+
+	reads := 0
+	for {
+		select {
+		case <-done:
+			assert.Positive(t, reads, "the reader never got to run, so the test proved nothing")
+			return
+		default:
+		}
```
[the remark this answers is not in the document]
jisung-02: Good catch—fixed in 2d4f2ed.

You are right that the test was vacuous in that case: if every `saveConfig` failed, the reader kept re-reading the file `CreateConfig` wrote during setup, which parses every time, so the test went green without a single replacement ever happening.

Each writer now records its first error in its own slot (one slot per goroutine, so no synchronization needed) and stops, and the reader asserts `errors.Join(writeErrs...)` before it checks its own read count.

Seen red first: with `saveConfig` returning an error only when the target already exists—so setup still succeeds and the reader still has a parseable file—the new assertion fires:

```
--- FAIL: TestSaveConfig_ConcurrentWritesNeverLeaveAnUnreadableFile (0.00s)
    config_test.go:340: Received unexpected error:
        	simulated replace failure
        	simulated replace failure
```

config/config.go
```diff
@@ -66,19 +74,37 @@ func saveConfig(config *Config) error {
...
 	if err = encoder.Encode(config); err != nil {
 		return fmt.Errorf("failed to encode config to JSON: %v", err)
 	}
 
+	if err = file.Close(); err != nil {
+		return fmt.Errorf("failed to write config file: %v", err)
+	}
+
+	configFile := filepath.Join(configDir, ConfigFileName)
+	if err = os.Rename(tempFile, configFile); err != nil {
```
junho226, line 107: Non-blocking, and I would ship it as is—but goreleaser builds `windows/amd64` and `windows/arm64` (`.goreleaser.yaml:14,23`), and `os.Rename` is a different operation there. Go compiles it to `MoveFileEx(from, to, MOVEFILE_REPLACE_EXISTING)` (`internal/syscall/windows/syscall_windows.go:364` in the 1.25.13 toolchain this module pins), which has to delete the destination, and Go opens files without `FILE_SHARE_DELETE`—`sharemode := FILE_SHARE_READ | FILE_SHARE_WRITE` at `syscall/syscall_windows.go:396`. `LoadConfig` holds its handle open across a streaming `json.NewDecoder` decode (`config.go:149-165`), so a second alpacon process reading the config in that window makes this rename fail with a sharing violation instead of replacing anything.

The old in-place write would have succeeded there, so on Windows this is a failure mode the PR introduces. It is still the right trade—the reader that used to see a torn file now makes the writer back off—but the cost lands one rung up: `RefreshAccessToken` returns the save error (`api/auth0/auth0.go:234-236`), discarding an access token Auth0 has already minted. Transient, recovered by the next command, and CI is ubuntu-only (`build-and-test.yaml:20`) so nothing in the pipeline will ever surface it. Raising it so it is written down, not because I want it changed.

config/config.go
```diff
@@ -66,19 +74,37 @@ func saveConfig(config *Config) error {
...
+		return fmt.Errorf("failed to set config file permissions: %v", err)
+	}
 
 	encoder := json.NewEncoder(file)
 	encoder.SetIndent("", "    ")
 	if err = encoder.Encode(config); err != nil {
 		return fmt.Errorf("failed to encode config to JSON: %v", err)
 	}
 
+	if err = file.Close(); err != nil {
```
junho226, line 102: Nit, non-blocking. No `file.Sync()` runs before this `Close`, so the rename buys atomicity and not durability. The doc comment above reads slightly wider than that—"a reader sees one whole config or the other, never a half of each" is true of a concurrent reader, but after a crash or power loss the file can still come back empty or reverted, which is the same "cannot parse, run alpacon login" outcome the fix exists to prevent, reached by the other road. ext4 papers over the common case with its auto-fsync-on-replace-via-rename heuristic; APFS journals the rename but not the data behind it.

I would not argue for it on durability grounds alone—the payload is a renewed access token, so losing it costs one refresh and nothing else. What makes it cheap is the write rate: this path runs at login, workspace switch, and token renewal, so a flush here is free. One line if you want it, and the concurrency claim the commit actually makes holds either way.

junho226 (dismissed):
## Verdict

Re-approving. The prior nit is closed—`6560b3fe` replaces the misattribution with exactly the property that holds, and `CLAUDE.md` moved with it. The two commits after it are not doc work, so I reviewed the atomic replace and its test from scratch; both hold up, and I mutated the production code to prove the test is not vacuous. Two non-blocking notes inline.

## Prior nit (closed)

`client/client.go:345-347` now reads "returns no user on every bearer rejection, whether it absorbs an exception or declines outright (auth0/auth.py)". That is accurate for all three examples, which I re-checked at the server:

- catch-all: `except Exception as exc: ... return None` (`auth0/auth.py:138-142`)
- workspace-claim mismatch: `return None` after the try (`:149-155`)
- authenticator-level MFA gate: `return None` (`:157-159`)
- uninvited user: `raise AuthenticationFailed(_('User is not invited'))` (`:554`), raised inside `super().authenticate()` and absorbed by the catch-all

So one of the three goes through the catch-all and two decline outright, and the new wording covers both without naming which is which—which is the right level, since the only property `isStaleCredential` reads is that the code slot ends up empty. `CLAUDE.md:84` carries the same correction.

## The atomic replace (`2bfd9089`)

This is the part that was not reviewed before, so I traced it rather than read it.

**Rename is same-directory.** `os.CreateTemp(configDir, ...)` (`config.go:77`) puts the temp file in `~/.alpacon`, so `os.Rename` at `:104` never crosses a filesystem. A reader gets the whole old file or the whole new one. `LoadConfig` streams with `json.NewDecoder` off a handle it holds open across the decode (`config.go:149-165`), and on POSIX that keeps the old inode, so an in-flight read finishes on a consistent snapshot rather than seeing the file change under it.

**Permissions do not widen—they narrow.** I probed this rather than reasoning about it, running `CreateConfig` under umask `0022`, `0000`, and `0077`, each time over a pre-existing `config.json` at mode `0666`. The result was `0600` in all six combinations, with the directory left at `0700`. `os.CreateTemp` already creates at `0600 &~ umask` and the explicit `file.Chmod(0600)` at `:89` pins it exactly, so no umask can widen it. Worth naming the direction: the old `os.OpenFile(configFile, ...)` left an existing file's mode alone, so a config that had somehow become group- or world-readable stayed that way. The replace now forces `0600` on every save. That is a small improvement, not a regression.

**Temp cleanup covers every error path.** The deferred `Close`+`Remove` at `:82-85` is registered immediately after `CreateTemp`, so it runs for the chmod, encode, close, and rename failures alike. I forced the rename failure (replaced the target path with a directory) and confirmed the temp file was gone afterward and only `config.json` remained; the error surfaced as `failed to replace config file: ... file exists`. After a success, `Remove` hits `ENOENT` and the ignored error is correct.

Two things I considered and am not raising: the deferred `Remove` also runs after a successful rename, where it could in principle delete a *different* process's temp file that `CreateTemp` had just handed the same name—`O_EXCL` plus the random suffix makes that negligible; and the replace turns a symlinked `config.json` into a regular file, which is the standard trade-off of write-temp-and-rename and the safer direction for a credential file.

## The test guard (`2d4f2ed2`)

The commit claims the test could not previously fail for the reason it exists. I checked both halves by mutating `config.go` and reverting after each.

- **Restore the pre-fix in-place `O_TRUNC` write** (the bug the PR closes): red on every one of 3 `-race` runs, within 0.00s, with `read a config that does not parse: ""`. So the test does catch the corruption it was written for, and catches it fast.
- **Drop the `os.Rename` and `return nil`**: red, but through `assert.Positive(t, reads, ...)` rather than the parse check—setup's `CreateConfig` goes through the same `saveConfig`, so the file never appears at all and the reader never runs.
- **Leave setup's file intact and make `SaveRefreshedAuth0Token` return nil without saving**: green. A write that succeeds without landing is still not detected.

That third one is the residual gap, and it is fine as scoped: every path in `saveConfig` that skips the rename returns an error, so a silently-succeeding no-op is unreachable in the code as written, and the joined-error assert at `config_test.go:340` covers the case the commit message actually names. The `errors.Join` collection is race-clean by construction—one slot per writer, read only after `writers.Wait()` closes `done`—and `-race` agrees.

## Checks run

- `go vet ./...`—clean.
- `go test -race -count=1 ./...`—exit 0, 39 packages `ok`, no race reports.
- `go test -race -count=30 -run TestSaveConfig_ConcurrentWritesNeverLeaveAnUnreadableFile ./config/`—`ok` in 5.2s, no flake.
- The same test at `GOMAXPROCS=1 -count=3`—`ok` in 2.5s, so the reader's busy spin does not starve the writers on a single-core runner.

## Not re-derived

The lock split and its ordering, the thundering-herd collapse via the `sent` comparison, the refresh-token handling and replay path, the bearer gate, and the uncoded-401 reasoning itself were all verified in my earlier reviews. This delta touches none of them, so I did not walk them again—the one comment that describes the uncoded 401 changed, and I checked the new wording rather than the mechanism behind it.

## Delta since `2d4f2ed2`

The head moved to `1e1f64c7` on a single commit—a merge of `origin/main` (`d60a2f0`) into the branch—so everything above stands unchanged and I re-derived none of it. What I checked is the merge itself.

It was not a clean auto-merge: replaying it with `git merge-tree` reports a content conflict in `client/client.go`, because main's `checkPrivilegesURL` → `currentUserURL` rename and trailing-slash fix (#392) land in the same `const` block this PR adds `bearerPrefix` to, and in the same `AlpaconClient` struct it adds `tokenMu` and `refreshMu` to. The hand resolution kept both sides verbatim. `git diff 2d4f2ed..1e1f64c` is exactly main's own `03cf818..d60a2f0` diff and nothing besides it, and this PR's delta measured against its new base (`d60a2f0..1e1f64c`) is byte-identical to what it was against the old one—same 7 files, same 736 insertions and 27 deletions, differing only in blob hashes and one hunk offset. So no evil merge, and the resolution dropped none of this PR's own code.

Main's change does land on code this PR touches: `LoadCurrentUser` now requests `/api/iam/users/-/` through `SendGetRequest`, which routes into the `sendRequest` this PR gives the stale-token retry. Nothing about that pairing is new—the retry is transport-level and reads only the `Authorization` header `setHTTPHeader` set—and the merged file still carries every symbol the review above traced (`bearerPrefix`, the `refreshAccessToken` seam, `refreshLocked`, both mutexes), with no conflict marker left anywhere in the tree. `config/` is untouched by main, so both inline notes are still added lines at the same numbers.

Checks re-run on the merged tree: `go build ./...` clean, `go vet ./...` clean, `go test -race -count=1 ./...` exit 0 with 39 packages `ok` and no race reports.

config/config.go
jisung-02: Fixed the comment in 94fc654, and left the code as it is.

You are right that the sentence claimed more than the rename buys. It said "a reader sees one whole config or the other" without saying which reader, and nothing calls Sync before the rename, so a crash can still bring the file back empty or reverted—the same unparseable-config outcome by the other road.

It now says the guarantee is atomicity against another process and not durability against a crash, and names what a power loss can still leave behind. No Sync call: as you say the payload is a renewed access token, so losing it costs one refresh, and the comment is what was wrong.

config/config.go
jisung-02: Leaving this as is, and leaving the thread open so the note stays findable rather than buried in a resolved thread.

The chain checks out: MoveFileEx has to delete the destination, Go opens without FILE_SHARE_DELETE, and LoadConfig holds its handle across the streaming decode, so a concurrent reader on Windows turns the replace into a sharing violation—and the cost lands on RefreshAccessToken discarding a token Auth0 already minted.

Not fixing it here because every fix is a design call this PR should not be making: reading the file whole and closing before decoding shrinks the window without closing it, and a Windows-only retry needs a Windows box to verify. CI is ubuntu-only, so nothing here would test either.

config/config.go
```diff
@@ -66,19 +77,37 @@ func saveConfig(config *Config) error {
...
+		return fmt.Errorf("failed to write config file: %v", err)
+	}
+
+	configFile := filepath.Join(configDir, ConfigFileName)
+	if err = os.Rename(tempFile, configFile); err != nil {
```
junho226, line 107: Carried forward from the last round and re-anchored—it sat on `:104` before the comment block grew. Nothing at this head changes it, and I would still ship as is.

`os.Rename` is a different operation on Windows, which goreleaser builds for: `windows/amd64`, `windows/arm`, and `windows/arm64` (`.goreleaser.yaml:11-24`, with only `windows/386` ignored). Go compiles it to `MoveFileEx(from, to, MOVEFILE_REPLACE_EXISTING)`, which has to delete the destination, and Go opens files without `FILE_SHARE_DELETE`—the share mode in `syscall/syscall_windows.go` is `FILE_SHARE_READ | FILE_SHARE_WRITE` in the 1.25.13 toolchain `go.mod` pins. `LoadConfig` holds its handle open across a streaming `json.NewDecoder` decode (`config.go:152-169`), so a second alpacon process reading the config in that window makes this rename fail with a sharing violation instead of replacing anything, and `RefreshAccessToken` returns that save error (`api/auth0/auth0.go:234-236`), discarding an access token Auth0 has already minted.

On the new wording specifically: the sentence above holds on Windows too. A failed rename leaves the reader the whole old file, so the concurrent-reader guarantee never breaks there—the failure lands on the writer, which the comment does not mention. It is the same magnitude as the lost-update case the comment does mention (one refresh, recovered by the next command), reached by a different road. CI is ubuntu-only (`build-and-test.yaml:20`), so nothing in the pipeline will ever surface it. Raising it again only so it stays written down, not because I want it changed.

junho226 (approved):
## Verdict

Still approving. The delta is one commit and it is comments only—proven mechanically rather than read: every added and removed line in `git diff 1e1f64c7..94fc654d` starts with `//`, and parsing both versions of `config/config.go` with `go/parser`, then printing them back with `f.Comments = nil`, gives byte-identical output. No behavior changed, so there is nothing below the doc comment to re-derive.

## What the new comment claims, and whether it holds

The narrowed comment (`config/config.go:63-68`) makes four claims. All four are true of the code at this head.

- "a concurrent reader sees one whole config or the other"—`os.CreateTemp(configDir, ...)` at `:80` keeps the temp file in `~/.alpacon`, so the `os.Rename` at `:107` is same-directory and atomic on POSIX, and `LoadConfig` streams off a handle it holds open across the decode (`:152-169`), so an in-flight read finishes on the old inode.
- "That is atomicity against another process and not durability against a crash"—correct scoping, and the right property to name: the concurrent-reader guarantee is the one this fix actually buys.
- "nothing flushes the temp file before the rename"—verified, not inferred. `grep -rn '\.Sync()' config/` returns nothing, and `file.Close()` on a regular file does not fsync. The encoder writes straight to the file with no `bufio` in between, so the bytes reach the kernel at `Encode` time and stop there.
- "a power loss can still bring the config back empty or reverted"—both outcomes are reachable: rename durable and data not gives the zero-length file, neither durable gives the old contents. "can" is the right modality; ext4's fsync-on-replace-by-rename heuristic starts writeback but does not wait on it, and APFS journals the rename and not the data behind it.

So the commit narrows the guarantee to exactly what holds, and the commit body names the trade it is declining instead of glossing over it.

## The Windows case is sidestepped, not resolved—which is defensible

Worth stating plainly, because it cuts in the comment's favor: the new sentence is true on Windows as written. If `MoveFileEx` fails with a sharing violation the reader still sees one whole config—the old one—so "one whole config or the other, never a half of each" never breaks there. What the comment does not say is that the *writer* fails on that path, which is the cost my earlier note described. The comment is silent, not wrong, so I am carrying the note forward rather than escalating it.

## Prior notes

- `:99` → now `:102` (`file.Close()`)—**closed**, and I am not re-posting it. Its first half, the doc comment reading wider than the property it had, is exactly what this commit fixes. Its second half, add `file.Sync()`, is consciously declined with the reasoning stated in the commit body: the payload is a renewed access token, so a lost flush costs one refresh. That was my own framing of the trade, so nothing is left to ask. Dropping this anchor is a judgement call and not a mechanical one—line 102 is still an added line in the diff.
- `:104` → now `:107` (`os.Rename`)—**still open**, re-anchored and carried forward inline in shorter form. Nothing at this head moves it: goreleaser still builds `windows/amd64`, `windows/arm`, and `windows/arm64` (`.goreleaser.yaml:11-24`, only `windows/386` ignored), and every workflow job is `ubuntu-latest`.

## Anchor re-check

The comment block grew by three lines, so everything under it shifts by +3. Against the merge base—`d60a2f0`, unchanged from the last round—both targets are still added lines on the RIGHT side: `102 + 	if err = file.Close(); err != nil {` and `107 + 	if err = os.Rename(tempFile, configFile); err != nil {`.

## Checks

- `go build ./...`—clean.
- `go vet ./...`—clean.
- `go test -race -count=1 ./...`—exit 0, 39 packages `ok`, zero `FAIL` lines and zero `DATA RACE` reports.
- `go test -race -count=10 -run TestSaveConfig_ConcurrentWritesNeverLeaveAnUnreadableFile ./config/`—`ok` in 2.6s.

## Doc consistency

`CLAUDE.md:192` describes the same mechanism and claims no durability property of its own, so narrowing the code comment does not leave the repo doc behind it.

## Not re-derived

The lock split and its ordering, the thundering-herd collapse, the refresh-token handling and replay path, the atomic-replace permission probe, and the uncoded-401 reasoning were all settled in earlier rounds. This delta is a comment, so I walked none of them again.

config/config.go
jisung-02: Agreed on both halves, and not changing it here.

The correction is fair: the comment says a concurrent reader sees one whole config or the other, which holds on Windows too, but it never says who pays when the rename fails there. That cost is the writer's—`RefreshAccessToken` returns the save error and drops a token Auth0 already minted—and it is the same magnitude as the lost-update case the comment does name.

Resolving both Windows threads now that the PR is approved. They stay readable in the PR either way, and re-litigating the fix (read-whole-then-close, or a Windows-only retry) needs a Windows box that CI does not have.

PR #393, opened 2026-08-25
fix(event): provision a fresh session per dial and re-subscribe on connect

### LINKED ISSUE #293: fix(event): SudoListener cannot survive a reconnect—stale single-use token and no re-subscribe
Found while reviewing #292. `SudoListener` predates the provisioned-listener work in #291 and was never migrated onto it.

## Problem

`api/event/sudolistener.go:60` builds its skeleton with `newWSListener(ac, wsURL, sudoHandshakeTimeout)`, which pins one fixed URL for every dial:

```go
// api/event/wslistener.go:47
func newWSListener(ac *client.AlpaconClient, wsURL string, handshakeTimeout time.Duration) *wsListener {
	return newProvisionedWSListener(ac, func() (string, error) { return wsURL, nil }, handshakeTimeout)
}
```

That URL comes from `event.CreateEventSession(ac)` (`cmd/websh/websh.go:339`)—the same event-channel session API `Watcher` uses, whose tokens are single-use. `api/event/wslistener.go:20` states the constraint outright: "Event channel tokens are single-use, so a reconnect needs a freshly provisioned URL."

So after any disconnect, `SudoListener` re-dials a URL whose token is already spent. Every reconnect attempt fails for the life of the process.

Second defect, independent of the first: `NewSudoListener` assigns only `handleFrame` (`api/event/sudolistener.go:65`). `Watcher` assigns three hooks:

```go
// api/event/watcher.go:53-55
w.handleFrame = w.handleMessage
w.onConnected = w.subscribe
w.onDialFailed = w.announceOutage
```

The subscription is issued once, out of band, at `cmd/websh/websh.go:356`. With `onConnected` nil, a reconnect that somehow did succeed would come back unsubscribed and silently deliver nothing. With `onDialFailed` nil, the outage is not reported either.

## Impact

An interactive `websh` session that loses the event channel stops receiving sudo MFA events for the rest of the session. The user sees a sudo prompt hang with no explanation—the browser MFA flow never fires because the frame never arrives.

`CommandOutputListener` (`api/event/commandoutput_listener.go:55`) has the same shape and may share the problem; it was not checked here.

## Proposal

Migrate `SudoListener` onto `newProvisionedWSListener` with a `provision` closure that calls `CreateEventSession`, and assign `onConnected` to re-issue the sudo subscription (which means moving the `SubscribeEvent` call from `cmd/websh/websh.go` into the listener, as `Watcher` does). Decide separately whether `onDialFailed` should warn, given this is a background listener under an interactive terminal—a warning written mid-session would corrupt the raw-mode display.

## Scope

- `api/event/sudolistener.go`
- `cmd/websh/websh.go:338-360` (`setupSudoListener`)
- Check `api/event/commandoutput_listener.go` for the same defect

jisung-02:
## Background

Event channel tokens are single-use, which `api/event/wslistener.go` has stated outright since #291: a reconnect needs a freshly provisioned URL. `SudoListener` predated that work and never moved onto it. It built its skeleton with `newWSListener(ac, wsURL, sudoHandshakeTimeout)`, which pinned one fixed URL for every dial, so after any disconnect it re-dialed a token that was already spent—and kept failing that way for the life of the process.

It also assigned only `handleFrame`. With `onConnected` nil, a dial that somehow did succeed came back unsubscribed and silently delivered nothing; the subscription was issued once, out of band, in `setupSudoListener`. With `onDialFailed` nil, the outage was never reported either. The user-visible result: an interactive `websh` session that lost the event channel stopped receiving sudo MFA events, and the next sudo prompt hung with no explanation, because the browser MFA flow never fired.

The issue flagged `CommandOutputListener` as possibly sharing the shape. It did. `NewCommandOutputListener(ac, session.WebsocketURL, cmdID)` took its URL from a session its caller had created, and `streamSubscribed` issued both subscriptions out of band, so the same two defects applied to `alpacon exec` streaming.

### Both listeners provision their own session

`NewSudoListener(ac, serverName, sessionID)` drops the URL parameter and takes the websh session ID instead, since the listener now owns the subscription. It builds on `newProvisionedWSListener` with a `provisionSession` closure that calls `CreateEventSession` per dial, and assigns all three hooks: `handleFrame`, `onConnected` to re-subscribe, `onDialFailed` to announce the outage. `setupSudoListener` no longer creates a session or calls `SubscribeEvent`.

`NewCommandOutputListener(ac)` does the same. Its first connect has nothing to subscribe to—the command is submitted only after the WebSocket is up—so `subscribeTo(commandID, serverID)` issues the first subscription once the command exists, and every later connect re-issues it through `onConnected`. `streamSubscribed` lost its `session` parameter and both `SubscribeEvent` calls along with it.

`newWSListener` is gone. Every listener in the package provisions now, so a fixed-URL constructor is only a way to reintroduce this bug.

### An outage is announced once, and again after it recovers

`SudoListener.fail` is the one decision a failed connect attempt goes through. Once subscribed, the failure is an outage: it prints one warning, and a successful resubscribe clears the flag, so a second outage later in the same session speaks again rather than staying silent behind a latched flag. Before the first subscribe it stays quiet, because `setupSudoListener` is already reporting that case to the user itself. The notice goes through `utils.Yellow`, which emits ANSI only when stderr is a terminal, so a redirected session collects the text without escape codes. `Stop` cancels neither an in-flight dial nor an in-flight session create, so the announce checks `done` first and a failure that lands after shutdown says nothing.

`fail` is called from all three failure points—session create, dial, and resubscribe—not just wired to `onDialFailed`. `onDialFailed` fires only when the WebSocket handshake fails, so an outage that dies at `CreateEventSession` or at the resubscribe would otherwise drop the sudo MFA flow with nothing on screen. `announceOutage` survives as the thin dial hook over `fail`, the shape `Watcher` already had.

### One fatal judgment, shared by all three listeners

`isFatalRequestError` (`api/event/wslistener.go:198`) is the single rule: a 4xx that refuses rather than asks to be retried—so neither a 408 nor a 429—means a bad request or an expired login when it lands before the first subscribe, which the next attempt would hit again, so the listener stops and surfaces the reason. The one exception is a 401: websh created its own session on that token milliseconds earlier, so `SudoListener.fail` renews it once and lets the attempt count as non-fatal. A 401 answering a token it just renewed is a real refusal and still stops it. `Watcher.sessionCreateFailed` now calls it instead of carrying its own copy of the same expression.

After the first subscribe, no status is fatal for either new listener. A 401 arriving mid-stream on a long command would otherwise cost the rest of that run its live output, and the reconnect loop is exactly the right place to wait it out.

A 401 is still the one status a retry cannot fix by itself, so `SudoListener.fail` renews the access token before the next dial, holding `mfaMu` so it cannot race the refresh `handleSudoMFA` already does. Without it, an access token that expires during a long interactive `websh` session is re-presented on every dial until the session ends, and sudo MFA never comes back.

`CommandOutputListener.subscribe` refuses to subscribe once the listener is stopped. Its first connect has nothing to subscribe to, so the caller holds a connected listener while `SubmitCommand` runs; a drop plus a fatal session create in that window stops the listener behind the caller's back, and subscribing on the dead session's channel would have let `streamSubscribed` proceed as if output were streaming. It still would not lose output, since the poll goroutine runs regardless, but the run would degrade to polling with nothing saying why. The refusal carries the reason the listener stopped, wrapped from `Err()`, so the fallback warning names the server's own 401 or 403 rather than the halt. The check runs a second time once both subscriptions are in, which narrows the window to their round trips; closing it properly would mean `streamSubscribed` selecting on the listener's liveness in its main loop, which is more than this PR wants.

### The failure reaches the caller instead of a flat timeout

Both listeners expose `Err()`. `listenerFailure` hands `runCommandFallback` the server's own error when the listener reached the server at all, and the connect budget only when it did not, so a rejected session no longer reports itself as a timeout. `sudoListenerWarning` replaces the old `isNotFoundError` string matching with `utils.HTTPStatusCode(cause) == http.StatusNotFound`.

`SudoListener.fail` records every cause, not only the fatal one, and a successful subscribe clears it. Otherwise a 500, a refused connection, a TLS failure, or a proxy 502 all leave `Err()` nil and `websh` prints "timed out connecting to the event channel" where it used to print the server's own message. `Err()` now promises the last reason the listener could not connect rather than the reason it stopped; the `Stop` decision stays gated on the fatal rule.

## Safety

The warning interpolates the server's own `detail` text into a terminal that `websh` is about to put into raw mode, so it goes through `utils.SanitizeTerminalText` first, following the sibling precedent at `cmd/websh/websh_watch.go:53`.

It is printed with explicit CRLF rather than through `utils.CliWarning`. `cliMessage` strips `\r`, and `term.MakeRaw` clears OPOST, so a bare `\n` in raw mode drops a line without returning the cursor and the next line starts mid-screen. `setupSudoListener` runs in a goroutine racing `OpenNewTerminal`, so it cannot assume the terminal is still cooked.

A 404 stays silent, as before: it means an older server without the events API, and the websh session runs fine without sudo MFA events.

Two things are known and deliberately left:

`channelID` is read under `stateMu` and then used for a `SubscribeEvent` POST outside it, so a reconnect landing between the read and the POST can subscribe on the previous channel. Closing the window would mean holding a mutex across network I/O. The consequence is bounded and self-healing—the next reconnect resubscribes on the current channel, and for command output the poll still delivers every chunk, so a blip downgrades live streaming to polling rather than losing output.

The `stateMu` / `channelID` / `subscribed` / `err` block now appears in all three listeners. Extracting it would share the fields and the mutex but not the notification, which differs in every case: `Watcher` pushes to a channel, `SudoListener` writes to stderr, `CommandOutputListener` says nothing. The judgment that actually repeated was the 4xx test, and that is extracted.

## Testing

Twenty-four new tests in `api/event`, five in `cmd/websh`. The ones that pin this PR's behavior:

- `TestSudoListener_CreatesANewSessionOnReconnect` and `TestSudoListener_ResubscribesOnReconnect` cover the two defects the issue names.
- `TestCommandOutputListener_CreatesANewSessionAndResubscribesOnReconnect` covers the same pair for exec streaming, and `TestCommandOutputListener_FirstConnectSubscribesToNothing` pins the empty first connect.
- `TestSudoListener_AnnouncesOutageOncePerOutage`, `..._WhenSessionCreateFails`, `..._WhenResubscribeFails`, and `TestSudoListener_AnnouncesTheNextOutageAfterRecovering` cover the announcement, each failure point, and the flag reset.
- `TestSudoListener_FirstSubscribeFailureStopsAndSurfaces` and `TestCommandOutputListener_KeepsRetryingAfterTheFirstSubscribe` pin the two halves of the fatal rule.
- `TestSudoListener_RefreshesTheAccessTokenOnUnauthorized` and `TestSudoListener_LeavesTheTokenAloneOnOtherStatuses` pin both halves of the 401 rule.
- `TestSudoListener_StaysQuietWhenAFailureLandsAfterStop` and `TestSudoListener_SurfacesANonFatalCauseAfterTheWaitEnds` cover the shutdown window and the widened `Err()`.
- `TestSudoListener_TriesOneRefreshBeforeGivingUpOnAnUnauthorizedFirstSession` pins both halves of the pre-subscribe exception: one refresh, then the refusal stands.
- `TestCommandOutputListener_SubscribeFailsOnAStoppedListener` covers the stop-then-subscribe window, `..._SubscribeCarriesWhyTheListenerStopped` pins the wrapped reason, and `..._SubscribeFailsWhenTheListenerStopsMidSubscription` drives the narrower window by stopping the listener from inside the subscription handler.
- `TestSudoListenerWarning_SanitizesTheServersText` and `TestSetupSudoListener_WarningReturnsTheCursor` cover the two raw-mode concerns above.

Every test was seen red first, and the assertions that carry a judgment were proven by mutation. Deleting `sl.warned = false` turns `TestSudoListener_AnnouncesTheNextOutageAfterRecovering` into `expected: 2, actual: 1`. Dropping the refresh call, refreshing on every status but 401, dropping the `done` guard, and recording only the fatal cause each turn exactly one test red. So do dropping the pre-subscribe refresh, dropping its one-shot guard, removing the stopped-listener check, dropping the `%w` wrap around the halt reason, and removing the second check.

On the final diff:

- `go test -race -count=1 ./...`: 39 `ok`, no `FAIL`.
- `go vet ./...`: no output.

Closes #293

### DIFF
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -3,6 +3,7 @@ package event
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
@@ -115,11 +116,11 @@ func (l *CommandOutputListener) subscribeTo(commandID, serverID string) error {
 func (l *CommandOutputListener) subscribe() error {
 	// A fatal session failure can stop the listener between the caller's WaitConnected
 	// and its subscribeTo. Subscribing then succeeds against a channel nothing reads,
-	// so the caller would stream from a dead listener instead of falling back.
-	select {
-	case <-l.done:
-		return errors.New("event listener stopped before the subscription")
-	default:
+	// so the caller would stream from a dead listener instead of falling back. Checking
+	// again once the subscriptions are in narrows the window to their round trips rather
+	// than closing it: a Stop landing after the second check still goes unnoticed.
+	if err := l.halted(); err != nil {
+		return err
 	}
 
 	l.cmdMu.Lock()
@@ -146,11 +147,31 @@ func (l *CommandOutputListener) subscribe() error {
 		_ = SubscribeEvent(l.ac, channelID, EventTypeCommandFin, serverID)
 	}
 
+	if err := l.halted(); err != nil {
+		return err
+	}
+
 	l.stateMu.Lock()
 	l.subscribed = true
 	l.stateMu.Unlock()
 
 	return nil
+}
+
+// halted reports why the listener stopped, or nil while it is live. provisionSession
+// records the reason on its way out, and carrying it here is what keeps the caller's
+// fallback warning naming the server's own refusal rather than the halt.
+func (l *CommandOutputListener) halted() error {
+	select {
+	case <-l.done:
+	default:
+		return nil
+	}
+
+	if cause := l.Err(); cause != nil {
+		return fmt.Errorf("event listener stopped while subscribing: %w", cause)
+	}
+	return errors.New("event listener stopped while subscribing")
 }
 
 // Chunks returns a receive-only channel of parsed chunk events.
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -56,9 +56,10 @@ type commandOutputEnvelope struct {
 	} `json:"payload"`
 }
 
-// NewCommandOutputListener constructs a listener without connecting. ac may be
-// nil (empty header, for tests); the targets are set later via setTargets, because
-// SubmitCommand must run once the WS is already connected.
+// NewCommandOutputListener constructs a listener without connecting. ac may be nil
+// only for a test that never calls Start, because provisioning a session dereferences
+// it. The targets are set later via setTargets, because SubmitCommand must run once
+// the WS is already connected.
 func NewCommandOutputListener(ac *client.AlpaconClient) *CommandOutputListener {
 	l := &CommandOutputListener{
 		ac:       ac,
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -84,10 +84,9 @@ func (l *CommandOutputListener) provisionSession() (string, error) {
 	if err != nil {
 		l.stateMu.Lock()
 		l.err = err
-		// Before the first subscribe a fatal 4xx means a bad request or an expired
-		// login, so retrying only burns the caller's connect budget before it falls
-		// back to polling anyway. Once the command is streaming, giving up would cost
-		// the rest of the run its live output, so every status is retried.
+		// Before the first subscribe a fatal 4xx only burns the caller's connect budget
+		// before it falls back to polling anyway. Once the command is streaming, giving
+		// up would cost the rest of the run its live output, so every status is retried.
 		fatal := !l.subscribed && isFatalRequestError(err)
 		l.stateMu.Unlock()
 
@@ -105,8 +104,7 @@ func (l *CommandOutputListener) provisionSession() (string, error) {
 	return session.WebsocketURL, nil
 }
 
-// subscribeTo issues the first subscription, once the command exists. Every later
-// connect re-issues it through onConnected.
+// subscribeTo issues the first subscription; every later connect re-issues it through onConnected.
 func (l *CommandOutputListener) subscribeTo(commandID, serverID string) error {
 	l.setTargets(commandID, serverID)
 	return l.subscribe()
@@ -184,8 +182,6 @@ func (l *CommandOutputListener) handleMessage(raw []byte) {
 	}
 }
 
-// setTargets names the command whose chunks to keep and the server whose fin event
-// ends the stream.
 func (l *CommandOutputListener) setTargets(commandID, serverID string) {
 	l.cmdMu.Lock()
 	l.commandID = commandID
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -26,12 +26,17 @@ type ChunkEvent struct {
 // event WebSocket, exposing chunks on Chunks() and the command's fin on
 // Finished().
 //
-// Lifecycle: NewCommandOutputListener -> Start -> (consume Chunks) -> Stop.
-// Stop is idempotent and safe to call from any goroutine.
+// Lifecycle: NewCommandOutputListener -> Start -> subscribeTo -> (consume Chunks) -> Stop.
+// Stop is idempotent and safe to call from any goroutine. Event channel tokens are
+// single-use, so every dial provisions its own session and re-subscribes.
 type CommandOutputListener struct {
 	*wsListener
-	cmdMu     sync.Mutex // guards commandID
+	ac        *client.AlpaconClient
+	cmdMu     sync.Mutex // guards commandID, serverID
 	commandID string
+	serverID  string
+	chanMu    sync.Mutex // guards channelID
+	channelID string
 	chunks    chan ChunkEvent
 	finished  chan struct{}
 }
@@ -50,16 +55,66 @@ type commandOutputEnvelope struct {
 }
 
 // NewCommandOutputListener constructs a listener without connecting. ac may be
-// nil (empty header, for tests); commandID may be set later via setCommandID.
-func NewCommandOutputListener(ac *client.AlpaconClient, wsURL, commandID string) *CommandOutputListener {
+// nil (empty header, for tests); the targets are set later via setTargets, because
+// SubmitCommand must run once the WS is already connected.
+func NewCommandOutputListener(ac *client.AlpaconClient) *CommandOutputListener {
 	l := &CommandOutputListener{
-		wsListener: newWSListener(ac, wsURL, commandOutputConnectTimeout),
-		commandID:  commandID,
-		chunks:     make(chan ChunkEvent, commandOutputChunkBuffer),
-		finished:   make(chan struct{}, 1),
+		ac:       ac,
+		chunks:   make(chan ChunkEvent, commandOutputChunkBuffer),
+		finished: make(chan struct{}, 1),
 	}
+	l.wsListener = newProvisionedWSListener(ac, l.provisionSession, commandOutputConnectTimeout)
 	l.handleFrame = l.handleMessage
+	l.onConnected = l.subscribe
 	return l
+}
+
+func (l *CommandOutputListener) provisionSession() (string, error) {
+	session, err := CreateEventSession(l.ac)
+	if err != nil {
+		return "", err
+	}
+
+	l.chanMu.Lock()
+	l.channelID = session.ChannelID
+	l.chanMu.Unlock()
+
+	return session.WebsocketURL, nil
+}
+
+// subscribeTo issues the first subscription, once the command exists. Every later
+// connect re-issues it through onConnected.
+func (l *CommandOutputListener) subscribeTo(commandID, serverID string) error {
+	l.setTargets(commandID, serverID)
+	return l.subscribe()
+}
+
+func (l *CommandOutputListener) subscribe() error {
+	l.cmdMu.Lock()
+	commandID, serverID := l.commandID, l.serverID
+	l.cmdMu.Unlock()
+
+	// The command is submitted only after the WS is up, so the first connect has
+	// nothing to subscribe to yet.
+	if commandID == "" {
+		return nil
+	}
+
+	l.chanMu.Lock()
+	channelID := l.channelID
+	l.chanMu.Unlock()
+
+	if err := SubscribeEvent(l.ac, channelID, EventTypeCommandOutput, commandID); err != nil {
+		return err
+	}
+
+	// Chunks carry no end marker, so without this the exit waits for the poll to
+	// notice. Best effort: on failure that poll stays the only terminal signal.
+	if serverID != "" {
+		_ = SubscribeEvent(l.ac, channelID, EventTypeCommandFin, serverID)
+	}
+
+	return nil
 }
 
 // Chunks returns a receive-only channel of parsed chunk events.
@@ -102,10 +157,11 @@ func (l *CommandOutputListener) handleMessage(raw []byte) {
 	}
 }
 
-// setCommandID assigns the commandID after construction. Used because
-// SubmitCommand must run after the WS is already connected.
-func (l *CommandOutputListener) setCommandID(id string) {
+// setTargets names the command whose chunks to keep and the server whose fin event
+// ends the stream.
+func (l *CommandOutputListener) setTargets(commandID, serverID string) {
 	l.cmdMu.Lock()
-	l.commandID = id
+	l.commandID = commandID
+	l.serverID = serverID
 	l.cmdMu.Unlock()
 }
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -31,15 +31,16 @@ type ChunkEvent struct {
 // single-use, so every dial provisions its own session and re-subscribes.
 type CommandOutputListener struct {
 	*wsListener
-	ac        *client.AlpaconClient
-	cmdMu     sync.Mutex // guards commandID, serverID
-	commandID string
-	serverID  string
-	stateMu   sync.Mutex // guards channelID, err
-	channelID string
-	err       error
-	chunks    chan ChunkEvent
-	finished  chan struct{}
+	ac         *client.AlpaconClient
+	cmdMu      sync.Mutex // guards commandID, serverID
+	commandID  string
+	serverID   string
+	stateMu    sync.Mutex // guards channelID, subscribed, err
+	channelID  string
+	subscribed bool
+	err        error
+	chunks     chan ChunkEvent
+	finished   chan struct{}
 }
 
 // commandOutputEnvelope is the WS message format emitted by alpacon-server. One
@@ -83,10 +84,14 @@ func (l *CommandOutputListener) provisionSession() (string, error) {
 	if err != nil {
 		l.stateMu.Lock()
 		l.err = err
+		// Before the first subscribe a 4xx means a bad request or an expired login,
+		// so retrying only burns the caller's connect budget before it falls back to
+		// polling anyway. Once the command is streaming, giving up would cost the
+		// rest of the run its live output, so every status is retried.
+		fatal := !l.subscribed && isFatalRequestError(err)
 		l.stateMu.Unlock()
-		// A 4xx means a bad request or an expired login, so retrying only burns the
-		// caller's connect budget before it falls back to polling anyway.
-		if isFatalRequestError(err) {
+
+		if fatal {
 			l.Stop()
 		}
 		return "", err
@@ -131,6 +136,10 @@ func (l *CommandOutputListener) subscribe() error {
 	if serverID != "" {
 		_ = SubscribeEvent(l.ac, channelID, EventTypeCommandFin, serverID)
 	}
+
+	l.stateMu.Lock()
+	l.subscribed = true
+	l.stateMu.Unlock()
 
 	return nil
 }
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -2,6 +2,7 @@ package event
 
 import (
 	"encoding/json"
+	"errors"
 	"sync"
 	"time"
 
@@ -112,6 +113,15 @@ func (l *CommandOutputListener) subscribeTo(commandID, serverID string) error {
 }
 
 func (l *CommandOutputListener) subscribe() error {
+	// A fatal session failure can stop the listener between the caller's WaitConnected
+	// and its subscribeTo. Subscribing then succeeds against a channel nothing reads,
+	// so the caller would stream from a dead listener instead of falling back.
+	select {
+	case <-l.done:
+		return errors.New("event listener stopped before the subscription")
+	default:
+	}
+
 	l.cmdMu.Lock()
 	commandID, serverID := l.commandID, l.serverID
 	l.cmdMu.Unlock()
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -35,8 +35,9 @@ type CommandOutputListener struct {
 	cmdMu     sync.Mutex // guards commandID, serverID
 	commandID string
 	serverID  string
-	chanMu    sync.Mutex // guards channelID
+	stateMu   sync.Mutex // guards channelID, err
 	channelID string
+	err       error
 	chunks    chan ChunkEvent
 	finished  chan struct{}
 }
@@ -69,15 +70,32 @@ func NewCommandOutputListener(ac *client.AlpaconClient) *CommandOutputListener {
 	return l
 }
 
+// Err returns why the listener could not connect, so a caller that only sees
+// WaitConnected time out can still report the server's own reason.
+func (l *CommandOutputListener) Err() error {
+	l.stateMu.Lock()
+	defer l.stateMu.Unlock()
+	return l.err
+}
+
 func (l *CommandOutputListener) provisionSession() (string, error) {
 	session, err := CreateEventSession(l.ac)
 	if err != nil {
+		l.stateMu.Lock()
+		l.err = err
+		l.stateMu.Unlock()
+		// A 4xx means a bad request or an expired login, so retrying only burns the
+		// caller's connect budget before it falls back to polling anyway.
+		if isFatalRequestError(err) {
+			l.Stop()
+		}
 		return "", err
 	}
 
-	l.chanMu.Lock()
+	l.stateMu.Lock()
 	l.channelID = session.ChannelID
-	l.chanMu.Unlock()
+	l.err = nil
+	l.stateMu.Unlock()
 
 	return session.WebsocketURL, nil
 }
@@ -100,9 +118,9 @@ func (l *CommandOutputListener) subscribe() error {
 		return nil
 	}
 
-	l.chanMu.Lock()
+	l.stateMu.Lock()
 	channelID := l.channelID
-	l.chanMu.Unlock()
+	l.stateMu.Unlock()
 
 	if err := SubscribeEvent(l.ac, channelID, EventTypeCommandOutput, commandID); err != nil {
 		return err
--- a/api/event/commandoutput_listener.go
+++ b/api/event/commandoutput_listener.go
@@ -84,10 +84,10 @@ func (l *CommandOutputListener) provisionSession() (string, error) {
 	if err != nil {
 		l.stateMu.Lock()
 		l.err = err
-		// Before the first subscribe a 4xx means a bad request or an expired login,
-		// so retrying only burns the caller's connect budget before it falls back to
-		// polling anyway. Once the command is streaming, giving up would cost the
-		// rest of the run its live output, so every status is retried.
+		// Before the first subscribe a fatal 4xx means a bad request or an expired
+		// login, so retrying only burns the caller's connect budget before it falls
+		// back to polling anyway. Once the command is streaming, giving up would cost
+		// the rest of the run its live output, so every status is retried.
 		fatal := !l.subscribed && isFatalRequestError(err)
 		l.stateMu.Unlock()
 
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -89,8 +89,7 @@ func (sl *SudoListener) Err() error {
 func (sl *SudoListener) provisionSession() (string, error) {
 	session, err := CreateEventSession(sl.ac)
 	if err != nil {
-		sl.announceOutage(err)
-		return "", sl.failIfFatal(err)
+		return "", sl.fail(err)
 	}
 
 	sl.stateMu.Lock()
@@ -106,8 +105,7 @@ func (sl *SudoListener) subscribe() error {
 	sl.stateMu.Unlock()
 
 	if err := SubscribeEvent(sl.ac, channelID, EventTypeSudo, sl.sessionID); err != nil {
-		sl.announceOutage(err)
-		return sl.failIfFatal(err)
+		return sl.fail(err)
 	}
 
 	sl.stateMu.Lock()
@@ -119,40 +117,37 @@ func (sl *SudoListener) subscribe() error {
 	return nil
 }
 
-// failIfFatal stops the listener when retrying cannot help: a 4xx before the first
-// subscribe means a bad session or an expired login, not an outage. Everything else
-// is left to the reconnect loop.
-func (sl *SudoListener) failIfFatal(cause error) error {
+// announceOutage is the dial-failure hook: a dial error reaches neither
+// provisionSession nor subscribe, so without it a transport outage is silent.
+func (sl *SudoListener) announceOutage(cause error) {
+	_ = sl.fail(cause)
+}
+
+// fail decides what a failed connect attempt means. A fatal 4xx before the first
+// subscribe is a bad session or an expired login, so the listener stops carrying the
+// reason; websh reports that one itself, which is why nothing is announced before then.
+// Once subscribed, the same failure is an outage: announced once, then left to the
+// reconnect loop, with its own CRLF because websh holds the terminal in raw mode.
+func (sl *SudoListener) fail(cause error) error {
 	sl.stateMu.Lock()
 	fatal := !sl.subscribed && isFatalRequestError(cause)
 	if fatal {
 		sl.err = cause
+	}
+	announce := sl.subscribed && !sl.warned
+	if announce {
+		sl.warned = true
 	}
 	sl.stateMu.Unlock()
 
 	if fatal {
 		sl.Stop()
 	}
+	if announce {
+		_, _ = fmt.Fprintf(os.Stderr, "\r\n%s\r\n", utils.Yellow("Sudo MFA listener disconnected; retrying..."))
+	}
 
 	return cause
-}
-
-// announceOutage reports a dropped event channel once per outage. Before the first
-// subscribe the caller (websh) already handles the failure, so this stays quiet.
-// The message is wrapped in CRLF because websh holds the terminal in raw mode.
-func (sl *SudoListener) announceOutage(error) {
-	sl.stateMu.Lock()
-	quiet := !sl.subscribed || sl.warned
-	if !quiet {
-		sl.warned = true
-	}
-	sl.stateMu.Unlock()
-
-	if quiet {
-		return
-	}
-
-	_, _ = fmt.Fprint(os.Stderr, "\r\n\033[33mSudo MFA listener disconnected; retrying...\033[0m\r\n")
 }
 
 func (sl *SudoListener) handleMessage(message []byte) {
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -122,7 +122,7 @@ func (sl *SudoListener) subscribe() error {
 // is left to the reconnect loop.
 func (sl *SudoListener) failIfFatal(cause error) error {
 	sl.stateMu.Lock()
-	fatal := !sl.subscribed && utils.IsFatalClientError(utils.HTTPStatusCode(cause))
+	fatal := !sl.subscribed && isFatalRequestError(cause)
 	if fatal {
 		sl.err = cause
 	}
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -47,23 +47,88 @@ type sudoMFAEvent struct {
 // The AlpaconClient (ac) is shared with the terminal WebSocket goroutines.
 // http.Client is concurrency-safe. Token refresh and grant verification are
 // serialized by mfaMu so only one MFA flow runs at a time.
+//
+// Event channel tokens are single-use, so every dial provisions its own session
+// and re-subscribes once connected.
 type SudoListener struct {
 	*wsListener
 	ac         *client.AlpaconClient
 	serverName string
+	sessionID  string
 	mfaMu      sync.Mutex // serializes handleSudoMFA so only one MFA flow runs at a time
+
+	stateMu    sync.Mutex // guards channelID, subscribed, err
+	channelID  string
+	subscribed bool
+	err        error
 }
 
 // NewSudoListener creates a SudoListener but does not connect yet. ac may be
 // nil in tests; MFA request frames are then dropped instead of dereferencing it.
-func NewSudoListener(ac *client.AlpaconClient, wsURL, serverName string) *SudoListener {
+func NewSudoListener(ac *client.AlpaconClient, serverName, sessionID string) *SudoListener {
 	sl := &SudoListener{
-		wsListener: newWSListener(ac, wsURL, sudoHandshakeTimeout),
 		ac:         ac,
 		serverName: serverName,
+		sessionID:  sessionID,
 	}
+	sl.wsListener = newProvisionedWSListener(ac, sl.provisionSession, sudoHandshakeTimeout)
 	sl.handleFrame = sl.handleMessage
+	sl.onConnected = sl.subscribe
 	return sl
+}
+
+// Err returns the error that stopped the listener before it ever subscribed.
+func (sl *SudoListener) Err() error {
+	sl.stateMu.Lock()
+	defer sl.stateMu.Unlock()
+	return sl.err
+}
+
+func (sl *SudoListener) provisionSession() (string, error) {
+	session, err := CreateEventSession(sl.ac)
+	if err != nil {
+		return "", sl.failIfFatal(err)
+	}
+
+	sl.stateMu.Lock()
+	sl.channelID = session.ChannelID
+	sl.stateMu.Unlock()
+
+	return session.WebsocketURL, nil
+}
+
+func (sl *SudoListener) subscribe() error {
+	sl.stateMu.Lock()
+	channelID := sl.channelID
+	sl.stateMu.Unlock()
+
+	if err := SubscribeEvent(sl.ac, channelID, EventTypeSudo, sl.sessionID); err != nil {
+		return sl.failIfFatal(err)
+	}
+
+	sl.stateMu.Lock()
+	sl.subscribed = true
+	sl.stateMu.Unlock()
+
+	return nil
+}
+
+// failIfFatal stops the listener when retrying cannot help: a 4xx before the first
+// subscribe means a bad session or an expired login, not an outage. Everything else
+// is left to the reconnect loop.
+func (sl *SudoListener) failIfFatal(cause error) error {
+	sl.stateMu.Lock()
+	fatal := !sl.subscribed && utils.IsFatalClientError(utils.HTTPStatusCode(cause))
+	if fatal {
+		sl.err = cause
+	}
+	sl.stateMu.Unlock()
+
+	if fatal {
+		sl.Stop()
+	}
+
+	return cause
 }
 
 func (sl *SudoListener) handleMessage(message []byte) {
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -3,6 +3,7 @@ package event
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"sync"
@@ -56,6 +57,8 @@ type SudoListener struct {
 	serverName string
 	sessionID  string
 	mfaMu      sync.Mutex // serializes handleSudoMFA so only one MFA flow runs at a time
+	// refreshToken is ac.RefreshToken, or nil when ac is nil. Replaced in tests.
+	refreshToken func() error
 
 	stateMu    sync.Mutex // guards channelID, subscribed, warned, err
 	channelID  string
@@ -64,13 +67,17 @@ type SudoListener struct {
 	err        error
 }
 
-// NewSudoListener creates a SudoListener but does not connect yet. ac may be
-// nil in tests; MFA request frames are then dropped instead of dereferencing it.
+// NewSudoListener creates a SudoListener but does not connect yet. ac may be nil
+// only for a test that never calls Start: MFA request frames are then dropped, but
+// provisioning a session on the listen goroutine would dereference it.
 func NewSudoListener(ac *client.AlpaconClient, serverName, sessionID string) *SudoListener {
 	sl := &SudoListener{
 		ac:         ac,
 		serverName: serverName,
 		sessionID:  sessionID,
+	}
+	if ac != nil {
+		sl.refreshToken = ac.RefreshToken
 	}
 	sl.wsListener = newProvisionedWSListener(ac, sl.provisionSession, sudoHandshakeTimeout)
 	sl.handleFrame = sl.handleMessage
@@ -79,7 +86,8 @@ func NewSudoListener(ac *client.AlpaconClient, serverName, sessionID string) *Su
 	return sl
 }
 
-// Err returns the error that stopped the listener before it ever subscribed.
+// Err returns the last reason the listener could not connect, so a caller that only
+// sees WaitConnected time out can report that instead of a bare timeout.
 func (sl *SudoListener) Err() error {
 	sl.stateMu.Lock()
 	defer sl.stateMu.Unlock()
@@ -112,6 +120,7 @@ func (sl *SudoListener) subscribe() error {
 	sl.subscribed = true
 	// A recovered channel means the next outage is a new one, worth announcing again.
 	sl.warned = false
+	sl.err = nil
 	sl.stateMu.Unlock()
 
 	return nil
@@ -125,13 +134,12 @@ func (sl *SudoListener) announceOutage(cause error) {
 // 4xx is a bad session or an expired login: the listener stops, carrying the reason for
 // websh to report, and says nothing itself. After it, the same failure is an outage
 // announced once, in CRLF because websh holds the terminal in raw mode, then left to
-// the reconnect loop.
+// the reconnect loop. Every cause is recorded either way, so a caller reading Err()
+// after a timeout gets the transport's own words rather than nothing.
 func (sl *SudoListener) fail(cause error) error {
 	sl.stateMu.Lock()
+	sl.err = cause
 	fatal := !sl.subscribed && isFatalRequestError(cause)
-	if fatal {
-		sl.err = cause
-	}
 	announce := sl.subscribed && !sl.warned
 	if announce {
 		sl.warned = true
@@ -140,12 +148,40 @@ func (sl *SudoListener) fail(cause error) error {
 
 	if fatal {
 		sl.Stop()
-	}
-	if announce {
-		_, _ = fmt.Fprintf(os.Stderr, "\r\n%s\r\n", utils.Yellow("Sudo MFA listener disconnected; retrying..."))
+		return cause
 	}
 
+	if announce {
+		select {
+		case <-sl.done:
+			// Stop cancels neither an in-flight dial nor a session create, so a
+			// failure landing after it must not claim anything is still retrying.
+		default:
+			_, _ = fmt.Fprintf(os.Stderr, "\r\n%s\r\n", utils.Yellow("Sudo MFA listener disconnected; retrying..."))
+		}
+	}
+
+	sl.refreshExpiredToken(cause)
+
 	return cause
+}
+
+// refreshExpiredToken renews the access token when cause is a 401. Past the first
+// subscribe nothing is fatal, so without this an access token that expires mid-session
+// is re-presented on every dial until websh exits and sudo MFA never comes back.
+func (sl *SudoListener) refreshExpiredToken(cause error) {
+	if sl.refreshToken == nil || utils.HTTPStatusCode(cause) != http.StatusUnauthorized {
+		return
+	}
+	// An MFA flow holds mfaMu across its own refresh; a second one would race it.
+	if !sl.mfaMu.TryLock() {
+		return
+	}
+	defer sl.mfaMu.Unlock()
+
+	// A failed refresh leaves the next attempt to the reconnect loop, which is
+	// where a recovery would land anyway.
+	_ = sl.refreshToken()
 }
 
 func (sl *SudoListener) handleMessage(message []byte) {
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -57,9 +57,10 @@ type SudoListener struct {
 	sessionID  string
 	mfaMu      sync.Mutex // serializes handleSudoMFA so only one MFA flow runs at a time
 
-	stateMu    sync.Mutex // guards channelID, subscribed, err
+	stateMu    sync.Mutex // guards channelID, subscribed, warned, err
 	channelID  string
 	subscribed bool
+	warned     bool
 	err        error
 }
 
@@ -74,6 +75,7 @@ func NewSudoListener(ac *client.AlpaconClient, serverName, sessionID string) *Su
 	sl.wsListener = newProvisionedWSListener(ac, sl.provisionSession, sudoHandshakeTimeout)
 	sl.handleFrame = sl.handleMessage
 	sl.onConnected = sl.subscribe
+	sl.onDialFailed = sl.announceOutage
 	return sl
 }
 
@@ -108,6 +110,8 @@ func (sl *SudoListener) subscribe() error {
 
 	sl.stateMu.Lock()
 	sl.subscribed = true
+	// A recovered channel means the next outage is a new one, worth announcing again.
+	sl.warned = false
 	sl.stateMu.Unlock()
 
 	return nil
@@ -129,6 +133,24 @@ func (sl *SudoListener) failIfFatal(cause error) error {
 	}
 
 	return cause
+}
+
+// announceOutage reports a dropped event channel once per outage. Before the first
+// subscribe the caller (websh) already handles the failure, so this stays quiet.
+// The message is wrapped in CRLF because websh holds the terminal in raw mode.
+func (sl *SudoListener) announceOutage(error) {
+	sl.stateMu.Lock()
+	quiet := !sl.subscribed || sl.warned
+	if !quiet {
+		sl.warned = true
+	}
+	sl.stateMu.Unlock()
+
+	if quiet {
+		return
+	}
+
+	_, _ = fmt.Fprint(os.Stderr, "\r\n\033[33mSudo MFA listener disconnected; retrying...\033[0m\r\n")
 }
 
 func (sl *SudoListener) handleMessage(message []byte) {
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -89,6 +89,7 @@ func (sl *SudoListener) Err() error {
 func (sl *SudoListener) provisionSession() (string, error) {
 	session, err := CreateEventSession(sl.ac)
 	if err != nil {
+		sl.announceOutage(err)
 		return "", sl.failIfFatal(err)
 	}
 
@@ -105,6 +106,7 @@ func (sl *SudoListener) subscribe() error {
 	sl.stateMu.Unlock()
 
 	if err := SubscribeEvent(sl.ac, channelID, EventTypeSudo, sl.sessionID); err != nil {
+		sl.announceOutage(err)
 		return sl.failIfFatal(err)
 	}
 
--- a/api/event/sudolistener.go
+++ b/api/event/sudolistener.go
@@ -28,6 +28,10 @@ const (
 	mfaPollingTimeout = 60 * time.Second
 
 	sudoHandshakeTimeout = 10 * time.Second
+
+	// sudoOutageWarning is asserted on by the tests, so it lives here rather than
+	// inline at the one place that prints it.
+	sudoOutageWarning = "Sudo MFA listener disconnected; retrying..."
 )
 
 // sudoMFAEvent represents the MFA request payload from the event WebSocket.
@@ -60,10 +64,11 @@ type SudoListener struct {
 	// refreshToken is ac.RefreshToken, or nil when ac is nil. Replaced in tests.
 	refreshToken func() error
 
-	stateMu    sync.Mutex // guards channelID, subscribed, warned, err
+	stateMu    sync.Mutex // guards channelID, subscribed, warned, refreshed, err
 	channelID  string
 	subscribed bool
 	warned     bool
+	refreshed  bool
 	err        error
 }
 
@@ -137,9 +142,19 @@ func (sl *SudoListener) announceOutage(cause error) {
 // the reconnect loop. Every cause is recorded either way, so a caller reading Err()
 // after a timeout gets the transport's own words rather than nothing.
 func (sl *SudoListener) fail(cause error) error {
+	// refreshToken is assigned before Start and never reassigned, so no lock here.
+	expired := sl.refreshToken != nil && utils.HTTPStatusCode(cause) == http.StatusUnauthorized
+
 	sl.stateMu.Lock()
 	sl.err = cause
 	fatal := !sl.subscribed && isFatalRequestError(cause)
+	// websh created its own session on this token moments ago, so a 401 on the first
+	// attempt means it expired in between and a refresh would fix it. One is enough:
+	// a 401 answering a token this listener just renewed is a real refusal.
+	if fatal && expired && !sl.refreshed {
+		sl.refreshed = true
+		fatal = false
+	}
 	announce := sl.subscribed && !sl.warned
 	if announce {
 		sl.warned = true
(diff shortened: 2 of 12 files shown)


api/event/sudolistener.go
```diff
@@ -47,25 +47,107 @@ type sudoMFAEvent struct {
...
+}
+
+// fail decides what a failed connect attempt means. Before the first subscribe a fatal
+// 4xx is a bad session or an expired login: the listener stops, carrying the reason for
+// websh to report, and says nothing itself. After it, the same failure is an outage
+// announced once, in CRLF because websh holds the terminal in raw mode, then left to
+// the reconnect loop.
+func (sl *SudoListener) fail(cause error) error {
+	sl.stateMu.Lock()
+	fatal := !sl.subscribed && isFatalRequestError(cause)
```
junho226, line 150: The post-subscribe rule has a hole: a 401 is retried forever with the token that produced it.

Once `subscribed` is set it is never cleared, so from the first successful subscribe onward `fatal` is permanently false and every status is retried—including 401. If the access token expires during a long interactive websh session, every later `CreateEventSession` answers 401, the loop retries at the 30s ceiling for the rest of the session, and sudo MFA never comes back. The user gets one "retrying..." line and then silence. Same shape at `commandoutput_listener.go:90`, though `execTimeout()` bounds it there.

The reasoning in the comment above is right for 5xx and for a transport drop—giving up would cost the run its live output. But a 401 is the one status where the retry provably cannot succeed until something changes, and the thing that would change it is already in this file: `handleSudoMFA` calls `sl.ac.RefreshToken()`. Attempting one refresh before the next dial when `utils.HTTPStatusCode(cause) == http.StatusUnauthorized`, and only then continuing to retry, would turn the dead loop into a recovery.

Not a blocker—this is still strictly better than main, where the channel never came back at all. But I would rather see it handled here than discovered from a "sudo stopped prompting halfway through my session" report.

api/event/sudolistener.go
```diff
@@ -47,25 +47,107 @@ type sudoMFAEvent struct {
...
+	if announce {
+		sl.warned = true
+	}
+	sl.stateMu.Unlock()
+
+	if fatal {
+		sl.Stop()
+	}
+	if announce {
+		_, _ = fmt.Fprintf(os.Stderr, "\r\n%s\r\n", utils.Yellow("Sudo MFA listener disconnected; retrying..."))
```
junho226: This can print after `Stop()`.

`Stop` closes `done` and closes the socket, but it cancels neither an in-flight `dialer.Dial` (up to `sudoHandshakeTimeout`, 10s) nor an in-flight `CreateEventSession` POST. When either finally fails, it lands in `fail`, `subscribed` is still true and `warned` may still be false, so the shutdown path can emit "Sudo MFA listener disconnected; retrying..." for a listener that is not retrying anything.

The window is narrow in websh, since the deferred `sl.Stop()` is followed almost immediately by process exit, but the message is wrong whenever it does land, and it lands on a terminal mid-restore. A `select { case <-sl.done: return cause; default: }` before the announce closes it.

api/event/sudolistener.go
```diff
@@ -47,25 +47,107 @@ type sudoMFAEvent struct {
...
+	channelID  string
+	subscribed bool
+	warned     bool
+	err        error
 }
 
 // NewSudoListener creates a SudoListener but does not connect yet. ac may be
 // nil in tests; MFA request frames are then dropped instead of dereferencing it.
-func NewSudoListener(ac *client.AlpaconClient, wsURL, serverName string) *SudoListener {
+func NewSudoListener(ac *client.AlpaconClient, serverName, sessionID string) *SudoListener {
```
junho226, line 78: The doc comment above is no longer true once `Start` is involved. With the listener provisioning its own session, a nil `ac` reaches `CreateEventSession` on the listen goroutine and dereferences it. Verified:

```
panic: runtime error: invalid memory address or nil pointer dereference
client.(*AlpaconClient).createRequest(0x0, ...) client.go:214
event.CreateEventSession(0x0?)                  subscribe.go:45
event.(*SudoListener).provisionSession(...)     sudolistener.go:90
```

Every current test that passes nil calls only `handleMessage`, `handleSudoMFA`, or `pollMFACompletion`, so nothing is broken today—but the comment invites the next test author to write `NewSudoListener(nil, ...)` plus `Start()`, and the panic surfaces on a background goroutine and takes the whole test binary with it. Worth narrowing to something like "ac may be nil only for tests that never call Start". `newProvisionedWSListener`'s "ac may be nil (empty header, for tests)" carries the same stale promise now that `provision` dereferences it.

cmd/websh/websh.go
```diff
@@ -347,49 +349,36 @@ func extractValue(args []string, i int) (string, int) {
...
+// sudoListenerWarning explains why the sudo MFA listener never connected, or returns ""
+// when the failure is not worth telling the user about: a 404 is an older server without
+// the events API, and the websh session runs fine without it. The server's own text is
+// sanitized because this command is about to put the terminal into raw mode.
+func sudoListenerWarning(cause error) string {
+	if utils.HTTPStatusCode(cause) == http.StatusNotFound {
+		return ""
+	}
+	if cause == nil {
+		return "Sudo MFA listener unavailable: timed out connecting to the event channel"
```
junho226: This branch now swallows the cause for the whole non-4xx class, which is most real outages.

`SudoListener` records into `sl.err` only on the fatal path, so a 500, a connection refused, a TLS failure, or a proxy 502 all leave `Err()` nil and land here as "timed out connecting to the event channel". On main those printed the server's or the transport's actual message. The retry behavior is a clear improvement—five seconds of attempts instead of one—but the diagnosability of the message went backward for exactly the failures an on-call engineer needs to read.

Recording the last cause in `fail` unconditionally, and keeping the `Stop()` decision gated on `fatal` as it is now, would restore it without touching the state machine: `Err()`'s contract widens from "what stopped it" to "the last reason it could not connect", which is what this caller actually wants.

junho226 (commented):
## Review summary

I traced the full reconnect cycle in `wsListener`, `SudoListener`, and `CommandOutputListener`, checked the outage/fatal split against what alpacon-server actually returns on `/api/events/sessions/` and `/api/events/subscriptions/`, ran the race suite, and mutation-tested the new assertions. The fix is correct and the structure is proportionate. Nothing here blocks the merge; the findings below are follow-ups, one of which I would like an answer on before it turns into a support ticket.

## What I verified

- `go test -race -count=1 ./...`—every package `ok`, no `FAIL`, no race report. `api/event` takes 72.7s.
- `go vet ./...`—clean. `golangci-lint` is not installed on this machine, so I am relying on the PR's own golangci-lint job, which passes.
- Mutation testing, six mutations, each reverted afterwards:
  - dropping `sl.onConnected` → `TestSudoListener_ResubscribesOnReconnect` fails on "a reconnect must re-issue the sudo subscription";
  - caching the URL after the first provision → `TestSudoListener_CreatesANewSessionOnReconnect` fails on "each attempt must create its own event session";
  - deleting `sl.warned = false` → `TestSudoListener_AnnouncesTheNextOutageAfterRecovering` fails with `expected: 2, actual: 1`, exactly as the PR body claims;
  - `fatal := false` in `SudoListener.fail` → `TestSudoListener_FirstSessionFailureStopsAndSurfaces` fails;
  - dropping the `!l.subscribed` guard in `CommandOutputListener.provisionSession` → `TestCommandOutputListener_KeepsRetryingAfterTheFirstSubscribe` fails;
  - dropping `l.onConnected` → `TestCommandOutputListener_CreatesANewSessionAndResubscribesOnReconnect` fails.
  - The one that stayed green: forcing `listenerFailure` to always return the flat timeout breaks no test. `TestListenerFailureFallsBackToTheConnectBudget` pins only the `Err() == nil` half, and `..._SurfacesAFatalSessionFailure` pins `Err()` in isolation, so nothing asserts end to end that a rejected session reaches `runCommandFallback` as the server's reason rather than as a timeout—which is the user-visible claim in the PR body. Three lines of function, so this is a note, not a demand.

## The state machine

Traced dial → provision → subscribe → drop → redial, and I could not find the classic failures:

- No goroutine or ticker per dial. `listenLoop` owns the only goroutine, `time.After` is the only timer, and both selects wake on `done`.
- No spin. Every iteration waits at least `reconnectBaseDelay` before redialing, including the reset path.
- No lost command output across a flap. Chunks published during an outage are recovered by the warm-fire read, `applyChunk`'s gap fill, and `drainRemainingChunks`, and `applyChunk` drops `chunk.Seq <= lastSeq`, so the duplicate-subscription window the PR documents costs a repeated chunk at worst, not a wrong one. Sudo MFA events during an outage are genuinely lost, which is what the new warning exists to say.
- Teardown ordering is right: `connectAndListen`'s deferred close clears `w.conn` and closes the socket before the next `provision()` runs, so the old session is never live alongside the new one.

## Findings

### The backoff resets on a connection that lasted zero milliseconds

`wsListener.listenLoop` (`api/event/wslistener.go:116`, unchanged by this PR, so no inline comment) resets `delay` whenever `connectAndListen` returns true, and true now means "the handshake succeeded and `onConnected` returned", not "the connection was worth anything". A server that accepts and immediately closes—a spent token, a capacity drop, a proxy that resets—keeps the loop pinned at `wsReconnectBaseDelay` forever.

Before this PR one such attempt cost one dial. Now it costs a `CreateEventSession` POST plus, for `SudoListener`, a `SubscribeEvent` POST, so the same flap is three requests per second for the life of an interactive websh session, and `DEFAULT_THROTTLE_CLASSES` on alpacon-server only throttles service-token auth, so a browser-login user is not bounded by anything. Gating the reset on a minimum connection lifetime—reset only if the connection outlived the base delay—would close it in about four lines. Worth a follow-up issue rather than this PR.

Related and smaller: `nextReconnectDelay` has no jitter, and the loop ignores `Retry-After` even though `utils.RetryAfter` exists and `isFatalRequestError` deliberately keeps 429 retryable.

### Every dial leaves a row behind on the server

`CreateEventSession` creates an `EventSession` and its `EventUserChannel`, and the CLI never issues the DELETE. On alpacon-server `EventSession.close()` runs only from `perform_destroy`, and no beat entry prunes them—`delete_old_events` covers `Event`, not sessions. The leak predates this PR, but its size used to be one row per websh session and is now one row per dial. Flagging it so somebody owns the cross-repo half; nothing to change here.

### Security

No concerns, and one thing worth stating because it is easy to get wrong: the WebSocket header carries no credential at all (`SetWebsocketHeader` sets only `Origin` and `User-Agent`), the credential lives in the per-dial provisioned URL, and `CreateEventSession` goes through `setHTTPHeader`, which reads `ac.AccessToken` at request time. So a token rotated by `RefreshToken` is picked up on the very next dial. Nothing logs the URL or the token: the outage notice is a fixed literal, and the websh warning runs the server's text through `utils.SanitizeTerminalText` before it reaches a terminal that is about to go raw.

### YAGNI

The 728 added lines are not 728 lines of machinery. Production code is +229/-112 across six files; the other ~499 lines are tests. Deleting `newWSListener` outright instead of keeping it "in case something wants a fixed URL" is the right call—it was the only way back into this bug. The one repeated block, `stateMu`/`channelID`/`subscribed`/`err` in three listeners, is correctly left alone: the notification differs in all three cases, and the judgment that actually repeated (the 4xx test) is the part that got extracted.

## Nits

- `setupSudoListener(ac, sessionID, serverName)` calls `event.NewSudoListener(ac, serverName, sessionID)`. The order is correct, but two adjacent string parameters swapping across one call boundary is a swap waiting to happen, and no test would catch it—the `TestSetupSudoListener_*` cases never reach a subscription, so neither value is ever asserted on.

api/event/sudolistener.go
jisung-02: Fixed in d71c192.

`SudoListener.fail` now calls `refreshExpiredToken`, which renews the access token when `utils.HTTPStatusCode(cause) == http.StatusUnauthorized` and then lets the reconnect loop carry on. It holds `mfaMu` via `TryLock`, so it never runs alongside the refresh `handleSudoMFA` already does, and never blocks the listen goroutine behind a 60s MFA poll. A failed refresh is left to the next attempt, which is where a recovery would land anyway.

Two tests pin both halves: `TestSudoListener_RefreshesTheAccessTokenOnUnauthorized` (subscribe, then 401 on every later session create) and `TestSudoListener_LeavesTheTokenAloneOnOtherStatuses` (same shape with a 500, refresh count must stay zero). Both were mutation-tested: dropping the call reds the first, inverting the status check reds the second.

Left `commandoutput_listener.go:90` alone. An expired token there fails the polling fallback too, so the run is broken regardless of the listener, and `execTimeout()` already bounds it. Happy to extend it if you would rather have the symmetry.

api/event/sudolistener.go
jisung-02: Fixed in d71c192. The announce branch now runs `select { case <-sl.done: default: ... }`, so a dial or session create that fails after `Stop` says nothing.

`TestSudoListener_StaysQuietWhenAFailureLandsAfterStop` drives it directly: mark the listener subscribed, `Stop`, then call `fail` with a synthetic error and assert stderr stays clean. Removing the guard reds it with the exact string.

api/event/sudolistener.go
jisung-02: Fixed in fc7e204 and d71c192. Three comments carried that stale promise, not two.

- `NewSudoListener`: "ac may be nil only for a test that never calls Start: MFA request frames are then dropped, but provisioning a session on the listen goroutine would dereference it."
- `newProvisionedWSListener`: "A nil ac only means an empty header here; whether it is safe at all is provision's call, since every attempt runs it."
- `NewCommandOutputListener` said the same thing and has the same `provisionSession`, so it got the same narrowing.

Kept it to the comment rather than guarding `provisionSession`, since a nil client reaching the network layer is a programming error and a silent early return would only hide it.

cmd/websh/websh.go
jisung-02: Fixed in d71c192, exactly as you suggested: `fail` records `sl.err` unconditionally and the `Stop` decision stays gated on `fatal`. A successful subscribe clears it, so `Err()` means "the last reason it could not connect" rather than a stale one from an outage the listener recovered from.

`TestSudoListener_SurfacesANonFatalCauseAfterTheWaitEnds` covers it: session create answers 500 forever, `WaitConnected` runs out, and `Err()` has to carry the server's message. Recording only the fatal cause reds it with "An error is expected but got nil." `TestSudoListener_SessionCreateRetryableStatusIsRetried` already asserted `NoError(sl.Err())` after a 503 recovery, so it doubles as the guard on the clear.

junho226 (commented):
## Delta re-review, `bb54719f..d4f1957e`

Four of my five findings are closed, and I verified each one in the files at head rather than from the commit messages. The backoff finding is untouched and stays open. The delta also opens one new question that I did not have to ask before this PR, because before it the listen goroutine issued no HTTP at all.

I did not re-derive what the first pass already settled: the reconnect state machine, the goroutine and subscription accounting, the teardown ordering, the security read of the per-dial URL, and the YAGNI verdict on the production/test line split. None of the three new commits touches any of that.

## Closed

**The mid-stream 401 (`sudolistener.go`).** `fail` now calls `refreshExpiredToken`, which renews the token exactly when `utils.HTTPStatusCode(cause)` is 401 and leaves it alone otherwise. It cannot hot-loop: a 401 comes from `provisionSession` or `subscribe`, both of which make `connectAndListen` return false, so the backoff keeps doubling to the 30s ceiling and a refresh that keeps failing costs one extra request per attempt at that cadence, not a spin. Taking `mfaMu` with `TryLock` is the right shape—it honors the invariant the struct doc already states and refuses to queue behind a 60s MFA poll, and the next failure retries anyway.

One scoping note rather than a request. The fix landed on `SudoListener` only. `CommandOutputListener` still retries a post-subscribe 401 forever with the same token, but `execTimeout()` bounds it, so I agree it does not need the same treatment. `Watcher` has the identical shape and no bound—`alpacon event watch` runs until Ctrl+C—but it predates this PR (the only change here is `isFatalRequestError`), and it at least prints the reason through `ReconnectFailed()` instead of going silent. Follow-up material, not this PR's.

**Non-4xx causes swallowed into a generic timeout.** `fail` records `sl.err = cause` unconditionally and `subscribe` clears it on success, so `sudoListenerWarning` gets the transport's own words. The 404 suppression is unaffected: a 404 is always fatal, so it still stops the listener before any later cause can overwrite it, and a gorilla dial error carries no status code, so nothing else can masquerade as one.

**The outage notice printing after `Stop()`.** The `select` on `sl.done` closes it, and `done` is built in `newProvisionedWSListener`, so `Stop()` before `Start()` is safe and the new test exercises exactly that path.

**The nil-client promise.** Narrowed in all three places—`NewSudoListener`, `NewCommandOutputListener`, and `newProvisionedWSListener`, which now correctly hands the question to `provision` instead of answering it. The panic is still reachable; the comment no longer invites it, which is what I asked for.

**The swapped adjacent string params.** Verified the swap is not silent: `websh.go:311` moved to `(alpaconClient, serverName, session.ID)` against the new `(ac, serverName, sessionID)` signature, and all three `websh_test.go` call sites moved with it. Values still map correctly at every site, and `grep` finds no others. I also confirmed the premise of the original nit still holds—swapping the two literals back in the test call sites leaves `TestSetupSudoListener_*` green—so matching the constructor's order is the mitigation, by construction rather than by coverage. That is the right call; a test that pins argument identity here would be testing the compiler.

## Still open

**The backoff resets on a connection that lasted zero milliseconds.** `wslistener.go:117` is unchanged; only its constructor's doc comment moved. A server that accepts, subscribes, and immediately closes still makes `connectAndListen` return true, so `delay` snaps back to `reconnectBaseDelay` forever and the flap costs a `CreateEventSession` POST plus a `SubscribeEvent` POST every second, unbounded for a browser-login user. The delta does not make it worse—that flap's cause carries no status code, so it never triggers the new refresh—but it does not make it better either. Follow-up issue, as before.

## New

**The listen goroutine now reads `ac.AccessToken` without synchronization.** Detail inline at `sudolistener.go:98`. This is the one thing in the delta I would not have raised on `bb54719f`, because it follows from the PR's core design rather than from the three new commits.

**A pre-subscribe 401 still hard-stops without trying the refresh that is now right there.** Inline at `sudolistener.go:151`. Narrow window, small note.

**`refreshExpiredToken`'s `TryLock` branch has no coverage.** Replacing the guard with an unconditional refresh leaves the whole `api/event` suite green. It is a correct defensive branch and I am not asking for a test that would need a 60s MFA poll to run concurrently with a reconnect—just recording that nothing pins it.

## What I ran

- `go test -race -count=1 ./...`—every package `ok`, no `FAIL`, no race report. `api/event` takes 74.1s.
- `go vet ./...`—clean. `golangci-lint` is still not installed here, so I am relying on the PR's own job.
- Mutation testing on the four new assertions, each reverted afterwards, and the tree is clean:
  - deleting the `sl.refreshExpiredToken(cause)` call → `TestSudoListener_RefreshesTheAccessTokenOnUnauthorized` fails;
  - refreshing on 500 instead of 401 → both `..._RefreshesTheAccessTokenOnUnauthorized` and `..._LeavesTheTokenAloneOnOtherStatuses` fail, so the status check is pinned in both directions;
  - dropping the `<-sl.done` guard → `TestSudoListener_StaysQuietWhenAFailureLandsAfterStop` fails;
  - restoring the fatal-only `sl.err` assignment → `TestSudoListener_SurfacesANonFatalCauseAfterTheWaitEnds` fails.
- Two extra probes: deleting `sl.err = nil` from `subscribe` fails `TestSudoListener_SessionCreateRetryableStatusIsRetried`, so the clear-on-recovery half is pinned too; removing the `TryLock` guard changes nothing, as noted above.

## Nits, carried forward unchanged

`nextReconnectDelay` still has no jitter and the loop still ignores `Retry-After` even though `utils.RetryAfter` exists and `isFatalRequestError` deliberately keeps 429 retryable. `EventSession` and `EventUserChannel` rows still accumulate one per dial with no server-side prune; that half lives in alpacon-server and needs an owner.

api/event/sudolistener.go
```diff
@@ -47,25 +48,142 @@ type sudoMFAEvent struct {
...
+// Err returns the last reason the listener could not connect, so a caller that only
+// sees WaitConnected time out can report that instead of a bare timeout.
+func (sl *SudoListener) Err() error {
+	sl.stateMu.Lock()
+	defer sl.stateMu.Unlock()
+	return sl.err
+}
+
+func (sl *SudoListener) provisionSession() (string, error) {
+	session, err := CreateEventSession(sl.ac)
```
junho226, line 103: This is the first thing in the CLI to issue an HTTP request from the listen goroutine, and it reads a field another goroutine writes without a lock.

`client.AlpaconClient.AccessToken` is a plain string. `RefreshToken` writes it at `client.go:406` and `createRequest` reads it at `client.go:204` to build the `Authorization` header. Until now every writer and every reader inside `SudoListener` sat under `mfaMu`: `handleSudoMFA` holds it across its own two `RefreshToken` calls, across `pollMFACompletion`, and across `verifySudoGrant`, and on `main` the listen goroutine dialed a URL that was provisioned before `Start`, so it made no requests at all.

After this PR every dial runs `CreateEventSession` here and `SubscribeEvent` in `subscribe`, both on the listen goroutine, and neither takes `mfaMu`. An MFA flow holds that lock for up to `mfaPollingTimeout` (60s) while the reconnect loop redials every 1-30s, so a drop during an MFA flow puts an unsynchronized read next to `handleSudoMFA`'s write. A torn string header yields a truncated or overlong `Bearer` value—usually a 401, which the new `refreshExpiredToken` then recovers from, so this mostly self-heals; the overlong case reads past the allocation.

`-race` will not see it: the new tests substitute `sl.refreshToken`, so no test drives a real refresh concurrently with a reconnect.

Not a blocker and not, I think, fixable here—the field belongs to `client`, and taking `mfaMu` around the provisioning call would park the reconnect behind a 60s poll, which is worse. A `sync.RWMutex` (or `atomic.Pointer[string]`) behind an `ac.accessToken()` accessor in `client` would close it for every caller at once. Worth an issue while the reasoning is fresh.

api/event/sudolistener.go
```diff
@@ -47,25 +48,142 @@ type sudoMFAEvent struct {
...
+	fatal := !sl.subscribed && isFatalRequestError(cause)
+	announce := sl.subscribed && !sl.warned
+	if announce {
+		sl.warned = true
+	}
+	sl.stateMu.Unlock()
+
+	if fatal {
+		sl.Stop()
+		return cause
```
junho226, line 168: Small one, and the mirror of the fix directly below: this early return means a 401 *before* the first subscribe never reaches `refreshExpiredToken`.

The return itself is right—it is behavior-preserving, since `announce` requires `subscribed` and `fatal` requires `!subscribed`, so the old fall-through could never announce anyway—and it makes the two outcomes read as two outcomes. But the pre-subscribe 401 is the same expired-token story the commit is about, and it now ends in "Sudo MFA listener unavailable: ..." with no sudo MFA for the rest of the websh session, while a refresh that would fix it sits four lines away.

The window is narrow: `setupSudoListener` runs milliseconds after `CreateWebshSession` succeeded on the same token, so the token expiring in between is unlikely. Mentioning it because someone reading `refreshExpiredToken`'s comment—"an access token that expires mid-session is re-presented on every dial"—will reasonably assume it covers the first dial too, and it does not.

api/event/sudolistener.go
jisung-02: Filed as #397, with your reasoning carried over: the field belongs to `client`, and taking `mfaMu` around the provisioning call would park the reconnect behind a 60s poll, which is worse than the race. The proposed shape there is an `RWMutex` or `atomic.Pointer[string]` behind an `ac.accessToken()` accessor, so it closes for every caller at once instead of per call site.

An independent review pass on the same diff found the same window from the other direction, which is a useful confirmation that it is reachable rather than theoretical.

Also filed #398 for the `Watcher` half of your scoping note: `alpacon event watch` runs until Ctrl+C, so its post-subscribe 401 has no bound at all. Sequenced after #397, since a shared accessor changes how the refresh should be serialized.

api/event/sudolistener.go
jisung-02: You are right that it reads as covering the first dial, so I closed the gap rather than narrowing the comment. Fixed in 01ec2cf.

`fail` now carries a `refreshed` flag under `stateMu`. A 401 before the first subscribe renews the token once and the attempt stops counting as fatal; a 401 answering a token this listener just renewed is a real refusal and still stops it. Post-subscribe behavior is unchanged, since nothing is fatal there and the refresh already ran on every 401.

`TestSudoListener_TriesOneRefreshBeforeGivingUpOnAnUnauthorizedFirstSession` pins both halves: exactly one refresh, exactly two session creates, then `stopped` closes with `Err()` set. Mutation-tested in both directions. Dropping the pre-subscribe refresh leaves session creation at one attempt; dropping the one-shot guard never stops the listener and the test times out.

On your other two notes: the `TryLock` branch still has no test, as you said it need not. The backoff finding stays open in #394.

api/event/sudolistener.go
```diff
@@ -47,25 +52,153 @@ type sudoMFAEvent struct {
...
+// Err returns the last reason the listener could not connect, so a caller that only
+// sees WaitConnected time out can report that instead of a bare timeout.
+func (sl *SudoListener) Err() error {
+	sl.stateMu.Lock()
+	defer sl.stateMu.Unlock()
+	return sl.err
+}
+
+func (sl *SudoListener) provisionSession() (string, error) {
+	session, err := CreateEventSession(sl.ac)
```
junho226, line 103: Re-anchoring my `d4f1957e` note, because `client/client.go` is not in this PR's diff at all: `AlpaconClient.AccessToken` is still a plain string, written at `client.go:406` and read at `client.go:205` with no mutex, accessor, or atomic.

`CreateEventSession` here—and `SubscribeEvent` in `subscribe` below—run on the listen goroutine and go through `setHTTPHeader`, so both read that field. On `main` the listen goroutine dialed a URL provisioned before `Start` and issued no requests at all, and every reader and writer inside `SudoListener` sat under `mfaMu`: `handleSudoMFA` holds it across its own two `RefreshToken` calls, across `pollMFACompletion`, and across `verifySudoGrant`. An MFA flow holds it for up to `mfaPollingTimeout` (60s) while the reconnect loop redials every 1-30s, so a drop during an MFA flow puts an unsynchronized read next to that write. A torn string header yields a truncated or overlong `Bearer` value—usually a 401, which `refreshExpiredToken` then recovers from, so it mostly self-heals; the overlong case reads past the allocation.

`01ec2cf` widens this slightly rather than narrowing it: `refreshExpiredToken` now calls `sl.refreshToken` (`= ac.RefreshToken`) from this same goroutine, so the listener both reads and writes `ac.AccessToken` here. `mfaMu`'s `TryLock` keeps the two *refreshes* from overlapping, which is what its comment claims and all it claims; it does nothing for the reads in `setHTTPHeader`.

`-race` still will not see it: the tests substitute `sl.refreshToken`, so nothing drives a real refresh concurrently with a reconnect.

Not a blocker and not, I think, fixable here—the field belongs to `client`, and taking `mfaMu` around the provisioning call would park the reconnect behind a 60s poll, which is worse. A `sync.RWMutex` (or `atomic.Pointer[string]`) behind an `ac.accessToken()` accessor in `client` would close it for every caller at once. Worth an issue while the reasoning is fresh.

api/event/commandoutput_listener.go
```diff
@@ -49,19 +57,102 @@ type commandOutputEnvelope struct {
...
+	return l.subscribe()
+}
+
+func (l *CommandOutputListener) subscribe() error {
+	// A fatal session failure can stop the listener between the caller's WaitConnected
+	// and its subscribeTo. Subscribing then succeeds against a channel nothing reads,
+	// so the caller would stream from a dead listener instead of falling back.
+	select {
+	case <-l.done:
+		return errors.New("event listener stopped before the subscription")
```
junho226: Two small things about these four lines, neither blocking.

**The reason is dropped.** In the case the comment describes, the listener stopped because `provisionSession` hit a fatal 4xx, and it recorded that error in `l.err` on the way out. `streamSubscribed` passes this error straight to `runCommandFallbackFromID`, which prints it: the user gets `real-time output unavailable (event listener stopped before the subscription); falling back to polling` instead of the server's own 401 or 403. Both sibling paths avoid exactly this by routing through `listenerFailure(listener)`. Swapping this for `listenerFailure` outright would be wrong—on the `SubscribeEvent` error path below, `l.err` is already nil from the last successful provision, so it would report a connect timeout that did not happen. Folding it in only here would not:

```go
if cause := l.Err(); cause != nil {
	return fmt.Errorf("event listener stopped before the subscription: %w", cause)
}
return errors.New("event listener stopped before the subscription")
```

**The window is narrowed, not closed.** `Stop` can still land between this check and `SubscribeEvent`'s return, and the caller then streams from a dead listener exactly as before. The narrowing is worth having—the window it closes spans `SubmitCommand`'s whole round trip, and the one left is a single HTTP call wide—but a reader could take the comment as a guarantee. Re-reading `done` after `SubscribeEvent` succeeds and before `l.subscribed = true` would shrink it further for four more lines; closing it properly means `streamSubscribed` selecting on the listener's liveness in its main loop, which is a bigger change than this PR wants.

Worth saying what the residual costs, since it bounds how much either is worth: on a dead listener the run still ends through `pollResult` and `finish`, and `drainRemainingChunks` writes everything. What is lost is liveness (output arrives in one block at the end) and the fallback warning—not output.

api/event/sudolistener.go
```diff
@@ -47,25 +52,153 @@ type sudoMFAEvent struct {
...
+	// refreshToken is assigned before Start and never reassigned, so no lock here.
+	expired := sl.refreshToken != nil && utils.HTTPStatusCode(cause) == http.StatusUnauthorized
+
+	sl.stateMu.Lock()
+	sl.err = cause
+	fatal := !sl.subscribed && isFatalRequestError(cause)
+	// websh created its own session on this token moments ago, so a 401 on the first
+	// attempt means it expired in between and a refresh would fix it. One is enough:
+	// a 401 answering a token this listener just renewed is a real refusal.
+	if fatal && expired && !sl.refreshed {
```
junho226, line 156: Robustness note, not a request—the invariant holds today.

`sl.refreshed` is spent here, before `refreshExpiredToken` has had a chance to run, so the allowance is burned even when no refresh happens. Two ways that can be true: `sl.refreshToken()` returns an error (swallowed by design), or `mfaMu.TryLock` fails and the function returns without refreshing.

The second one is currently unreachable, and that is load-bearing: `mfaMu` is only held by `handleSudoMFA`, which only runs on a frame, and frames are only read after `onConnected` returned nil—which is what sets `subscribed`. Since this branch requires `!subscribed`, no MFA flow can be in flight when it runs. So the allowance is never silently burned in practice.

It is worth a word in the comment, though, because the branch reads as if it were guarding the refresh when it is really guarding the downgrade, and the reason those coincide lives two files away in `connectAndListen`'s ordering. Something like "spent whether or not the refresh lands: a refresh that fails leaves the retry presenting the same token, and that second 401 is the one that stops us" would make the intent local.

junho226 (commented):
## Delta re-review, `d4f1957e..6c85ff25`

Four commits landed here, not three: `01ec2cf` ("try one token refresh before giving up on the first session") sits between `d4f1957e` and `44d0a743`, and it is the one that closes a finding. As before I verified each item in the files at head rather than from the commit messages.

**Closed:** the pre-subscribe 401 low.
**Still open:** the `ac.AccessToken` race, the zero-millisecond backoff reset, and the `TryLock` coverage nit—all three untouched by the delta.
**New:** two lows, both on `44d0a743`. Nothing blocking.

I did not re-derive what the earlier passes settled: the reconnect state machine, the goroutine and subscription accounting, the param-order call sites, and the YAGNI verdict on the production/test line split. None of the four commits touches any of that.

## Closed

**A pre-subscribe 401 now reaches the refresh (`01ec2cf`).** `fail` computes `expired` before taking `stateMu`, and when a 4xx would be fatal before the first subscribe and it is a 401 and the allowance is unspent, it flips `fatal` to false, spends the allowance, and lets the reconnect loop take the retry. The shape is right in three ways I checked rather than assumed:

- **The refresh happens before the retry, not after it.** `refreshExpiredToken` runs at the tail of `fail`, on the listen goroutine, and `fail`'s return propagates through `provisionSession` to `connectAndListen`, which returns false and parks the loop in its backoff select. The token is renewed before the next dial reads it.
- **The allowance cannot leak into the mid-session case.** `refreshed` is only consulted under `fatal`, and `fatal` requires `!subscribed`, which latches true at the first successful subscribe and never goes back. A post-subscribe 401 still refreshes on every occurrence, which is what the mid-session recovery needs.
- **It cannot loop.** The second 401 takes the `if fatal` early return, so it stops without even attempting a second refresh. The connect budget absorbs the extra round: `setupSudoListener` waits 5s, and the retry costs one refresh plus one `reconnectBaseDelay` (1s).

The new test pins both directions—I confirmed by mutation, below.

## Still open

**The listen goroutine still reads `ac.AccessToken` unsynchronized.** Untouched, and not deferred so much as simply not addressed: `client/client.go` does not appear in this PR's diff at all. The field is still a plain string, still written at `client.go:406` by `RefreshToken` and read at `client.go:205` by `setHTTPHeader`, with no mutex, accessor, or atomic anywhere on `AlpaconClient`. Everything in my original note holds unchanged, and `01ec2cf` adds one more writer on that goroutine—`refreshExpiredToken` now calls `sl.refreshToken` (`= ac.RefreshToken`) from the listen goroutine, so the listener both reads and writes the field there while `handleSudoMFA` may hold `mfaMu` across its own two writes on another. `mfaMu`'s `TryLock` keeps the two *refreshes* apart; it does nothing for the reads in `setHTTPHeader`. Detail re-anchored inline. Still `client`'s to fix, still worth an issue.

**Backoff resets on a connection that lasted zero milliseconds.** `wslistener.go:117` is byte-for-byte what it was at `bb54719f`. A server that accepts, subscribes, and immediately closes still makes `connectAndListen` return true, so `delay` snaps back to `reconnectBaseDelay` forever. Follow-up issue, as before.

**`refreshExpiredToken`'s `TryLock` branch has no coverage.** Re-verified at head: deleting the guard and its `defer` leaves the whole `./api/event` suite green (73.0s, `ok`). Recording it, not asking for a test that would need a 60s MFA poll running concurrently with a reconnect.

## `44d0a743`, reviewed as new production logic

The change is a `done` check at the top of `CommandOutputListener.subscribe`. Tracing it out:

*What "halted" means.* `l.done` is closed in exactly one place, `wsListener.Stop`, under `closeOnce`. Three callers reach it while a stream is being set up: `runCommandStreamingWithWriter` and `StreamApprovedCommand` on their own failure paths, and—the one that matters—`provisionSession`'s fatal branch, which runs on the listen goroutine and can therefore fire at any point in the caller's `SubmitCommand` round trip.

*Where subscription state is read.* `l.subscribed` under `stateMu` is only ever read by `provisionSession` to decide fatality; the caller never sees it. What the caller actually acts on is `subscribeTo`'s return value, and that is what was lying: on a stopped listener, `SubscribeEvent` still succeeds against the server (the channel is real, the token was valid), so `subscribe` returned nil and `streamSubscribed` went on to select over a `Chunks()` channel nothing would ever write to. The check fixes the right thing.

*Can it report a live subscription as halted?* I do not think so. The check reads `done`, and `done` closed means `listenLoop` returns at its next iteration guard and `connectAndListen`'s read loop returns on its own `done` select—so by the time the check can observe it, there is no live subscription left to misreport. The one behavior change on the `onConnected` path is that a `Stop` racing an in-flight dial now makes `connectAndListen` return false instead of closing `connected`; that *removes* a coin flip rather than adding one, since `WaitConnected` selects over `connected` and `done` together and Go picks randomly when both are ready.

*Race between the halt and the subscribe path.* There is a residual one, inline below. It is a narrowing, not an elimination, and the blast radius is small enough that I would not hold the PR for it.

## New

**The stopped-listener error discards the reason the listener stopped.** Inline at `commandoutput_listener.go:121`.

**The check-to-subscribe window is still open.** Same inline comment—same four lines, so I kept them together.

## What I ran

- `go test -race -count=1 ./...`—every package `ok`, no `FAIL`, no race report. `api/event` takes 74.6s.
- `go vet ./...` and `gofmt -l .`—both clean. `golangci-lint` is still not installed here, so I am relying on the PR's own job.
- Mutation testing on the two new assertions, each reverted afterwards, with `git status` clean at the end:
  - deleting the `<-l.done` guard from `subscribe` → `TestCommandOutputListener_SubscribeFailsOnAStoppedListener` fails on "An error is expected but got nil";
  - deleting the `fatal = false` downgrade → `TestSudoListener_TriesOneRefreshBeforeGivingUpOnAnUnauthorizedFirstSession` fails on both counters (`refreshes` 0, `sessions` 1);
  - relaxing `!sl.refreshed` to an unbounded downgrade → the same test fails on the stop assertion, so the *bound* is pinned and not just the refresh.
- Re-probed the `TryLock` guard, as noted above: still nothing pins it.

## Small notes, not requests

`b88888e` is accurate: `provisionSession` is the only writer of `l.err`, it clears to nil on a successful session create, and `subscribe`'s `SubscribeEvent` error deliberately does not land there—so "the last error a session attempt recorded, or the connect budget when no attempt got that far" says exactly what `Err()` holds. `6c85ff2` is behavior-preserving—the 404 early return is untouched, and `cause == nil` still produces the same sentence.

One robustness note on `01ec2cf`, inline at `sudolistener.go:154`.

## Nits, carried forward unchanged

`nextReconnectDelay` still has no jitter and the loop still ignores `Retry-After` even though `utils.RetryAfter` exists and `isFatalRequestError` deliberately keeps 429 retryable. `EventSession` and `EventUserChannel` rows still accumulate one per dial with no server-side prune; that half lives in alpacon-server and needs an owner. `commandoutput_listener.go` and `api/event/watcher.go` still share the unbounded-retry shape, and the watcher predates this PR.

api/event/commandoutput_listener.go
jisung-02: Both halves fixed in 72f44ce.

The reason is carried now. `subscribe` routes its halted check through a `halted()` helper that wraps `l.Err()`: `event listener stopped while subscribing: <the server's own refusal>`. Kept it local as you said rather than swapping in `listenerFailure`, because the `SubscribeEvent` error path below never reaches `halted()`, so it cannot name a connect timeout that did not happen.

The window is narrowed further, not claimed closed. `halted()` runs again once both subscriptions are in and before `l.subscribed = true`, so the residual is a `Stop` landing after that second check rather than anywhere in the two round trips. The comment now says it narrows rather than closes, and names what closing it properly would take—`streamSubscribed` selecting on the listener's liveness in its main loop.

Two tests, both mutation-tested. Dropping the `%w` wrap reds `TestCommandOutputListener_SubscribeCarriesWhyTheListenerStopped` on "Target error should be in err chain"; dropping the second `halted()` reds `TestCommandOutputListener_SubscribeFailsWhenTheListenerStopsMidSubscription` on "An error is expected but got nil". The second one drives the real window: the subscription handler calls `Stop` before answering 201.

Your scoping note is what made this cheap to judge—output still arrives through `drainRemainingChunks` either way, so what these two buy is liveness and the fallback warning, and neither was worth a bigger change.

api/event/sudolistener.go
jisung-02: Taken as written, in a5f1a0f. The comment now says the allowance is spent whether or not the refresh lands, because what it bounds is the downgrade rather than the refresh, and that a refresh which fails leaves the retry presenting the same token—so the second 401 is the one that has to stop us.

Left the mechanism alone. Spending it at the downgrade is what keeps a failing refresh from retrying the same token forever, so the coincidence you traced is worth relying on rather than coding around. Recorded that as a `Directive` trailer on the commit so the next person moving this line sees why it sits before `refreshExpiredToken` and not after.

api/event/sudolistener.go
jisung-02: No code change here, and I agree with your read that there should not be one in this PR.

This is tracked as #397, filed after your `d4f1957e` note with the reasoning carried over: the field belongs to `client`, and taking `mfaMu` around the provisioning call would park the reconnect behind a 60s poll, which is worse than the race. The proposed shape there is an `RWMutex` or `atomic.Pointer[string]` behind an `ac.accessToken()` accessor, so it closes for every caller at once instead of per call site.

One correction to the issue rather than to the code: your re-anchor is right that `01ec2cf` widens it, since `refreshExpiredToken` now writes `ac.AccessToken` from the listen goroutine too. #397 was filed before that commit and describes the read side only, so I have added the write side to it.

#398 covers the `Watcher` half—`alpacon event watch` runs until Ctrl+C, so its post-subscribe 401 has no bound at all—sequenced after #397, since a shared accessor changes how the refresh should be serialized.

api/event/sudolistener.go
```diff
@@ -47,25 +52,155 @@ type sudoMFAEvent struct {
...
+// Err returns the last reason the listener could not connect, so a caller that only
+// sees WaitConnected time out can report that instead of a bare timeout.
+func (sl *SudoListener) Err() error {
+	sl.stateMu.Lock()
+	defer sl.stateMu.Unlock()
+	return sl.err
+}
+
+func (sl *SudoListener) provisionSession() (string, error) {
+	session, err := CreateEventSession(sl.ac)
```
junho226, line 103: Re-anchoring the open medium so it does not get lost in the delta: it did not move.

`client/types.go:12` still declares `AccessToken string`, `client/client.go:204-205` still reads it in `setHTTPHeader`, `client/client.go:406` still writes it in `RefreshToken`, and `loadOnce` is still the only `sync` member on `AlpaconClient`. `client/client.go` does appear in `6c85ff25..a5f1a0f2`, but only through the `origin/main` merge—PR #392's `checkPrivilegesURL` to `currentUserURL` rename—so that churn is not movement on this.

`CreateEventSession` here, and `SubscribeEvent` in `subscribe`, run on the listen goroutine and go through `setHTTPHeader`, so both read the field while `handleSudoMFA` may hold `mfaMu` across its own two `RefreshToken` calls on another goroutine. `refreshExpiredToken` adds a write on this same goroutine. `mfaMu`'s `TryLock` keeps the two refreshes apart, which is all it claims; it does nothing for the reads. `-race` still cannot see it, because the tests substitute `sl.refreshToken`.

Unchanged conclusion: not a blocker, not fixable here—the field belongs to `client`, and taking `mfaMu` around the provisioning call would park the reconnect behind a sixty-second poll. A `sync.RWMutex` or `atomic.Pointer[string]` behind an `ac.accessToken()` accessor in `client` closes it for every caller at once. Worth an issue.

junho226 (commented):
## Delta re-review, `6c85ff25..a5f1a0f2`

Three commits: the `origin/main` merge (`74776ab8`) plus the branch's own `72f44ce6` and `a5f1a0f2`. As before I checked each item in the files at head rather than from the commit titles.

**Closed:** the discarded-reason low. The check-to-subscribe low too, though as documented rather than eliminated—see below.
**Still open, both untouched:** the `ac.AccessToken` race and the zero-millisecond backoff reset.
**Still open, nit:** `refreshExpiredToken`'s `TryLock` branch has no coverage.
**New:** nothing. One precision note on the wording `a5f1a0f2` added, no action.

I did not re-derive what earlier rounds settled: the reconnect state machine, the goroutine and subscription accounting, the param-order call sites, and the YAGNI verdict on the production/test line split. Nothing in this delta touches any of them.

## The merge is clean

Merge base `03cf8180`; the main side is PR #392 (`c546da6`, `bd6be5f`, `d60a2f0`)—`checkPrivilegesURL` to `currentUserURL` with the trailing slash, and `CheckPrivilegesResponse` to `CurrentUserResponse`.

- `git diff-tree --cc 74776ab8` prints no hunks at all, so nothing was resolved by hand.
- Replayed both sides: `diff(merge-base, 6c85ff25)` equals `diff(d60a2f0, 74776ab8)`, and `diff(merge-base, d60a2f0)` equals `diff(6c85ff25, 74776ab8)`, ignoring index lines. Both identical, so not an evil merge.

Of the six files in the delta stat, `client/client.go`, `client/client_test.go`, and `client/types.go` are entirely main's. The branch's own delta is `commandoutput_listener.go`, its test, and a comment in `sudolistener.go`. Worth saying explicitly because `client/client.go` appearing in the delta could otherwise read as movement on the race below: it is a rename, and it does not touch the field.

## Closed: a halted listener now names the server's refusal (`72f44ce6`)

I traced the whole chain rather than trusting the commit message:

- `provisionSession` records `l.err`, which is `CreateEventSession`'s `fmt.Errorf("failed to create event session: %w", err)` over the client's `*apiError`.
- `halted()` wraps it as `event listener stopped while subscribing: %w`.
- `subscribe` returns it to `subscribeTo`, `streamSubscribed`'s error branch passes it to `runCommandFallbackFromID`, and that prints `real-time output unavailable (%v); falling back to polling`.

So the user reads something like `real-time output unavailable (event listener stopped while subscribing: failed to create event session: permission denied: you do not have the required privileges for this action); falling back to polling`—the server's own refusal, not a second generic string.

One precision worth stating, since it bounds what "the status reaches the user" means here: `apiError.Error()` returns only `message`, so the numeric 401 or 403 never prints. What reaches the user is the sentence `client.authStatusMessage` renders from that status, plus the server's `detail` when it sent one. That is exactly what the sibling `listenerFailure` path already produced, so this is parity and not a shortfall. If the number is wanted, it belongs in `client`, once, for every caller.

**The siblings.** Both `listenerFailure` call sites (`event.go:383`, `event.go:405`) already carried the cause—that asymmetry is what made this a gap—so nothing there needed changing. The other two listeners genuinely do not have the problem: `SudoListener.subscribe` and `Watcher.subscribe` are reached only through `onConnected`, so no external caller reads their return value and then streams on the strength of it. `CommandOutputListener.subscribeTo` is the package's only external subscribe entry point. That is also why keeping `halted()` on `CommandOutputListener` instead of lifting it into `wsListener` is the right call: `wsListener` has no `Err()` to carry, and each embedder's `err` is its own.

## Closed as documented: the check-to-subscribe window

The second `halted()` at line 150 moves the residual from "the whole `SubmitCommand` round trip plus both subscribe calls" down to "between that check and `l.subscribed = true`", and the comment now says out loud that it narrows rather than closes. That is what the note asked for. Closing it properly still means `streamSubscribed` selecting on the listener's liveness in its main loop, which remains a bigger change than this PR wants, so I am not carrying it forward.

## Still open, untouched

**The listen goroutine still reads `ac.AccessToken` unsynchronized.** `client/types.go:12` still declares `AccessToken string`; `client/client.go:204-205` reads it in `setHTTPHeader` and `client/client.go:406` writes it in `RefreshToken`; the only `sync` member on `AlpaconClient` is `loadOnce`. `git diff 6c85ff25..a5f1a0f2 -- client/` is main's rename and nothing else. Everything in the earlier note holds unchanged. Still `client`'s to fix, still worth an issue. Re-anchored inline.

**Backoff resets on a connection that lasted zero milliseconds.** `git diff 6c85ff25..a5f1a0f2 -- api/event/wslistener.go` is empty. New this round, and it downgrades the item rather than moving it: `origin/main`'s `listenLoop` is character-identical, so this predates the branch entirely instead of being something the branch introduced. `connectAndListen` returns true for a socket that accepted and closed in the same instant, so `delay` snaps back to `reconnectBaseDelay` forever. A follow-up issue against `wslistener.go`, not work for this PR.

**`refreshExpiredToken`'s `TryLock` branch still has no coverage.** Re-probed at head: deleting the guard and its `defer` leaves the whole `./api/event` suite green (71.4s, `ok`). Recorded, not requested—pinning it needs a sixty-second MFA poll running concurrently with a reconnect.

## `a5f1a0f2`, judged against what it claims

The comment states the real bound. "The allowance is spent here whether or not the refresh below lands" covers both ways the refresh can fail to happen—`sl.refreshToken()` returning an error, and `mfaMu.TryLock()` failing—and "what it bounds is the downgrade, not the refresh" names the thing the branch actually guards, which is what the old wording obscured. I confirmed the consequence it asserts: on the second 401, `!sl.refreshed` is false, `fatal` stays true, and `fail` takes the `if fatal` early return, so it stops without even attempting another refresh.

One precision, no action, inline below: "a refresh that fails leaves the retry presenting the same token" is exact for the error mode and merely conservative for the `TryLock` mode. There the lock is held by `handleSudoMFA`, which is itself refreshing, so the retry may well present a new token and the allowance is spent for nothing. That mode stays unreachable before the first subscribe, which is why it costs nothing—and the commit message says so, which is a reasonable place for it.

## What I ran

- `go test -race -count=1 ./...`—every package `ok`, no `FAIL`, no race report. `api/event` takes 74.3s.
- `go vet ./...` and `gofmt -l .`—both clean. `golangci-lint` is still not installed here, so the PR's own job stays the authority on it.
- Mutation testing on the two new assertions, each reverted afterwards, with `git status --porcelain` empty at the end:
  - dropping the `%w` wrap from `halted()` so it always returns the flat sentence gives `TestCommandOutputListener_SubscribeCarriesWhyTheListenerStopped` failing on "Target error should be in err chain";
  - deleting the second `halted()` call from `subscribe` gives `TestCommandOutputListener_SubscribeFailsWhenTheListenerStopsMidSubscription` failing on "An error is expected but got nil";
  - deleting the `TryLock` guard from `refreshExpiredToken` leaves the suite green, as noted above.

So both new assertions pin what they claim, and the third probe confirms the nit still stands.

## Small note on the new tests

`TestCommandOutputListener_SubscribeCarriesWhyTheListenerStopped` writes `l.err` directly rather than driving a real fatal provision, so the chain is pinned in two halves: `TestCommandOutputListener_SurfacesAFatalSessionFailure` shows `provisionSession` records `failed to create event session` end to end, and the new one shows `subscribe` carries whatever is there. That is a fair trade—an end-to-end version would have to win a race against `WaitConnected`—but the seam is worth knowing about if either half moves.

`TestCommandOutputListener_SubscribeFailsWhenTheListenerStopsMidSubscription` closes over a `*CommandOutputListener` that is still nil when `newWatcherTestServer` builds the handler. It is safe as written because the listener is never `Start`ed, so the only thing that can reach the handler is the caller's own `subscribeTo` after the assignment. A future edit adding `l.Start()` to that test would turn it into a nil dereference on the server goroutine.

## Nits, carried forward unchanged

`nextReconnectDelay` still has no jitter and the loop still ignores `Retry-After` even though `utils.RetryAfter` exists and `isFatalRequestError` deliberately keeps 429 retryable. `EventSession` and `EventUserChannel` rows still accumulate one per dial with no server-side prune, and that half lives in alpacon-server. `commandoutput_listener.go` and `api/event/watcher.go` still share the unbounded-retry shape, and the watcher predates this PR.

## Verdict

Nothing here blocks the merge, and the branch's own delta is clean in every way I could check. I am staying at COMMENT only because the two mediums are still open and I am carrying them rather than dropping them; neither is work this PR should absorb. If you would rather land them as issues and move on, say so and I will flip to approve on the next pass.

api/event/commandoutput_listener.go
```diff
@@ -49,19 +58,122 @@ type commandOutputEnvelope struct {
...
+// fallback warning naming the server's own refusal rather than the halt.
+func (l *CommandOutputListener) halted() error {
+	select {
+	case <-l.done:
+	default:
+		return nil
+	}
+
+	if cause := l.Err(); cause != nil {
+		return fmt.Errorf("event listener stopped while subscribing: %w", cause)
```
junho226, line 172: This closes the low from the last round—confirmed by following the chain rather than the commit message. `l.err` is `CreateEventSession`'s `failed to create event session: %w` over the client's `*apiError`, this wraps it, `streamSubscribed` hands it to `runCommandFallbackFromID`, and that prints it through `%v`. So the fallback warning now carries the server's own refusal instead of a flat sentence, and I verified by mutation that dropping the `%w` turns `TestCommandOutputListener_SubscribeCarriesWhyTheListenerStopped` red.

One thing to be precise about, since it bounds what the fix buys: `apiError.Error()` returns only `message`, so the numeric 401 or 403 never prints. What the user sees is the sentence `client.authStatusMessage` renders from the status, plus the server's `detail` when it sent one. That is exactly what the sibling `listenerFailure` path already produced, so this is parity rather than a shortfall—if the status number is wanted, it belongs in `client`, once, for every caller, not here.

Keeping `halted()` on `CommandOutputListener` instead of lifting it into `wsListener` is right, for the record: `wsListener` has no `Err()` to carry, and `SudoListener.subscribe` and `Watcher.subscribe` are reached only through `onConnected`, so neither has an external caller that could stream on the strength of a stale success. `subscribeTo` is the package's only external subscribe entry point.

api/event/sudolistener.go
```diff
@@ -47,25 +52,155 @@ type sudoMFAEvent struct {
...
+	return cause
+}
+
+// refreshExpiredToken renews the access token so the next dial stops presenting the one
+// that just came back 401. Past the first subscribe nothing else would: no status is
+// fatal there, so an access token that expires mid-session would be re-presented on
+// every dial until websh exits and sudo MFA would never come back.
+func (sl *SudoListener) refreshExpiredToken() {
+	// An MFA flow holds mfaMu across its own refresh; a second one would race it.
+	if !sl.mfaMu.TryLock() {
```
junho226, line 194: Two carried items land on this branch, neither a request.

**The comment above `fatal && expired && !sl.refreshed` now states the real bound.** "Spent here whether or not the refresh below lands" covers both ways the refresh can fail to happen: `sl.refreshToken()` returning an error, and this `TryLock` missing. And "what it bounds is the downgrade, not the refresh" names what the branch actually guards, which is what the old wording obscured. The consequence it asserts holds—on the second 401 `!sl.refreshed` is false, `fatal` stays true, and `fail` takes the `if fatal` early return without attempting another refresh.

One precision: "a refresh that fails leaves the retry presenting the same token" is exact for the error mode and conservative for this one. Here the lock is held by `handleSudoMFA`, which is itself refreshing, so the retry may well present a new token and the allowance is spent for nothing. That costs nothing because the branch is unreachable before the first subscribe—no frame is read until `onConnected` returned nil, and that is what sets `subscribed`—and the commit message already says so, which is a fine place for it. No change wanted.

**The coverage nit still stands.** Re-probed at head: deleting this guard and its `defer` leaves the whole `./api/event` suite green (71.4s, `ok`). Recording it rather than asking for a test, since pinning it needs a sixty-second MFA poll running concurrently with a reconnect.

api/event/sudolistener.go
jisung-02: Both mediums are already filed, so the re-anchor lands on something that exists rather than on this PR.

#397 covers this one. It names `client/types.go:12`, the `setHTTPHeader` read, and the `RefreshToken` write, and proposes exactly what you describe—a `sync.RWMutex` or `atomic.Pointer[string]` behind an `ac.accessToken()` accessor in `client`, so every caller is covered at once instead of one listener at a time. After your round-3 note I extended the body with a "The listen goroutine writes the field too" section, because `refreshExpiredToken` made the listen goroutine a writer and not only a reader. The `-race` blindness is recorded there too, with the reason: the tests substitute `sl.refreshToken`.

#398 covers the other one—`alpacon event watch` retrying a mid-stream 401 forever with the same token, in the shape `commandoutput_listener.go` and `api/event/watcher.go` share.

So, taking you up on the offer: neither is work this PR should absorb, both are landed as issues, and we would like the approve flip on the next pass.

api/event/commandoutput_listener.go
jisung-02: Agreed on all three points, and thanks for tracing the chain rather than taking the commit message for it.

On the status number: I read it the same way. `apiError.Error()` returning only `message` is a `client` concern, and putting the number in here would give one caller a richer string than the sibling `listenerFailure` path that already renders the same thing. Parity is the right outcome for this PR. If you want the status carried, say so and I will file it against `client` so every caller gets it at once—right now it is not filed.

On `halted()` staying on `CommandOutputListener`: that is exactly the reasoning I used. `wsListener` has no `Err()` to carry and each embedder owns its own `err`, so lifting it would mean inventing a field on the skeleton for the one embedder that needs it. `subscribeTo` being the only external subscribe entry point is what keeps that from becoming a gap in the other two listeners.

api/event/sudolistener.go
jisung-02: Taking both as recorded, with one of them acted on.

On the precision: you are right that the sentence is conservative for the `TryLock` mode, and I am leaving the wording as it is. The comment sits above the allowance to explain why the downgrade is spent once, and qualifying it with a mode that cannot be reached before the first subscribe would cost more reading than it buys. The commit message carries the qualification, which is where a reader who wants it will be looking.

On the coverage nit: I am not adding the test in this PR, for the reason you gave—pinning it needs a sixty-second MFA poll running concurrently with a reconnect, and building that harness is a larger change than the guard it would cover. Recording it here rather than filing it, since it is a missing test on a branch that is unreachable in the failure mode that matters. If you would rather it were tracked, say so and I will open it.

junho226 (approved):
## Verdict: approve

Last round I held at COMMENT only because the two mediums were open, and said I would flip if you would rather land them as issues and move on. You did, so I am flipping. The branch has not moved since `a5f1a0f2`, so this is the same code I already cleared, with the carried items now owned somewhere that outlives the PR.

I checked the issues rather than taking the reply for them:

- **#397** is open, and it is the note itself rather than a pointer at it: `client/types.go:12`, the `setHTTPHeader` read, the `RefreshToken` write, and the `sync.RWMutex` or `atomic.Pointer[string]` behind an `ac.accessToken()` accessor. The section you added after round 3 is the part I cared about most, because `refreshExpiredToken` is what turned the listen goroutine into a writer and not only a reader, and it records the `-race` blindness with its cause.
- **#398** is open, and it scopes the unbounded retry correctly: `CommandOutputListener` has the same hole with `execTimeout()` bounding it, `alpacon event watch` has no bound at all, and it sequences itself after #397 for the right reason.

#394, #395 and #396 are open too.

## Your two questions

**The status number: do not file it.** I am narrowing what I said, because `client` has already decided this and I had not read that far. `authStatusMessage` (`client/client.go:141`) separates the cases a human needs separated—a bare 401 renders `authentication failed: please run 'alpacon login' again`, a coded 401 renders `request denied by server`, and a 403 renders `permission denied: you do not have the required privileges for this action`. The number would only restate what the sentence already carries. More to the point, the comment above the coded-401 branch says the information-lightness is deliberate and names `TestSendRequest_403CodeWithoutDetailKeepsCodeSource` as the guard on the no-raw-code contract. A caller wanting the machine-readable half already has `ErrorCode()`, `ErrorSource()` and `HTTPStatusCode()`. So "it belongs in `client`" was the wrong half of my sentence: it belongs nowhere, and the parity you landed is the whole answer.

**The coverage nit: I was wrong about the cost, and it still should not gate this PR.** I said pinning `refreshExpiredToken`'s `TryLock` branch needs a sixty-second MFA poll running concurrently with a reconnect. It does not. The branch asks only that `mfaMu` be held by someone else, and the test can be that someone:

```go
type unauthorized struct{}

func (unauthorized) Error() string       { return "session create refused" }
func (unauthorized) HTTPStatusCode() int { return http.StatusUnauthorized }

func TestSudoListener_LeavesTheTokenAloneWhileAnMFAFlowIsRefreshing(t *testing.T) {
	sl := NewSudoListener(nil, "my-server", "session-1")
	var refreshes atomic.Int32
	sl.refreshToken = func() error { refreshes.Add(1); return nil }

	sl.mfaMu.Lock() // stands in for handleSudoMFA holding it across its own refresh
	defer sl.mfaMu.Unlock()

	_ = sl.fail(unauthorized{})

	assert.Zero(t, refreshes.Load(), "an MFA flow already refreshing owns the token")
}
```

No server, no second goroutine, no timing. `TestSudoListener_StaysQuietWhenAFailureLandsAfterStop` already drives `fail` directly with hand-set state, so the shape is the file's own, and the path is short: `expired` is true, `fatal && expired && !sl.refreshed` spends the allowance and clears `fatal`, `announce` is false because nothing subscribed, and `refreshExpiredToken` finds the lock taken. The one thing the package lacks is a status-carrying error—`client.statusError` and `withStatus` are both unexported—so it wants a test-local type, and `utils.HTTPStatusCode` matches structurally on an exported `HTTPStatusCode() int`, which is why the four lines above are enough. Twelve lines with the type.

The mutation you would want also reds cleanly: delete the guard and its `defer`, as I did when probing, and `refreshToken` runs, so `assert.Zero` fails on the count rather than hanging.

Given that, do not open an issue for it—an issue would cost more than the test. Whether it goes in now is yours: the guard is unreachable before the first subscribe either way, and I am not holding the approve on it.

## What I ran

Read at `a5f1a0f2f0a4ebcc8f45760751ac2981b49701e1` rather than from the commit messages: `api/event/sudolistener.go` (`fail`, `refreshExpiredToken`, the `mfaMu`/`stateMu` split), `api/event/sudolistener_test.go`, `client/client.go` (`authStatusMessage`, `apiError`, `statusError`, `withStatus`), `utils/error.go` (`HTTPStatusCode`). Confirmed no commit landed after `a5f1a0f2` and that #394 through #398 are all open, with #397 carrying the write-side section.

I did not re-derive what the earlier rounds settled—the reconnect state machine, the goroutine and subscription accounting, the parameter-order call sites, the halted-listener reason chain. Nothing has changed under them.

PR #399, opened 2026-08-27
refactor(mfa): share the standard error callbacks across commands

jisung-02:
## Background

Seven call sites built a `utils.ErrorHandlerCallbacks` value for `utils.HandleCommonErrors`, and five of them wrote the same four fields out by hand, differing only in `RetryOperation`:

| Site | Operation retried |
|---|---|
| `cmd/exec/exec.go:158` | `event.SubmitCommand` (detached run) |
| `cmd/exec/run.go:649` | `event.RunCommandStreaming` |
| `cmd/websh/websh.go:281` | `websh.CreateWebshSession` |
| `cmd/ftp/cp.go:112` | `uploadObject` |
| `cmd/ftp/cp.go:141` | `downloadObject` |

`OnMFARequired`, `OnUsernameRequired`, `CheckMFACompleted`, and `RefreshToken` were byte-identical across all five, apart from the name of the client variable the closures captured. Changing one of them meant remembering to change four others.

The sixth site had already solved it. `cmd/edit/edit.go` factored the same four fields into a local `editErrorCallbacks(ac, retry)` helper and called it from both its download and upload paths, so the shape was settled—it just was not shared.

## Changes

`mfa.ErrorCallbacks(ac, retry)` in `api/mfa/mfa.go` is `editErrorCallbacks` promoted to a package-level function. All seven call sites in the five files now go through it, and `cmd/edit`'s local copy is gone.

On placement: `utils` cannot host the helper. `api/mfa` and `api/iam` both import `utils`, so `utils` importing them back is a cycle. `api/mfa` is the closest home the code already had—`HandleMFAError` and `CheckMFACompletion` are three of the four callbacks, and every one of these commands already imports the package. The one new edge is `api/mfa` importing `api/iam` for `HandleUsernameRequired`; `api/iam` does not import `api/mfa`, so that adds no cycle. A separate `pkg/errhandler` package was the first attempt and was dropped as an unnecessary file for fifteen lines of wiring.

Three commands still construct the struct themselves, because their callbacks differ for reasons that are not cosmetic:

| Site | Why it stays |
|---|---|
| `cmd/tunnel/tunnel.go:172` | Creates its client lazily, only once MFA turns out to be needed, and leaves `CheckMFACompleted` nil so it takes `HandleCommonErrors`' legacy retry path |
| `cmd/server/server_action.go:32` | Sets no `OnUsernameRequired` |
| `cmd/workspace/workspace_authentication_update.go`, `cmd/workspace/workspace_access_control_update.go` | `OnMFARequired` opens a workspace-scoped MFA link via `GetWorkspaceSecurityMFALink`, not a server-scoped one, and neither takes a username |

Folding any of them into `mfa.ErrorCallbacks` would change behavior. The two `cmd/workspace` sites do now share their callbacks with each other, through `errorCallbacks` in the new `cmd/workspace/error_callbacks.go`—see the review follow-up below.

The five command files each drop their `api/iam` import; they all still import `api/mfa`, now for `ErrorCallbacks` rather than the individual functions. The CLAUDE.md "Error handling" bullet records the new helper, which commands are deliberately exempt, and why `utils` is not an option.

## Review follow-up

Round one, three commits:

| Item | Commit |
|---|---|
| The `ErrorCallbacks` doc comment claimed "a token refresh before each retry", which holds on neither path the helper enables—the MFA path refreshes once after completion and retries once, and the username-required path never calls `RefreshToken` at all | `4882e03` |
| The CLAUDE.md bullet stopped mid-argument at "`api/mfa` and `api/iam` both import `utils`" without stating the conclusion, counted "the four fields" of a five-field struct, and joined `server` and `workspace` with an "or" that reads as either/or across both | `bdf1bec` |
| Nothing pinned the helper's field set, so one edit could silently move four commands onto the legacy retry loop | `c39bdc6` |

Round two, three more:

| Item | Commit |
|---|---|
| `TestErrorCallbacks_WiresEveryField` called `cb.RetryOperation()` without a `NotNil` guard first, so dropping that field panicked instead of failing—aborting the rest of the package and printing a stack trace where the field name belongs | `8df4696` |
| The two `cmd/workspace` update commands held near-verbatim copies of the same callback literal and had already drifted: authentication read the `cfg` its `IsSaaS` gate loads up front, access control re-loaded config inside the MFA callback. `errorCallbacks(ac, retry)` in `cmd/workspace` now serves both | `a55c486` |
| That extraction moved the same silent-switch risk from one update command to both, and neither had any coverage | `4462c15` |

Round three, two more:

| Item | Commit |
|---|---|
| That extraction standardized on the wrong copy. `authentication update` used to name the MFA link from the `cfg` its `IsSaaS` gate had already loaded; the shared helper re-read config inside the callback instead. The client pins `BaseURL` at construction from the first read (`client/client.go:89`) and a blocking editor session sits between the two, so a concurrent `alpacon ws use` sends the PATCH to one workspace and the MFA link to another | `9a4dbbd` |
| Nothing stopped a later edit from putting the reload back—the callback would still build a link and the drift only shows under a race | `3bef0fb` |

An earlier revision of this body defended the reload, arguing that taking `workspaceName` as a parameter would force access control to read config on the happy path where it has no gate to piggyback on. That was wrong on both counts. The read costs nothing, since `NewAlpaconAPIClient` loads config anyway and the added check only moves the accurate "not logged in" message ahead of a vaguer "connection failed" one, and keeping the read inside the callback is exactly what let `authentication update` regress. Both commands now load config once, up front, before the client that pins the target exists.

Net: 165 insertions, 159 deletions across 12 files.

## Testing

Three tests were added: one per helper, plus one pinning the workspace helper's new parameter.

`api/mfa/mfa_test.go` gains `TestErrorCallbacks_WiresEveryField`. It asserts that every field of the returned struct is non-nil and that `RetryOperation` is the closure handed in. The field that matters is `CheckMFACompleted`: leaving it nil is not a compile error, it is a silent switch from `HandleCommonErrors`' poll-then-retry-once flow to the legacy refresh-and-retry loop, and that switch now reaches `exec`, `websh`, `cp`, and `edit` from one place instead of one command from one copy. Seen red by dropping `CheckMFACompleted` from the returned literal: `Expected value not to be nil.`

`cmd/workspace/error_callbacks_test.go` is the same guard for the workspace helper, and it also pins `OnUsernameRequired` as nil, since that absent field is one of the two reasons these commands stay off `mfa.ErrorCallbacks`. Seen red twice, by dropping `CheckMFACompleted` and `RetryOperation` in turn:

```
--- FAIL: TestErrorCallbacks_WiresEveryField (0.00s)
    error_callbacks_test.go:20: Expected value not to be nil.
    error_callbacks_test.go:25: Expected value not to be nil.
```

The same file gains `TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn`, which stands up an httptest server and asserts the MFA link request carries the name handed to the helper rather than whatever config holds. Seen red by restoring the reload the fix removed:

```
--- FAIL: TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn (0.00s)
    error_callbacks_test.go:50:
        	expected: "my-workspace"
        	actual  : "alpacax"
```

"alpacax" is the workspace in the local config the mutant read, which is the regression itself rather than a proxy for it. `ALPACON_NO_BROWSER=1` keeps `utils.OpenBrowser` from launching one during the run.

The command paths themselves stay uncovered, and nothing else covers them either. I claimed otherwise in the original commit body and was wrong; here is what the suites actually do:

- `cmd/edit/edit_test.go` injects a fake `editDeps` and calls `runEdit`, so `realEditDeps`—the function that uses the helper—is never constructed.
- `cmd/ftp/cp_test.go` covers local/remote path classification and argument parsing only. It never reaches `uploadObject` or `downloadObject`.
- Nothing anywhere calls `RunCommandWithRetry`, `RunRemoteExec`, or `websh.CreateWebshSession`.
- `cmd/workspace` had no test file at all before this PR, and the new one covers the helper rather than either `Run` function.
- The only test that drives `utils.HandleCommonErrors` is `utils/error_handler_test.go`, and it builds the callback struct inline rather than going through any command.

So the new tests pin the helpers, and the compiler pins the substitution—each site's `RetryOperation` closure is carried over verbatim, including the variables it assigns to. Neither pins the commands end to end. Covering those means giving `exec`, `websh`, `cp`, and the two workspace commands the same seam `cmd/edit` already has, which is a larger change than this one and belongs in its own PR.

## Known, not addressed here

`mfa.ErrorCallbacks` has the same shape the workspace helper just shed. Its `OnMFARequired` reaches `GetMFALinkByServerName`, which reads config at `api/mfa/mfa.go:86` to name the workspace in the link—inside the callback, on a client whose `BaseURL` was pinned earlier. `cmd/edit` is the closest match to the bug fixed above, because `editDeps.upload` runs after the editor closes, so the same unbounded human time sits between the two reads.

This PR does not change it. The reload predates the refactor and no call site moved onto or off it here. Correcting it means `GetMFALinkByServerName` taking the workspace name as a parameter, which pulls in `mfa.ErrorCallbacks` and through it all seven call sites, plus `StepUpForSudo` (`api/mfa/mfa.go:113`), `api/event/sudolistener.go:126`, `cmd/exec/run.go:65`, and the two direct `HandleMFAError` callers in `cmd/server/server_action.go:34` and `cmd/tunnel/tunnel.go:189`. That is wider than what this PR has already grown to hold, so it will go in its own PR. No issue is filed for it yet.

The missing `IsSaaS()` guard on `access-control update` is also still missing. Adding the config read at the top of its `Run` puts the guard's natural home in place, but the guard itself is a behavior change and stays out of this PR.

## Verification

On the final diff:

- `go build ./...`: no output.
- `go test -race ./...`: 40 `ok`, no `FAIL`.
- `go vet ./...`: no output.
- `gofmt -l .`: no output.


api/mfa/mfa.go
```diff
@@ -58,6 +59,27 @@ func CheckMFACompletion(ac *client.AlpaconClient) (bool, error) {
 	return resp.Completed, nil
 }
 
+// ErrorCallbacks returns the standard callback set commands hand to
+// utils.HandleCommonErrors: MFA and username-required handling, MFA completion
+// polling, and a token refresh before each retry. retry re-runs the operation
+// that failed.
```
geunwoonoh: **Suggestion: the doc comment describes a retry policy the code does not have.**

"a token refresh before each retry" is wrong on both paths this helper enables:

- MFA path (`CheckMFACompleted` non-nil): refresh happens *once*, after completion is observed, followed by exactly one retry (`utils/error_handler.go:84-95`). There is no "each".
- `UsernameRequired` path: `RefreshToken` is never called — `HandleCommonErrors` goes straight to `RetryOperation` (`utils/error_handler.go:129-142`).

"before each retry" is accurate only for the legacy loop, which is the one branch this helper never takes. The rewrite below also fixes the sentence starting with a bare lowercase identifier (`retry re-runs…`).

```suggestion
// ErrorCallbacks returns the standard callback set commands hand to
// utils.HandleCommonErrors: MFA and username-required handling, MFA completion
// polling, and a token refresh once MFA completes. The retry function re-runs
// the operation that failed.
```

api/mfa/mfa.go
```diff
@@ -58,6 +59,27 @@ func CheckMFACompletion(ac *client.AlpaconClient) (bool, error) {
...
+func ErrorCallbacks(ac *client.AlpaconClient, retry func() error) utils.ErrorHandlerCallbacks {
```
geunwoonoh, line 66: **Warning: centralizing changes the blast radius, and nothing pins the field set.**

`CheckMFACompleted` being nil is not a compile error. It is a silent switch: `utils/error_handler.go:56` takes the poll-then-retry-once flow when it is set, and the `else` at `utils/error_handler.go:99` falls back to the legacy "refresh + retry every second for up to 3 minutes" loop when it is not.

Before this PR, dropping that field in one place broke one command. Now one edit to `ErrorCallbacks` silently changes `exec`, `websh`, `cp`, and `edit` at once — four commands whose MFA paths have no test coverage at all, as the PR body honestly documents.

That is an argument for a test, not against the refactor. `api/mfa/mfa_test.go` already has the httptest harness, so this is cheap:

```go
func TestErrorCallbacks_WiresEveryField(t *testing.T) {
	ac := &client.AlpaconClient{}
	retried := false

	cb := ErrorCallbacks(ac, func() error { retried = true; return nil })

	assert.NotNil(t, cb.OnMFARequired)
	assert.NotNil(t, cb.OnUsernameRequired)
	// A nil CheckMFACompleted silently drops every caller onto the legacy
	// retry loop in utils.HandleCommonErrors — no compile error, no failure.
	assert.NotNil(t, cb.CheckMFACompleted)
	assert.NotNil(t, cb.RefreshToken)

	assert.NoError(t, cb.RetryOperation())
	assert.True(t, retried, "RetryOperation must be the closure passed in")
}
```

Not a merge blocker — the change is provably mechanical. But this is the test that guards the *next* change, and it did not exist before because the risk was distributed across five copies.

CLAUDE.md
```diff
@@ -72,7 +72,7 @@ utils/               # Shared utilities (output, prompts, errors, SSH parsing)
 - **Command registration**: All commands are registered in `cmd/root.go` via `RootCmd.AddCommand()`
 - **API layer**: Each `cmd/` package calls corresponding `api/` package for HTTP requests
 - **SSH-like syntax**: `websh`, `exec`, `cp` support `user@host` syntax via `utils.ParseSSHTarget()`
-- **Error handling**: Common errors (MFA required, username required) are handled via `utils.HandleCommonErrors()` with retry callbacks
+- **Error handling**: Common errors (MFA required, username required) are handled via `utils.HandleCommonErrors()` with retry callbacks. `mfa.ErrorCallbacks(ac, retry)` builds the standard callback set for a command that already holds a client—use it instead of writing the four fields out again. Commands whose callbacks genuinely differ construct `utils.ErrorHandlerCallbacks` themselves: `tunnel` creates its client lazily and takes the legacy retry path, and `server`/`workspace` need no username handling or open a workspace-scoped MFA link. `utils` cannot host the helper—`api/mfa` and `api/iam` both import `utils`
```
geunwoonoh: **Suggestion: the cycle argument is cut off, plus two smaller wording issues.**

1. The bullet ends on "`utils` cannot host the helper—`api/mfa` and `api/iam` both import `utils`" and never states the conclusion. A reader who does not already know the answer is left to infer it.
2. "`server`/`workspace` need no username handling **or** open a workspace-scoped MFA link" distributes ambiguously. `server` skips username handling; `workspace` skips username handling *and* opens a workspace-scoped link. As written, `or` reads as an either/or across both commands.
3. "writing the four fields out again" is precise inside this PR's framing (the four non-`RetryOperation` fields) but `ErrorHandlerCallbacks` has five, and CLAUDE.md is read without that context.

All three in one line:

```suggestion
- **Error handling**: Common errors (MFA required, username required) are handled via `utils.HandleCommonErrors()` with retry callbacks. `mfa.ErrorCallbacks(ac, retry)` builds the standard callback set for a command that already holds a client—use it instead of writing the whole set out again. Commands whose callbacks genuinely differ construct `utils.ErrorHandlerCallbacks` themselves: `tunnel` creates its client lazily and takes the legacy retry path; `server` needs no username handling; `workspace` needs none either and opens a workspace-scoped MFA link. `utils` cannot host the helper—`api/mfa` and `api/iam` both import `utils`, so `utils` importing either back would be an import cycle
```

geunwoonoh (dismissed):
## Review summary

Seven call sites built `utils.ErrorHandlerCallbacks` by hand; five of them wrote the same four non-retry fields verbatim. This promotes `cmd/edit`'s existing local `editErrorCallbacks` helper to `mfa.ErrorCallbacks(ac, retry)` in `api/mfa/mfa.go` and routes all seven through it. Three commands (`tunnel`, `server`, `workspace`×2) keep their own literals.

I verified the substitution site by site: every `RetryOperation` closure is carried over character-for-character, including the ones that assign to outer variables (`session, err =` in websh, `resp, err =` in exec). The outer `err = utils.HandleCommonErrors(...)` still wins over the closure's assignment because Go evaluates the RHS before assigning — unchanged from before. **No behavior change.**

Verified locally on `040294b`:

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l .` | clean |
| `go test ./...` | 39 `ok`, 0 `FAIL` |
| `golangci-lint` | not installed locally — CI is the gate |

Import cycle claim confirmed with `go list -deps`: `api/iam` reaches `config`, `utils`, `api/auth0`, `client`, `api`, `api/types` — never `api/mfa`. The new edge is safe.

All four exemption claims in `CLAUDE.md:75` also check out: `tunnel` really does lazy-create its client and leave `CheckMFACompleted` nil (`cmd/tunnel/tunnel.go:172-203`), `server_action.go:32` really sets no `OnUsernameRequired`, and both workspace sites really open a workspace-scoped link via `GetWorkspaceSecurityMFALink`. Folding any of them in would change behavior, as stated.

## Critical issues (0)

None.

## Warnings (1)

**1. `api/mfa/mfa.go:66` — centralizing changes the blast radius, and nothing pins the field set.** (inline)

`CheckMFACompleted` being nil is not a compile error. It is a silent switch between two different retry policies. Before this PR, dropping that field in one place broke one command; now one edit silently changes `exec`, `websh`, `cp`, and `edit` at once — four commands whose MFA paths have no test coverage at all, as the PR body honestly documents. That is an argument for a test, not against the refactor. Details and a suggested ~15-line test inline.

## Suggestions (3)

**1. `api/mfa/mfa.go:62-65` — the doc comment describes a retry policy the code does not have.** (inline)

"a token refresh before each retry" is wrong on both paths this helper enables. Suggested rewrite inline.

**2. `CLAUDE.md:75` — the cycle argument is cut off, and two smaller wording issues.** (inline)

The bullet never states the conclusion of its own import-cycle argument; `server`/`workspace` are described with an ambiguous `or`; and "the four fields" reads as off-by-one against a five-field struct. One suggestion inline fixes all three.

**3. Follow-up, out of this PR's scope: `cmd/workspace` still holds the duplication this PR set out to remove.**

`cmd/workspace/workspace_authentication_update.go:49` and `cmd/workspace/workspace_access_control_update.go:41` are near-verbatim copies of each other — same four fields, same stderr format string, same `OpenBrowser` call, ~18 lines each. The exemption in `CLAUDE.md` explains why they cannot use `mfa.ErrorCallbacks`; it does not explain why they cannot share a `workspaceErrorCallbacks(ac, workspaceName, retry)` between *themselves*. The same argument that justified this PR applies to them.

They have already drifted, which is the usual symptom: `workspace_authentication_update.go` reads an outer `cfg`, while `workspace_access_control_update.go:47` re-loads config *inside* the MFA callback. One of those is doing unnecessary work; more to the point, nobody decided that — it is copy-paste divergence.

`cmd/server/server_action.go:32` is a weaker case: three of its five fields are byte-identical to the helper's, but a variant taking an `includeUsername bool` would be worse than the duplication. Leave it.

## On placement (not a finding)

`api/mfa` importing `api/iam` for an interactive TTY prompt would normally read as a layering violation for a package tree documented as "API client functions per domain". It is not one here, and the PR body undersells its own case: `HandleMFAError` at `api/mfa/mfa.go:35` *already* writes to stderr and calls `utils.OpenBrowser`. `api/mfa` was never a pure HTTP wrapper, so `ErrorCallbacks` is consistent with what the package already does rather than a new kind of dependency.

For the record, a `cmd/*` shared package would also have worked — `cmd/worksession` is exactly that precedent and is already imported by all five touched command packages plus `tunnel`. The author considered a new package and rejected it as overhead for fifteen lines, which is a defensible call, and CLAUDE.md records the reasoning either way.

## Performance

Neutral, and it is worth stating why rather than assuming: `ErrorCallbacks` allocates the same struct and the same three closures the inline literals did, and every one of these calls sits on an error path that is already blocked on network I/O. The one thing to note is that the argument is built eagerly even when `HandleCommonErrors` falls through to `default:` and discards it — but that was equally true of the struct literals it replaces, so there is no regression, and no case for changing it.

## Highlights

- Net −65 lines with seven call sites collapsed to one definition, and the shape was not invented — it was `cmd/edit`'s already-proven helper promoted.
- The PR body retracts a false test claim from the commit message and replaces it with a file-by-file account of what the suites actually cover. That is the right way to handle it, and it is what let me verify the coverage gap in one pass instead of discovering it.
- The exemption table is not hand-waving — all four sites hold up under inspection, and the reason each gives is the real reason.
- Every removed `api/iam` import was genuinely dead afterward (no `iam.` references remain in any of the five files), and no `editErrorCallbacks` reference survives.

---

**Verdict**: **APPROVED**

Nothing here blocks the merge. The one warning is a test that guards future edits to the new shared helper, and the suggestions are doc-comment and CLAUDE.md wording. The refactor itself is correct.

**Checklist**:
- [x] Code follows stack conventions (gofmt clean, doc comment starts with the function name, declaration order intact)
- [ ] Tests included — none added; the affected paths were and remain uncovered (acknowledged in the PR body)
- [x] No security vulnerabilities — no auth logic changed; the same callbacks reach the same handlers
- [x] Documentation updated — `CLAUDE.md:75` records the helper and the exemptions, with the wording fixes suggested inline

api/mfa/mfa.go
jisung-02: Applied as suggested in 4882e03. You are right on both paths: the MFA branch refreshes once after completion and retries once, and the username-required branch reaches `RetryOperation` without ever touching `RefreshToken`. The comment now reads "a token refresh once MFA completes. The retry function re-runs the operation that failed."

api/mfa/mfa.go
jisung-02: Added in c39bdc6, as written. Seen red by dropping `CheckMFACompleted` from the returned literal: `mfa_test.go:149: Expected value not to be nil.`, and green again once restored.

The framing is the part worth keeping: the test is not there because the substitution is risky, it is there because centralizing moved a silent failure mode from one command per copy to four commands at once, and nothing downstream would catch it. The PR body Testing section now says that instead of just "no test was added".

CLAUDE.md
jisung-02: Applied in bdf1bec, all three. The bullet now closes the argument with "so `utils` importing either back would be an import cycle", says "the whole set" instead of counting fields, and splits `server` and `workspace` onto their own clauses so the "or" cannot distribute.

api/mfa/mfa_test.go
```diff
@@ -135,3 +135,20 @@ func TestGetMFALink_ServerErrorIsAnError(t *testing.T) {
...
+	cb := ErrorCallbacks(ac, func() error { retried = true; return nil })
+
+	assert.NotNil(t, cb.OnMFARequired)
+	assert.NotNil(t, cb.OnUsernameRequired)
+	// A nil CheckMFACompleted silently drops every caller onto the legacy
+	// retry loop in utils.HandleCommonErrors—no compile error, no failure.
+	assert.NotNil(t, cb.CheckMFACompleted)
+	assert.NotNil(t, cb.RefreshToken)
+
+	assert.NoError(t, cb.RetryOperation())
```
geunwoonoh, line 154: **Suggestion: `RetryOperation` is the one field called without a `NotNil` guard first.**

The other four fields are checked with `assert.NotNil`, but `RetryOperation` is invoked directly on line 152. I mutated the helper to drop `RetryOperation: retry,` to see what that costs:

```
panic: runtime error: invalid memory address or nil pointer dereference
	mfa_test.go:152
```

So the regression is still caught — but as a panic, which aborts the remaining tests in the package and reports a stack trace instead of naming the missing field. Compare with the `CheckMFACompleted` mutation, which fails as `mfa_test.go:149: Expected value not to be nil.`

`CLAUDE.md`'s assertion conventions cover this: "`require` when later lines depend on the assertion, `assert` when checks are independent." Calling `cb.RetryOperation()` depends on it being non-nil, so this is the `require` case.

```suggestion
	require.NotNil(t, cb.RetryOperation)
	assert.NoError(t, cb.RetryOperation())
```

Note this also needs `"github.com/stretchr/testify/require"` added to the import block — the suggestion block cannot reach it. Entirely optional; the test guards correctly either way.

geunwoonoh (dismissed):
## Follow-up review — all three findings applied

Re-reviewed at `c39bdc6` (was `040294b` at my previous approval). Three commits, one per finding, and nothing else touched — the original refactor is byte-for-byte unchanged.

| Previous finding | Commit | Status |
|---|---|---|
| Warning — nothing pins the callback field set | `c39bdc6` | Applied |
| Suggestion — doc comment describes a retry policy the code does not have | `4882e03` | Applied |
| Suggestion — `CLAUDE.md` cycle argument cut off, plus two wording issues | `bdf1bec` | Applied, all three |
| Suggestion (follow-up) — `cmd/workspace` duplication | — | Untouched, correctly: I scoped it out of this PR |

## Verification at `c39bdc6`

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l .` | clean |
| `go test -race ./...` | 39 `ok`, 0 `FAIL` |

**I confirmed the guard test actually guards, by mutation rather than by reading it.** A test that does not fail on the mutation it targets is worth nothing, so I did not take the "seen red" note in the thread at face value — I reproduced it.

Dropping `CheckMFACompleted` from the returned literal:

```
mfa_test.go:149: Expected value not to be nil.
--- FAIL: TestErrorCallbacks_WiresEveryField (0.00s)
```

Same line, same message as reported. On that same mutant, `go build ./...` **succeeded** — which is the premise of the original warning demonstrated rather than asserted: the failure mode is a silent switch, not a compile error. Without this test, one edit could have moved `exec`, `websh`, `cp`, and `edit` onto the legacy retry loop with nothing downstream catching it. Working tree restored afterward.

## On how the feedback was applied

The suggested test was not pasted in verbatim — it was corrected to house style on the way in. My version wrote `HandleCommonErrors — no compile error` with spaces around the em dash; the committed version reads `HandleCommonErrors—no compile error`, which is the `word—word` rule in `CLAUDE.md` that my own suggestion had broken. Good catch against the reviewer.

The PR body's Testing section was also rewritten from "no test was added" to state what the new test pins *and* that the command paths remain uncovered. It claims exactly what was added and no more.

## Suggestions (1) — optional, not blocking

**`api/mfa/mfa_test.go:152` — `RetryOperation` is the one field called without a `NotNil` guard first.** (inline)

The other four fields are checked with `assert.NotNil`; `RetryOperation` is invoked directly. If that field is dropped, the test panics instead of failing cleanly — confirmed by mutation:

```
panic: runtime error: invalid memory address or nil pointer dereference
	mfa_test.go:152
```

It still fails, so the guard holds. But a panic aborts the rest of the package's tests and reports a stack trace instead of a cause. `CLAUDE.md`'s "`require` when later lines depend on the assertion" is written for exactly this case. Details inline; fine to merge as is.

---

**Verdict**: **APPROVED**

My approval on `040294b` still stands and was never dismissed, so this is not required to unblock the merge — it re-approves on the new head and puts the mutation-testing evidence on the record. Nothing in the three follow-up commits weakens the original approval; the one warning that existed is now closed by a test I verified independently.

**Checklist**:
- [x] Code follows stack conventions (gofmt clean, em dash rule respected, `assert.True` message carries the operand)
- [x] Tests included and passing — `TestErrorCallbacks_WiresEveryField` verified red on mutation, green on `c39bdc6`
- [x] No security vulnerabilities — no auth logic changed across any of the follow-up commits
- [x] Documentation updated — `CLAUDE.md:75` and the `ErrorCallbacks` doc comment both corrected

api/mfa/mfa_test.go
jisung-02: Applied in `8df4696`, as suggested, with the `require` import added.

You were right that the cost is not whether the regression is caught but how it reports. I reproduced both sides rather than taking the panic on trust. Before, dropping `RetryOperation` gave the nil-pointer panic you quoted at `mfa_test.go:152`. After, the same mutation ends at `mfa_test.go:153` with `Expected value not to be nil.`, and the rest of the package still runs.

Two other things came out of it. The line number moved because `require.NotNil` sits on its own line above the call, so a future reader diffing your comment against the file will land one line later than the id says. And the same '`require` when later lines depend on the assertion' rule applies to the workspace helper I extracted in `a55c486`, so its test at `cmd/workspace/error_callbacks_test.go:25` was written that way from the start rather than fixed afterward.

cmd/workspace/error_callbacks.go
```diff
@@ -0,0 +1,38 @@
...
+// commands take no username and their MFA link is workspace-scoped rather than
+// server-scoped. The retry function re-runs the operation that failed.
+func errorCallbacks(ac *client.AlpaconClient, retry func() error) utils.ErrorHandlerCallbacks {
+	return utils.ErrorHandlerCallbacks{
+		OnMFARequired: func(_ string) error {
+			cfg, err := config.LoadConfig()
+			if err != nil {
+				return fmt.Errorf("failed to load configuration: %w", err)
+			}
+			mfaURL, err := mfa.GetWorkspaceSecurityMFALink(ac, cfg.WorkspaceName)
```
geunwoonoh: **Warning: this drops the `cfg` the caller already loaded and validated.**

`workspace authentication update` changed behavior here. It previously built the MFA link from the outer `cfg` loaded at `workspace_authentication_update.go:24` and validated by the `IsSaaS()` guard at `:28`. This helper calls `config.LoadConfig()` again inside the callback.

Between those two points sits a blocking editor session (`workspace.EditAuthentication`, `workspace_authentication_update.go:37`) — unbounded human time. If another shell runs `alpacon ws use` while the editor is open, config's workspace changes underneath.

The PATCH target does not follow it. The client pins `BaseURL` at construction (`client/client.go:89`, used at `:243`), which this codebase states outright at `client/client.go:537`:

> Uses ac.BaseURL (not config's WorkspaceURL) to stay consistent with the client's target.

So the request goes to the original workspace while the MFA link points at the new one. The user completes MFA in the wrong place, `CheckMFACompletion(ac)` on the pinned client never sees it, and the command times out after 3 minutes explaining none of this. Nothing is written to the wrong workspace and no privilege is wrongly granted — a confusing failure, not a security issue.

`access-control update` already re-loaded inside the callback, so it is unchanged; only `authentication update` regressed.

Taking the name as a parameter removes the regression and makes the dependency visible:

```suggestion
func errorCallbacks(ac *client.AlpaconClient, workspaceName string, retry func() error) utils.ErrorHandlerCallbacks {
	return utils.ErrorHandlerCallbacks{
		OnMFARequired: func(_ string) error {
			mfaURL, err := mfa.GetWorkspaceSecurityMFALink(ac, workspaceName)
```

Three things the suggestion block cannot reach:

1. `config` becomes an unused import in this file and must be removed (`fmt` stays — `fmt.Fprintf` still uses it).
2. `workspace_authentication_update.go:45` passes the `cfg` it already has: `errorCallbacks(alpaconClient, cfg.WorkspaceName, func() error {`.
3. `workspace_access_control_update.go:36` has no `cfg` in scope and would need to load one near the top of `Run` — which is also where the missing `IsSaaS()` guard would go, if you want to close that separately.

Narrow enough that I would merge without it: it needs a concurrent workspace switch, during an editor session, on a request that also returns MFA-required.

geunwoonoh (dismissed):
## Follow-up review — `4462c15`

Both remaining items from my last review are in, and the scope grew with them: 8 files to 12.

| Previous item | Commit | Status |
|---|---|---|
| Suggestion — `RetryOperation` called without a `NotNil` guard | `8df4696` | Applied, with the `require` import |
| Suggestion (follow-up) — `cmd/workspace` duplication | `a55c486` + `4462c15` | Done here, though I had scoped it to its own PR |

On the scope: I said the workspace extraction belonged in a separate PR and it landed in this one instead. The result stays on a single theme and each helper arrived with its own test, so it is still reviewable at this size. No objection.

## Verification at `4462c15`

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l .` | clean |
| `go test -race ./...` | 40 `ok` (39 → 40, new `cmd/workspace`), 0 `FAIL` |

**I reproduced all three mutations reported in the threads rather than taking them on trust.**

| Mutation | Reported | Reproduced |
|---|---|---|
| Drop `RetryOperation` from the mfa helper | `mfa_test.go:153`, no panic | Matches — `require` catches it as a clean assertion |
| Drop `CheckMFACompleted` from the workspace helper | `error_callbacks_test.go:20` | Matches |
| Drop `RetryOperation` from the workspace helper | `error_callbacks_test.go:25` | Matches |

The `require.NotNil` change does what it was meant to: the same mutation that previously ended in a nil-func panic now fails as an assertion, and the rest of the package still runs.

## Warnings (1)

**`cmd/workspace/error_callbacks.go:19` — the extraction dropped the caller's already-validated `cfg`.** (inline)

`workspace authentication update` changed behavior. It used to build the MFA link from the outer `cfg` loaded at `workspace_authentication_update.go:24` and validated by the `IsSaaS()` guard four lines later. The shared helper calls `config.LoadConfig()` again *inside* `OnMFARequired`.

What sits between those two points is a blocking editor session (`workspace.EditAuthentication`, same file line 37) — unbounded human time. If another shell runs `alpacon ws use` during it, config's workspace changes underneath.

The PATCH target does not follow. The client pins `BaseURL` at construction (`client/client.go:89`, used at `:243`), a principle this codebase states outright at `client/client.go:537`: *"Uses ac.BaseURL (not config's WorkspaceURL) to stay consistent with the client's target."* So the request goes to the original workspace while the MFA link points at the new one. The user completes MFA in the wrong place, `CheckMFACompletion(ac)` on the pinned client never sees it, and the command times out after 3 minutes with a message that explains none of this.

To be clear about the blast radius: nothing is written to the wrong workspace and no privilege is wrongly granted. It is a confusing failure, not a security issue. `access-control update` already re-loaded inside the callback, so it is unchanged there — only `authentication update` regressed.

Passing the workspace name in removes the regression and makes the dependency visible. Suggested signature inline.

This needs a concurrent workspace switch, during an editor session, on a request that also comes back MFA-required. **It does not block the merge.**

## Note, not a finding — outside this PR

`workspace_access_control_update.go` has no `IsSaaS()` guard. `workspace_authentication_update.go:28`, `workspace_authentication.go:20`, `workspace_mfa_methods.go:20`, and `workspace_usage.go:25` all have one. A command using a workspace-scoped MFA link looks SaaS-only, so the omission reads as a gap — but it predates this PR and is not its job to fix. It is worth noting only because the two commands now share a helper, which makes the asymmetry visible, and because adding the `workspaceName` parameter above would move config loading to the top of the command, where that guard naturally belongs.

## Highlights

- The workspace test pins the *deliberate* difference, not just the wiring: `assert.Nil(t, cb.OnUsernameRequired)`. That absent field is one of the two reasons these commands stay off `mfa.ErrorCallbacks`, and it is now held by a test instead of a comment.
- The workspace test was written with `require` from the start rather than fixed after the fact — the feedback was applied as a rule, not patched at the one line where it was raised.
- The PR body's Testing section carries the red evidence for both helpers, with line numbers, and still states plainly which command paths remain uncovered.
- Both workspace call sites carry their `RetryOperation` closures over verbatim, assignments to outer variables included.

---

**Verdict**: **APPROVED**

The one warning is a narrow race whose worst outcome is a confusing timeout. Everything else is verified.

Note that my approval on `c39bdc6` was dismissed when these commits landed — this repo dismisses stale approvals — so this review is what puts an active approval back on the PR.

**Checklist**:
- [x] Code follows stack conventions (gofmt clean, em dash rule respected, `require` vs `assert` per `CLAUDE.md`)
- [x] Tests included and passing — both helpers guarded, all three mutations verified red independently
- [x] No security vulnerabilities — the warning above is a UX failure, not a privilege or data-integrity issue
- [x] Documentation updated — `CLAUDE.md:75` now records the workspace helper alongside the shared one

cmd/workspace/error_callbacks.go
jisung-02: Applied in `9a4dbbd`, with the test in `3bef0fb`.

You were right that this was a behavior change and not just a stylistic one. I checked `c39bdc6` before touching anything: `authentication update` did build the link from the outer `cfg`, and the extraction replaced that with a fresh read. I standardized on the wrong one of the two copies, and the reason it is the wrong one is the one you gave—`ac.BaseURL` is fixed at construction from the first read, so any later read can name a workspace the request is not going to.

All three points the suggestion block could not reach are in: `config` is gone from the helper file (`fmt` stays), `workspace_authentication_update.go:45` passes the `cfg` it already holds, and `access_control_update.go:24` loads one at the top of `Run`. That read costs nothing, since `NewAlpaconAPIClient` reads config anyway—it just moves the accurate "not logged in" message ahead of the vaguer "connection failed" one.

One thing beyond what you asked for: the signature change alone leaves nothing stopping a later edit from putting `config.LoadConfig()` back inside the callback, and the drift only shows under a race. `TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn` stands up an httptest server and asserts the link request carries the name passed in. Seen red by restoring the reload:

```
--- FAIL: TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn (0.00s)
    error_callbacks_test.go:50:
        	expected: "my-workspace"
        	actual  : "alpacax"
```

`alpacax` is the workspace in my local config, so the mutant reproduced your scenario rather than a proxy for it. Working tree restored afterward.

I left the `IsSaaS()` guard out, as you scoped it. The config read now sits where it would go.

The same shape survives on the `mfa.ErrorCallbacks` path—details in the round summary below, not fixed here.

cmd/workspace/workspace_authentication_update.go
```diff
@@ -46,25 +42,10 @@ Modify the desired fields, save, and close the editor to apply changes.`,
...
-				CheckMFACompleted: func() (bool, error) {
-					return mfa.CheckMFACompletion(alpaconClient)
-				},
-				RefreshToken: alpaconClient.RefreshToken,
-				RetryOperation: func() error {
-					authenticationDetail, err = workspace.PatchAuthentication(alpaconClient, data)
-					return err
-				},
-			})
+			err = utils.HandleCommonErrors(err, "", errorCallbacks(alpaconClient, cfg.WorkspaceName, func() error {
```
geunwoonoh, line 45: Suggestion: nothing pins this argument here, unlike its sibling.

The two callers are guarded unequally, and the difference is incidental. In `workspace_access_control_update.go` the `cfg` loaded at line 24 exists only to supply the workspace name, so the compiler enforces the fix:

```
$ sed -i '' 's/cfg.WorkspaceName/""/' cmd/workspace/workspace_access_control_update.go
$ go build ./...
cmd/workspace/workspace_access_control_update.go:24:3: declared and not used: cfg
```

Here `cfg` has a second consumer at line 28 (`cfg.IsSaaS()`), so the same mutation compiles and the suite stays green:

```
$ sed -i '' 's/cfg.WorkspaceName/"wrong-workspace"/' cmd/workspace/workspace_authentication_update.go
$ go build ./... && go test -race ./cmd/workspace/...
ok  github.com/alpacax/alpacon-cli/cmd/workspace  1.411s
```

Both callers are correct as written—this is not a defect and I would not change code for it. Two things worth recording: the stronger guard is an accident of the `IsSaaS()` guard being absent next door, so closing that gap would remove it; and `TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn` pins the helper's behavior but nothing pins what the commands hand it.

geunwoonoh (approved):
## Round 4 — everything from round 3 is in, verified independently at `3bef0fb`

### Round 3 follow-up

| Item | Commit | Verified |
|---|---|---|
| Warning — `errorCallbacks` re-read config inside the callback | `9a4dbbd` | Applied, with the suggested signature |
| `config` import removed, `fmt` kept | `9a4dbbd` | `error_callbacks.go:4` still needs `fmt` for line 28; no dead import |
| `authentication_update.go:45` passes the `cfg` it holds | `9a4dbbd` | Confirmed |
| `access_control_update.go` loads `cfg` at the top of `Run` | `9a4dbbd` | Confirmed at line 24 |
| `IsSaaS()` guard left out, as scoped | — | Confirmed: still absent |
| Guard test (not requested) | `3bef0fb` | Added |

### Independent verification

| Check | Result |
|---|---|
| `go build ./...` | no output |
| `go vet ./...` | no output |
| `gofmt -l .` | no output |
| `go test -race ./...` | 40 `ok`, 0 `FAIL`, 16 no-test-files |
| CI on `3bef0fb` | 6/6 pass, including `golangci-lint` |
| `9a4dbbd` standalone | build / vet / `-race` suite all OK |

golangci-lint is not installed on my machine—the passing CI job on `3bef0fb` is the gate for it.

### Mutations reproduced

| Mutation | Expected | Observed |
|---|---|---|
| Restore `config.LoadConfig()` inside `OnMFARequired` | `TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn` red | Red: `expected "my-workspace", actual "gwtest"` |
| Drop `CheckMFACompleted` | build OK, `..._WiresEveryField` red at `:23` | Exactly that |
| `access_control_update.go` drops `cfg.WorkspaceName` | — | **Compile error**: `:24:3: declared and not used: cfg` |
| `authentication_update.go` passes `"wrong-workspace"` | — | Build OK, suite green — nothing catches it |

The first mutation reproduced the reported bug rather than a proxy for it: my local config names `gwtest`, so the callback would have sent `gwtest` while the client targeted whatever the caller pinned. Your run produced `alpacax` from your own config—same failure, two machines.

---

## Critical issues (0)

None.

## Warnings (0)

None.

## Suggestions (2)

**1. `cmd/workspace/workspace_authentication_update.go:45` — the argument is pinned in one caller and not the other**

The last two mutations above are the finding. In `access_control_update.go` the compiler enforces the fix: `cfg` exists only to supply the workspace name, so removing it fails the build. In `authentication_update.go` `cfg` has a second consumer (`cfg.IsSaaS()` at line 28), so a wrong name compiles and the suite stays green.

The asymmetry is incidental, not a defect—both callers are correct today. It is worth knowing that the stronger of the two guards is an accident of the `IsSaaS()` guard's absence next door, and that closing the guard gap would remove it. Not worth a code change on its own. Inline below.

**2. The `cmd/edit` follow-up — file the issue**

Answering the question at the end of your round summary directly: yes, open it, and no, it should not hold this PR.

I traced the same path before reading your note and landed on the same place. `realEditDeps.upload` (`cmd/edit/edit.go:150`) runs at `runEdit` line 286, after `deps.runEditor` blocks at line 261, on the `ac` constructed at line 76—so `mfa.ErrorCallbacks` → `HandleMFAError` → `GetMFALinkByServerName` → `config.LoadConfig()` at `api/mfa/mfa.go:86` reads config across the same unbounded human window the workspace commands had. `download` at line 142 runs before the editor and is unaffected.

Your scope argument holds and every line number in it is exact—`api/mfa/mfa.go:113`, `api/event/sudolistener.go:126`, `cmd/exec/run.go:65`, `cmd/server/server_action.go:34`, `cmd/tunnel/tunnel.go:189`. `git diff origin/main...3bef0fb -- api/mfa/mfa.go` confirms this PR added `ErrorCallbacks` and touched nothing else in that file, so the refactor is behavior-neutral there. Its own PR is the right call.

## Note outside this PR's scope

The `main` ruleset sets `require_code_owner_review: true` and `required_approving_review_count: 1`, but the repo has no `CODEOWNERS` file, and no classic branch protection exists (`/branches/main/protection` returns 404). `reviewDecision` reads empty both when an approval is active and when none is, so it is not a usable signal on this repo—`mergeStateStatus` is the thing to watch. This also explains the round-3 puzzle: `4462c15` sat at `BLOCKED` with an active approval and CI that had gone green six hours earlier. Worth a look separately from this PR.

## Highlights

- **The fix is the right one, and the doc comment says why rather than what.** Lines 17–20 of `error_callbacks.go` record the reasoning—`ac` pins `BaseURL` at construction, an editor session sits between that and the prompt—so the next person to reach for `config.LoadConfig()` here reads the reason before making the change.
- **You checked the claim against `c39bdc6` before acting on it.** A reviewer's account of which of two copies was correct is exactly the kind of thing worth confirming, and you confirmed it and then said plainly that the regression was yours.
- **The guard test goes beyond the ask and earns it.** A signature change alone does not stop a later edit from putting the read back, and the drift only appears under a race—so the httptest assertion on the outgoing `workspace` param is the only thing that would catch it.
- **`t.Setenv("ALPACON_NO_BROWSER", "1")` is load-bearing, not decoration.** `OpenBrowser` returns at `utils/browser.go:24` before `acquireBrowserLock()`, so the test neither launches a browser on a developer's machine nor writes `~/.alpacon/.browser_lock`—the lock file's mtime is unchanged after my run.
- **The `access_control_update.go` config read costs nothing and improves the failure.** `NewAlpaconAPIClient` reads config anyway; moving the read up front replaces "Connection to Alpacon API failed… Consider re-logging" with "Not logged in. Run 'alpacon login' first."
- **`assert.Nil(t, cb.OnUsernameRequired)` pins the deliberate difference from `mfa.ErrorCallbacks`,** so the divergence reads as a documented decision rather than an omission someone later "fixes."
- **Scope discipline held all four rounds.** The `IsSaaS()` guard stayed out, the `mfa.ErrorCallbacks` drift stayed out with a written reason, and the PR body was rewritten against the final diff rather than left describing an earlier one.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Code follows stack conventions (gofmt, vet, golangci-lint green in CI)
- [x] Tests included and passing (40 packages, `-race`; new guard test seen red under mutation)
- [x] No security vulnerabilities (no credential handling changed; the fixed drift was a UX defect, not an authz one)
- [x] Documentation updated (`CLAUDE.md:75` already covers `errorCallbacks` without naming a signature, so `9a4dbbd` introduced no drift)

cmd/workspace/workspace_authentication_update.go
jisung-02: Agreed, and no code change. Recording it here is the right outcome.

I ran both mutations myself rather than taking the asymmetry on trust, and got exactly what you got at `3bef0fb`:

```
$ sed -i '' 's/cfg\.WorkspaceName/"wrong-workspace"/' cmd/workspace/workspace_authentication_update.go
$ go build ./... && go test -race ./cmd/workspace/...
ok  github.com/alpacax/alpacon-cli/cmd/workspace  1.540s

$ sed -i '' 's/cfg\.WorkspaceName/"wrong-workspace"/' cmd/workspace/workspace_access_control_update.go
$ go build ./...
cmd/workspace/workspace_access_control_update.go:24:3: declared and not used: cfg
```

Working tree restored after each.

Two things follow from it, both worth having in the thread.

The stronger guard is not free to keep. Adding the `IsSaaS()` guard to `access-control update` gives `cfg` a second consumer there too, and the compile error goes away. So the guard and the guard gap are the same fact seen twice, and closing the gap is still the right call: an unused-variable error is a byproduct, not a test.

The real gap is that nothing drives either command's `Run`. `TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn` pins what the helper does with the name it is handed, and nothing pins what the callers hand it, because there is no seam to inject through. That is the same uncovered-command-path hole I have been reporting since round one, now with a concrete thing it would have caught. Still no issue filed for it.

Separately, I filed #401 for the `cmd/edit` follow-up, per your suggestion 2.

PR #402, opened 2026-08-28
fix(mfa): build the MFA link from the client's own workspace

### LINKED ISSUE #401: GetMFALinkByServerName reads config inside the MFA callback, after the client pinned its target
Raised by @geunwoonoh in review on #399, as the surviving sibling of the `cmd/workspace` drift that PR fixes in `9a4dbbd`.

## Background

`client.NewAlpaconAPIClient()` reads config once and pins `BaseURL` at construction (`client/client.go:89`, used at `:243`). Every request that client sends afterward goes to that original workspace, whatever config says later. The codebase states the principle outright at `client/client.go:537`: "Uses ac.BaseURL (not config's WorkspaceURL) to stay consistent with the client's target."

`mfa.ErrorCallbacks`' `OnMFARequired` reaches `HandleMFAError` and then `GetMFALinkByServerName`, which calls `config.LoadConfig()` a second time (`api/mfa/mfa.go:86`). That second read decides the `workspace` parameter on the MFA link, and any amount of time can pass between the two reads.

## Consequence

If another shell runs `alpacon ws use` between client construction and the MFA prompt, the request still goes to the original workspace while the MFA link points at the new one. The user completes MFA in the wrong place, `CheckMFACompletion(ac)` on the pinned client never observes it, and the command times out after 3 minutes with a message that explains none of this.

Nothing is written to the wrong workspace and no privilege is wrongly granted. It is a confusing failure, not a security issue.

`cmd/edit` is the worst case. The client is constructed at `cmd/edit/edit.go:76`, and `realEditDeps.upload` (`:150`) runs at `:286`, after `deps.runEditor` blocks at `:261`. Unbounded human time sits between the two reads. `download` (`:139`) runs before the editor and is unaffected.

`cp`, `websh`, and `exec` can also wait a long time, but not on a human, so their window is much narrower.

## Proposed change

Take the workspace name as a parameter on `GetMFALinkByServerName`, supplied by the caller from the same config read that built its client. This is the fix already applied to `cmd/workspace/error_callbacks.go`.

Changing that signature pulls in:

| Site | What |
|---|---|
| `api/mfa/mfa.go:37` | `HandleMFAError` |
| `api/mfa/mfa.go:69` | `ErrorCallbacks`, whose own signature then changes, reaching all seven call sites below |
| `api/mfa/mfa.go:113` | `StepUpForSudo` |
| `api/event/sudolistener.go:126` | direct `GetMFALinkByServerName` call |
| `cmd/exec/run.go:65` | the `mfaLinkByServerName` test seam |
| `cmd/server/server_action.go:34` | direct `HandleMFAError` call |
| `cmd/tunnel/tunnel.go:189` | direct `HandleMFAError` call |

The seven `ErrorCallbacks` call sites are `cmd/websh/websh.go:280`, `cmd/ftp/cp.go:111`, `cmd/ftp/cp.go:127`, `cmd/edit/edit.go:142`, `cmd/edit/edit.go:153`, `cmd/exec/run.go:648`, and `cmd/exec/exec.go:157`.

`tunnel` needs separate thought: it creates its client lazily, so where the workspace name comes from at callback-construction time is not the same question it is for the other commands.

## Testing

Mirror `TestErrorCallbacks_MFALinkNamesTheWorkspacePassedIn` (`cmd/workspace/error_callbacks_test.go:33`): stand up an httptest server and assert the outgoing MFA link request carries the name the caller passed. The signature change alone does not stop a later edit from putting `config.LoadConfig()` back inside the callback, and that drift only surfaces under a race, so the guard is the part that lasts.

## Scope

This predates #399. `GetMFALinkByServerName` always read config itself, and that PR's refactor moved call sites onto `mfa.ErrorCallbacks` without changing what the callback does. Confirmed with `git diff origin/main...3bef0fb -- api/mfa/mfa.go`, and independently by the reviewer.

jisung-02:
## Background

`client.NewAlpaconAPIClient()` reads config once and pins `BaseURL` from it, so every request a client sends goes to that workspace whatever config says later. `GetMFALinkByServerName` called `config.LoadConfig()` a second time to name the workspace on the MFA link, and it runs at the MFA prompt—which an editor session or a long transfer can put minutes after the client was built.

If another shell runs `alpacon ws use` in those minutes, the user verifies MFA in one workspace while the request keeps going to the other. `CheckMFACompletion` polls the pinned client, never sees the completion, and the command times out after three minutes explaining none of it. `cmd/edit` is the worst case: `runEditor` blocks on a human between the two reads.

## Changes

The workspace name is pinned on the client next to the URL it belongs to, and nothing below `cmd/` reads config to learn it.

| Layer | What changed |
|---|---|
| `client/types.go`, `client/client.go` | `AlpaconClient` gains `WorkspaceName`, assigned in the constructor from the same `config.LoadConfig()` that supplies `BaseURL` |
| `api/mfa/mfa.go` | `GetMFALinkByServerName` no longer imports config; `GetMFALink` and `GetWorkspaceSecurityMFALink` drop their `workspaceName` parameter and read `ac.WorkspaceName` |
| `cmd/workspace/error_callbacks.go` | `errorCallbacks` loses the parameter it only forwarded |
| `cmd/workspace/workspace_usage.go` | Takes the payment URL and workspace name off the client instead of its own earlier config read |

Issue #401 proposed passing the workspace name as a parameter from the caller's own config read. That was rejected: a parameter moves the bug one layer up rather than removing it, since a caller can still pass a name that disagrees with `ac.BaseURL`. It also has a cost the pinning approach does not—the issue lists seven `ErrorCallbacks` call sites plus `HandleMFAError`, `StepUpForSudo`, `api/event/sudolistener.go`, `cmd/server`, `cmd/tunnel`, and the `cmd/exec` test seam, all of which would have to thread the name through. None of those files are touched here. Every client that reaches an MFA function comes from `NewAlpaconAPIClient()`; the one hand-built client (`api/auth/auth.go:60`) is never passed to one.

An empty workspace is now an error instead of a link carrying `workspace=`. `requireWorkspace` runs at the top of `GetMFALinkByServerName` as well as inside `fetchMFALink`, so a client with no workspace costs no round trip and the error names the workspace rather than a name lookup that failed after it. A config written before `workspace_name` existed and never re-logged-in now fails at the MFA prompt with `no workspace name on the client; run 'alpacon login' first`, which names its own fix. All three login paths fill the field, falling back to `utils.ExtractWorkspaceName(workspaceURL)`, so no current login produces such a config.

`TestAPILayerNeverReadsConfig` keeps the rule from decaying. It walks the AST of every non-test file under `api/` and fails on `LoadConfig`, `IsSaaS`, `GetActiveWorkSession`, or `ResolveAuthMethod`—the last three read the file inside and answer out of it, so reaching for one is the same second read renamed. Writes stay allowed, since login, logout, and token renewal are why the file exists. Walking the AST rather than the text means a comment naming `config.LoadConfig` is not an offender, an aliased import still is, and a dot import is reported on sight because it leaves the call a bare identifier no selector walk can attribute. What it does not reach is written down rather than left to be discovered: it walks `api/` only, so a read added under `client/`—the package that holds the pin—stays green (#404), and no text-level guard can follow `api/` into a helper package that reads config on its behalf. Review catches that one.

Net: 275 insertions, 58 deletions across 12 files, of which 208 insertions are tests and the boundary guard.

## Testing

`go test -race ./...` passes. Every test below was seen red before it was trusted.

| Test | Seen red by |
|---|---|
| `TestNewAlpaconAPIClient_PinsWorkspaceIdentityFromConfig` | Dropping the new assignment: `expected: "my-workspace" / actual: ""` |
| `TestGetMFALinkByServerName_NamesTheWorkspaceTheClientIsPinnedTo` | Restoring the config read—it then read `alpacax` from the local config under an empty HOME |
| `TestGetMFALink_EmptyWorkspaceIsAnErrorBeforeAnyRequest`, `TestGetMFALinkByServerName_EmptyWorkspaceCostsNoRoundTrip` | Removing `requireWorkspace`—the fake servers recorded a request |
| `TestGetMFALink_ServerErrorIsAnError` | Reached no handler, guarding against a guard that fires too early |
| `TestErrorCallbacks_EmptyWorkspaceFailsInsteadOfPrintingALink` | Removing `requireWorkspace` from `fetchMFALink`. The fake server answers with a usable `mfa_url` on purpose, so both assertions fail for the right reason instead of tripping the parse error an empty body raises with or without the guard |
| `TestAPILayerNeverReadsConfig` | A real `config.IsSaaS()` call, the same call through an aliased import, and a file dot-importing config; green for a comment-only mention |

`cmd/workspace/workspace_usage.go` has no test. Its logic is inline in `RunE` with no seam to call from, and the `api/` boundary guard cannot be extended over `cmd/`, which reads config by design. The substitution there is exact—the constructor assigns `BaseURL` from `WorkspaceURL` verbatim—and nothing breaks today, since its two config loads sit microseconds apart with no wait between them. It changed because it was the last place below `cmd/` still answering "which workspace" from config while holding a client that already knew.

## Follow-ups

Two items from review are filed rather than fixed here, both pre-existing and on a different axis than the MFA link.

| Issue | What it is |
|---|---|
| #403 | `work-session create --wait --use` files the session under whatever workspace config names after the approval wait, not the one it was created in. Same "act here, record there" shape, different command. |
| #404 | The boundary guard cannot see `client/`, so widening the walk needs an allowlist for `client/client.go`'s two known reads. |

Closes #401


eunyoung14 (dismissed):
**The fix is complete for the class it targets. Approving.**

I did not take the diagnosis on trust — I reproduced the race and then confirmed it is closed, running the same probe against both trees. Fixture config names workspace `alpha`, build the client through the real `NewAlpaconAPIClient()`, rewrite the config to `beta` (the `alpacon ws use` from another shell), then call `GetMFALinkByServerName`.

| Tree | client pinned to | config now says | `workspace=` on the request |
| --- | --- | --- | --- |
| `origin/main` (`b5cb452`) | alpha | beta | **beta** — the link names a workspace the request is not going to |
| this head (`8524a15`) | alpha | beta | **alpha** |

## The census holds

I enumerated every `config.LoadConfig()` in the repo rather than spot-checking the MFA path.

**`api/` has zero.** It touches config only through writes, device-id helpers, and a `config.Config` passed in as a parameter. `pkg/` and `utils/` are also zero — `pkg/tunnel/runtime/runtime.go:104` calls `GetSmuxConfig()`, which is a pure constructor with no file read.

**`client/` has two:** `:73`, the pin, which is correct and is the whole point; and `:564` in `refreshLocked` — see below.

All eleven reads in `cmd/` run at command start, with one exception flagged below.

**Every construction path sets `WorkspaceName`.** Two non-test `AlpaconClient` literals exist: the constructor, which now assigns it, and `api/auth/auth.go:60`, the pre-login status probe — that one calls only `SendGetRequest(statusURL)` and is never handed to an MFA function. All 136 production call sites go through `NewAlpaconAPIClient()`. An empty name is reachable only from a config predating the field, which `requireWorkspace` handles.

## The boundary guard is real — mutation-tested

`api/config_boundary_test.go` is the durability mechanism, so I broke it seven ways against a clean export:

| Mutation | Result |
| --- | --- |
| `config.LoadConfig()` re-added in `api/` | fails |
| aliased import `cfgpkg.IsSaaS()` | fails |
| dot import + bare `LoadConfig()` | fails |
| read in a nested package under `api/` | fails (the walk recurses) |
| **control — a comment mentioning `config.LoadConfig()`** | **green** |
| same read placed in `client/` | green — see SHOULD-FIX 2 |
| `api/` reads config via a helper package | green — see NIT 1 |

## SHOULD-FIX

**1. `cmd/worksession/worksession_use.go:118` is a live counterexample to "`cmd/` is fine, it runs at command start".**

`alpacon work-session create --wait --use` blocks in `pollForApproval` (`worksession_create.go:258`) waiting on a human, then calls `RunUse(ac, session.ID)` → `config.SetActiveWorkSession(...)`, which does a fresh `LoadConfig()` and writes into `cfg.ActiveWorkSessions[cfg.WorkspaceName]` (`config/config.go:188-200`).

So: run `create --wait --use` against A; while you wait for approval, another shell runs `alpacon ws use B`; the session minted in A is filed under B's key. A later `alpacon exec` in B then sends A's session id to B's server and is rejected with nothing explaining why — the same "act here, record there" shape this PR is fixing, on a different axis.

`config.SetActiveWorkSessionFor(ac.WorkspaceName, uuid)` takes the key off the client that already knows it, which is exactly the pattern this PR establishes. Out of scope here; worth a follow-up issue.

**2. `client/client.go:564` is the one remaining read below `cmd/`, and the guard cannot reach it.**

The walk starts in `api/`, so an identical read added to `client/` keeps the suite green — I measured that. `refreshLocked` reads `RefreshToken` live rather than from the pin.

I am reporting this as pre-existing and outside the PR's declared scope, and the live read is arguably intentional. But the PR's whole argument is "one read, pinned on the client", so the guard should cover the package that holds the pin. Either point `WalkDir` at the repo root with an explicit allowlist for `client/client.go`'s two known reads, or add a sibling guard in `client/`.

## NITS

**1.** Indirect reads are invisible to the guard: `api/` code calling a helper package that reads config stays green. A text-level guard cannot chase that, which is fine — but the CLAUDE.md paragraph presents the guard as airtight and does not mention it.

**2.** `requireWorkspace` runs twice on the `GetMFALinkByServerName` path (`api/mfa/mfa.go:89` and `:204`). Deliberate and documented; noting it only so a future reader does not "clean it up".

**3.** `cmd/workspace/workspace_access_control_update.go:27` calls `config.LoadConfig()` purely as a login precheck and discards the result — `NewAlpaconAPIClient()` on the next line does the same load and errors on the same condition. It buys a friendlier message. Keeping it is fine; so is deleting it.

**4.** The added CLAUDE.md bullet is ~180 words of prose in a bullet list. Content is fine; it is just heavy for the format.

## Signature changes

Clean. `GetMFALink` lost `workspaceName` — one non-test caller, updated. `GetWorkspaceSecurityMFALink` lost it — one caller, updated. `errorCallbacks` lost it — two callers, both updated.

No adjacent-`string` hazard in any of the three: after the removal `GetMFALink(ac, serverID)` has one string left, `GetWorkspaceSecurityMFALink(ac)` has none, and `errorCallbacks`' surviving second parameter is a `func() error`. The `cmd/exec/run.go:65` test seam is unaffected — that signature did not change. No generated code in the repo.

## What the drift actually did

I checked the server side rather than inferring. `auth0/api/serializers.py:143` validates the `workspace` param against the caller's own workspaces claim, so the param is **not** an existence oracle for a workspace you are not in — membership gates it and the error is generic.

So the pre-fix drift affected exactly one population: a member of **both** workspaces. For them the check passes, and the server stamps the pinned client's host into a redirect whose `workspace=` names the other one. The success page reports completion to A while presenting B's context — which is the three-minute silent timeout described in the PR body, corroborated end to end. **No cross-tenant boundary was crossed either before or after. This was a correctness and UX defect.**

One property worth knowing, unchanged by this PR: the link is written to stderr and passed as argv to the platform's URL opener, so it is visible in a process listing on a shared host and in captured CI logs. It never reaches shell history or a file. Not a regression here.

## Verification

`go build ./...`, `go vet ./...`, `go test ./... -p 1`, and `-race` on the four changed packages all clean on Go 1.26.1. `golangci-lint` at **v2.9.0**, the version `.github/workflows/lint.yaml` pins — I installed that exact version rather than substituting one — **0 issues**. `go mod tidy` produces no diff.

CI on the exact head `8524a1590`: all eight check-runs green, completed **today 06:47–06:51 UTC**. `Build and Test`, `golangci-lint`, `govulncheck`, CodeQL and the Copilot reviewer, which generated no comments.

The `alpacon-server` validation above was read from a local mirror at `06fcba3b5` that I did not re-fetch, so treat that as a couple of days stale. Nothing was exercised against a real workspace.

## One note on merge mechanics

This repo has **no `required_status_checks` rule in any ruleset**, and the `workflows` rule carries an empty list — the same shape as the sibling repos. `Build and Test`, `golangci-lint` and `govulncheck` are visible but not merge-blocking; what blocks is the pull-request rule, code scanning, code quality and the Copilot review. Worth knowing when reading a green tick.

## Recommendation

**Approve.** File the two SHOULD-FIX items as follow-ups rather than holding this — the work-session key is a different command's post-wait read, and the `client/` guard gap is pre-existing.

eunyoung14 (approved):
수정 방향이 맞습니다. 값을 올바르게 넘기는 게 아니라 **틀리게 넘길 수 있는 능력 자체를 제거**한 것이 이 PR의 핵심이고, #401의 파라미터 안을 거절한 근거도 그것이라 일관됩니다. 시그니처에서 `workspaceName`이 사라졌으니 `ac.BaseURL`과 어긋나는 이름을 넘기는 코드는 컴파일되지 않습니다.

## 확인한 것

**`api/` 아래 config 읽기가 실제로 0입니다.** 브랜치에서 `LoadConfig`·`IsSaaS`·`GetActiveWorkSession`·`ResolveAuthMethod`를 직접 grep했고 비테스트 파일에 하나도 없습니다. 경계 가드가 지키려는 상태가 지금 실제 상태입니다.

**경계 가드가 텍스트가 아니라 AST를 봅니다.** 별칭 임포트는 잡히고 주석은 안 잡히며, dot import는 셀렉터로 귀속할 수 없으니 발견 즉시 보고합니다. 이 종류의 가드가 정규식으로 짜여 주석 한 줄에 빨개지거나 별칭 하나에 눈머는 경우가 흔한데 둘 다 피했습니다. 못 미치는 범위(`client/` 아래, `api/`가 부르는 헬퍼 패키지)를 발견에 맡기지 않고 코드 옆에 적어둔 것도 좋습니다 — #404로 이어놓으셨네요.

**`requireWorkspace`가 호출 그래프보다 강한 보증을 줍니다.** 본문은 "수제 client(`api/auth/auth.go:60`)는 MFA 함수에 전달되지 않는다"로 논증하는데, 확인해보니 그 client는 `WorkspaceName`을 채우지 않습니다. 즉 **나중에 누가 그걸 MFA 함수에 넘기더라도 틀린 워크스페이스를 가리키는 링크가 아니라 에러가 납니다.** 불변식이 호출 그래프가 아니라 가드로 지켜지고 있고, 이 PR이 주장하는 것보다 한 단계 강합니다.

**빈 워크스페이스를 라운드트립 전에 거절한 것**도 맞습니다. `GetMFALinkByServerName` 진입점과 `fetchMFALink` 양쪽에 두어, 서버 이름 조회가 먼저 실패해서 진짜 원인을 덮는 일이 없습니다.

## 테스트

"seen red by" 표가 이 PR에서 제일 값어치 있는 부분입니다. 특히 `TestGetMFALinkByServerName_NamesTheWorkspaceTheClientIsPinnedTo`를 빈 HOME 아래에서 config 읽기를 복원해 빨갛게 만든 방식 — 그게 아니면 이 테스트는 픽스 없이도 초록일 수 있습니다. `TestErrorCallbacks_...`에서 가짜 서버가 **일부러 쓸 만한 `mfa_url`을 답하게** 해서, 빈 본문이 유발하는 파싱 에러에 걸려 우연히 빨개지는 게 아니라 옳은 이유로 실패하게 만든 것도 정확합니다.

`TestGetMFALink_ServerErrorIsAnError`가 "가드가 너무 일찍 발사되는 것"을 막는 대조군인 것도 좋습니다. 가드를 넣을 때 같이 넣어야 하는 테스트인데 자주 빠집니다.

## 선택 사항 하나

`api/auth/auth.go`의 수제 client가 `WorkspaceName`을 비워두는데, 바로 위에서 `loginTargetMetadata`가 이미 `workspaceName`을 손에 쥐고 있습니다. 채워도 무해하고 구조체가 어디서 만들어지든 일관되게 채워진다는 성질이 생깁니다.

다만 지금도 안전합니다 — 위에 적었듯 비어 있으면 링크를 만드는 대신 에러가 나니까요. 그래서 이건 결함이 아니라 취향이고, 굳이 안 채우는 쪽을 택했다면 "이 client는 MFA 경로에 들어가지 않으므로 채우지 않는다"는 한 줄이 있으면 다음 사람이 빠뜨린 것으로 읽지 않겠습니다.

## `workspace_usage.go`에 테스트가 없는 것

납득했습니다. 로직이 `RunE` 안에 인라인이라 부를 seam이 없고, `api/` 경계 가드는 설계상 config를 읽는 `cmd/`로 확장할 수 없습니다. 치환이 정확하고(생성자가 `WorkspaceURL`을 그대로 `BaseURL`에 넣음) 두 config 읽기 사이에 대기가 없어 오늘 깨질 것도 없다는 설명이 사실과 맞습니다. 테스트가 없다는 것과 그 이유를 본문에 적어둔 것 자체가 리뷰어에게 필요한 정보였습니다.

PR #408, opened 2026-09-01
feat(update): add alpacon update to self-update the CLI

### LINKED ISSUE #378: [FEATURE] Add "alpacon update" to self-update the CLI binary
## Summary

Add an `alpacon update` command that checks the latest GitHub release and replaces the running binary in place.

## Problem Statement

The CLI gives no signal when it is outdated, and the update procedure differs by install path (brew, apt/yum, manual download). On Windows there is no package manager distribution at all, so manual replacement is the only path. Users on old versions silently miss fixes and new commands.

## Proposed Solution

Add `alpacon update`:

- Query the GitHub Releases API for the latest version and compare against the current build (`utils.Version`).
- If already current, print "already up to date" and exit 0.
- Otherwise download the archive matching the running OS/architecture, verify it against the release checksum file (`alpacon-*-checksums.sha256`), and atomically replace the running binary.

Considerations:

- Package-manager installs: overwriting a binary owned by brew or apt/yum desyncs the package manager's state. Detect the install path (e.g. Homebrew `Cellar`, nfpm's `/usr/local/bin` install) and, when detected, print the correct package-manager update command instead of self-replacing.
- Windows self-replacement: a running exe cannot be overwritten, so use rename-then-replace.
- Permissions: system-path installs may not be writable; on failure, suggest re-running with sudo.
- Local `dev` builds are excluded from self-update.
- A once-a-day new-version notice on other commands can be split into a follow-up issue.

Done when `alpacon update` replaces a manually installed binary with the latest release, guides package-manager installs to the right command instead of overwriting, refuses to replace on checksum mismatch with exit 1, and the README install section is updated.

## Priority

medium

jisung-02:
Closes #378.

`alpacon update` brings the CLI to the latest GitHub release. `alpacon update --check` reports whether one exists and installs nothing.

If a package manager or version manager installed the binary, that tool owns it: the command prints how to upgrade there and exits `1` without touching the file. Homebrew, deb, and rpm get the exact command; mise and asdf are named without one. Ownership is asked of `dpkg` and `rpm`, and the CLI never runs `sudo` or a package manager itself.

For a binary you placed yourself, the archive is downloaded, verified against the release's SHA-256 checksums, and only then does the running executable get replaced. On unix a temp file is renamed over the target, so a failure leaves the old binary intact. Windows forbids overwriting a running executable but allows renaming it, so the new binary is copied in and two renames follow, and the displaced copy stays as `alpacon.exe.old.<timestamp>`. A process killed between those renames leaves no `alpacon.exe`, and nothing recovers it on its own, because the binary that would run the next update is the missing one. The README says how to rename the preserved copy back.

The update holds `.alpacon-update.lock` beside the binary: `flock` on unix, `LockFileEx` on Windows. The kernel drops it however the process exits, so there is no staleness rule to get wrong.

## What the update trusts

Verification happens before extraction, and both asset URLs are pinned to the origins GitHub serves releases from, so an `http://` or off-GitHub URL is refused rather than fetched. Redirects past the first hop stay unpinned because the CDN host changes; the https requirement carries.

What that does not cover: the checksums come from the same release as the archive, and `.goreleaser.yaml` publishes no signature, so the hash proves the bytes arrived intact, not who published them. Signing the checksums and verifying against a pinned identity is follow-up work (#412). The README says `unsigned` plainly.

Extraction lets the archive decide nothing. `ExtractBinary` builds the destination itself and reads entry names only through `filepath.Base`, pinned by a test because a refactor that starts honoring the name would pass every other one. The bytes gzip inflates are bounded separately from the entry written and the archive downloaded.

Two things are deliberately left alone. A group- or world-writable binary keeps its mode and earns a warning rather than a `chmod`, since re-permissioning a file the operator set up is not the CLI's call. And a build whose version matches no release is refused inside `pkg/selfupdate` as well as in the command, because the package is public and an importer may skip the command's guard.

## Exit codes

`--check` adds exit code `8`, meaning a newer release exists; the README table and `--help` both state it. Everything else is the existing `0` and `1`. The branch first claimed `7`, which #410 took for `purpose_required` on main, so it was rebased and renumbered.

## Version handling

Comparison follows semver ordering including pre-releases, and only moves forward. The release workflow publishes pre-release tags with `--latest=false`, so `/releases/latest` answers an rc build with the previous stable release, and treating any difference as outdated would install that older binary and call it an update. A leading `v` is stripped from both sides, and a version this CLI cannot place, such as `dev`, is refused rather than compared.

## alpacon version

`alpacon version` shares the release lookup with the update path and names `alpacon update` in its notice, printing another tool's command where that tool owns the install. It does not query ownership, which would spawn `dpkg` and `rpm` on every call.

## CI

A `windows-latest` leg runs `./pkg/selfupdate/... ./cmd`. Replacing a running executable is the one thing only Windows forbids, so its build-tagged files compile nowhere else and would ship untested. The leg is scoped to those packages because the rest of the suite asserts unix file modes. `golangci-lint` runs for `linux` and `windows`.

The matrix job is `test`, and a `build-and-test` job that `needs` it reports the flat context an `alpacax` org ruleset requires on main. A matrix emits one context per leg, which left this PR waiting for a status that never arrived. An earlier round read main as unprotected because the repository protection endpoint answers 404; the rule lives in the org ruleset instead. `release.yaml` names a job in its own file and is unaffected.

## Follow-up

- Sign the release checksums (cosign keyless or minisign) and verify against an identity compiled into the binary rather than one the release names (#412). Until then the update is integrity-protected but not authenticated.
- Release metadata freshness. `LatestRelease` reads only `tag_name`, `html_url` and `assets`. There is no `published_at`, no ETag and no persisted last-seen version, so a pinned stale-but-newer response is undetectable. Fold this into #412 rather than opening another issue.
- Apply the same aggregator shape to `lint.yaml` so the flat `golangci-lint` context comes back. Harmless while the ruleset requires `build-and-test` alone, but it breaks silently the moment `golangci-lint` is added, and both main and alpamon emit the flat one.
- `README.md:75-82` fetches the archive and the checksums from the release the CLI would have used, so it re-checks transport integrity rather than the publisher question the paragraph above it raises. Reword it honestly, and add a PowerShell equivalent for the `.zip` Windows gets.
- `warnKeptWiderMode` and `keptModeIsWider` still cannot fire for a `0755` caller. `warnWorldWritableInstall` bypassed them rather than fixing them, so repair or delete them before someone assumes they work.
- A SIGKILL leaks `/tmp/alpacon-update-*` holding the archive, the checksums and the staged binary, and nothing sweeps that path. The install directory is handled by `SweepPreserved`.
- `replace_windows.go:31` removes the staged file before attempting the restore, so a failed restore leaves neither.
- Say in one line why `maxReleaseFileSize` and `maxArchiveStreamSize` are `var` rather than `const`, the way `DefaultReleaseAPIURL` does.
- The Windows rollback branch, where the second rename fails, has no test. `osRename` is a seam that makes one possible, but it cannot be watched failing from a non-Windows machine.
- The symlink defence narrows its window rather than closing it. Closing it needs a writer that opens `O_NOFOLLOW`, which `utils.SaveStreamAtomic` deliberately does not.
- Trim the impact framing in #412, which is a public issue, and keep the detail in the private repo.
- `README.md:321` cites an internal decision-record number a public reader cannot follow. Pre-existing, so a separate cleanup issue.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -15,9 +15,9 @@ permissions:
   contents: read
 
 jobs:
-  # No `name:` on purpose: the check context is then the job id, which the
-  # matrix splits in two where a flat `build-and-test` context expects one.
-  build-and-test:
+  # A matrix reports one check context per leg, and the org ruleset on main
+  # requires the flat `build-and-test`. The job below is what reports it.
+  test:
     runs-on: ${{ matrix.os }}
 
     # Windows is not a portability formality: the files that replace a running
@@ -60,3 +60,13 @@ jobs:
 
       - name: Run Tests
         run: go test -race -v ${{ matrix.packages }}
+
+  # success only if every leg was.
+  build-and-test:
+    needs: [test]
+    if: always()
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fail unless every matrix leg passed
+        run: '[ "${{ needs.test.result }}" = "success" ]'
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -15,10 +15,27 @@ permissions:
   contents: read
 
 jobs:
-  # No `name:` on purpose. The check context is then the job id, which is the
-  # canonical single-segment `build-and-test` the other Go repos already emit.
+  # No `name:` on purpose: the check context is then the job id, which the
+  # matrix splits in two where a flat `build-and-test` context expects one.
   build-and-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+
+    # Windows is not a portability formality: the files that replace a running
+    # executable are compiled out everywhere else and would ship untested without
+    # this runner. It runs only the packages that carry them—the rest of the
+    # suite asserts Unix file modes in tests that predate this job.
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            packages: ./...
+          - os: windows-latest
+            packages: ./pkg/selfupdate/... ./cmd
+
+    defaults:
+      run:
+        shell: bash
 
     steps:
       - name: Check out code
@@ -33,6 +50,7 @@ jobs:
         run: echo "GOTOOLCHAIN=auto" >> "$GITHUB_ENV"
 
       - name: Verify go.mod and go.sum are tidy
+        if: runner.os == 'Linux'
         run: |
           go mod tidy
           git diff --exit-code go.mod go.sum
@@ -41,4 +59,4 @@ jobs:
         run: go build -v .
 
       - name: Run Tests
-        run: go test -race -v ./...
+        run: go test -race -v ${{ matrix.packages }}
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -14,8 +14,18 @@ permissions:
 
 jobs:
   golangci-lint:
-    name: golangci-lint
+    name: golangci-lint (${{ matrix.goos }})
     runs-on: ubuntu-latest
+
+    # The self-update path has a file per platform; linting one GOOS leaves the
+    # other's unread.
+    strategy:
+      fail-fast: false
+      matrix:
+        goos: [linux, windows]
+
+    env:
+      GOOS: ${{ matrix.goos }}
 
     steps:
       - name: Check out code
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -63,7 +63,7 @@ cmd/                 # Cobra command definitions
 api/                 # API client functions per domain
 client/              # HTTP client wrapper for Alpacon API
 config/              # Configuration management (credentials, workspace)
-pkg/                 # Internal packages (cert, tunnel)
+pkg/                 # Internal packages (cert, selfupdate, tunnel)
 utils/               # Shared utilities (output, prompts, errors, SSH parsing)
 ```
 
@@ -193,5 +193,5 @@ _ = json.NewEncoder(w).Encode(resp)
 - **Config file**: `~/.alpacon/config.json` (dir `0700`, file `0600`). `saveConfig` writes a sibling temp file and renames it over the target—two alpacon processes can be saving a renewed access token at the same moment, and a plain truncate-and-write leaves a config the next command cannot parse. Any new writer must go through `saveConfig`
 - **Alias**: `alpacon` can also be invoked as `ac`
 - **File transfer**: The `cp` command lives in `cmd/ftp/` (package name `ftp`)
-- **Exit codes**: `0` success, `1` general error, `2` usage error (only `work-session`, `event wait`/`watch`, and `utils.RequirePositiveInt`; every other command's own validation and Cobra parse errors still exit `1`), `3` WorkSession gate denied, `4` pending approval with the outcome still open, `5` server busy, `6` approval not granted, `7` purpose required—an agent's command held while the gate asks what it is for, answerable by the caller rather than by an approver (constants in `utils/error.go`). Keep these stable—scripts, CI, and AI agents branch on them. See README "Exit codes".
+- **Exit codes**: `0` success, `1` general error, `2` usage error (only `work-session`, `event wait`/`watch`, and `utils.RequirePositiveInt`; every other command's own validation and Cobra parse errors still exit `1`), `3` WorkSession gate denied, `4` pending approval with the outcome still open, `5` server busy, `6` approval not granted, `7` purpose required—an agent's command held while the gate asks what it is for, answerable by the caller rather than by an approver, `8` a newer release exists (only `alpacon update --check`) (constants in `utils/error.go`). Keep these stable—scripts, CI, and AI agents branch on them. See README "Exit codes".
 - **IAM**: `user` and `group` commands both live in `cmd/iam/`
--- a/README.md
+++ b/README.md
@@ -64,9 +64,9 @@ go build && sudo mv alpacon-cli /usr/local/bin/alpacon
 
 ### Updating
 
-`alpacon update` brings the CLI to the latest release. A binary built from source carries the version `dev`, which matches no release, so the command refuses it and says so—build with GoReleaser or install a released build to use it. A binary you placed yourself is downloaded, verified against the release's SHA-256 checksums, and replaced in place; the old binary is kept beside it until the replacement succeeds—on Windows, where a running executable stays locked, it is kept as `alpacon.exe.old.<timestamp>` and removed by the next update. A binary a package manager or a version manager owns is never overwritten—the command prints how to update through that tool and exits `1`, leaving the upgrade to you. Homebrew, deb and rpm get the exact command; mise and asdf are named without one, since the command depends on how the tool was set up. While it runs it holds `.alpacon-update.lock` beside the binary, so two updates cannot replace the same file at once; the file stays there afterwards and is safe to leave alone. The command only moves forward: a build ahead of the latest release, such as a release candidate, is left alone rather than downgraded. `alpacon update --check` reports whether a newer release exists and installs nothing, exiting `8` when there is one.
+`alpacon update` brings the CLI to the latest release. A binary built from source carries the version `dev`, which matches no release, so the command refuses it and says so—build with GoReleaser or install a released build to use it. A binary you placed yourself is downloaded, verified against the release's SHA-256 checksums, and replaced in place; the old binary is kept beside it until the replacement succeeds—on Windows, where a running executable stays locked, it is kept as `alpacon.exe.old.<timestamp>` and removed by the next update. A binary a package manager or a version manager owns is never overwritten—the command prints how to update through that tool and exits `1`, leaving the upgrade to you. If the ownership question cannot be answered at all—`rpm -qf` takes a read lock on the rpm database, so it is killed when another package transaction holds it—the binary is left alone rather than assumed unowned, and the command says to retry once that transaction has finished. Homebrew, deb and rpm get the exact command; mise and asdf are named without one, since the command depends on how the tool was set up. While it runs it holds `.alpacon-update.lock` beside the binary, so two updates cannot replace the same file at once; the file stays there afterwards and is safe to leave alone. The command only moves forward: a build ahead of the latest release, such as a release candidate, is left alone rather than downgraded. `alpacon update --check` reports whether a newer release exists and installs nothing, exiting `8` when there is one.
 
-If the binary is group- or world-writable, the update keeps that mode and says so on stderr rather than narrowing it—re-permissioning a file you set up is your call. Close it with `chmod go-w`, since any local account can otherwise replace the binary you run.
+If the binary or its directory is group- or world-writable, the update says so on stderr rather than narrowing anything—re-permissioning what you set up is your call. The directory is the one that usually matters: replacing a file is a rename, which needs no permission on the file itself, so a `0755` root-owned `alpacon` inside a `0775` directory can still be swapped by anyone in that group. Close it with `chmod go-w`.
 
 #### What the checksum does and does not prove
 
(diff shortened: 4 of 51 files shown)


eunyoung14 (changes requested):
This is the most carefully built self-updater I have reviewed here. Verify-before-extract, fail-closed hashing, no path traversal, an atomic same-directory replace, no elevation, and a lock that refuses a symlinked path — the mechanics below the trust anchor are right, and several of them are the exact things this feature class usually gets wrong.

**The trust anchor itself is the problem, and it needs an explicit decision rather than a code change.**

## The update path is integrity-protected, not authenticated

`FetchVerifiedBinary` picks `alpacon-<v>-checksums.sha256` out of `release.Assets` (`pkg/selfupdate/download.go:26,37,45`) — the same asset list the archive came from. And `.goreleaser.yaml` publishes no signature (no `signs:` block).

So the hash proves the bytes on the wire match the bytes the release store holds. It proves nothing about who put them there. Anyone who can write to a release — a leaked release token, a compromised maintainer account, a poisoned release workflow — uploads a trojaned archive *and* a matching checksums file, verification passes, and the CLI prints "Successfully updated". Since this CLI administers production servers, that is code execution on every operator workstation with that operator's privileges.

Signing the checksums file (cosign keyless, or minisign) and verifying it against a pinned identity before any hash is trusted is what closes it. If that is follow-up work rather than this PR — a defensible call — then please file the signing issue and say plainly in the README that updates are **checksum-verified but unsigned**. Right now the README documents the checksum and a reader would reasonably infer more than it provides.

Related, and cheap enough to fold in:

**The asset URL is used exactly as the API returned it** (`release.go:80` → `download.go:32,37,51`). `refuseSchemeDowngrade` (`download.go:91-98`) only fires when the *first* request was https, so an `http://` asset URL is fetched in the clear and the guard never runs. Not reachable while the API is honest and TLS holds — this is defense in depth — but pinning the scheme and host on both assets is a few lines.

## A group- or world-writable install silently stays that way

`utils/file_save.go:62-73,152-157` keeps the destination's mode, and `warnKeptWiderMode` cannot fire on this path: the caller passes `0755`, and `keptModeIsWider` requires `newFilePerm&0044 == 0`. So an `alpacon` sitting at `0777` from a sloppy earlier install is still `0777` after the update, silently. Any local user can then overwrite the binary the operator uses to reach production.

Keeping the mode is the right default. But a **warning** when the target is writable by group or other — not a chmod — tells the operator something worth knowing at exactly the moment they are already thinking about this binary. `TestReplaceBinaryKeepsTheDestinationMode` currently locks in the silence.

## The workflow conflict is a decision, not a rebase

The PR conflicts with main on one file, `.github/workflows/build-and-test.yaml`, against `11804e6` ("ci: report the build job under the canonical build-and-test context"). That commit **deleted** `name: Build and Test` so the check context becomes the job id `build-and-test`, matching the sibling Go repos. This branch instead sets `name: Build and Test (${{ matrix.os }})`. The two intents are directly opposed.

My suggestion is to keep main's rule and let the matrix emit `build-and-test (ubuntu-latest)` / `build-and-test (windows-latest)` — but flag that anything keying on a flat `build-and-test` context then sees two. `main` is not branch-protected today so nothing is blocking on the rename, and `release.yaml`'s `workflow_call` targets the file rather than the job name, so a tag still resolves and simply also runs the Windows leg.

## Smaller items

- **Freeze, not downgrade.** `IsOutdated` only moves forward (`version.go:35-37`), so an older release answered to the version check yields "already up to date" — rollback is correctly refused. What remains possible is holding an operator on a known-vulnerable build forever, since there is no metadata freshness. Requires breaking TLS; worth naming as residual, not fixing here.
- **Tar scanning is unbounded in decompressed bytes** (`archive.go:14,42-54`). The 512 MiB cap bounds the written entry and the download, not the inflated stream. DoS only, and only for someone who already controls the release; an `io.LimitReader` around the gzip reader closes it.
- **A killed update litters the install directory.** `SaveStreamAtomic`'s `.alpacon-<pid>-<nanos>-<i>.tmp` (`file_save.go:190`) is removed by `defer` on an error return but not on SIGKILL, and `SweepPreserved` matches only `<binary>.(old|staged).<digits>` (`preserve.go:65`).
- **Windows has a window with no unaided recovery.** Between the two renames (`replace_windows.go:27-34`) a killed process leaves no `alpacon.exe`, only `alpacon.exe.old.<ts>`. The comment says it is "swept on the next update" — but there is no next update, because the binary is gone. The operator has to rename it back by hand, and `README.md:65` does not say so.
- **Stale exit-code-7 references** after the renumber to 8: the comment at `release.go:17` and the test name `TestUpdateCommandCheckExitsSevenWhenANewerReleaseExists` (`cmd/update_test.go:188`). The assertions use the constant and are correct — only the names are stale, and confusingly so now that 7 means something else.
- **The unknown-version guard lives in the command, not the package** (`cmd/update.go:35` vs `pkg/selfupdate/update.go:28`). `compareVersions` reads an unreadable core as `0.0.0`, so an importer of this public package that skips `IsUnknownVersion` would have a `dev` build pull the latest release over itself.
- `maxReleaseFileSize` is a `var` in a public package (`archive.go:14`), mutable by any importer; `const` plus a test-only seam is tighter.

## What I checked and found clean

The host is compile-time pinned, and I grepped for `Getenv`/`LookupEnv` across `pkg/selfupdate` and `cmd/update.go` — the only hits are inside a `_test.go` file gated on a helper env var. **No env-var or flag seam ships in the binary.**

Verification happens before extraction (`download.go:45-48`), which is the ordering most self-updaters get backwards. A checksums file that does not name this archive yields an empty expectation and errors rather than skipping the check (`checksum.go:27-29`).

**No zip-slip, and structurally so** — `ExtractBinary` computes `destPath` itself (`archive.go:17`) and never uses an entry name as a path; names are only compared via `filepath.Base` for selection. Symlink and non-regular entries are skipped on both formats.

TLS is never relaxed here, and notably the repo's own `--insecure` escape hatch does **not** reach the updater. Redirects are bounded at 10, there is no `Authorization` header to leak across one, and every call has a timeout and a size cap with an over-limit error rather than a silent truncation.

The replace is atomic and done properly: temp file created `O_EXCL` **in the install directory** rather than `/tmp`, staged at `0600`, chmod'd through the handle rather than the path (with a comment saying why), then renamed on the same filesystem. Download scratch is `os.MkdirTemp`. No sudo anywhere — `permissionHint` prints the command for the operator and explicitly does not re-exec, and the only subprocesses are `dpkg -S` / `rpm -qf` with args as a slice and no shell. The lock is `flock`/`LockFileEx` beside the binary, opened `O_NOFOLLOW` as an explicit guard.

No new dependencies. And the change to `cmd/version.go` fixes a real pre-existing bug: the old code compared for inequality and called any difference an upgrade, so an rc build published with `--latest=false` was reported as needing a downgrade.

## Tests

Strong, and the failure paths are the ones covered. A bad checksum is rejected four ways, and downgrade refusal has a 15-case table including rc ordering and `rc10 > rc2`. Four gaps worth closing:

1. No test that an entry named `../../evil/alpacon` lands inside `destDir`. It is structurally impossible today — which is exactly why it deserves a regression guard, since a refactor that starts honoring `header.Name` would pass every existing test.
2. No test that a non-pinned-host or `http://` URL is refused, because that check does not exist yet.
3. No test that the installed binary is not group/world-writable.
4. `TestFetchVerifiedBinaryRefusesAMismatchedChecksum` proves the error but not the ordering — asserting the extracted file does not exist afterwards would pin "never extracted" as the invariant.

## Docs

`README.md:65` is a solid Updating section and the exit-code row is there. Three additions I would make: say the update is unsigned, tell the operator how to verify out of band (compare `alpacon version` against the Releases page, or hand-check the archive against the published checksums file), and tell a Windows operator what to do if the update dies between the renames.

Separately, and pre-existing rather than yours: `README.md:321` cites an internal decision-record number that a public reader cannot follow. Worth a cleanup issue.

CI is green on `199827c9a` across all nine checks including both `Build and Test` legs and both `golangci-lint` legs.

---

To merge: settle the signing question explicitly (fix or file-and-document), and resolve the workflow conflict as a deliberate choice about the check context. The scheme/host pin and the mode warning are small enough to fold in now.

eunyoung14 (approved):
Both merge conditions are settled, and most of the smaller items landed. **One thing I did not raise last round now matters more than anything I did**, so I am leading with it.

## The squash body would publish internal session URLs

`squash_merge_commit_message` on this repo is `COMMIT_MESSAGES`, so the squash body is pre-filled with every commit message concatenated. **Eight of the twenty commits here carry a `Claude-Session:` trailer with an internal session URL** — `6642560`, `a59d323`, `ed6575c`, `c2b0a27`, `e54f51d`, `1713bed`, `50f6a53`, `a0c9656`. Merging on the default prefill lands those permanently in a public repository's history.

`#409` has **zero** such trailers. So this is not a new standard I am inventing at you — it is one the sibling PR already meets, and this branch drifted from it.

Same body, one more thing: **`e54f51d`'s message still says the displaced Windows copy "is swept by the next update."** You corrected that claim in the code comment in `734005e`, correctly, because there is no next update when the binary is gone. The message was not corrected, so the false sentence rides into the squash body.

Replace the body at merge time — `gh pr merge 408 --squash --body "…"` — or flip the repo setting first.

## Condition 2 — resolved correctly, and it matches main's intent

`42cb3ee` is the real resolution, and it is the right shape: the matrix job is renamed `test`, and a separate aggregator reports the canonical flat context.

```yaml
build-and-test:
  needs: [test]
  if: always()
  steps:
    - run: '[ "${{ needs.test.result }}" = "success" ]'
```

That satisfies main's rule rather than papering over the conflict, and I confirmed `build-and-test` is a reported, successful check-run on `8337a6a4f` — it is the org ruleset's only required context. `release.yaml`'s `needs:` names a job in its own file and is unaffected. Your commit message also corrects the earlier "main is not branch-protected" reading, which was mine to begin with.

**One residual:** the same aggregator shape was not applied to `lint.yaml`, so the flat `golangci-lint` context disappears on this branch — `golangci-lint (linux)` / `(windows)` instead. Harmless today because the ruleset requires `build-and-test` alone, but it breaks silently the moment anyone adds `golangci-lint` to the ruleset, and both `main` and `alpamon` emit the flat one. Your PR body acknowledges it; worth a follow-up rather than a round.

## Condition 1 — settled by the documented-decision route

#412 states the threat model accurately — same-release provenance, the attacker classes, the right fix **in the right order** (signature, then checksum, then extraction, against an identity compiled into the binary rather than one the release names), and it correctly scopes out binary signing as a different problem.

The README says unsigned plainly and explains both halves of what the checksum proves. That is what I asked for.

**Two things to weigh, neither blocking:**

**The out-of-band procedure is not out of band.** `README.md:75-82` fetches the archive *and* the checksums from the same GitHub release the CLI would have used, so it re-checks transport integrity — precisely not the publisher-compromise threat the paragraph immediately above it describes. It is runnable and the asset names are right, but "to check out of band" overstates it. "Verify a second copy by hand" would be honest; a real out-of-band anchor does not exist until #412 lands. There is also no PowerShell equivalent, despite the comment noting Windows ships `.zip`.

**#412 is a public issue.** It spells out the exploit path and the operator-workstation impact for a weakness that is unfixed and shipping. Our rule keeps exploitability in the advisory rather than the public tracker. Consider trimming the impact framing there and keeping the mechanics in the private repo.

## Your other items

Verified: the origin pin (exact equality on `scheme://host`, checked as the first statement of `downloadTo`, so `github.com.evil.test` and `http://` both fail up front); the inflate bound; the Windows comment now stating plainly that nothing recovers it; the exit-code renumber with no stale `7` anywhere; the unknown-version guard moved into `Run` as its first statement with a test driving it directly; and all four test gaps I named — including the literal `"../../evil/alpacon"` entry and the lookalike-host case.

**Three worth a word:**

**The world-writable check is a new one, not a repair.** `warnWorldWritableInstall` inspects both the file and the directory, and a `0755` file in a `0777` dir does warn — good, and better than what I asked for. But the old `warnKeptWiderMode` / `keptModeIsWider` is unchanged and **still cannot fire for a `0755` caller**. You bypassed it rather than fixed it, which was the right call; I am noting it so nobody later assumes that helper works.

**Test 12 asserts the opposite of hardening, deliberately.** `replace_unix_test.go:34` warns at `0777` and then asserts the mode is *still* `0777` — "the warning replaces a chmod; it must not become one." That is the stated policy and I agree with it. Saying so here so the intent is on the record rather than read as a miss.

**Freshness is still absent.** Freeze-not-downgrade is verified. But `LatestRelease` reads only `tag_name`, `html_url` and `assets` — no `published_at`, no ETag, no persisted last-seen version — so a pinned stale-but-newer response is undetectable, and #412 does not mention it. Fold it into #412 rather than opening another.

Two small residues: a SIGKILL leaks `/tmp/alpacon-update-*` holding the archive, checksums and the staged binary, and nothing sweeps that path — so "nothing is left behind" is not quite true even though the install directory is now handled well. And `replace_windows.go:31` removes the staged file *before* attempting the restore, so a failed restore leaves the operator with neither. Nit, and the README covers both leftovers.

On `maxReleaseFileSize`: my phrasing was off. Both it and the new `maxArchiveStreamSize` are **unexported**, so no importer can mutate them and the exposure is smaller than I implied. Only tests assign them, each with `t.Cleanup`. A one-line comment saying they are vars for testability — the way `DefaultReleaseAPIURL` does — closes it.

CI is green on `8337a6a4f` across all ten checks, including both `test` legs and both lint legs.

Approving. Merge it with the squash body replaced.

PR #409, opened 2026-09-01
feat(iam): move user promote/demote onto the RBAC role API (#270)

### LINKED ISSUE #270: [RBAC Hardening] Verify/migrate CLI user promote/demote off PATCH flags
Part of the RBAC Hardening epic (alpacax/alpacon-server#2416) — follow-up to R1-08 (alpacax/alpacon-server#2417).

## Background

alpacon-server is unifying admin/superuser promotion/demotion onto the **RBAC role API** (`/api/rbac/user-roles/`); the user-edit flags (`is_staff` / `is_superuser` on `PATCH /api/iam/users/{id}/`) will become read-only (Option A). During the transition (Option B) the flag PATCH still works.

## Scope (alpacon-cli)

- **Verify** whether the CLI sets `is_staff` / `is_superuser` via the user endpoint (e.g. a user promote/demote command or a `--superuser` / `--staff`-style flag).
- If it does → migrate to the RBAC role API (grant/revoke the `admin` / `superuser` roles) before Option A lands.
- If it does not → close as not-applicable.

## Refs

- Epic: alpacax/alpacon-server#2416
- R1-08: alpacax/alpacon-server#2417

geunwoonoh:
Closes #270. Context: alpacax/alpacon-server#3324

## What this changes

`is_staff` and `is_superuser` on a user are read-only projections of the account's RBAC roles: the server drops them from a `PATCH` and answers `200` with the flags unchanged, never an error. The CLI had no way to change a privilege, and `alpacon user update` reported success for edits that were never applied — it round-tripped the whole user detail body through `$EDITOR` and PATCHed it back, flags included.

```bash
alpacon user role ls [USER]                        # bindings held, with scope
alpacon user role catalog                          # roles this workspace defines
alpacon user role describe ROLE                    # what it grants, and who holds it
alpacon user role grant  USER ROLE                 # --reason --dry-run -y
alpacon user role revoke USER ROLE                 # --cascade --reason --dry-run -y
alpacon user role history [USER]                   # who changed what, and why
alpacon user permission ls [USER]                  # effective access  [--patterns]
alpacon user permission can-i [USER] PERMISSION    # -q --explain
```

`alpacon user update` now builds a sparse patch of only the fields the editor actually changed. A privilege-flag edit is held back and reported with the exact `alpacon user role` command that performs it, and such an edit never exits 0 — whether or not the rest of it applied. The sparse patch also stops `is_ldap_user` being re-submitted on every edit, which made the server run a live LDAP lookup for a field nobody touched.

`alpacon user create`'s Staff and Superuser prompts are unchanged. The create path is a different contract and writes the role rows correctly.

## Why this command shape

Surveyed kubectl, Teleport (`tsh`/`tctl`), AWS/gcloud/Azure IAM, HashiCorp Boundary/Vault/Consul, Auth0/Okta/1Password/GitHub/Snowflake, plus clig.dev, the Heroku style guide, the kubectl sig-cli conventions and the gcloud/Azure CLI guidelines.

Counted across every tool, the surface hangs off the **role** noun 6 to 3. But those six are cloud IaaS CLIs whose principals are usually machines. Restricted to tools whose subject is a **human** in an identity provider or SaaS workspace, the convention inverts: Auth0 ships `auth0 users roles assign` and Teleport puts the account first. This API binds exactly one principal kind, so gcloud's typed `--member=user:` prefix and kubectl's repeatable `--user/--group/--serviceaccount` would be ceremony here.

| Name | Precedent | Why |
|---|---|---|
| `grant` / `revoke` | SQL, 1Password | The oldest and most widely understood pair in access control (`GRANT ROLE analyst TO USER jane`). The `TO`/`FROM` flip keeps word order identical. |
| `can-i` | kubectl | Its exact verb, including `-q`. The subject is optional, so one argument asks about yourself. |
| `catalog` | Azure, Auth0, gcloud | Every surveyed tool separates the catalog from a subject's bindings. `ls` is already the binding list on the same group. |
| `describe ROLE` | gcloud, kubectl | Carries the holder list too, so one command answers both "what does this role grant" and "who holds it" — the second is the question kubectl's RBAC surface cannot ask at all. |

No new top-level noun, no `cmd/root.go` change, and **no new `--role` flag** — `--role` occurs exactly once in `cmd/` today (`cmd/iam/member_add.go`, the group-membership tier), so after this it still means one thing binary-wide. A test walks the whole `UserCmd` tree to keep it that way.

Decisions the survey settled:

- **`--cascade` is opt-in.** The one rule the survey is unanimous on: breadth is added only on the revoke side and must be typed (gcloud's `--all` on `remove-iam-policy-binding`, absent from `add-`). Never delete a row the operator did not name.
- **`--reason` is optional, with a warning at the platform tier.** A justification flag is idiomatic; a required one is not — gcloud PAM brackets `[--justification]` in its flagship JIT product, Teleport defers requiredness to server-side policy, and NIST SP 800-53 Rev.5 AC-6(9) demands *logging*, not operator-typed justification. `user role history` is what keeps it from being write-only decoration.
- **No declarative `set` verb.** Boundary can ship `set-principals` only because its server takes the whole desired set in one request; the atomicity is the server's. A client-side diff over a non-transactional API has none of it, and a `set` whose diff computes an unmentioned DELETE demotes a colleague as a side effect of an incomplete command line.
- **Object-scoped writes deferred, the API layer scope-typed from day one.** `BindingCreateRequest` already carries the scope pair, so adding a flag later is purely additive.
- **No branch on deployment mode.** kubectl bans version sniffing, and the CLI cannot locally distinguish the two deployments that behave differently here anyway. The difference surfaces only as a rewritten runtime error keyed off the actual response.
- **Idempotent convergence.** A grant of a role already held, and a revoke of one not held, both change nothing and exit 0. Neither verb claims success from a 2xx alone: both re-read the bindings and report the rows actually observed.

## Behaviour worth knowing before merge

- Granting `superuser` creates a companion `admin` binding server-side, so revoking `superuser` demotes to admin and leaves that companion; `--cascade` removes both, superuser first. With no superuser binding to start from, `--cascade` revokes nothing rather than reaching for the companion.
- Revoking `admin` from someone who still holds `superuser` is refused before anything is sent — the account would keep every platform flag while no longer registering as an admin. Revoking a role a user does not hold is never refused.
- Changing a binding needs a workspace superuser and recent MFA. On Alpacon Cloud workspaces an API token is refused on the `alpacon user role` commands, so they need a browser login; `user role history` and the `user permission` commands are the carve-outs, and the README says which.
- `alpacon user update` exits **1** when the edit touched a privilege flag. That is the only behaviour change a wrapper script would notice, and it is in the exit-code table.

## Verification

- `go build ./...`, `go test -race ./...`, `go vet ./...`, `gofmt -l .` — all clean; CI green.
- Every command exercised end to end against a stub implementing the verified contract, including both `user update` regressions, the `--cascade` guard, the superuser companion, the last-holder refusals, the self-form request paths and the `-q` exit codes.
- Contract tests pin the grant body with one `JSONEq` over the whole document, the revoke reason never reaching the query string, the absence of an `ordering` parameter, the multi-page walk, the IAM-hosted read paths, and the full refusal-message matrix per gate and credential with negative assertions.
- Exit-code contract unchanged: every own-validation failure exits `1`.

## Commits

18 commits, each building on its own (checked by replaying each into a scratch worktree), plus the review round. Read-only surface first, then the write verbs, then the `user update` fix, then permission introspection, then tests and docs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cmd/iam/user_role_list.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/alpacax/alpacon-cli/api/rbac"
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+var userRoleListCmd = &cobra.Command{
+	Use:     "ls [USER]",
+	Aliases: []string{"list"},
+	Short:   "List the role bindings a user holds",
```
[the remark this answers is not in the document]
geunwoonoh: Not applied — the convention is `["list"]` only.

`CLAUDE.md:100` states it as: *"**Subcommand aliases**: list → `["list"]`, delete → `["rm"]`, describe → `["desc"]`"*. Every one of the 24 existing `ls` commands in the repo matches that exactly (`cmd/iam/user_list.go`, `cmd/server/server_list.go`, `cmd/token/token_list.go`, `cmd/approval/approval_list.go`, …) and none carries an `"all"` alias.

The only `"all"` in the tree is a **flag value**, not a command alias: `work-session ls --status all` / `--user all`, resolved in `cmd/worksession/worksession_list.go:18,30`. Adding `"all"` here would depart from the documented convention rather than match it, and would make these two commands the only `ls` in the binary that answers to a third name.

cmd/iam/user_permission_list.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/alpacax/alpacon-cli/api/rbac"
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+var userPermissionListCmd = &cobra.Command{
+	Use:     "ls [USER]",
+	Aliases: []string{"list"},
+	Short:   "List a user's effective permissions",
```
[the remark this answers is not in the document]
geunwoonoh: Not applied — the convention is `["list"]` only.

`CLAUDE.md:100` states it as: *"**Subcommand aliases**: list → `["list"]`, delete → `["rm"]`, describe → `["desc"]`"*. Every one of the 24 existing `ls` commands in the repo matches that exactly (`cmd/iam/user_list.go`, `cmd/server/server_list.go`, `cmd/token/token_list.go`, `cmd/approval/approval_list.go`, …) and none carries an `"all"` alias.

The only `"all"` in the tree is a **flag value**, not a command alias: `work-session ls --status all` / `--user all`, resolved in `cmd/worksession/worksession_list.go:18,30`. Adding `"all"` here would depart from the documented convention rather than match it, and would make these two commands the only `ls` in the binary that answers to a third name.

cmd/iam/user_role_grant.go
```diff
@@ -0,0 +1,144 @@
...
+catalog' to see the names this workspace defines.`,
+	Example: `  alpacon user role grant john operator --reason "on-call rotation Q3"
+  alpacon user role grant john superuser --reason "SEC-1421" -y
+  alpacon user role grant john auditor --dry-run`,
+	Args: cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		reason, _ := cmd.Flags().GetString("reason")
+		dryRun, _ := cmd.Flags().GetBool("dry-run")
+		yes, _ := cmd.Flags().GetBool("yes")
+
```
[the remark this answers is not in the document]
geunwoonoh: Good catch — fixed in a831518.

The two were genuinely out of step: `warnMissingReason` judged `strings.TrimSpace(reason)` while `request.Reason` carried the untrimmed value, so `--reason "   "` warned that the audit entry would carry no justification *and* filled it with blanks. That is worse than the omission the operator meant, because a non-empty `reason` also keeps the server-side `warn_if_unjustified` quiet.

The flag is now read and trimmed in one place, `reasonFlag`, shared by `grant` and `revoke`, and `warnMissingReason` compares against `""`. Since `BindingCreateRequest.Reason` is `omitempty`, an omission now reaches the server as a genuinely absent field.

Verified against a stub implementing the endpoint: `--reason "   "` puts `{"user": …, "role": …}` on the wire with no `reason` key, and `--reason "  SEC-1421  "` sends `{"reason":"SEC-1421"}`. `TestReasonFlagTrimsOnce` pins all four cases.

cmd/iam/user_role_revoke.go
```diff
@@ -0,0 +1,155 @@
...
+Revoking a role the user does not hold changes nothing and succeeds.`,
+	Example: `  alpacon user role revoke john operator
+  alpacon user role revoke john superuser --cascade --reason "rotated off on-call"
+  alpacon user role revoke john admin --dry-run`,
+	Args: cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		reason, _ := cmd.Flags().GetString("reason")
+		dryRun, _ := cmd.Flags().GetBool("dry-run")
+		yes, _ := cmd.Flags().GetBool("yes")
+		cascade, _ := cmd.Flags().GetBool("cascade")
```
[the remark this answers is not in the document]
geunwoonoh: Good catch — fixed in a831518.

The two were genuinely out of step: `warnMissingReason` judged `strings.TrimSpace(reason)` while `request.Reason` carried the untrimmed value, so `--reason "   "` warned that the audit entry would carry no justification *and* filled it with blanks. That is worse than the omission the operator meant, because a non-empty `reason` also keeps the server-side `warn_if_unjustified` quiet.

The flag is now read and trimmed in one place, `reasonFlag`, shared by `grant` and `revoke`, and `warnMissingReason` compares against `""`. Since `BindingCreateRequest.Reason` is `omitempty`, an omission now reaches the server as a genuinely absent field.

Verified against a stub implementing the endpoint: `--reason "   "` puts `{"user": …, "role": …}` on the wire with no `reason` key, and `--reason "  SEC-1421  "` sends `{"reason":"SEC-1421"}`. `TestReasonFlagTrimsOnce` pins all four cases.

cmd/iam/rbac_errors.go
```diff
@@ -0,0 +1,95 @@
...
+		switch {
+		case gate == gateUserRead:
+			return errors.New("reading another account's permissions requires the user:read permission on that account; your own are always readable")
+		case gate == gateRoleWrite && !ac.IsBearerAuth():
+			return errors.New("a role-binding write requires the superuser role, and this credential may be refused outright: the RBAC API accepts no API token on an Alpacon Cloud workspace. Run 'alpacon login' to authenticate through the browser")
+		case gate == gateRoleWrite:
+			return errors.New("a role-binding write requires the superuser role")
+		case !ac.IsBearerAuth():
+			return errors.New("the RBAC API refuses API tokens on Alpacon Cloud workspaces, reads included. Run 'alpacon login' to authenticate through the browser; 'alpacon user role history' is the exception and needs a token carrying the role_audit_log:read scope")
+		default:
```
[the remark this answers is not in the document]
geunwoonoh: Valid — fixed in the push above.

You are right that it reads as a contradiction, and the cause is that I had one message covering two *opposite* failures. Worth spelling out, because your suggested split is close but inverts one detail:

- `role-audit-logs` is the one `/api/rbac/` route carrying `auth0_guard_exempt`, so an API token **reaches it on every deployment**, Alpacon Cloud included. Its gate is `HasTokenScope.for_resource("role_audit_log")`, which is not deployment-specific — so the scope requirement is not a self-hosted thing.
- Every other `/api/rbac/` read is behind `AUTH0_GATE`, where `IsAuth0Authenticated` refuses an `APIToken` outright on Alpacon Cloud.

So one route refuses a token for *lacking a scope*, and the rest refuse it for *being a token at all*. No single sentence covers both without an "except" clause, which is what you were reading.

Rather than reword, the audit log now has its own gate (`gateAuditRead`) and `alpacon user role history` passes it. Neither message mentions the other case:

- role read + token → "the RBAC API refuses API tokens on Alpacon Cloud workspaces, reads included. Run `alpacon login`…"
- audit read + token → "this API token is missing the `role_audit_log:read` scope, which the role history requires. Widen the token's scopes, or run `alpacon login`…"

`TestDescribeRBACError_CodelessForbidden` now asserts the negatives too: the role-read message must not contain `role_audit_log:read` or "exception", and the audit-read message must not contain "refuses API tokens".

README.md
```diff
@@ -199,6 +199,21 @@ $ alpacon group member add --group <group> -u <user> --role <role>
...
+$ alpacon user role ls <username>                   # roles a user holds, and at what scope
+$ alpacon user role catalog                         # roles this workspace defines
+$ alpacon user role describe <role>                 # what it grants, and who holds it
+$ alpacon user role grant  <username> <role> --reason "<why>"
+$ alpacon user role revoke <username> superuser --cascade
+$ alpacon user role history <username>              # who changed what, and why
+$ alpacon user permission can-i <username> server:update
+```
+
+Granting `superuser` also creates a companion `admin` binding, so revoking `superuser` demotes to admin and leaves that companion in place; `--cascade` removes both. Changing a binding requires a workspace superuser and recent MFA. On Alpacon Cloud workspaces every RBAC request needs an interactive browser login—an API token is refused, reads included.
```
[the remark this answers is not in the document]
geunwoonoh: Both valid, and both mine — fixed in bac0673.

You caught the same thing in two places, and the cause is worth naming: the previous round split `gateAuditRead` off in *code* precisely because the audit log is the exception, and I did not carry that correction into the prose. The claim was blanket in three sites, not two — `alpacon user role history` never said it is the exception, which is where an operator hitting the refusal would most want to read it.

The accurate split, for the record:

- `RoleAuditLogViewSet` carries `auth0_guard_exempt`, so an API token reaches `/api/rbac/role-audit-logs/` on **every** deployment. Its gate is `HasTokenScope.for_resource("role_audit_log")`.
- Every other `/api/rbac/` route sits behind `AUTH0_GATE`, where `IsAuth0Authenticated` refuses an `APIToken` outright on Alpacon Cloud.

All three now say which is which:

- **`README.md`** — "an API token is refused on these commands, reads included … `alpacon user role history` is the one exception: it reads the audit log, which accepts a token carrying the `role_audit_log:read` scope on either deployment."
- **`cmd/iam/user_role.go`** — same, as its own paragraph so it is not buried in the MFA sentence.
- **`cmd/iam/user_role_history.go`** — states it directly: "This is the one `alpacon user role` command an API token can reach on an Alpacon Cloud workspace."

`grep -rn "every RBAC request"` over the tree now returns nothing outside `CLAUDE.md`, which does not carry the claim.

cmd/iam/user_role.go
```diff
@@ -0,0 +1,84 @@
...
+platform flags set.
+
+These are not group membership roles. 'alpacon group member add --role' sets a
+member's tier within a group (owner, manager, member), a different axis from a
+workspace RBAC role.
+
+Changing a role binding requires a workspace superuser and recent multi-factor
+authentication. On self-hosted workspaces an API token is accepted and skips the
+MFA step. On Alpacon Cloud workspaces every RBAC request—reads included—requires
+an interactive browser login; run 'alpacon login' first.`,
```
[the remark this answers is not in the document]
geunwoonoh: Both valid, and both mine — fixed in bac0673.

You caught the same thing in two places, and the cause is worth naming: the previous round split `gateAuditRead` off in *code* precisely because the audit log is the exception, and I did not carry that correction into the prose. The claim was blanket in three sites, not two — `alpacon user role history` never said it is the exception, which is where an operator hitting the refusal would most want to read it.

The accurate split, for the record:

- `RoleAuditLogViewSet` carries `auth0_guard_exempt`, so an API token reaches `/api/rbac/role-audit-logs/` on **every** deployment. Its gate is `HasTokenScope.for_resource("role_audit_log")`.
- Every other `/api/rbac/` route sits behind `AUTH0_GATE`, where `IsAuth0Authenticated` refuses an `APIToken` outright on Alpacon Cloud.

All three now say which is which:

- **`README.md`** — "an API token is refused on these commands, reads included … `alpacon user role history` is the one exception: it reads the audit log, which accepts a token carrying the `role_audit_log:read` scope on either deployment."
- **`cmd/iam/user_role.go`** — same, as its own paragraph so it is not buried in the MFA sentence.
- **`cmd/iam/user_role_history.go`** — states it directly: "This is the one `alpacon user role` command an API token can reach on an Alpacon Cloud workspace."

`grep -rn "every RBAC request"` over the tree now returns nothing outside `CLAUDE.md`, which does not carry the claim.

cmd/iam/user_role_history.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"github.com/alpacax/alpacon-cli/api/rbac"
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+var userRoleHistoryCmd = &cobra.Command{
+	Use:   "history [USER]",
+	Short: "Show the role grants and revocations recorded for a user",
```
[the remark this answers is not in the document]
geunwoonoh: Valid — fixed in the push above. It was 56 characters, and `CLAUDE.md` says *"`Short`: Concise, starts with a verb, under 50 chars"*.

New: `Show who changed a user's roles, and why` (40). Shorter, still verb-first, and it names what the command is actually for — a binding read carries neither the actor nor the reason, so this is the only surface that has them.

I measured every `Short` I added rather than just this one; the other nine are 32–40 characters, so this was the only breach.

While I was in there I cross-checked each behavioural claim in these commands against the code implementing it, and that turned up three more of my own that no review had flagged yet:

- The `--cascade` flag help said the flag is "ignored" when there is no superuser binding. It is not — `plannedRevocations` returns an empty plan, so *nothing at all* is revoked. The wrong wording reads as "the admin row still goes", which is the opposite of the safety property the last round added.
- `revoke`'s `Long` stated both "revoking an unheld role succeeds" and "revoking admin while superuser is held is refused" with nothing to say which wins. A reader in the half-state could not tell. The no-op now leads and says it outranks the rest, matching the code order.
- The group help said the admin-while-superuser revoke is refused, full stop. It is refused only when there *is* an admin binding to delete.

eunyoung14 (changes requested):
The contract work here is the most thoroughly verified I have reviewed in a while. I checked every endpoint, query parameter and response field against the server rather than against the description, and found **no path the server does not serve, no field it does not emit, and no required field omitted**. Pagination is handled on all three paginated routes, the bulk-vs-scalar behavior is described exactly as the server implements it, and the `DELETE`-with-body reason matches the server's own reasoning for why it is not a query parameter. CI is green on `3bff94e56`, and all eight Copilot threads were resolved with real code — I verified each against the diff rather than trusting the replies.

**One thing has to change before this merges, and it is not in the code.**

## The PR description is public, and it should not be

This repository is open source. The description currently carries several categories of content our rules keep internal:

- issue references into two private repositories
- a private handbook document, cited by path and by finding number
- ADR numbers, together with an open governance question and its own assessment of the control it covers
- server source file paths with line numbers
- internal authorization mechanics named directly

The governance paragraph is the one I would move first. It states publicly that the control currently guarding a privileged write is weaker than what our own decision record requires, and that the question is unresolved. That is a security-posture disclosure, and it is exactly what "no internal detail in public" is for.

The technical content is good and I want it kept — please move the epic, ADR, handbook and server-citation context into an `alpacon-server` issue and reference that issue here without describing it. What belongs in the public description is what the diff shows: what the commands do, the CLI-side design rationale, and the verification you ran.

One caveat so the fix is not mistaken for a redaction: GitHub retains PR body edit history and it stays visible to anyone who can see the PR. Editing narrows further exposure; it does not undo it. Please handle it as a disclosure that already happened.

The code, help text, README and commit messages are clean — no non-English text, no internal hostnames, no customer names, no attack mechanics. The two comments that name a server file and an ADR follow long-standing practice already on `main`, so this PR introduces nothing new there.

## Correctness

### 1. `permission_denied` is unmapped, and it is the code `can-i --explain` gets

`describeRBACError` (`cmd/iam/rbac_errors.go:57-72`) maps five coded refusals plus the codeless 403, and the codeless branch at `:74` is gated on `code == ""`. A **coded** 403 therefore falls through to the raw error.

The troubleshoot endpoint raises exactly such a code when the target is not the caller, and `cmd/iam/user_permission_cani.go:51` routes that response through `describeRBACError`. The `gateRoleRead` comment at `rbac_errors.go:26` already names troubleshoot as the coded exception — the case is acknowledged and then not handled.

A workspace admin without superuser running `alpacon user permission can-i john rbac:manage --explain` gets a raw error envelope, while every sibling path gives guidance.

### 2. `user role history` cannot return the rows the help promises

The server's `?user=` filter pins the **subject** axis only, while the non-privileged visibility queryset is subject **or** actor. `api/rbac/rbac.go:98-104` always sends `?user=<id>`, which intersects the actor half away.

`cmd/iam/user_role_history.go:20-21` tells the operator they will see rows "naming them as the subject or the actor". A non-auditor looking for the grant they made yesterday will never find it. Either drop the filter on the self form, or narrow the help to the rows recorded against them.

### 3. A self-hosted token user is sent to a browser login that cannot help

The token arms at `rbac_errors.go:80-81` and `:88-89` are unconditional on credential kind, but the refusal they describe only exists when the Auth0 gate is enabled. On a self-hosted workspace that gate is not installed at all, so a codeless 403 with an API token is a visibility or scope limit, not a credential refusal.

Two consequences: the operator is told to run `alpacon login`, which changes nothing, and the accurate default at `:91` is unreachable for every token session. At `:80-81` the same shape tells a token that *has* the audit scope that it lacks it, when the right answer is the message at `:83`.

The PR reasons correctly that `config.IsSaaS()` cannot separate the two deployments. Consider leading with the visibility reading and offering the credential refusal as the alternative, rather than the reverse.

## Smaller items

- **`role describe` fetches the whole user directory to label holders**, and does it even when the role has none. `cmd/iam/user_role_describe.go:55-60` calls `iam.GetUsernamesByID`, which pages `/api/iam/users/` at 100 per page — roughly 100 sequential round trips on a large workspace. Skipping on `len(holders) == 0` is one line.
- **`GrantRole` decodes a body no caller uses** (`api/rbac/rbac.go:230-241`; the caller discards it at `cmd/iam/user_role_grant.go:75`). The server answers the bulk branch with 201 and an empty body, where `json.Unmarshal` fails — latent today because the CLI never sends that shape, but the decode buys nothing and carries the risk.
- **`--reason` on a `DELETE` body has no delivery confirmation.** The contract is right and the server reads it, but an intermediary that strips a `DELETE` entity-body drops the justification silently: the revoke still returns 204 and the audit row records an empty reason. Worth a comment at `client/client.go:444`, since `user role history` is the reason the flag exists.
- **`can-i -q --output json` prints nothing** — `cmd/iam/user_permission_cani.go:63-68` returns before the JSON branch. The help at `:27-29` reads as though the two combine.
- `describeRBACError` returns fresh `errors.New` values rather than wrapping with `%w`, discarding the server's message and the status from every rewritten path.

## Tests

What is pinned is pinned well: the grant body via `JSONEq` over the whole document, the reason never reaching the query string, the absence of an `ordering` parameter, the multi-page walk, and the full codeless-403 message matrix per gate and credential **with negative assertions**. That is contract-pinning rather than happy-path exercise.

The gap worth closing: **nothing touches the three IAM-hosted reads.** No test hits the permissions or effective-permissions endpoints. `PermissionCheckResponse.Allowed` is the only field read from the check endpoint, so a rename on the server side decodes to `false` — `can-i` prints "no" and `-q` exits 1. Fail-closed, but silently wrong, and the rest of this surface would catch the same rename. `GET /roles/{id}/scopes/` builds its path by string concatenation (`rbac.go:72`) with nothing asserting the result, unlike `TestResolveRole_ByUUID`.

## One README line

There was never a `promote`/`demote` command, so nothing is removed or renamed — the title describes moving the flow, which is accurate.

The one script-visible change is that `alpacon user update` now exits **1** when the edit touched a privilege flag, where it previously exited 0 and printed success. That is the correct fix, and it is documented in the command's `Long` — but not in `README.md`. The exit-code table's `1` row was updated only for `can-i -q`. Please add a clause there; it is the only behavior change a wrapper script would notice, and the README is where its author would look.

`README.md:212` also has one extra space before its trailing comment, breaking the block's alignment.

---

The description is the blocker. Everything else is small, and the contract verification behind this is genuinely good work.

jisung-02 (commented):
## Review at `d6d4373`

I checked the three commits that landed after the last review (`07d112d`, `0a1386a`, `d6d4373`) against the diff, and every item raised there is addressed. This review covers only what is new on top of that.

What I ran locally, all clean: `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues), `go test ./api/... ./cmd/...`. CI is green on all six checks. The new tests read no clock and no timezone, and no httptest handler carries an assertion, so `-race -count=20 -shuffle=on` and a `TZ=Pacific/Kiritimati` run both stayed stable.

### Critical issues (0)

None.

### Warnings (1)

**1. `cmd/iam/user_role.go:89-98` — naming yourself is not the self form.**

`resolveSubject` takes the `"-"` alias only on the no-argument branch. Pass your own username or UUID and `PK` becomes the UUID, which is exactly the addressing `gatePermissionIntrospect` documents as refused: `/api/iam/users/{id}/permissions/` pins no permission of its own, so a by-UUID read auto-resolves to an orphan `user:permissions` that only the superuser wildcard grants.

For a workspace admin without superuser:

```
alpacon user permission ls --patterns              # works
alpacon user permission ls jane --patterns         # 403, jane is the caller
alpacon user permission can-i server:update        # works
alpacon user permission can-i jane server:update   # 403, same account
```

The refusal message already says "run the command without a USER argument", so the case looks understood. It has the same shape as the `permission_denied` gap fixed in `07d112d`, and it would be good to close it here too. `resolveSubject` already calls `iam.GetCurrentUser` on one branch, so comparing the resolved id against the caller's and swapping in `selfUserPK` would cover it.

`TestResolveSubject` has three cases and none is "the caller names themselves", so a fourth would keep it closed. The README also presents `alpacon user permission ls <username>` as the general form.

### Suggestions (6)

**1. `cmd/iam/rbac_errors.go:77-78` — `invalid_input` is generic but rewritten into a write-only sentence.**

Running the coded switch ahead of the gate is deliberate, and `TestDescribeRBACError_CodedRefusalIgnoresTheGate` pins it. That is right for `rbac_superuser_last_removal_forbidden`, a code only a write can produce. `invalid_input` is different: eight of the ten `describeRBACError` call sites are reads, and none of them binds a scope. `can-i` lets any argument containing `:` or `*` through `looksLikePermission`, so a malformed permission reaches the server, and a rejection there would print "the server rejected the binding scope" from a command that binds nothing.

Scoping the case to `gateRoleWrite`, or giving it the same kind of comment the duplicate case above it has, would settle it. `workspace_suspended` ("so it accepts no changes") reads the same way on `user role ls`.

**2. `cmd/iam/user_role_grant.go:58-61`, `cmd/iam/user_role_revoke.go:72-77` — `--dry-run` discloses less than the prompt it replaces.**

The dry-run branch returns before both `warnMissingReason` and the confirmation. Previewing a superuser grant therefore never mentions the companion `admin` binding the server creates, which is the one fact the interactive path goes out of its way to state, and it never warns that the audit entry will carry no justification. Those two are exactly what someone reaches for `--dry-run` to see before a privileged write.

**3. `api/rbac/rbac.go:169`, `:194`, `:232` — nothing asserts the time columns.**

`BindingAttributesFrom`, `HolderAttributesFrom` and `AuditAttributesFrom` all build their time string with `AddedAt.Local().Format(timeLayout)`, and no test reads the resulting field. `TestAuditAttributesFrom_ToleratesMissingActorAndRole` checks `ChangedBy`, `Role` and `Reason`; `TestHolderAttributesFrom_NilMapPrintsIDs` checks `User`.

So a changed `timeLayout`, a dropped `.Local()`, or a row without `added_at` rendering as `0001-01-01 00:00` would all pass. The audit log exists to answer "who changed what, when", and the "when" is the unpinned part.

One fixed UTC instant asserted against the rendered string would cover it. Because of `.Local()` the expected value depends on the runner's timezone, so it needs either a pinned `time.Local` or a comparison on the `time.Time` itself. The current suite is stable precisely because it never touches timezones, and it would be good to keep that property while adding this.

**4. `api/iam/types.go:152-160` — the JSON tags on `UserEdit` and `PrivilegeEdit` are never used.**

Neither struct is ever marshalled. `cmd/iam/user_update.go:56` passes `edit.Changes`, a map, and `PatchUser` marshals that map. Every other type in this file is a wire type whose tags are live, so the next reader will take these four as evidence that the server accepts `{"changes": ..., "privileges": ...}`.

Leaving off the `~Response`/`~Request` suffix reads as deliberate and I think it is right: these are not wire types. One small thing, `PrivilegeEdit.Want` is a thin name for a bool. `Enable` or `Desired` says what is wanted at the call site.

**5. `api/iam/user_edit_test.go` — the filename points at a file that does not exist.**

All three tests exercise `diffEditedUser` only, and `diffEditedUser` lives in `api/iam/iam.go:358`. There is no `user_edit.go`. Meanwhile `api/iam/iam_test.go` already covers other functions from the same source file (`GetUserList`, `GetUserIDByName`, `CreateUser`, `InviteUser`, `AddMember`), so one source file is split across two test files with nothing in the names marking the split. Merging into `iam_test.go`, or renaming to something like `user_diff_test.go`, would make it findable.

**6. `cmd/workspace/error_callbacks.go` — the alias is now a shell worth removing.**

Moving the body into `api/mfa` left this file as a single delegating line, while `cmd/iam` in the same PR calls `mfa.WorkspaceErrorCallbacks` directly. Two ways to reach one thing, and CLAUDE.md documents the split rather than removing it.

There are only two call sites (`workspace_access_control_update.go:44`, `workspace_authentication_update.go:45`), so the substitution is small. The part that needs a decision is `cmd/workspace/error_callbacks_test.go`, which calls `errorCallbacks` three times; that test now exercises `api/mfa` behavior, so moving it there would also put `cmd/iam`'s path under the same coverage.

Related and much smaller: `api/rbac/rbac.go:63`. `ResolveRole` uses `FetchAllPages` to turn a name into a role, while the existing precedent for that operation, `GetUserIDByName` (`api/iam/iam.go:217`), does one `SendGetRequest` and takes `Results[0]`. The name filter is exact so behavior is identical, but two patterns now coexist for the same job. `FetchPagesUpTo(..., 1)` would put the intent in the code.

### Out of scope for this branch (2)

Noting these only so they are on record; neither needs to happen here.

- `api/rbac/types.go:186` — putting `IsWorkspaceWide` and `ScopeLabel` next to `UserRoleResponse` is closer to ordinary Go practice, but CLAUDE.md's "Go declaration order" fixes top-level declarations at `const → var → type → func`, and methods are `func`. The current placement follows the rule, so changing it is a repo-rule question rather than a file one.
- `cmd/iam/json_slice.go` — swapping a nil slice for an empty one keeps `--output json` emitting `[]` instead of `null`. List commands get this from `utils.PrintTable`, but the JSON branches of `user role describe` and `user permission ls` hand-build a `map[string]any` and skip that path, so the helper earns its place. If other `cmd/` packages hand-build JSON documents they are likely solving the same thing separately, which makes this a candidate for `utils` later.

### Highlights

- The tests pin contracts rather than exercise paths: one `JSONEq` over the whole grant document, the revoke reason asserted absent from the query string, `assert.NotContains(query, "ordering")` for a parameter's absence, and a codeless-403 matrix carrying `wantNotSaid` per gate and credential so a message cannot quietly start giving the wrong advice. The two follow-ups held that bar, with `TestGrantRole_ToleratesAnEmptyCreatedBody` and `TestIAMHostedReadsHitTheRightPaths`.
- `rewritten` with `Unwrap` is the right shape. The operator reads only the actionable half, and `utils.HTTPStatusCode` and `errors.Is` still reach the server's refusal underneath.
- Neither write verb claims success from a 2xx. Both re-read the bindings, which is what makes the superuser companion and a half-finished cascade visible instead of inferred.
- `plannedRevocations` refusing to treat an admin-only account as a resumable cascade is a subtle call, and the comment explains why the obvious shortcut would strip admin from someone who was never a superuser.
- `cmd/iam/member_add.go:43` separating the group-membership `--role` from the new RBAC workspace role in the help text, with `TestRoleFlagKeepsOneMeaning` keeping `--role` to one meaning repo-wide.
- `utils.PrintHeader` writing to stderr, so table mode leaves stdout parseable.

### Checklist

- [x] Tests included and passing, verified locally and in CI
- [x] Documentation updated (README, `Long` text, CLAUDE.md, exit-code table)

cmd/iam/user_role.go
```diff
@@ -0,0 +1,99 @@
...
+func resolveSubject(ac *client.AlpaconClient, args []string) subject {
+	if len(args) == 0 {
+		current, err := iam.GetCurrentUser(ac)
+		if err != nil {
+			utils.CliErrorWithExit("Failed to identify the current user: %s.", err)
+		}
+		return subject{PK: selfUserPK, ID: current.ID, Label: current.Username}
+	}
+
+	if utils.IsUUID(args[0]) {
```
jisung-02, line 90: Warning: naming yourself is not the self form.

`resolveSubject` takes the `"-"` alias only on the no-argument branch. Pass your own username or UUID and `PK` becomes the UUID, which is the addressing `gatePermissionIntrospect` documents as refused: `/api/iam/users/{id}/permissions/` pins no permission of its own, so a by-UUID read auto-resolves to an orphan `user:permissions` that only the superuser wildcard grants.

For a workspace admin without superuser, `alpacon user permission ls --patterns` works but `alpacon user permission ls jane --patterns` is refused when jane is the caller. Same for `can-i`.

The refusal message already says "run the command without a USER argument", so the case looks understood. Comparing the resolved id against the caller's and swapping in `selfUserPK` would close it, and `TestResolveSubject` could use a fourth case to keep it closed.

cmd/iam/rbac_errors.go
```diff
@@ -0,0 +1,119 @@
...
+	case codeSuperuserLastRemoval:
+		return rewrite(err, "this is the workspace's last superuser, and a workspace cannot be left without one; grant 'superuser' to someone else first")
+	// Unreachable from 'grant', which absorbs duplicates as convergence; kept for any later
+	// caller that writes a binding without doing the same.
+	case rbac.CodeRoleAssignmentDuplicate:
+		return rewrite(err, "that role is already bound to the user at this scope")
+	case codeBulkLimitExceeded:
+		return rewrite(err, "the server refused the request as a bulk operation; this is a bug in the CLI, which binds one role to one user per request")
+	case codeInvalidInput:
+		return rewrite(err, "the server rejected the binding scope")
```
jisung-02: Suggestion: `invalid_input` is a generic code rewritten into a write-only sentence.

Running the coded switch ahead of the gate is deliberate and pinned by `TestDescribeRBACError_CodedRefusalIgnoresTheGate`, which is right for `rbac_superuser_last_removal_forbidden`. `invalid_input` is different: eight of the ten `describeRBACError` call sites are reads and none binds a scope. `can-i` lets any argument containing `:` or `*` through `looksLikePermission`, so a malformed permission reaches the server and a rejection there prints "the server rejected the binding scope" from a command that binds nothing.

Scoping the case to `gateRoleWrite`, or adding the kind of comment the duplicate case above it carries, would settle it. `workspace_suspended` reads the same way on `user role ls`.

cmd/iam/user_role_grant.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		if rbac.FindWorkspaceBinding(bindings, role.ID) != nil {
+			utils.CliInfo("%s already holds %s workspace-wide. Nothing to do.", subj.Label, role.Name)
+			return
+		}
+
+		request := rbac.BindingCreateRequest{User: subj.ID, Role: role.ID, Reason: reason}
+
+		if dryRun {
```
jisung-02, line 66: Suggestion: `--dry-run` discloses less than the prompt it replaces.

This branch returns before both `warnMissingReason` and the confirmation, so previewing a superuser grant never mentions the companion `admin` binding the server creates, which is the one fact the interactive path goes out of its way to state, and never warns that the audit entry will carry no justification. Those two are exactly what someone reaches for `--dry-run` to see before a privileged write.

`cmd/iam/user_role_revoke.go:72-77` has the same ordering.

api/rbac/rbac.go
```diff
@@ -0,0 +1,346 @@
...
+			changedBy = entry.ChangedBy.Label
+		}
+
+		result = append(result, RoleAuditAttributes{
+			Action:    entry.Action,
+			Role:      entry.RoleName,
+			Scope:     entry.Scope,
+			ChangedBy: changedBy,
+			Reason:    entry.Reason,
+			At:        entry.AddedAt.Local().Format(timeLayout),
```
jisung-02: Suggestion: nothing asserts the time columns.

`AuditAttributesFrom` here, plus `BindingAttributesFrom` (`:169`) and `HolderAttributesFrom` (`:194`), all build their time string with `AddedAt.Local().Format(timeLayout)`, and no test reads the resulting field. `TestAuditAttributesFrom_ToleratesMissingActorAndRole` checks `ChangedBy`, `Role` and `Reason` only.

So a changed `timeLayout`, a dropped `.Local()`, or a row without `added_at` rendering as `0001-01-01 00:00` would all pass. The audit log exists to answer "who changed what, when", and the "when" is the unpinned part.

One fixed UTC instant asserted against the rendered string covers it, but because of `.Local()` the expected value depends on the runner's timezone. The current suite is stable precisely because it never touches timezones, so it would be worth pinning `time.Local` or comparing the `time.Time` itself rather than the formatted string.

api/rbac/rbac.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+
+		var role RoleResponse
+		if err = json.Unmarshal(responseBody, &role); err != nil {
+			return nil, err
+		}
+		return &role, nil
+	}
+
+	roles, err := api.FetchAllPages[RoleResponse](ac, rolesURL, map[string]string{"name": nameOrID})
```
jisung-02: Suggestion (small): `ResolveRole` uses `FetchAllPages` to turn a name into a role, while the existing precedent for that operation, `GetUserIDByName` (`api/iam/iam.go:217`), does one `SendGetRequest` and takes `Results[0]`.

The name filter is exact so behavior is identical, but two patterns now coexist for the same job. `FetchPagesUpTo(..., 1)` would put the intent in the code.

The rest of the pagination matches the other packages: `FetchAllPages` where the whole list is needed, `FetchPagesUpTo` only where `--tail` bounds it, and no cursor helper since these routes are PageNumber.

api/iam/types.go
```diff
@@ -148,3 +148,13 @@ type SetUsernameResponse struct {
 	ID       string `json:"id"`
 	Username string `json:"username"`
 }
+
+type UserEdit struct {
```
jisung-02, line 153: Suggestion: the JSON tags here are never used.

Neither `UserEdit` nor `PrivilegeEdit` is ever marshalled. `cmd/iam/user_update.go:56` passes `edit.Changes`, a map, and `PatchUser` marshals that map. Every other type in this file is a wire type whose tags are live, so the next reader will take these four as evidence that the server accepts `{"changes": ..., "privileges": ...}`.

Leaving off the `~Response`/`~Request` suffix reads as deliberate and I think it is right, since these are not wire types. One small thing: `PrivilegeEdit.Want` is a thin name for a bool, and `Enable` or `Desired` would say what is wanted at the call site.

api/iam/user_diff_test.go
```diff
@@ -0,0 +1,103 @@
+package iam
```
jisung-02, line 1: Suggestion: the filename points at a file that does not exist.

All three tests exercise `diffEditedUser` only, and `diffEditedUser` lives in `api/iam/iam.go:358`. There is no `user_edit.go`.

`api/iam/iam_test.go` already covers other functions from the same source file (`GetUserList`, `GetUserIDByName`, `CreateUser`, `InviteUser`, `AddMember`), so one source file is now split across two test files with nothing in the names marking the split. Merging into `iam_test.go`, or renaming to something like `user_diff_test.go`, would make it findable.

cmd/workspace/error_callbacks.go
```diff
@@ -1,35 +1,11 @@
...
-			utils.OpenBrowser(mfaURL)
-			return nil
-		},
-		CheckMFACompleted: func() (bool, error) {
-			return mfa.CheckMFACompletion(ac)
-		},
-		RefreshToken:   ac.RefreshToken,
-		RetryOperation: retry,
-	}
+	return mfa.WorkspaceErrorCallbacks(ac, retry)
```
jisung-02: Suggestion: the alias is now a shell worth removing.

Moving the body into `api/mfa` left this file as a single delegating line, while `cmd/iam` in the same PR calls `mfa.WorkspaceErrorCallbacks` directly. Two ways to reach one thing, and CLAUDE.md documents the split rather than removing it.

There are only two call sites (`workspace_access_control_update.go:44`, `workspace_authentication_update.go:45`), so the substitution is small. The part that needs a decision is `cmd/workspace/error_callbacks_test.go`, which calls `errorCallbacks` three times; that test now exercises `api/mfa` behavior, so moving it there would also put `cmd/iam`'s path under the same coverage.

eunyoung14 (changes requested):
The description is cleaned and the context moved where it belongs — thank you. Every code item I raised is fixed, and two of the fixes are better than what I asked for. **But the disclosure surface is wider than the description was, and one part of it becomes unfixable the moment this merges.**

## Where the content still sits

**The description is clean.** The only private-repo reference left is a bare issue number with no description, and the issue that now holds the context exists. That is exactly the shape I asked for.

**Three other surfaces still carry the same categories**, and none of them were covered by editing the body:

1. **PR comments.** Several comments on this PR carry the same classes the description did — internal authorization class names, an internal mechanic described by behavior, and a meta-comment enumerating which sibling repositories are private. One comment reproduces roughly forty words of a private repository's source file **verbatim**. That is the sharpest item currently live: it is internal source content, quoted literally, in a public thread. Comments are as public and as permanent as the body, and editing them leaves the same audit trail the body edit does.

2. **Shipped source comments.** Eight added comment lines across `cmd/iam/rbac_errors.go`, `cmd/iam/user_role.go` and `api/iam/iam.go` name a private repository's source path, a decision-record number, a server view class and method, an internal field class together with its database column type, and a deployment-conditional gate by name. These ship in the public tree permanently, and unlike a comment they are read by anyone who opens the file.

3. **Commit messages, and this is the time-critical one.** Six of the twenty-four commits on this branch carry the same class of detail. **After a merge commit they cannot be edited without a force-push.** A squash merge with a rewritten subject and body collapses all twenty-four into one and is the cheap way out — but it has to be decided before the merge, not after.

I want to be clear about what is and is not achievable here. The pre-edit body remains retrievable through GitHub's edit history by anyone who can see the PR, so the cleanup narrows further exposure and does not undo what was already published. Your framing of it as a disclosure that already happened is the right one. What is still fully in our control is (1) not adding more, (2) the source comments, and (3) the commit messages.

For the source comments specifically: the useful half of each is the *behavior* the CLI has to accommodate. "The server narrows this silently rather than refusing" is worth writing down; naming the class that does it is what carries across the boundary. Rewriting them to describe the observable contract keeps everything a future maintainer of this repo actually needs.

Nothing else turned up — no internal hostnames, no customer names, no attack mechanics.

## Code — all fixed, and two better than asked

`permission_denied` is mapped with a coded-403 test. `role describe` no longer pages the directory when a role has no holders. `GrantRole` returns an error and tolerates the empty 201 the bulk branch produces. `describeRBACError` now wraps, and the new `rewritten` type genuinely restores the chain — I checked that both `utils.HTTPStatusCode` and `utils.ParseErrorResponse` walk it with `errors.Unwrap`, so the status and the coded body survive a rewrite. The README exit-code row names the `user update` change.

Two I want to credit specifically:

**`user role history`** — narrowing the help rather than dropping the filter was the better call. The new wording ("the changes recorded against USER, so a grant you made to someone else appears in their history rather than yours") matches what the endpoint actually returns.

**`TestIAMHostedReadsHitTheRightPaths`** is exactly the guard the gap needed — a four-case table over all three IAM-hosted reads plus the concatenated scopes path, asserting the path and the query parameter, and failing if the permission field decodes to `false`. That closes the silent-wrong-denial case I was most worried about.

`canonicalUUID` is a real find on your own second pass: `IsUUID` accepts the un-dashed form while the audit filter is an exact string compare, so a typed UUID silently returned an empty table and exit 0.

## One thing the fix traded

The `gateAuditRead` messages now lead with the auditor reach limit — but the server applies that limit as silent queryset narrowing and **never returns a 403 for it**. So both audit arms now lead with a cause that cannot produce the refusal they are describing, and by your own note the bearer arm is unreachable besides.

You documented both in the commit message and scoped this pass to comments only, which is a defensible line. I am flagging it because my item 6 asked you to fix an inaccurate message and the fix moved the inaccuracy one gate over rather than removing it. Worth a follow-up issue so it does not become the next reader's puzzle.

## Before this can merge

There is a sixteen-thread review from jisung-02 with **eight inline threads still unresolved and not outdated**. Some are substantive — `--dry-run` returning before the reason warning and the companion-role disclosure, so the preview omits the two facts the interactive path exists to state; `invalid_input` rewriting to a binding-scope sentence on read call sites reachable from `can-i` with a malformed permission; and a test naming a file that does not exist. His first thread's mechanism is stale (a later commit corrected exactly that claim), but the observation under it still stands.

I am not going to merge over an open review from another reviewer. Please work those threads, and settle the commit-message question before the merge rather than after.

CI is green on `83e8c555`. One correction to something I implied earlier: this repo runs a single `ubuntu-latest` leg for both `Build and Test` and `golangci-lint`, not a matrix.

cmd/iam/user_role_grant.go
geunwoonoh: Fixed in 95be0be, in grant and revoke alike.

The companion-binding notice and `warnMissingReason` now run ahead of the `dryRun` return, so `--dry-run` previews everything the real run discloses. The confirmation prompt stays after it—a dry run has no business reaching it.

```
$ alpacon user role grant john superuser --dry-run
Info: The server also creates a companion workspace-wide admin binding for a superuser grant.
Warning: No --reason given, so the audit entry for this change will carry no justification.
Info: Would grant superuser to john (user 2222…, role aaaa…), workspace-wide.

$ alpacon user role revoke john superuser --dry-run --reason x
Info: The companion admin binding stays, so john remains a workspace administrator; --cascade removes both.
Info: Would revoke superuser from john (binding b-new).
```

api/rbac/rbac.go
geunwoonoh: Fixed in ee8c5ad—`FetchPagesUpTo(..., 1)`, with the reason in a comment: the name filter is exact, so the match is on the first page or nowhere.

api/iam/types.go
geunwoonoh: Fixed in b8ed5a7. The tags are gone and a comment says why they are absent, so the next reader does not add them back:

```go
// Not a wire type: nothing marshals it, so it carries no JSON tags. cmd/iam passes
// Changes to PatchUser, which marshals the map alone.
```

`Want` is now `Enable`. It holds the value the operator typed into the editor, and "enable" says which way it points where "want" only says someone wanted something.

api/iam/user_diff_test.go
geunwoonoh: Fixed in b8ed5a7—renamed to `api/iam/user_diff_test.go` with `git mv`, so the file names the function all three tests exercise. Kept separate from `iam_test.go` rather than merged: `diffEditedUser` is the one unexported helper in that file and the only one with no HTTP round trip, so the split is now visible in the name.

cmd/workspace/error_callbacks.go
geunwoonoh: Fixed in 5da91d8. The wrapper is gone and both update commands call `mfa.WorkspaceErrorCallbacks` directly, so there is one spelling again.

On the part you flagged as needing a decision: the test moved with it, to `api/mfa/workspace_error_callbacks_test.go`. The behavior it pins is `api/mfa`'s, and from there it covers `cmd/iam`'s callers too. The CLAUDE.md bullet that documented the split is updated to describe the single helper.

api/rbac/rbac.go
geunwoonoh: Fixed in ecd45fb, and you found a second defect underneath it.

The test pins `time.Local` rather than reading it, per your note—one fixed UTC instant has exactly one correct rendering that way, on every runner:

```go
restore := time.Local
t.Cleanup(func() { time.Local = restore })
time.Local = time.FixedZone("KST", 9*60*60)

granted := time.Date(2026, 3, 4, 3, 4, 5, 0, time.UTC)
const rendered = "2026-03-04 12:04"
```

It covers all three projections. Mutation-checked both ways: dropping `.Local()` fails it (03:04 instead of 12:04), and so does changing the layout.

The zero case turned out to be worth changing rather than only asserting. `0001-01-01 00:00` reads like an answer to "when was this granted" and is not one, so the three call sites now share `localTimestamp`, which returns "" for a zero `time.Time`. The test pins the blank too.

Same commit fixes a scope-column defect the audit rows had: `auditScopeLabel` translated only `global`, so a row at either other tier printed the server's raw `content_type` beside binding rows that say `type:42/web-01`. It is now `tierLabel`, covering all three, and `EffectiveRoleAttributesFrom`—which prints the same vocabulary—routes through it as well. `TestTierLabel` pins each tier.

cmd/iam/rbac_errors.go
geunwoonoh: Half fixed, half not reachable.

**`invalid_input`**—fixed in be1bf47. You are right that it reaches every call site and cannot name a scope. The malformed-permission path through `looksLikePermission` is the concrete case. It now reads "the server rejected one of the request's values", with the reason in a comment beside it rather than scoping the case to `gateRoleWrite`: the code is generic on the server side too, so a write-only rewrite would move the mismatch to the next reader rather than remove it.

**`workspace_suspended`**—this one cannot reach a read. Workspace suspension is enforced read-keep/write-block: safe methods are exempt by construction, so `user role ls` never sees the code. The message stays write-shaped because the only requests that receive it are writes.

Same commit narrows one more message you did not flag, found while re-reading these: the `gateAuditRead` 403 led with the auditor reach limit, which the server applies as silent narrowing of the result set—a short list, never a refusal—so a missing `role_audit_log:read` scope is the only cause a 403 there can have. The bearer arm of that case was unreachable for the same reason and is gone.

cmd/iam/user_role.go
geunwoonoh: Checked this against the server rather than against the message, and the premise does not hold: a by-UUID read of your own row is not refused.

The observable behavior: the user detail routes authorize a detail action at the object level, and every account is auto-granted a self-scoped role at creation that carries a wildcard over the user resource. That wildcard covers the permission this endpoint resolves to, so the object check passes when the target is the caller. The sibling effective-permissions endpoint has a server-side test asserting **200** for exactly this shape—an own-UUID read by a non-privileged account—and names that self-scoped role as the reason. The server-side test for this endpoint pins only the cross-account refusal, which is the case the CLI message describes.

So `alpacon user permission ls jane --patterns` with jane as the caller succeeds, and so does `can-i`.

Not adding the id comparison: it would cost a `GetCurrentUser` round trip on every named invocation to cover one residual case—an account that somehow lacks the self-scoped role, since its assignment swallows failures—and for exactly that account the printed advice "run the command without a USER argument" is already the fix.

An earlier comment of mine on this PR asserted the opposite; that was wrong, and this is the correction. Citations are in alpacax/alpacon-server#3324.

eunyoung14 (dismissed):
The code is done. I verified all four new fixes against the server's current `main` rather than against the commit messages, and every earlier item is still fixed.

**`be1bf47` is the one I want to credit.** I asked you to fix a message that named a cause which cannot produce the refusal — and the fix removes the unreachable arm entirely rather than rewording it. I checked `RoleAuditLogViewSet` on the server: the auditor reach is applied wholly in `get_queryset`, so a non-privileged read returns rows and never a 403. Both old messages led with that, and both are gone. The test now asserts what is *not* said, which is the right shape for a message-accuracy guard.

`95be0be` puts both disclosures ahead of the `--dry-run` return on grant and revoke, with the confirmation prompt correctly still after. `ee8c5ad` is safe because the server declares `name` as an exact filter over a unique column, so one row is the whole answer. `ecd45fb` covers all three scope tiers the server emits and renders a zero timestamp blank; pinning the zone in the test rather than reading it was the right instinct.

`7f255fd` explains itself: the three files were deleted earlier on this branch and a stale index re-added them. Nothing from `main` was resurrected. Worth knowing that `39aa536` therefore does not compile — irrelevant under squash, relevant if anyone bisects.

All eight of @jisung-02's threads are answered in code, including the one where the rejection is correct — I confirmed against a server test that an own-UUID read resolves to self and is not refused. They still need a click to resolve; I am not going to merge over open threads from another reviewer.

## Correcting my own advice, before the merge

I told you a squash merge with a rewritten body would collapse the commit messages. **That is only true if the body is actually replaced, and this repository is configured so it is not.** `squash_merge_commit_message` is `COMMIT_MESSAGES`, so the squash body is **pre-filled with all 32 messages concatenated**. Merging on that default would put them onto `main` permanently — the opposite of what I described. My mistake, and it would have caused the exact thing I asked you to avoid.

Ten of the 32 still carry the categories I flagged. So the merge itself is the decision point:

- merge with an explicit body — `gh pr merge 409 --squash --body "…"`, or
- change the repository's squash-message setting to `PR_BODY` (the body is clean) or `BLANK` first.

Either works. The default does not.

Note also that a correct squash keeps them out of `main`'s history and nothing more — `refs/pull/409/head` stays fetchable regardless.

## What is already published

The comment bodies are cleaned and the shipped source comments are fixed — six of six, and the replacements say the observable contract rather than naming what implements it, which is what a maintainer of this repo actually needs.

The earlier revisions of the edited text remain retrievable. That is not something an edit undoes, and for one of the surfaces it is not something you can remove at all. Your framing of it as a disclosure that already happened is right. I would rather work out how to handle it off this thread than describe it further here — a public post-mortem of where the material sits is one more pointer to it. Let's take that part to the server-side issue.

One thing to do here regardless: the italic notes you appended to the edited comments describe what the earlier version contained. Please strip those.

## Not asking for these

Three comments in the diff use the server's framework and ORM vocabulary and name the identity provider. That vocabulary is already the baseline on this repo's `main`, in six places — one of which names a private repository's source file path outright. **This PR is stricter than the branch it merges into.** Holding it to a bar `main` does not meet would be the wrong call, so I am leaving them and will open a separate cleanup for the pre-existing one.

Approving on the code. Merge it once @jisung-02's threads are resolved, and with the squash body handled.

cmd/iam/user_role_test.go
```diff
@@ -0,0 +1,445 @@
...
+
+func TestResolveSubjectCanonicalizesAUUID(t *testing.T) {
+	const canonical = "2222222a-2222-2222-2222-22222222222b"
+
+	for _, typed := range []string{
+		canonical,
+		"2222222A-2222-2222-2222-22222222222B",
+		"2222222a222222222222222222222 22b",
+	} {
+		typed := strings.ReplaceAll(typed, " ", "")
```
jisung-02: Nit: the space in the fixture above does nothing.

`"2222222a222222222222222222222 22b"` is 33 characters, and this line immediately strips the space back out to reach the 32 the un-dashed case is testing. The reader has to do that arithmetic to see what the case actually covers.

Writing the 32 hex characters directly and dropping this line reads the same and is one fewer step. If the space is load-bearing for something I am not seeing, a word of comment would settle it.

api/rbac/rbac.go
```diff
@@ -0,0 +1,365 @@
...
+}
+
+// tierLabel names a scope tier for the sources that carry no id pair and so cannot use
+// ScopeLabel: "global" prints as the "workspace" every command in this group shows.
+func tierLabel(scope string) string {
+	switch scope {
+	case "global":
+		return "workspace"
+	case "content_type":
+		return "content-type"
```
jisung-02: Nit: the content-type tier now has two spellings in the same binary.

`tierLabel` prints `content-type`, while `ScopeLabel` (`api/rbac/types.go:187`) prints the same tier as `type:42` or `type:42/web-01`. So `alpacon user role ls` and `alpacon user role history` name one tier two ways, side by side in an operator's terminal.

The comment on `tierLabel` explains why the id pair is missing here, and that reasoning holds—the audit row and the effective-role row genuinely carry no ids. What it does not settle is the word: `type` would match the column the operator already knows, at no cost to accuracy. Worth a moment's thought, not a change I would hold the PR for.

jisung-02 (dismissed):
Approving. Every point from my last round is answered, and the two I did not expect to be answered the way they were are the better outcomes: the timestamp comment turned up the zero-time rendering underneath it, and the `permission_denied` self-read was checked against the server rather than against my claim. The correction on that one is right and I withdraw it.

What I checked on this push: `go test ./...` is green across every package and `go vet ./...` is silent.

Two nits below, neither blocking.

eunyoung14 (dismissed):
Re-approving on `cdf6fbe66`. My earlier approval was dismissed mechanically by `dismiss_stale_reviews_on_push`, not by anything substantive.

**`cdf6fbe` is comment prose and nothing else.** I checked it two ways rather than taking the subject line for it: every changed line's first non-whitespace token is `//`, and all sixteen files are byte-identical after comment stripping. No struct tag, string literal, identifier, signature or control flow moved — I looked specifically at the `json:"..."` tags in `api/rbac/types.go` and `api/iam/types.go` and at the user-facing strings in `cmd/iam/rbac_errors.go`, since those are where a "comment-only" pass most often isn't.

It also removes six *more* internal-detail clauses from shipped comments, on top of the six `39aa536` had already cleaned. That was your own initiative, not something I asked for.

Thank you for working @jisung-02's threads to resolution rather than leaving them answered-but-open — and he has now approved this exact head.

## Two things stand between this and the merge button

**Two new threads arrived with his approval** — the test fixture carrying a character the next line strips, and the tier label spelled two ways between two commands' output. He flagged both as nits he would not hold the PR for, but **this repo's ruleset sets `required_review_thread_resolution`**, so an unresolved thread is what is holding `mergeStateStatus` at BLOCKED regardless of severity. Answer or resolve them and the button clears.

**The squash body still needs a hand edit.** `squash_merge_commit_message` is `COMMIT_MESSAGES`, so the body is pre-filled with all 33 messages concatenated. Eight of them carry the categories I flagged — and that count is unchanged rather than improved, which is expected and not a criticism: the previously-audited commits are SHA-identical, so no message *could* have changed. The only fixes are replacing the body at merge time or rewriting history, and replacing the body is the cheap one.

**Separately, and fixable right now: the PR body's first line carries a private-repo cross-reference.** That is on the public page today, independent of the squash. Worth editing before this merges.

## Context that is not this PR's problem

While checking the shipped comments I looked at what `main` already carries, so the bar here is clear. `main` ships this same category widely — roughly two dozen lines naming the private server repo, thirteen backend source paths, and a couple of dozen decision-record references, including in thirteen of the files this branch touches. One instance is in the **public README on main**, in the exit-code table, and it arrived with a different PR.

This branch neither adds to that nor removes it, and it is now stricter than the branch it merges into. That cleanup is real but it is a separate, larger piece of work and should not be hung on this PR. I will open it separately.

Nothing in the 2974 added lines across 47 files reintroduces any of the six categories — I read all 174 added Go comment lines.

CI is green on `cdf6fbe66`, and the required `build-and-test` context is present. Being nine commits behind `main` is harmless here: the ruleset's strict policy is off, and because this branch does not touch `.github/` at all, the merge ref picks up main's fixed workflow — which is also why the flat context reports correctly despite this branch's own copy predating that change.

cmd/iam/user_role_test.go
geunwoonoh: Fixed in dc668e6. The fixture is the 32 characters directly now, and both the stripping line and the `strings` import are gone with it.

The space was not load-bearing. Nothing else in that loop needed the replace either, so the two entries above it were being run through a no-op.

api/rbac/rbac.go
geunwoonoh: Fixed in dc668e6, and you undersold it: `type` is not just the word an operator already knows, it is the string `ScopeLabel` builds the rest of its output on top of. Two spellings for one tier was a defect in the surface, not a preference.

`tierLabel` returns `type` now, and `TestTierLabel` says why in one line, so a future change to either function has to keep the pair in step rather than rediscover this.

Confirmed against a stub with a content-type-scoped audit row, which is the case where the two columns sit closest together:

```
$ alpacon user role ls john
ROLE           SCOPE            GRANTED AT
superuser      workspace        2026-08-01 09:00
server:owner   type:42/web-01   2026-08-02 09:00

$ alpacon user role history john
ACTION    ROLE    SCOPE   CHANGED BY   REASON       AT
granted   admin   type    Root User    onboarding   2026-08-01 09:00
```

`object` is left as it is: `ScopeLabel` renders that tier as `type:42/web-01` and has no competing word for it, so there is no second spelling to reconcile.

jisung-02 (approved):
Approving. Both nits from the last round are fixed in dc668e6, and the tier one came back better than I asked for.

`tierLabel` returns `type`, which is the string `ScopeLabel` heads its own output with, so `user role ls` and `user role history` name the tier the same way in the same terminal. `TestTierLabel` now carries the reason in one line, so the next change to either function has to keep the pair in step instead of rediscovering why they match. The UUID fixture is the 32 characters directly, and the `ReplaceAll` line and the `strings` import went with it.

Verified locally on dc668e6: `golangci-lint run ./...` reports `0 issues.`, and `go test -race ./...` passes with no FAIL, `api/rbac` and `cmd/iam` included. All six CI checks are green.

Nothing blocking left. One label that is not worth a commit on its own: the test case at `api/rbac/rbac_test.go:88` is still named `"content-type wide"`, the last place the old spelling survives. It never reaches an operator, so fold it into whatever touches that file next.

Issue #364, opened 2026-08-18
exec: command-result lines print server-controlled status and error_phase unsanitized


jisung-02:
## Background

#324 closed the path where server-supplied text reaches the terminal through the `Cli*` helpers: `utils.cliMessage` now sanitizes every rendered message per line (PR #359). #360 tracks the registration guide strings that bypass that choke point in `cmd/server/server_create.go` and `cmd/login.go`.

`cmd/exec` bypasses it the same way, on a different surface: the command-result lines. They are built with `fmt.Sprintf` and written straight to stderr with `fmt.Fprint`, so nothing on that path is sanitized.

## Where

`cmd/exec/logs.go` — `logsCommandOutcome` builds `stderrLine`, the `Run` body writes it:

```
 57: fmt.Fprint(os.Stderr, stderrLine)
 72: stderrLine = fmt.Sprintf(... details.Status, details.ID)          // still running
 83: stderrLine = fmt.Sprintf(... details.Status, details.ID)          // awaiting approval
 91: stderrLine = fmt.Sprintf(... details.Status)                      // rejected
 98: stderrLine = fmt.Sprintf(... *details.ErrorPhase, details.Status) // stuck/error/cancelled
101: stderrLine = fmt.Sprintf(... details.Status)
113: stderrLine = fmt.Sprintf(... *details.ErrorPhase)
128: stderrLine = fmt.Sprintf(... details.Status)                      // unrecognised status
```

`cmd/exec/run.go` — `remoteCommandOutcome` (623) builds the phase line, `HandleCommandResult` writes it:

```
555: fmt.Fprint(os.Stderr, stderrLine)
625: stderrLine = fmt.Sprintf("%s: [%s] %s\n", utils.Red("Error"), remoteErr.ErrorPhase, event.DescribePhase(remoteErr.ErrorPhase))
```

## Why it is reachable

Both interpolated values come from the server, unconstrained:

- `EventDetails.Status` (`api/event/types.go`, `json:"status"`) is a free-form string. Every branch above prints it verbatim, including the `unrecognised status` fallback, which exists precisely to render a value the CLI does not know.
- `RemoteCommandError.ErrorPhase` / `EventDetails.ErrorPhase` go through `event.DescribePhase`, which **returns the raw identifier when the phase is unknown** (`api/event/types.go`). An unknown phase is therefore an attacker-chosen string on the operator's terminal.

[...]

PR #410, opened 2026-09-01
feat(command): answer a purpose demand (#368)

### LINKED ISSUE #368: feat(command): answer a purpose demand (client half of the ADR 0052 gate)
## Background

Handbook [ADR 0052](https://github.com/alpacax/alpacon-handbook/blob/main/adr/0052-purpose-rejudgment-before-the-approval-queue.md) is merged, and both the server and the assessor halves are up: alpacon-server#3089 and alpacon-ai-server#378.

When an agent's command draws a verdict that would queue for a human, the server no longer opens an approval request straight away. It parks the command and **demands the command's purpose**. The agent answers, the assessor re-judges once with the purpose in hand, and a changed verdict produces an ordinary signed approval. Only a verdict that survives opens an approval request, and then the stated purpose rides the approver's card.

**Without this issue the server feature is dead code.** The gate arms only for a client that declared `purpose_demand_supported`. The CLI does not send it, so enabling the server flag executes the path zero times.

This is the CLI counterpart of alpacon-mcp#186. Both are needed: agents drive Alpacon through the CLI as well as through MCP, and the gate is per-submission, so whichever surface omits the capability keeps today's behaviour on that surface only.

## Why this is agent-facing, not a human prompt

ADR 0052 is deliberately agent-only, and the server enforces it — the gate does not arm unless the work session's `requester_type` is `agent`. A human asked to fill in a purpose for a held command will not; the friction would land on the requester the way notification fatigue lands on the approver. An agent generates the sentence for free, was already reasoning from one, and is already watching the machine channel for the command's outcome.

So this is not a new interactive prompt for `alpacon command exec`. It is a capability flag plus a response path that an agent driving the CLI can use.

## What is needed

1. **Declare the capability on submit.** `CommandRequest` (`api/event/types.go:68`) gains `PurposeDemandSupported bool` → `purpose_demand_supported`, set where the request is built (`api/event/event.go:240`). A client that never declares it is never parked, so the declaration *is* the opt-in.

2. **Optional up-front purpose.** `CommandRequest` gains `Purpose string` → `purpose,omitempty` (server ceiling: 2000 chars). When present the assessor judges with it on the first pass and the demand round trip never happens. **This is the steady state ADR 0052 actually intends** — state the purpose up front on any command expected to be non-trivial; the demand is the backstop that enforces it.

3. **Detect the demand.** The command parks at `verification_status = "awaiting_purpose"`, and `status` reports `awaiting_purpose`. The server also publishes a live event. Either channel works; polling `status` is the simpler path for the CLI.

4. **Answer it.** `POST /api/events/commands/{id}/purpose/` with `{"purpose": "..."}` → `202`. **Only the requester may answer** — the server refuses anyone else with the same code it uses for a settled command.

5. **Respect the deadline.** Default 60s (`COMMAND_PURPOSE_DEADLINE`). Silence falls through to exactly today's path, so the command is not blocked — but the chance to state a purpose is gone, and **there is one demand per command**, so a late answer is a `400`.

## What to put in the help text

A caveat worth stating explicitly, because it decides whether this feature does anything: **the response is a more reliable channel than help text.** An agent reads the response to its own call far more reliably than it recalls documentation — docs get compacted out of context, the response is right there. So the payload returned when a demand is open should carry what to do; help text carries the contract.

Guidance on *what makes a good purpose* matters as much as the mechanics. The assessor is looking for a connection the session description does not already imply. Measured against the real model: when a description states a broad outcome the judge rationalises any plausible ops action into alignment and supplies the bridge itself, but when it states a concrete deliverable the judge cites the bridge the requester stated. So a useful purpose is **a fact local to this host** — clock skew against a certificate's `notBefore` window, a duplicate config block overriding the edited value, contention for a single JVM attach slot. A bridge made of general knowledge is one the judge already has, so it was never the requester's to supply.

## What not to suggest

- **Do not present purpose as a way around the verdict.** The assessor's prompt states three limits: a purpose cannot lower intrinsic risk, cannot outrank the session description, and cannot make an unmeasurable command measurable. A purpose that tries to manipulate the judge is reported in its own right (`Purpose attempts to manipulate the verdict:`) and denied.
- **Do not promise that a purpose unblocks an opaque script.** Measured: for `bash /tmp/x.sh` a purpose can improve alignment, but confidence stays in the low band, so it still becomes a grey-zone hold. That axis belongs to ADR 0053, not to this feature.

## Prerequisite

alpacon-ai-server must promote the prompt (`promote_prompts`) **before** any client starts sending a purpose. Until then the live prompt has no `## Purpose` section and how the field is read is undefined.

## Related

- alpacon-mcp#186 (the MCP half of the same client work)
- alpacon-server#3089 (gate, `awaiting_purpose` state, purpose endpoint, approver-card exposure)
- alpacon-ai-server#378 (verify contract, prompt, cache key, eval)
- ADR 0053 / alpacon-ai-server#377 (judging script content — the sibling that opens the opaque half)

junho226:
Closes #368

## What this changes

The verification gate holds an agent's command and asks what it is for rather than opening an approval request straight away (ADR 0052). It arms only for a client that declares `purpose_demand_supported`, and the CLI sent nothing—so enabling the server flag ran the path zero times here.

- `alpacon exec --purpose "..."` states it on submission (2000-char ceiling, checked locally).
- A demand is reported as **exit code 7** and, under `--output json`, a `{"status":"purpose_required", ...}` envelope naming the command, the deadline, and what makes a purpose useful.
- `alpacon exec purpose JOB_ID "..."` answers it; the command re-enters judgment once.
- `alpacon exec logs JOB_ID` reports a demand the same way—the only sight `--detach` has of one, since `SubmitCommand` returns before the verdict.

## Why exit 7 and not the pending-approval 4

Nothing is pending on a human while the demand is open. No approval request exists, nobody has been notified, and the next move belongs to the caller that submitted the command. A script or agent that reads this as a pending approval waits for someone who was never asked—and the window is about a minute, so it waits the demand away.

The same reasoning decides two more things: `HandlePurposeDemand` runs ahead of `HandlePendingApproval` (a parked command has no approval request, so reporting one would name a queue it is not in), and `--wait` does not apply. `--wait` exists because an approval is somebody else's to grant and the caller can only block; a purpose is this caller's own to give, so blocking would spend the command's single chance on sleep.

## Why the guidance rides in the envelope

An agent reads the answer to its own call far more reliably than help text, which context compaction drops. So the envelope carries what makes a purpose useful—a fact local to this host that the session description does not already imply, plus the three things a purpose cannot do—and the help text carries the contract.

## Threading, not a global

`purpose` goes through the same chain `workSessionID` already travels: `RunExecWithApprovalWait` → `runPresenceStepUp` → `RunCommandWithRetry` → `RunCommandStreaming` → `SubmitCommand`. That precedent is why it is threaded rather than parked in a package variable. Test call sites gained the argument; no assertion, fixture, or expectation changed.

`purpose_demand_supported` is declared unconditionally, `--detach` included, because no path here waits a demand out—exec reports it at once, `exec logs` reports it on retrieval.

## Notes

- Stating the purpose up front skips the demand entirely. That is the steady state the ADR intends; the demand is the backstop.
- One demand per command. On silence the command is not blocked—it takes the ordinary path—but the chance to explain it is gone, and a late or second answer is refused.
- Only the principal that submitted the command may answer. A settled command and a bystander's answer share one server error, so the CLI cannot report which happened; the message points at `exec logs` rather than at a resubmit.

## Prerequisite

alpacon-ai-server must promote the prompt (`promote_prompts`) before any client starts sending a purpose. Until then the live prompt has no `## Purpose` section and how the field is read is undefined.

## Verification

- `gofmt -l .` clean, `go vet ./...` clean
- `go test ./... -count=1` — all packages pass, 8 new tests in `cmd/exec/purpose_demand_test.go`

## Related

- alpacon-server#3089 (gate, `awaiting_purpose` state, purpose endpoint, approver-card exposure)
- alpacon-ai-server#378 (verify contract, prompt, cache key, eval)
- alpacon-mcp#186 / alpacax/alpacon-mcp#208 (the MCP half of the same client work)
- alpacon-web#1954 (showing the stated purpose on the approver's card)


cmd/exec/purpose.go
```diff
@@ -0,0 +1,101 @@
...
+			return
+		}
+		if strings.TrimSpace(purpose) == "" {
+			utils.CliErrorWithExitCode(utils.ExitCodeUsageError, "PURPOSE cannot be empty: the server refuses a blank answer and the command only gets one demand")
+			return
+		}
+		if len(purpose) > PurposeMaxLength {
+			utils.CliErrorWithExitCode(utils.ExitCodeUsageError, "PURPOSE is limited to %d characters; the server refuses a longer one", PurposeMaxLength)
+			return
+		}
```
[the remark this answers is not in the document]
junho226: Correct, fixed in 31967aa. The README does reserve `2` for `work-session`, `event wait`/`watch`, and `utils.RequirePositiveInt`, and `exec logs` exits `1` on an invalid JOB_ID two files away—so this was inventing a contract for one subcommand. All three refusals now go through `CliErrorWithExit`, and `TestExecPurposeRejectsBadArgsWithExit1` pins each of them plus the success path.

cmd/exec/purpose.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/alpacax/alpacon-cli/api/event"
+	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/config"
+	"github.com/alpacax/alpacon-cli/utils"
+	"github.com/spf13/cobra"
+)
+
+var purposeCmd = &cobra.Command{
+	Use:   "purpose JOB_ID PURPOSE",
+	Short: "State what a held command is for, and send it back for judgment",
```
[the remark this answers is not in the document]
junho226: Fixed in 31967aa. Shortened to "State what a held command is for" (32 chars), in line with `logs` ("Fetch the result of a detached command"). The clause about sending it back for judgment moved into `Long`, which already explains the re-judgment in full.

cmd/exec/parse.go
```diff
@@ -99,6 +109,19 @@ func ParseRemoteExecArgs(args []string) RemoteExecArgs {
...
+			var errMsg string
+			// No short form: pass the long name as sentinel to disable attached-value parsing.
+			purpose, i, errMsg = extractFlagValue(args, i, "--purpose")
+			if errMsg != "" {
+				return RemoteExecArgs{Err: errMsg}
+			}
+			if strings.TrimSpace(purpose) == "" {
+				return RemoteExecArgs{Err: "--purpose requires a value describing what this command is for"}
+			}
+			if len(purpose) > PurposeMaxLength {
```
jisung-02: Critical: this counts bytes, not characters.

`len(purpose)` is a byte count. The server validates with `serializers.CharField(max_length=PURPOSE_MAX_LENGTH)` (`events/api/serializers.py:52,390`), and DRF's `max_length` counts characters. So a Korean purpose is refused here at roughly 666 characters, and the message asserts something untrue about the server at that length.

```suggestion
			if utf8.RuneCountInString(purpose) > PurposeMaxLength {
```

Needs `"unicode/utf8"` in the import block. Precedent for rune counting in this repo: `utils/utils.go:93`. The same fix is needed at `cmd/exec/purpose.go:72`.

cmd/exec/purpose.go
```diff
@@ -0,0 +1,105 @@
...
+		// included, and a script must not read 2 here as a usage-only surface.
+		if !utils.IsUUID(jobID) {
+			utils.CliErrorWithExit("invalid JOB_ID %q: must be a UUID (e.g. a1b2c3d4-5678-abcd-ef01-234567890abc)", jobID)
+			return
+		}
+		if strings.TrimSpace(purpose) == "" {
+			utils.CliErrorWithExit("PURPOSE cannot be empty: the server refuses a blank answer and the command only gets one demand")
+			return
+		}
+		if len(purpose) > PurposeMaxLength {
```
jisung-02: Critical: same byte-vs-character problem as `cmd/exec/parse.go:122`.

The server counts characters (DRF `CharField(max_length=2000)`), this counts bytes, and the message claims the server would refuse the value.

```suggestion
		if utf8.RuneCountInString(purpose) > PurposeMaxLength {
```

Since the check and its wording now exist in two files, consider a shared helper in `cmd/exec` that returns the error string, so the ceiling and the message stay in one place.

cmd/exec/run.go
```diff
@@ -602,6 +602,30 @@ func isPollFailure(err error) bool {
 // true. Neither path carries an approval request id, so the machine signal omits
 // it. reRunHint is the exact command the caller invoked (with any --env caveat
 // in its Description), so a human can copy-paste it once the request is approved.
+// HandlePurposeDemand reports a command the gate parked for its purpose and
```
jisung-02, line 597: Warning: the comment block above now documents the wrong function.

Lines 597-603 are `HandlePendingApproval`'s doc comment—they describe pending-approval feedback and the `reRunHint` parameter, neither of which belongs to `HandlePurposeDemand`. Inserting the new function between that block and its function makes godoc render `HandlePurposeDemand` as "HandlePendingApproval emits the structured pending-approval feedback...", and leaves `HandlePendingApproval` at line 629 undocumented.

Move 597-603 down to sit directly above `func HandlePendingApproval`, so each comment attaches to its own function.

api/event/event.go
```diff
@@ -235,6 +235,11 @@ func SubmitCommand(ac *client.AlpaconClient, serverName, command string, usernam
 		Server:      serverID,
 		RunAfter:    []string{},
 		WorkSession: workSessionID,
+		Purpose:     purpose,
+		// Declared on every submission, including --detach: the demand is
+		// reported to the caller the moment it is seen rather than waited out,
+		// so nothing here can be left stalling for an answer it cannot give.
+		PurposeDemandSupported: true,
```
jisung-02, line 242: Warning: this adds up to 60s of latency to `--detach`, and nothing says so.

The comment is right that no CLI path stalls. But the command does get parked: `events/services.py:184` schedules the expiry countdown, and the command only takes the ordinary path once `COMMAND_PURPOSE_DEADLINE` elapses. So a gated command submitted with `--detach` from an agent-requester session can sit for a minute, and the caller who never runs `exec logs` inside that window sees only the delay.

Human sessions are unaffected—parking is gated on `work_session.requester_type == 'agent'` (`events/services.py:174`).

Worth a line in the `--detach` documentation in `cmd/exec/exec.go` and in the README, since it changes the timing of every gated agent command that detaches.

cmd/exec/purpose_demand_test.go
```diff
@@ -0,0 +1,302 @@
...
+	helper.Stdout = &outBuf
+	helper.Stderr = &errBuf
+	require.NoError(t, helper.Run(), "stderr: %s", errBuf.String())
+	return errBuf.String()
+}
+
+// purposeRequiredSignal is the machine-readable half of the purpose-demand
+// contract these tests assert on. Its flags are what separate it from a pending
+// approval: nothing is waiting on a human, and the answer is the caller's.
+type purposeRequiredSignal struct {
```
jisung-02: Warning: three fields of the envelope contract are never asserted.

This struct omits `Message` and `Context` entirely, and no test reads `NextActions[1]`. `context.deadline_seconds` is documented as part of the contract in `README.md:314` ("carrying the command id, the deadline"), so it is a promised field with nothing behind it—and it is also the field most likely to be wrong, since the server setting is env-overridable.

As the tests stand, deleting `message` and `context` from `purposeRequiredJSON` keeps everything green.

```suggestion
type purposeRequiredSignal struct {
	OK                    bool               `json:"ok"`
	Status                string             `json:"status"`
	ExitCode              int                `json:"exit_code"`
	Message               string             `json:"message"`
	CommandID             string             `json:"command_id"`
	RequiresHumanApproval bool               `json:"requires_human_approval"`
	AnswerableByCaller    bool               `json:"answerable_by_caller"`
	Guidance              string             `json:"guidance"`
	Context               struct {
		CommandID       string `json:"command_id"`
		DeadlineSeconds int    `json:"deadline_seconds"`
	} `json:"context"`
	NextActions           []utils.NextAction `json:"next_actions"`
}
```

Separately, the table-mode branch of `PrintPurposeDemand`—the default path with no `--output json`—has no test at all.

cmd/exec/purpose_demand_test.go
```diff
@@ -0,0 +1,302 @@
...
+	assert.Equal(t, utils.PurposeRequiredStatus, got.Status)
+	assert.Equal(t, purposeJobID, got.CommandID)
+	assert.False(t, got.RequiresHumanApproval)
+}
+
+const purposeJobID = "a1b2c3d4-5678-abcd-ef01-234567890abc"
+
+// TestAwaitingPurposeStatusIsNotAnApproval keeps the two holds apart at the
+// status predicates, which is where a rename would collapse them.
+func TestAwaitingPurposeStatusIsNotAnApproval(t *testing.T) {
```
jisung-02, line 185: Warning: nothing pins the "`--wait` does not apply" decision.

`RunExecWithApprovalWait` returns an `AwaitingPurposeError` straight to the caller only because the `PendingApprovalError` branch and `isApprovalDenial` both miss it. Fold the demand into either one later and `--wait` will sleep through the ~60s window instead of reporting it—with the whole suite still green.

The description argues this point over two paragraphs, which is a good sign it should be a test: stub `runPresenceStepUp` to return `&event.AwaitingPurposeError{CommandID: ...}`, call `RunExecWithApprovalWait` with a non-zero `waitTimeout`, and assert the call returns that error immediately (`assert.Same`) and that the stub was called exactly once. `run_test.go` already has the seam and the `stubApprovalWaitSeams` helper for this shape.

cmd/exec/purpose_demand_test.go
```diff
@@ -0,0 +1,302 @@
+package exec
```
jisung-02, line 1: Warning: the file name points at another package, and the file it names has no test.

`purpose_demand.go` is in `utils/`, but this file is in `cmd/exec/` and holds 11 tests spanning `purpose.go`, `parse.go`, `exec.go`, `run.go`, `logs.go`, and `api/event/types.go`. At the same time `utils/purpose_demand.go` has no direct test, while its sibling `utils/pending_approval.go` has `utils/pending_approval_test.go`.

Suggested split:

- `utils/purpose_demand_test.go` — the envelope: JSON shape including `message` and `context.deadline_seconds`, and the table-mode stderr output, which is currently untested.
- `cmd/exec/purpose_test.go` — `purposeCmd`: the helper process, `TestExecPurposeRejectsBadArgsWithExit1`, `TestExecPurposeAnswersTheDemand`.
- `cmd/exec/parse_test.go` (already exists) — the three `--purpose` parsing cases, ideally as one table (see the separate comment).
- What is left is the end-to-end demand reporting for `exec` and `exec logs`, which sits naturally beside the approval half in `pending_approval_test.go` or in a file named for that.

Filling in the missing assertions while the tests move is the cheap moment to do it.

README.md
```diff
@@ -311,6 +311,7 @@ Also separate from the work session gate: when you authenticate with an API toke
...
+| `7`  | Purpose required—the verification gate held an agent's command and is asking what it is for, before any approval request exists (ADR 0052). Distinct from `4`: nothing is pending on a human, nobody has been notified, and the next move belongs to the caller that submitted the command. Answer with `alpacon exec purpose <JOB_ID> "..."` within about a minute; `--wait` does not apply, because the answer is yours to give rather than somebody else's to grant. Pass `--purpose` on the original `exec` to state it up front and skip the demand entirely. One demand per command: on silence the command takes the ordinary path, and a late or second answer is refused. Under `--output json`, a `{"status":"purpose_required", ...}` object is emitted carrying the command id, the deadline, and what makes a purpose useful. Returned by `exec` and by `alpacon exec logs` (the only sight `--detach` has of a demand) |
```
jisung-02: Warning: this row omits websh command mode.

websh command mode goes through `RunRemoteExec` (`cmd/websh/websh.go:253`), which now calls `HandlePurposeDemand`, so it returns 7 too. And `buildRemoteExecArgs` (`cmd/websh/websh.go:116`) never sets `Purpose`, so websh has no `--purpose` to state one up front—the demand can only be answered with `alpacon exec purpose`.

The exit-4 row above already names websh and spells out that it has no `--wait`; exit 7 deserves the same. `cmd/websh/websh.go:147` also enumerates exit codes 3 and 4 in its `Long` and should gain 7.

cmd/exec/purpose_demand_test.go
```diff
@@ -0,0 +1,302 @@
...
+	assert.False(t, event.IsAwaitingApprovalStatus("awaiting_purpose"))
+	assert.False(t, event.IsAwaitingPurposeStatus("awaiting_approval"))
+	// Not a running state either: treating it as one would poll a command that
+	// is waiting on this caller, until the window closed.
+	assert.False(t, event.IsRunningStatus("awaiting_purpose"))
+}
+
+// TestPurposeFlagRejectsAnEmptyValue keeps a blank answer a usage error rather
+// than a 400 that spends the command's one demand.
+func TestPurposeFlagRejectsAnEmptyValue(t *testing.T) {
```
jisung-02: Suggestion: fold these three into one table.

`TestPurposeFlagRejectsAnEmptyValue` (210), `TestPurposeFlagRejectsAnOverLongValue` (218), and `TestPurposeFlagIsParsedInBothForms` (230) all feed `args` to `ParseRemoteExecArgs` and check `Err` and `Purpose`, so `{name, args, wantErr, wantPurpose}` covers all four cases. `CLAUDE.md` names table-driven tests as the convention, and `TestExecPurposeRejectsBadArgsWithExit1` at line 245 is already one.

cmd/exec/purpose_demand_test.go
```diff
@@ -0,0 +1,302 @@
...
+	assert.Equal(t, utils.ExitCodePurposeRequired, exitCode)
+
+	var got purposeRequiredSignal
+	require.NoError(t, json.Unmarshal([]byte(stdout), &got), "stdout: %s", stdout)
+	assert.Equal(t, utils.PurposeRequiredStatus, got.Status)
+	assert.Equal(t, purposeJobID, got.CommandID)
+	assert.False(t, got.RequiresHumanApproval)
+}
+
+const purposeJobID = "a1b2c3d4-5678-abcd-ef01-234567890abc"
```
jisung-02: Suggestion: keep the declaration order `CLAUDE.md` states—`const` → `var` → `type` → `func`.

Test files in this package do not follow it today, but a new file is worth aligning. The concrete cost here: `purposeJobID` is declared at 195 and first used at 174, so a reader meets the identifier before its definition. Move it up next to `execPurposeHelperMarker` at line 20, and group `type purposeRequiredSignal` (61) after the consts rather than between functions.

cmd/exec/logs.go
```diff
@@ -43,6 +43,19 @@ Run the command again later to check for completion.`,
 			return
 		}
 
+		// Ahead of the approval hold: a parked command has no approval request,
+		// so reporting one would name a queue it is not in (ADR 0052). This is
+		// the detach path's only sight of the demand—SubmitCommand returns before
+		// the verdict, so --detach cannot see it at submission time.
+		if event.IsAwaitingPurposeStatus(details.Status) {
+			utils.PrintPurposeDemand(
+				"Purpose required—the verification gate held this command and is asking what it is for. "+
```
jisung-02: Suggestion: this message is duplicated in `cmd/exec/run.go:620`.

The first two sentences are identical there; only the trailing "Stay silent..." sentence differs. Extracting the shared prefix as a const in `utils/purpose_demand.go` keeps one copy from being reworded without the other.

cmd/exec/purpose_demand_test.go
```diff
@@ -0,0 +1,302 @@
...
+func TestPurposeFlagRejectsAnEmptyValue(t *testing.T) {
+	parsed := ParseRemoteExecArgs([]string{"--purpose", "   ", "prod", "uptime"})
+
+	assert.Contains(t, parsed.Err, "--purpose requires a value")
+}
+
+// TestPurposeFlagRejectsAnOverLongValue mirrors the server ceiling locally, for
+// the same reason.
+func TestPurposeFlagRejectsAnOverLongValue(t *testing.T) {
+	long := make([]byte, PurposeMaxLength+1)
```
jisung-02: Suggestion: `longPurpose()` at line 296 already does this.

```suggestion
	parsed := ParseRemoteExecArgs([]string{"--purpose", longPurpose(), "prod", "uptime"})
```

utils/purpose_demand.go
```diff
@@ -0,0 +1,97 @@
...
+			// always sees the status field.
+			_, _ = fmt.Fprintf(os.Stdout, `{"ok":false,"status":%q}`+"\n", PurposeRequiredStatus)
+		}
+		return
+	}
+
+	CliWarning("%s", message)
+	fmt.Fprintf(os.Stderr, "  %s\n", purposeGuidance)
+	for _, action := range purposeNextActions(commandID) {
+		fmt.Fprintf(os.Stderr, "  %s\n", action.PlainText())
```
jisung-02, line 125: Suggestion: the server-supplied command id reaches the terminal unsanitized.

`CliWarning` on line 92 sanitizes through `cliMessage`, but line 93 and this loop use a bare `fmt.Fprintf` to stderr, and `purposeNextActions` interpolates the server's command id into the text. `cmd/exec/purpose.go:99` wraps the same kind of id in `utils.SanitizeTerminalText`, so the intent is clearly there.

`PrintPendingApproval` has the identical shape, so this is not a regression introduced here—but it is a second site of the pattern #364 addressed, and the cheapest fix is to sanitize `commandID` once at the top of `PrintPurposeDemand`.

utils/purpose_demand.go
```diff
@@ -0,0 +1,97 @@
...
+	AnswerableByCaller    bool               `json:"answerable_by_caller"`
+	Guidance              string             `json:"guidance"`
+	Context               purposeRequiredCtx `json:"context"`
+	NextActions           []NextAction       `json:"next_actions,omitempty"`
+}
+
+// PurposeDeadlineSeconds mirrors the server's default COMMAND_PURPOSE_DEADLINE.
+// Reported so a consumer knows the answer is urgent; the server owns the real
+// clock, so this is copy and never a timer the CLI enforces.
+const PurposeDeadlineSeconds = 60
```
jisung-02: Suggestion: this copies a server setting that is env-overridable.

`COMMAND_PURPOSE_DEADLINE` is read from the environment (`alpacon/settings.py:1469`), so on a workspace that raised it, `deadline_seconds` in the envelope is simply wrong—and a consumer branching on it would act on a deadline that does not exist. The comment is honest that this is a copy, which raises the question of whether reporting a number the CLI cannot know is better than reporting none.

`purpose_requested_at` exists on the server model but is not in the command serializer. Exposing it would let the CLI compute the real deadline instead.

jisung-02 (changes requested):
## Summary

Wires the CLI into the purpose-demand half of the verification gate (ADR 0052): the gate holds an agent's command and asks what it is for instead of opening an approval request straight away. The gate arms only for a client that declares `purpose_demand_supported`, and the CLI sent nothing, so enabling the server flag ran this path zero times before this PR.

I checked the description's claims against the server implementation (`events/api/views.py:256`, `events/services.py:170`, `events/api/serializers.py:52`, `alpacon/settings.py:1460`). They hold: the endpoint answers 202, only the requester may answer, a settled command and a wrong principal share one error code, and the deadline default is 60s. The exit-7-not-4 reasoning is sound and is pinned at the status predicates.

Two things need fixing before merge—the local length ceiling counts bytes where the server counts characters, and a doc comment now documents the wrong function. The rest is test assertions, test placement, and docs that did not move with the code.

## Critical (1)

**C1 — the 2000 ceiling is counted in bytes, but the server counts characters.** `cmd/exec/parse.go:122`, `cmd/exec/purpose.go:72`

Both sites use `len(purpose)`, which counts bytes. The server validates with `serializers.CharField(max_length=PURPOSE_MAX_LENGTH)` (`events/api/serializers.py:52,390,750`), and DRF's max_length counts characters. A Korean purpose is refused locally at roughly 666 characters, and the message it prints—"the server refuses a longer one"—is false at that length. Use `utf8.RuneCountInString`; the precedent is `utils/utils.go:93`. The check and its message are copied into both files, so both need the change.

## Warnings (6)

**W1 — a doc comment now documents the wrong function.** `cmd/exec/run.go:597-628`

The comment block at 597-603 belongs to `HandlePendingApproval`. The new function was inserted between that block and its function, so godoc renders `HandlePurposeDemand` with "HandlePendingApproval emits the structured pending-approval feedback..."—including a description of `reRunHint`, a parameter the new function does not have. `HandlePendingApproval` is left with no doc comment. Move 597-603 down to just above line 629.

**W2 — three contract fields of the JSON envelope are never asserted.** `cmd/exec/purpose_demand_test.go:61`

`purposeRequiredSignal` does not declare `Message` or `Context` at all, and no test reads `NextActions[1]` (the `exec logs` fallback). `context.deadline_seconds` is named as part of the contract in `README.md:314`, so it is a documented field with no test behind it. As it stands, deleting `message` and `context` from the envelope keeps every test green. Separately, the table-mode branch of `PrintPurposeDemand` (the default path, no `--output json`) has no coverage at all.

**W3 — nothing pins "`--wait` does not apply to a purpose demand".** `cmd/exec/purpose_demand_test.go:199`

`RunExecWithApprovalWait` returns the error immediately only because the `PendingApprovalError` branch and `isApprovalDenial` both happen to miss `AwaitingPurposeError`. A later change that folds the demand into either one would make `--wait` sleep through the ~60s window, and no test would fail. The description spends two paragraphs on why this must not wait, which is exactly the kind of decision that earns a test.

**W4 — the test file's name points at a package it is not in, and the file it names has no test.** `cmd/exec/purpose_demand_test.go:1`

`purpose_demand.go` lives in `utils/`, but this file sits in `cmd/exec/` and holds 11 tests spanning `purpose.go`, `parse.go`, `exec.go`, `run.go`, `logs.go`, and `api/event/types.go`. Meanwhile `utils/purpose_demand.go` has no direct test, while its sibling `utils/pending_approval.go` has `utils/pending_approval_test.go`. Suggested split: envelope tests to `utils/purpose_demand_test.go`, `purposeCmd` tests to `cmd/exec/purpose_test.go`, and the `--purpose` parsing cases into the existing `cmd/exec/parse_test.go`. Filling in the W2 assertions while moving them is the natural moment.

**W5 — the exit-7 docs omit websh command mode.** `README.md:314`, `cmd/websh/websh.go:147`

websh command mode reaches `HandlePurposeDemand` through `RunRemoteExec` (`cmd/websh/websh.go:253`), so it exits 7. But `buildRemoteExecArgs` (`cmd/websh/websh.go:116`) never sets `Purpose`, so websh has no way to state one up front. The exit-4 row already names websh and spells out that it has no `--wait`; exit 7 needs the same treatment, in the README row and in websh's own `Long`.

**W6 — declaring support on `--detach` adds up to 60s of silent latency, undocumented.** `api/event/event.go:242`

The comment argues that nothing here can be left stalling, and that is true of the CLI process. The command itself is parked: the server sets a countdown at `events/services.py:184` and releases it after `COMMAND_PURPOSE_DEADLINE`. So every gated command from an agent-requester session run with `--detach` can sit for a minute before taking the ordinary path, and the caller has no way to know. Human sessions are unaffected—parking is gated on `work_session.requester_type == 'agent'` (`events/services.py:174`). This deserves a line in the `--detach` docs.

## Suggestions (6)

**S1 — fold the three `--purpose` parsing tests into one table.** `cmd/exec/purpose_demand_test.go:210,218,230`

All three take `args` and check `Err` and `Purpose`, so `{name, args, wantErr, wantPurpose}` covers them. `CLAUDE.md` names table-driven tests as the convention, and line 245 in the same file is already one.

**S2 — keep the `const` → `var` → `type` → `func` order.** `cmd/exec/purpose_demand_test.go:195`

`CLAUDE.md` states the order for top-level declarations. Test files in this package do not follow it today, but a new file is worth aligning. `purposeJobID` at 195 is used by the test at 174, so a reader meets the identifier before its declaration—move it next to `execPurposeHelperMarker` at line 20, and group `type purposeRequiredSignal` (61) after the consts.

**S3 — the demand message is duplicated.** `cmd/exec/logs.go:52`, `cmd/exec/run.go:620`

The first two sentences are identical. A const in `utils/purpose_demand.go` keeps one from being edited without the other.

**S4 — the over-long-string builder is duplicated.** `cmd/exec/purpose_demand_test.go:219`

`longPurpose()` already exists at line 296.

**S5 — the server-supplied command id is printed raw.** `utils/purpose_demand.go:95`

The guidance line and the next-action lines are written with a bare `fmt.Fprintf` to stderr, outside the sanitizing `Cli*` helpers, and the next-action text embeds the server's command id. `cmd/exec/purpose.go:99` wraps the same id in `utils.SanitizeTerminalText`. `PrintPendingApproval` has the same shape, so this is not a regression from this PR—but it is a second site of the pattern #364 was about.

**S6 — `PurposeDeadlineSeconds` copies a server setting that is env-overridable.** `utils/purpose_demand.go:40`

`COMMAND_PURPOSE_DEADLINE` is read from the environment (`alpacon/settings.py:1469`), so `deadline_seconds` in the envelope is wrong on any workspace that raised it. The comment acknowledges the copy. The server could expose `purpose_requested_at`—it is a model field but not in the serializer—and let the CLI compute it, or the field could be dropped rather than reported inaccurately.

## Highlights

- The exit 4 / exit 7 separation is reasoned in the description and pinned down at the status predicates, including that `IsRunningStatus("awaiting_purpose")` is false—the assertion that stops a parked command from being polled until its window closes.
- Adding the demand to `propagateCommandError` is the right call: without it the MFA retry path would swallow it as a transport failure.
- Threading `purpose` along the chain `workSessionID` already travels, rather than parking it in a package variable, kept every existing test expectation unchanged.
- `AnswerPurposeDemand`'s doc comment is honest about what a failure cannot tell you, and the error message points at `exec logs` instead of a resubmit—which matches what the server actually does.

## Checklist

- [ ] Tests included and passing—CI is green, but the envelope's documented fields and the no-wait decision are unasserted (W2, W3)
- [ ] Documentation updated if needed—websh and `--detach` behavior are missing (W5, W6)

cmd/exec/parse.go
junho226: Correct, fixed in 1161166. I read the Python `max_length=2000` and never checked what `len` means in Go.

The check and its wording were copied into two files, so both went into one `checkPurpose(subject, purpose)` helper—the ceiling and the message cannot drift apart again. Rune counting follows the `utils/utils.go:93` precedent you named.

Regression coverage on both write paths: `TestParseRemoteExecArgsPurpose` has a "2000-character Korean purpose is within the ceiling" case, and `TestExecPurposeAcceptsAMultibytePurposeAtTheCeiling` asserts one actually reaches the server.

cmd/exec/purpose.go
junho226: Fixed alongside it. Both sites now call `checkPurpose`, differing only in the subject they pass (`--purpose` vs `PURPOSE`).

cmd/exec/run.go
junho226: Correct, fixed. The block moved down to sit directly above `func HandlePendingApproval`. Inserting the new function left no blank line between the two comments, so they merged into one block attached to the wrong function—which is how the `reRunHint` prose ended up on a function that does not take it.

api/event/event.go
junho226: Correct. My comment said nothing here stalls, which was true of the CLI process and not of the command. The command stays parked until the countdown at `events/services.py:184` elapses, and the caller has no way to see it.

Added a line to the `--detach` docs in `cmd/exec/exec.go` and to the README exit-7 row. I confirmed your point that human sessions are unaffected—the park is gated on `requester_type == 'agent'`.

cmd/exec/purpose_demand_test.go
junho226: Correct—as it stood, deleting `message` and `context` from the envelope kept the whole suite green.

Moved to `utils/purpose_demand_test.go` and now asserting the full contract: `message`, `context.command_id`, `context.deadline_seconds`, and `next_actions[1]` (the `exec logs` fallback and its "do not re-submit" wording). Table mode—the default path—is covered by `TestPrintPurposeDemand_TableModeIsActionableOnStderr`, which also checks stdout stays clean.

cmd/exec/purpose_demand_test.go
junho226: Correct. A decision I spent two paragraphs on in the description had nothing holding it up.

Added `TestRunExecWithApprovalWait_DoesNotWaitOnAPurposeDemand`, in the shape you described: `stubApprovalWaitSeams`, `assert.Same` on the returned error, the stub called exactly once (so the wait loop never retries it), and elapsed time under a second (so the window is not entered at all).

cmd/exec/purpose_demand_test.go
junho226: Split as you suggested.

- `utils/purpose_demand_test.go` — the envelope contract plus table mode, beside `utils/pending_approval_test.go`
- `cmd/exec/purpose_test.go` — `purposeCmd`: helper process, argument refusals, success path, multibyte ceiling
- `cmd/exec/parse_test.go` — the `--purpose` parsing cases, folded into one table
- `cmd/exec/purpose_demand_test.go` — what is left: end-to-end reporting for `exec` and `exec logs`, and the `--wait` decision. A header comment names where the rest went

The missing assertions went in while the tests moved, as you suggested.

README.md
junho226: Correct. I confirmed both halves: websh reaches exit 7 through `RunRemoteExec`, and `buildRemoteExecArgs` never sets `Purpose`, so it has no way to state one up front.

The README exit-7 row now names websh and says the demand can only be answered with `alpacon exec purpose`. `cmd/websh/websh.go`'s `Long` gained an exit-7 paragraph beside the existing 3 and 4.

cmd/exec/purpose_demand_test.go
junho226: Folded into one table—`{name, args, wantErr, wantPurpose}`, five cases now that the multibyte ceiling is one of them. It lives in the existing `cmd/exec/parse_test.go`.

cmd/exec/purpose_demand_test.go
junho226: Both new files now follow `const` → `type` → `func`. `purposeJobID` and `execPurposeHelperMarker` share the const block at the top of `cmd/exec/purpose_test.go`.

cmd/exec/logs.go
junho226: Extracted as `utils.PurposeDemandLead`. `exec` appends its "Stay silent..." sentence; `exec logs` uses the lead as-is.

cmd/exec/purpose_demand_test.go
junho226: Switched to `longPurpose()`. The parsing tests moved to `parse_test.go` and build their rune slice there, and `longPurpose()` in `cmd/exec/purpose_test.go` now counts runes rather than bytes.

utils/purpose_demand.go
junho226: Correct, fixed—`commandID` goes through `SanitizeTerminalText` once at the top of `PrintPurposeDemand`, since every line below interpolates the same value. The guidance string goes through it too.

`TestPrintPurposeDemand_SanitizesTheServerSuppliedID` asserts ANSI escapes and BEL do not reach the terminal.

You are right that `PrintPendingApproval` has the same shape; I left it alone since this PR did not introduce it. Worth its own issue.

utils/purpose_demand.go
junho226: A fair point, and it turned into a server change. alpacax/alpacon-server#3323 now exposes `purpose_requested_at` on the detail serializer, read-only.

The timestamp alone does not close it, though—the deadline length is still unknown to the client. So:

- With a timestamp, the envelope reports the window minus the elapsed time. **Elapsed time is the larger of the two errors**: without it, a demand noticed four minutes later still reads "60s left".
- Without one, `deadline_seconds` is **omitted entirely** (`omitempty`). That is the answer to your question about whether reporting a number the CLI cannot know beats reporting none.
- The residual error runs one way only. A workspace that raised the deadline makes this understate the window, which hurries the caller; nothing here can overstate one. That reasoning is in the `PurposeDeadlineAssumed` comment.

An already-elapsed window reports 0 rather than a negative. Four tests pin the cases.

api/event/types.go
```diff
@@ -69,6 +86,12 @@ type EventDetails struct {
 	AddedAt       time.Time           `json:"added_at"`
 	Server        types.ServerSummary `json:"server"`
 	RequestedBy   types.UserSummary   `json:"requested_by"`
+	// What the requester said this command is for, and when the gate asked
+	// (ADR 0052). Both are absent on a server predating the read exposure, and
+	// on any command nobody was asked—which is every command until the gate is
+	// enabled, so nil and empty are the ordinary shapes here.
+	Purpose            string     `json:"purpose"`
```
jisung-02: Warning: nothing reads this field.

`EventDetails.Purpose` is added here, and no non-test code references it. `exec logs` prints only the outcome lines, and `exec ls` projects through `EventAttributes`, which has no purpose column. So the CLI can send a purpose but has no way to read one back.

The comment above the field describes what the requester said the command is for, which makes it a promise no surface keeps.

If it should be visible, `exec logs` is the natural place, since that is already where a caller goes to see what became of a held command. Sanitize it at that print site: a purpose is text one principal wrote that then reaches another principal's terminal, which is exactly the shape #364 addressed. If it should not be visible, drop the field rather than parse it into nothing.

cmd/exec/purpose.go
```diff
@@ -0,0 +1,101 @@
...
+					"or this credential may not be the one that submitted the command. Read the command's state with "+
+					"`alpacon exec logs %s`—do not re-submit it, which would create a second command.",
+				jobID, err, jobID,
+			)
+			return
+		}
+
+		// The command re-enters judgment on the worker, so there is no verdict to
+		// report here; naming where to read it keeps the caller off a resubmit.
+		fmt.Fprintf(os.Stderr, "Purpose recorded. The command is being judged again with it in hand.\n")
```
jisung-02: Warning: this command ignores `--output json`.

`--output` is a persistent flag on `RootCmd`, so `alpacon exec purpose --output json JOB_ID '...'` is accepted here. But `purposeCmd` never reads `utils.OutputFormat`: success writes two lines to stderr, failure goes through `CliErrorWithExit`, and stdout stays empty either way.

ADR 0052 is an agent-facing flow end to end. `exec` and `exec logs` both emit a `purpose_required` envelope on stdout, and `exec purpose` is the command that answers it—yet it is the only one with no machine-readable output. An agent that branched on exit 7, answered, and now wants to know whether the answer landed has the exit code and nothing else; the reason for a refusal is only available as English prose on stderr.

The refusal case is where this costs the most. The message here deliberately cannot distinguish an expired demand from an answer by the wrong principal, and tells the caller to read the command's state instead—which is precisely the kind of branch a machine consumer needs a field for.

Following the envelope shape already in `utils/pending_approval.go` and `utils/purpose_demand.go` would be enough: `ok`, `command_id`, and a `next_actions` entry pointing at `alpacon exec logs JOB_ID`.

utils/purpose_demand.go
```diff
@@ -0,0 +1,135 @@
...
+			_, _ = fmt.Fprintf(os.Stdout, `{"ok":false,"status":%q}`+"\n", PurposeRequiredStatus)
+		}
+		return
+	}
+
+	CliWarning("%s", message)
+	if remaining != nil {
+		fmt.Fprintf(os.Stderr, "  About %ds left to answer.\n", *remaining)
+	}
+	fmt.Fprintf(os.Stderr, "  %s\n", SanitizeTerminalText(purposeGuidance))
```
jisung-02: Suggestion: this sanitizes a compile-time constant.

`purposeGuidance` is a package const declared at line 26, so it cannot carry a control character or an escape sequence. `SanitizeTerminalText` walks the string on every print and returns it unchanged.

No harm today, but in this repo a sanitizing call is a signal that the value came from the server (#364, #324). Wrapping a literal we wrote ourselves weakens that signal for the next reader, who then has to check where `purposeGuidance` comes from to know whether it matters.

The `commandID` sanitizing at line 103 is the real one and should stay.

```suggestion
	fmt.Fprintf(os.Stderr, "  %s\n", purposeGuidance)
```

cmd/exec/logs.go
```diff
@@ -43,6 +43,17 @@ Run the command again later to check for completion.`,
 			return
 		}
 
+		// Ahead of the approval hold: a parked command has no approval request,
+		// so reporting one would name a queue it is not in (ADR 0052). This is
+		// the detach path's only sight of the demand—SubmitCommand returns before
+		// the verdict, so --detach cannot see it at submission time.
+		if event.IsAwaitingPurposeStatus(details.Status) {
+			utils.PrintPurposeDemand(
+				utils.PurposeDemandLead, details.ID, details.PurposeRequestedAt,
```
jisung-02: Suggestion: this drops the one sentence that says what happens on silence.

`HandlePurposeDemand` (`cmd/exec/run.go:612`) prints `PurposeDemandLead` plus "Stay silent and it takes the ordinary path when the demand expires." This call site prints the lead alone. Everything else the two surfaces emit—the remaining seconds, the guidance, both next actions—is identical, so that sentence is the whole difference.

It is also the fact a caller who arrives through `exec logs` needs most. They found the demand late by definition, so whether the window closing blocks the command or lets it run is the question in front of them. The second next action says to read the outcome if the demand already expired, which implies there is an outcome, but never states that the command proceeds on the ordinary path rather than being held.

Since the behavior belongs to the server and not to whichever command reported it, I would fold the sentence into `PurposeDemandLead` itself and drop the concatenation in `run.go`, so the two surfaces cannot describe expiry differently again.

jisung-02 (commented):
A second pass over the branch as it stands now, after the fixes from the previous round. Everything I raised last time is addressed, so this is a comment rather than another change request.

Verified locally at a7141c9: `go vet ./...` is clean, `golangci-lint run` over `cmd/exec`, `utils` and `api/event` reports 0 issues, and `go test -race` passes for all three (`cmd/exec` 14.6s, `utils` 2.5s, `api/event` 73.1s).

Two of the five below are worth acting on before merge: a field the diff adds that nothing reads, and the one agent-facing command in this flow that writes nothing to stdout under `--output json`. The other three are small.

One thing I checked that could have gone wrong and did not: `awaiting_purpose` is absent from `IsRunningStatus` (`api/event/types.go:177`), so `isPollWaitStatus` returns false and the poll goroutine in `streamSubscribed` ends on its first tick instead of holding the caller for the whole timeout. The demand reaches the caller promptly on the streaming path and the fallback path alike, which is the behavior the ~60s window depends on.

cmd/exec/parse.go
```diff
@@ -1,13 +1,39 @@
...
+// an empty string when it passes. subject names the thing being validated—the
+// flag for `exec --purpose`, the argument for `exec purpose`—so one ceiling and
+// one wording serve both.
+//
+// The ceiling is counted in runes, not bytes. The server validates with DRF's
+// CharField(max_length=...), which counts characters, so len() would refuse a
+// Korean purpose at roughly 666 of them—while telling the caller the server
+// would refuse it, which at that length is untrue.
+func checkPurpose(subject, purpose string) string {
+	if strings.TrimSpace(purpose) == "" {
```
jisung-02: Suggestion: the purpose is sent with its surrounding whitespace.

`strings.TrimSpace` is used to decide whether the value is empty, and then the untrimmed original is what travels: `--purpose '  clock skew  '` is stored, judged, and written to the card with the padding intact.

Trimming is not sanitizing, and the distinction matters here. Stripping control characters out of a purpose would falsify a statement the assessor and the audit record attribute to the requester, so the value should reach the server as typed—but leading and trailing whitespace is not part of what anyone typed on purpose.

If you trim, count the ceiling on the trimmed value too, so the length that is checked is the length that is sent.

api/event/types.go
junho226: Correct, and "a promise no surface keeps" is the right way to put it — I added the field for the type to be complete and never asked who reads it.

Given it a reader in 53e60e8. `exec logs` prints `Stated purpose: ...` ahead of the outcome, on stderr so stdout stays the command own output, and sanitized at the print site for the reason you name: it is text one principal wrote reaching another principal terminal.

`TestExecLogsShowsTheStatedPurpose` feeds a purpose carrying an ANSI escape and asserts it does not survive, and that stdout is untouched.

cmd/exec/purpose.go
junho226: Correct. `exec purpose` was the one command in an agent-facing flow writing nothing to stdout, and it is the command that answers an envelope the other two emit.

Fixed in 53e60e8 with the shape you point at: `purpose_recorded` on success and `purpose_refused` on failure, both carrying `ok`, `command_id`, and a `next_actions` entry pointing at `alpacon exec logs JOB_ID`.

Your reading of where it costs most is right. The refusal cannot distinguish an expired demand from an answer by the wrong principal, so the caller is told to read the command state — and turning that uncertainty into prose left a machine consumer with nothing to branch on. `TestPrintPurposeRefused_JSONDoesNotSendTheCallerToResubmit` pins that the one next action is `exec logs` and never a resubmission.

utils/purpose_demand.go
junho226: Applied. The point about the signal is the part worth having: a sanitizing call in this repo means the value came from the server, and wrapping a literal we wrote ourselves makes the next reader check where it came from before knowing whether it matters.

`commandID` at line 103 stays, and it is now the only one in this function.

cmd/exec/logs.go
junho226: Agreed, and I took the version you suggested — the sentence is folded into `PurposeDemandLead` and the concatenation in `run.go` is gone, so the two surfaces cannot describe expiry differently again.

Your reason for it being the sentence that matters most on this path is the one I had missed: someone arriving through `exec logs` found the demand late by definition, so whether the window closing blocks the command is the question actually in front of them. I also made the wording explicit about it — "the command is not blocked" rather than leaving it to be inferred from "takes the ordinary path". `TestPurposeDemandLead_SaysWhatSilenceCosts` pins both halves.

cmd/exec/parse.go
junho226: Correct — checked trimmed, sent untrimmed. The padding was being stored, judged, and written to the approver card.

Fixed in 53e60e8: `checkPurpose` returns the value to send, and the ceiling is counted on that, so the length checked is the length that travels.

The distinction you drew is the one that decided how far to go. Trimming surrounding whitespace is not part of what anyone typed on purpose; stripping control characters out of the body would falsify a statement the assessor and the audit record attribute to the requester. So the text itself still reaches the server as typed, and sanitizing happens at the terminal print site instead — which is the same conclusion as the `exec logs` thread above.

`TestExecPurposeTrimsBeforeSending` asserts what lands on the wire.

jisung-02 (dismissed):
Approving. All five points from the second pass are addressed at `53e60e8`, and the expiry rework in `64e1248` is a better answer than the one I asked for.

What landed:

- `EventDetails.Purpose` has a reader. `exec logs` prints it as `Stated purpose:` on stderr, sanitized at the print site rather than at the API boundary, which is the rule #364 set.
- `alpacon exec purpose` answers on the same contract its callers emit. `purpose_recorded` and `purpose_refused` carry `command_id` and `next_actions`, so an agent that branched on exit 7 and answered now gets a field instead of English prose—and the refusal case, where the server collapses "already settled" and "wrong principal" into one code, is exactly where that matters.
- `purposeGuidance` is no longer sanitized. A compile-time constant passing through the sanitizer only weakened the signal that a sanitized value came from the server.
- The "stay silent" sentence moved into `PurposeDemandLead`. It describes server behavior, not the behavior of whichever command reported it, and `exec logs`—the caller who by definition found the demand late—was the one dropping it.
- `checkPurpose` returns the value it validated. `--purpose '  x  '` no longer stores padding on the approver's card, and the ceiling is counted on what travels.

Reading the server's `purpose_expires_at` instead of subtracting elapsed time from an assumed 60 seconds removes the guess entirely, and a workspace that raises `COMMAND_PURPOSE_DEADLINE` needs no release of this client. Worth the extra round.

Two leftovers, neither blocking:

- `utils/purpose_demand.go:19`—`PurposeDeadlineAssumed` is still declared and now has no callers. The commit message for `64e1248` says it was removed entirely, so this looks like an oversight rather than a decision. It is exported, so `unused` will not flag it, and the doc comment above it still explains the elapsed-time subtraction that commit replaced—which will read as current guidance to whoever finds it next.
- `api/event/types.go:96`—`PurposeRequestedAt` is now the field with no reader. The reporter switched to `ExpiresAt`, so this one is parsed and dropped. Either keep it for a surface that will show when the gate asked, or drop it alongside the constant.

Verified locally at `53e60e8`: `go vet ./...` silent, `golangci-lint run` on `cmd/exec`, `utils`, and `api/event` reports `0 issues.`, and `go test -race` is green on all three (`cmd/exec` 16.617s, `utils` 2.899s, `api/event` 72.664s).

jisung-02 (approved):

PR #416, opened 2026-09-02
Poll the sudo grant instead of re-running the command to wait for approval

### LINKED ISSUE #312: Approval-wait loops poll on a fixed interval and abort on 429

jisung-02:
## What

`exec --wait` polled for a sudo approval by re-submitting the whole command every 5 seconds. It now reads the denied command's own detail and decides on `sudo_grant_status`. The other three poll loops in the CLI move onto the same pacing rules, which now live in `utils/poll.go` for all of them.

## Why

A denial from the sudo plugin carries the denial code but no approval request id, and this credential channel has no approval-status endpoint (ADR 0015), so re-running the command was the only poll available. One tick is not one request: `RunCommandStreaming` performs an event-session create, a server lookup, a submit, two subscribes, a chunk read, and at least one status GET. A 30-minute `--wait-approval` therefore spent past alpacon-server's default `service_token: 1000/hour` quota, and the 429 it earned was neither a denial nor an approval, so the wait ended on the throttle error while the approval request was still perfectly valid.

alpacax/alpacon-server#2911 closed that gap. Since server 2.28.0 the command detail carries `sudo_approval_request_id` and `sudo_grant_status`, which is a status the CLI can read at one GET per tick.

## Changes

`exec --wait` reads the grant. `authorized` is approval, `rejected` and `expired` both settle without one, and every other value keeps waiting — a status nobody can read as an answer must not end a wait, so `used` and an absent field stay undecided with the timeout as the backstop.

The two settled outcomes share an exit code but not a sentence. Telling a user whose approval window simply lapsed that a reviewer said no points them at the wrong next move, so `CommandRejectedError` carries which one settled it and words the message accordingly.

The command being watched already ran and was blocked, so re-subscribing to it would replay the old denial. Approval runs the command once more through the presence-aware path, and that run is what spends the grant. A run denied or held again does not open a second wait; it reports that the grant looks already used.

| Outcome | Before | After |
| --- | --- | --- |
| Approved | ran on the next tick | runs once, after the poll reads `authorized` |
| Rejected | waited out the window, exit 4 | exit 6 as soon as the grant reads `rejected`, naming a reviewer's refusal |
| Expired | waited out the window, exit 4 | exit 6, naming the expiry rather than a refusal |
| Timed out | exit 4 | exit 4, now naming the approval request id |
| Throttled (429) | ended the wait | keeps waiting, and spends throttle budget to hold the deadline |

`work-session create --wait` paced its failure path already but slept a fixed 10 seconds on a successful tick; it now takes `NextPollTick` like the exec wait. A wait that settles without a grant used to print `work session was rejected` and nothing else, leaving a caller with a session it had no id for. It now names the session in the message and in `context.work_session_id`, with `work-session describe` and `work-session complete` as `next_actions`. Not `work-session use`: `terminalWaitError` is built only for rejected, expired, revoked, cancelled and completed, and `RunUseSession` refuses anything but an active session, so that command could never have worked from there. `next_actions` is read by agents that run what it lists, which makes a guaranteed failure worse than no suggestion.

The two MFA step-up polls keep their fixed 500ms tick. They were briefly moved onto `NextPollTick` with the rest, and review showed why that is wrong here. At a 500ms base the schedule holds 500ms for five seconds, widens to 2.5s out to 30s, then to 5s: about 26 polls across the 60-second window where the fixed ticker did 120. Worst-case detection lag after the user finishes in the browser goes from half a second to five, and a completion at t=56-59s is missed outright, since the last poll lands at 55s and the next races the deadline. Nothing is bought for it—the quota this pacing exists for is a service token's, and by `StepUpForSudo`'s own contract no script, CI job or agent reaches a step-up. In the listener it is worse than neutral: the buffer `defaultMFAPollTimeout` keeps over the server's pending-grant expiry is exactly what a widened tail spends. Backoff is left out for the same reason—`NextPollBackoff` caps at 60 times the base tick, which is 30 seconds here.

The two MFA polls also stop reading their pacing off package vars. Those had been `var` only so tests could shorten them, and `SudoListener` reads its interval from a background goroutine, so a listener outliving its own test raced the next one's assignment. The listener takes fields and `StepUpForSudo` takes arguments; the exported signature is unchanged.

The 429 deadline allowance was four local variables inside `pollCommandExecution`. It is now `utils.ThrottleBudget` with the same rule, shared by all three approval waits. The allowance is spent rather than re-earned: `Reset` is the progress signal and only a status change counts as progress, so reading the same pending status again buys nothing back. Without that, one non-429 response refilled the whole allowance and the wait had no ceiling at all. `utils.ThrottleCeiling` states the bound that follows — one timeout, one allowance, and the overshoot of the grant that crosses it — and both command waits are tested against it.

Extending the deadline broke the work-session wait's progress line, which read elapsed as `timeout - remaining`—correct only while the deadline stands still. Once a 429 moves it, the reported time shrinks by the extension, and `NextPollTick` widens the gap off that same value. Elapsed now reads `time.Since(start)`, which is what the other three waits already did, and the window it is reported against moves with the deadline for the same reason: held at the flag's timeout, an extended wait printed "32m0s elapsed of 30m0s" and read as the CLI blowing through its own timeout.

The deadline test itself had grown to three copies, two of them from this branch. `pollSleep` takes the deadline now and reports whether the wait may go on, so the check lives with the sleep it bounds. The failed-poll branch keeps its own one-line guard: it has to run before the failure cap, because a wait whose window has closed timed out rather than gave up.

For that budget to mean anything, a 429 no longer counts toward `MaxConsecutivePollFailures`. That cap is for a server the CLI cannot reach; a 429 is the server answering, and counting it ended both command waits at five throttled polls no matter how far the budget had just carried the deadline. Throttled polls run on their own counter, which still grows the backoff, and stop when `Extend` refuses. All three waits say so once per throttled stretch rather than per poll, since the cap is no longer there to bound it. The notice lives on `ThrottleBudget.WarnThrottled`, which is `ShouldWarn` plus the wording: the line existed twice already and `exec --wait` was the third site, so it moved onto the type that decides whether it prints. `exec --wait` is where it is new behavior—it spent the budget silently before, leaving a user unable to tell a slow approver from a CLI the server is rate limiting.

A run denied or held again after approval no longer shares one sentence with the other. A job the server parked for a fresh approval is not a denial, and telling that caller to re-run contradicts the pending text `HandlePendingApproval` prints moments later on the same stream, which says the command runs on its own once approved. Each shape now says what its own way out is.

A fatal 4xx on the status read now ends the wait on the pending contract instead of returning the raw error. The poll submits nothing, so a 403 or a 404 on the detail GET says nothing about the request's validity, and exit 1 reads as retryable to an agent that answers it by filing a second request for the same command — the same harm the give-up branch beside it already avoids. The cause still reaches stderr through the warning that branch uses.

The deadline timer drops its `if !timer.Stop() { <-timer.C }` drain—but not because `Stop` cannot return false here. It does, for any timer that has already fired, which is precisely the interleaving this line is reached in. Since Go 1.23 the timer channel is unbuffered and `Stop` guarantees no stale value is delivered, so the drain is a receive that can only block forever. The old code was not dead; it was a hang waiting for the both-ready case.

## Not in this PR

Subscribing to the sudo decision event. alpacax/alpacon-server#3320 publishes `approval_granted` / `approval_rejected` / `approval_expired`, but it has no release tag yet, an `exec` without `--work-session` has no channel to subscribe to, and a dropped frame is silent — so the poll is needed underneath it either way. That work will build on `api/event/waiter.go`, which already subscribes with a REST catch-up.

## Verification

Rebased onto `main` at `8356098`, which brought in #417 (parallel tests and the testifylint adoption). Two test files conflicted and were resolved by keeping this branch's rewritten poll tests under #417's `t.Parallel()` and `require` conversions; `api/event/sudolistener_test.go` merged cleanly but stopped compiling, because #417 added a test reading `mfaPollingInterval` and this branch had moved that constant onto a listener field. Both fixes are folded into the commits that caused them.

Re-run on the pushed head (`f9240f4`), go1.25.13:

```
go build -o /dev/null .          ok
go vet ./...                     no output
go test -race -shuffle=on ./...  no FAIL; 42 packages ok, 16 with no test files
golangci-lint run ./...          0 issues.
```

Every new or changed test was seen red. Putting `NextPollTick` back turns the two step-up pacing tests red, reporting an expected 10ms against an actual 100ms. Rewording the wrapped refresh error turns the step-up completion test red. Printing `timeout` again turns the work-session elapsed test red with "1s elapsed of 2s" where the extended window is 3s. Putting the two post-approval branches back under one condition turns that warning test red, reporting the parked case printing "denied again" and the re-run advice it must not carry. Blanking the session id turns both work-session refusal tests red.

The four tests that assert poll timing run under `testing/synctest`, so the gaps and elapsed times come off a fake clock and are exact rather than thresholds carrying a margin for a loaded runner. That rules out `httptest` — a real network read is never durably blocked, so the bubble's clock would not advance — hence the stub `http.RoundTripper` in `pkg/testutil`. Restoring the per-poll budget reset turns the two ceiling tests red with `"1.33s" is not less than "260ms"` and `"1.453s" is not less than "260ms"`.

A later self-review round added four checks and widened one window. `TestRunExecWithApprovalWait_ThrottleWarnsOnce` pins the new exec notice at one line per stretch; removing the `WarnThrottled` call turns it red with an expected 1 against an actual 0. `TestThrottleBudget_ResetRestoresTheAllowance` covers the two fields `Reset` clears that no test touched—dropping either `b.spent = 0` or `b.extensions = 0` turns its matching subtest red. `TestRunUseSessionRefusesASettledSession` drives the real `RunUseSession` against a rejected session and requires no next action names `work-session use`; it and the two refusal tests all turn red with the old next-action list restored. The refusal test that reads the JSON envelope now compares the whole body with `JSONEq` rather than a declared field list, which drops what it forgets to name. The subprocess timeout test's window goes from 30ms to 300ms: three ticks of a real HTTP round trip inside a subprocess is not enough margin for the poll that carries the request id.

Two comments in the fixed-tick tests claimed `waitFor = 125 * tick` was "not a whole multiple of the tick". It is exactly 125 of them, so the boundary the comment said it avoided was the case being run. The value becomes `125*tick + tick/2`, which makes the sentence true; no assertion changed.

Three tests were removed with the mechanism they tested. Two asserted the spinner restarting on every tick, and one a mid-wait switch from a denial code to a status hold, which only a re-submitting tick could produce. The spinner comment that promised a per-tick heartbeat in a redirected log is corrected in the same diff: no caller restarts it on a timer any more, so a long wait in a redirected log now prints one line rather than repeating.

Closes #312


utils/poll.go
```diff
@@ -13,16 +13,76 @@ const (
...
+
+// Extend grows deadline by delay when the count budget allows it and the
+// cumulative duration either still has room for delay or has not been spent at
+// all yet, and reports whether it did.
+//
+// The very first extension in a window is granted whole even when delay alone
+// overshoots limit: a zero-length grant would leave the poll no time to finish
+// the wait a throttled server itself just handed it. Every extension after
+// that must fit inside what remains, so one oversized grant cannot be
+// followed by an unbounded run of further ones.
```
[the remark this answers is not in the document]
jisung-02: Good catch—the docstring was wrong, not the code. The rule `Extend` actually implements is the one lifted verbatim from `api/event`: an extension is granted while the time spent so far is still under `limit` and the count is under `PollMaxThrottleExtensions`, so whichever grant carries the total past `limit` is taken whole and nothing follows it. There is nothing special about the first one.

Trimming that crossing grant is what I want to avoid: the delay comes straight from the server via `Retry-After`, and handing the poll less time than the server just asked for makes the next request another 429. The overshoot is one delay wide, and the caller caps that at `PollMaxBackoffTick` times its base tick.

Rewrote the comment to say that, and fixed the same claim in `CLAUDE.md`. `TestThrottleBudget_ExtendsUntilTheDurationRunsOut` already pins the behavior—three 4s grants against a 10s budget all succeed whole, the fourth is refused.

cmd/exec/run.go
```diff
@@ -532,49 +572,76 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
...
+			case err != nil:
 				spinner.Stop()
 				return err
 			}
-			restartSpinner()
-			// Each tick re-submits, so a fixed 5s gap over a 30m wait outspends the
-			// default 1000/hour service-token quota.
+			failures = 0
+			throttles = 0
+			budget.Reset()
```
junho226: Critical: resetting the budget on every successful poll defeats the bound it exists for.

`ThrottleBudget.Reset` is documented as the progress signal—"called when the poll sees progress" (`utils/poll.go`)—and `pollCommandExecution` honors that by resetting only inside `IsRunningStatus` (`api/event/event.go:355-358`), where the deadline is refreshed alongside it. This site resets on any `err == nil` poll instead. But a successful read of `pending_approval` is not progress: the grant is exactly where it was, and the deadline was not refreshed.

The effect is that the extension allowance is re-earned by a single non-429 response. One over-quota stretch spends up to `waitTimeout`, one 200 refills it, the next stretch spends another—so `--wait-approval 5m` against a token hovering at its quota boundary has no stated ceiling, and the PR body's "extend by at most one timeout" no longer describes this wait.

Gate the reset on a status change, or drop it and let the budget run for the life of the wait—the deadline already bounds that. Same fix needed at `cmd/worksession/worksession_create.go:560`.

cmd/worksession/worksession_create.go
```diff
@@ -460,10 +550,14 @@ func pollForApproval(ac *client.AlpaconClient, id string, untilActive bool, inte
...
 			utils.CliWarning("Poll failed (%s); still waiting.", err)
-			pollSleep(remaining, utils.NextPollBackoff(interval, failures-1, utils.RetryAfter(err)))
+			if !pollSleep(deadline, utils.NextPollBackoff(interval, failures-1, utils.RetryAfter(err))) {
+				return nil, timedOut
+			}
 			continue
 		}
 		failures = 0
+		throttles = 0
+		budget.Reset()
```
junho226: Critical: same as `cmd/exec/run.go:585`—a successful poll that reads `pending` is not progress, so resetting the throttle budget here lets the deadline be extended by another whole `timeout` after every non-429 response. `pollForApproval` has no other ceiling, so the wait's bound becomes "one timeout per successful poll" rather than the one timeout the docstring above promises.

cmd/exec/run.go
```diff
@@ -532,49 +572,76 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
...
+			}
+			switch outcome {
+			case outcomeApproved:
+				spinner.Stop()
+				return runAfterApproval(ac, serverName, command, username, groupname, env, workSessionID, purpose, out)
+			case outcomeNotGranted:
+				spinner.Stop()
+				// Settled without a grant. CommandRejectedError is what already carries
+				// this to ExitCodeNotApproved, so the two denial shapes exit alike.
+				return &event.CommandRejectedError{CommandID: cmdID}
```
junho226: Warning: an expired grant is reported as a reviewer's rejection.

`approvalOutcomeOf` folds `rejected` and `expired` into `outcomeNotGranted`, and both then return `CommandRejectedError`, whose message is fixed text: `command cmd-1 was rejected by a reviewer` (`api/event/types.go:174`). Exit 6 is right for both—the README lists expiry under it—but the sentence tells a user whose approval window simply lapsed that a human said no, which points them at the wrong next move: go ask the reviewer, rather than re-request.

`TestRunExecWithApprovalWait_RejectionEndsTheWaitWithoutAGrant` asserts only the error type for both statuses, so the wrong wording is invisible to the suite. Carry the settled status through—either a field on the error, or a `CliWarning` naming the status before this return.

cmd/exec/run.go
```diff
@@ -532,49 +572,76 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
...
 					utils.CliWarning("Approval wait gave up after %d failed polls (%s); the command is still pending.", failures, err)
-					return pendingDenial
+					return pendingWithRequestID(pendingDenial, lastRequestID)
 				}
-				restartSpinner()
 				poll.Reset(utils.NextPollBackoff(approvalWaitPollInterval, failures-1, utils.RetryAfter(err)))
 				continue
-			default:
-				// Answered: it ran, or a terminal denial replaced the pending one.
+			case err != nil:
```
junho226, line 587: Suggestion: a fatal 4xx on the *status read* returns the raw error, so `exec` exits 1 while the approval request is still open. That is the harm the give-up branch three lines up deliberately avoids—its own comment says "Exit 1 reads as retryable, and an agent answers it by re-running exec and filing a second request for the same command."

The reasoning applies identically here. The poll no longer submits anything, so a 403 or a 404 on the detail GET says nothing about the request's validity. `return pendingWithRequestID(pendingDenial, lastRequestID)` would keep both ends of the same wait on one contract.

Behavior is unchanged from before the PR (a fatal 4xx from the re-submission exited 1 too), so this is a consistency call rather than a regression.

cmd/exec/run.go
```diff
@@ -496,34 +533,37 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
...
+				// instead, and counting it would end the wait before the budget it
+				// just spent could carry it anywhere.
+				if utils.HTTPStatusCode(err) == http.StatusTooManyRequests {
+					delay := utils.NextPollBackoff(approvalWaitPollInterval, throttles, utils.RetryAfter(err))
+					throttles++
+					if newDeadline, extended := budget.Extend(deadline, delay); extended {
+						deadline = newDeadline
+						if !timer.Stop() {
+							<-timer.C
+						}
```
junho226: Suggestion: this drain is dead code under this module's timer semantics, and would deadlock if it ever ran.

`go.mod` declares `go 1.25.13`, so Go 1.23+ synchronous timer channels are in force. `Stop` reports **true** for a timer that expired but whose value was never received—which is precisely the state here: nothing has received from `timer.C`, or we would be in the other select arm. I measured 0/300 false returns across four timings under go1.25.12 with this exact shape (a `NewTimer` left unreceived while another timer's arm runs).

And per `Timer.Stop`'s contract, "any receive from t.C after Stop has returned is guaranteed to block"—so on the one path where `Stop` could report false, `<-timer.C` hangs forever instead of draining.

```suggestion
						timer.Stop()
						timer.Reset(time.Until(deadline))
```

This is the only occurrence of the pre-1.23 idiom in the repo.

api/mfa/mfa.go
```diff
@@ -17,10 +17,12 @@ import (
...
+)
 
-	// stepUpPollInterval and stepUpTimeout bound the StepUpForSudo wait. They
-	// mirror the websh sudo listener (api/event/sudolistener.go) so the step-up
-	// feel is consistent across the two terminals.
+// stepUpPollInterval and stepUpTimeout bound the StepUpForSudo wait. They
+// mirror the websh sudo listener (api/event/sudolistener.go) so the step-up
+// feel is consistent across the two terminals. Vars so a test can shorten them.
+var (
 	stepUpPollInterval = 500 * time.Millisecond
```
junho226, line 26: Suggestion: four poll knobs move from `const` to `var` so tests can shorten them, and the tests assign them with no synchronization. `SudoListener` reads `mfaPollingInterval` from `pollMFACompletion` on a background goroutine, and `api/event` starts several listeners across its suite—so a listener that outlives its own test makes this a `-race` flake waiting for a slow runner.

A field on `SudoListener` (and a parameter for `StepUpForSudo`) removes the shared mutable state instead of timing around it.

api/mfa/mfa_test.go
```diff
@@ -221,3 +223,40 @@ func TestGetMFALinkByServerName_EmptyWorkspaceCostsNoRoundTrip(t *testing.T) {
...
+	ac := &client.AlpaconClient{
+		HTTPClient:    ts.Client(),
+		BaseURL:       ts.URL,
+		WorkspaceName: "pinned-ws",
+	}
+
+	err := StepUpForSudo(ac, "my-server")
+
+	require.Error(t, err)
+	got := atomic.LoadInt32(&polls)
```
junho226: Suggestion: the threshold is loose enough that the test can pass with the widening removed.

The widened schedule over a 120ms window at a 1ms base gives roughly 10 + 10 + 6 = 26 polls, so `< 60` has a wide margin—but on a loaded runner where each httptest round trip costs 2ms, a *fixed* 1ms tick also lands under 60 and the assertion holds for the wrong reason. Asserting the observed gap, or counting only the polls inside the first fast window, would pin the behavior rather than its side effect. Same shape in `api/event/sudolistener_test.go:224`.

junho226 (changes requested):
## Summary

`exec --wait` stops re-submitting the command every tick and instead reads the denied command's own detail for `sudo_grant_status`, which server 2.28.0 now supplies along with `sudo_approval_request_id`. The pacing rules the other three poll loops were duplicating move into `utils/poll.go` as `NextPollTick` / `NextPollBackoff` / `ThrottleBudget`. A 429 stops counting toward `MaxConsecutivePollFailures` in both command waits, `work-session create --wait` names the session it leaves behind on a refusal, and the two MFA step-up polls widen with age instead of ticking at a flat 500ms.

I checked the quota arithmetic that motivates the change: at a 5s base tick a 30-minute wait now costs ~50 GETs (10 at 5s, 10 at 25s, 30 at 50s) against the 1000/hour service-token quota, versus the previous ~360 ticks each costing several requests. The core claim holds.

## Critical issues (0)

## Warnings (2)

1. **[cmd/exec/run.go:585]** and **[cmd/worksession/worksession_create.go:560]** `budget.Reset()` on every successful poll defeats the bound the budget exists for. `Reset` is documented as the *progress* signal, and `pollCommandExecution` honors that by resetting only inside `IsRunningStatus`. A successful read of `pending_approval` is not progress, so the allowance is re-earned by any non-429 response and the PR body's "extend by at most one timeout" no longer describes either wait.
2. **[cmd/exec/run.go:597]** An expired grant is reported as a reviewer's rejection. Exit 6 is right for both outcomes, but `CommandRejectedError`'s fixed text tells a user whose approval window simply lapsed that a human said no, which points them at the wrong next move.

## Suggestions (4)

1. **[cmd/exec/run.go:579]** A fatal 4xx on the status read exits 1 while the approval request is still open—the exact harm the give-up branch three lines up avoids by design.
2. **[cmd/exec/run.go:559]** The `if !timer.Stop() { <-timer.C }` drain is dead code under this module's Go 1.25 timer semantics, and would block forever if it ever ran.
3. **[api/mfa/mfa.go:26]** Four poll knobs move from `const` to `var` and are assigned from tests with no synchronization while a listener goroutine reads them.
4. **[api/mfa/mfa_test.go:257]** Both widening tests can pass with the widening removed on a loaded runner.

## Highlights

- `approvalOutcomeOf`'s default-to-pending rule is the right default and the table test states it explicitly, `used` and the absent field included. A status nobody can read as an answer ending a wait would be the worse failure.
- The `runAfterApproval` boundary is well drawn: the granted run happens once, a second denial or a status hold warns and reports instead of opening a wait the user never asked for, and both shapes are tested.
- `pollSleep` taking the deadline and returning "may the wait go on" collapses three copies of the same check onto the sleep it bounds—and the one branch that keeps its own guard has a stated reason (the deadline outranks the failure cap, because a closed window timed out rather than gave up).
- The elapsed-time fix in the work-session progress line is a real bug found by the refactor, not a cosmetic one: reading `timeout - remaining` against a deadline a 429 moves made the reported time shrink as the wait grew, and the test pins wall clock across an extension.
- `TestExecWaitTimeoutCarriesTheApprovalRequestID` and `TestExecRejectedMidWaitExits6` drive the real subprocess against a real server, so the envelope and the exit code are covered end to end rather than at the return value.
- Removing the three tests along with the mechanism they tested—rather than keeping them green against nothing—is the right call, and the spinner comment that promised a per-tick heartbeat is corrected in the same diff.

## On the open Copilot comment

The remaining "Changes recommended" about `ThrottleBudget.Extend`'s documentation not matching its cap semantics looks like a false positive. The doc says a grant is given "while the count is still under `PollMaxThrottleExtensions`", the code refuses at `>=`, and `TestThrottleBudget_ExtendsUntilTheCountRunsOut` takes exactly 60 grants and is refused the 61st. All three agree.

## Verdict: changes requested

- [x] Tests included, and thorough
- [ ] The stated throttle bound holds (warning 1)
- [ ] The refusal message matches the outcome (warning 2)

cmd/exec/run.go
jisung-02: Right on both counts: `Reset` is documented as the progress signal, and a successful read of `pending_approval` is not progress. Fixed in 7fc5736.

`exec` drops the reset outright rather than gating it, because this loop returns the moment the status becomes an answer, so there is no progress left for it to observe. The allowance now runs for the life of the wait, like the deadline it extends.

`utils.ThrottleCeiling` states the bound that follows, one timeout plus one allowance plus the overshoot of the grant that crosses it, so both waits are tested against the same expression. The new `TestRunExecWithApprovalWait_APendingReadDoesNotRefillTheThrottleBudget` drives a throttled stretch long enough to spend the whole allowance, then one clean read of a grant sitting where it was. Seen RED by restoring the per-poll reset: `"1.33s" is not less than "260ms"`. Green it settles at 223ms, and it runs under `testing/synctest`, so that number is exact rather than a threshold carrying a margin.

cmd/worksession/worksession_create.go
jisung-02: Same fix, same commit 7fc5736. `pollForApproval` resets only when `s.Status` actually changes.

One thing your comment made me check: the first status read is not a change either. Nothing was observed before it, and the throttles that preceded it are exactly what the allowance is meant to bound, so treating it as progress would hand back a whole allowance those throttles had already spent. It records the status without resetting.

`TestPollForApproval_APendingReadDoesNotRefillTheThrottleBudget` was seen RED at `"1.453s" is not less than "260ms"` with the per-poll reset restored.

cmd/exec/run.go
jisung-02: Agreed. Exit 6 is right for both, but the fixed sentence sends a user whose window simply lapsed to the reviewer instead of to a fresh request. bd49fb9.

`approvalOutcome` splits into `outcomeRejected` and `outcomeExpired`, and `CommandRejectedError` carries an `Expired` flag, so an expired grant reads `command cmd-1 was not approved before the request expired` while a refusal keeps the wording it had. The zero value still renders the refusal, so the other two construction sites (`exec logs`, `errorFromDetails`) are untouched.

You were right that the test was the reason this got through: `TestRunExecWithApprovalWait_RejectionEndsTheWaitWithoutAGrant` now asserts the sentence for both statuses, not the type alone.

cmd/exec/run.go
jisung-02: Taken. fec2af7 returns `pendingWithRequestID(pendingDenial, lastRequestID)` here, so both ends of the same wait exit on one contract.

The cause is not lost: it goes to stderr through the same warning the give-up branch uses, which is where the ANSI sanitizer already sits. The renamed `TestRunExecWithApprovalWait_FatalClientErrorEndsTheWaitOnThePendingContract` pins the pending error, the `server said 401` line, and the single poll.

While in there: the README's exit 6 entry described exec's share of it as "when a reviewer rejected the command itself", which was already too narrow, since the same exit covers a lapsed window. Corrected in the same commit.

cmd/exec/run.go
jisung-02: Confirmed and applied as written, in f86fde8.

The hang is what settled it for me rather than the dead code: on the one path where `Stop` could report false, `<-timer.C` blocks forever per `Timer.Stop`'s own contract, so the branch was not merely unreachable but wrong if reached.

api/mfa/mfa.go
jisung-02: Fixed in 125d724, along the line you drew.

`SudoListener` takes `pollInterval` and `pollTimeout` fields, defaulted in `NewSudoListener` from `defaultMFAPollInterval` / `defaultMFAPollTimeout`. `api/mfa`'s two go back to `const`, and `StepUpForSudo` delegates to an unexported `stepUpForSudo` taking the pacing as arguments, so the exported signature is unchanged and nothing but a test needs the shorter wait. No test assigns a package var any more.

The two widening tests had to move in the same commit, since they assign the vars this removes and would not compile apart from it.

api/mfa/mfa_test.go
jisung-02: You were right that the count could hold for the wrong reason, and the fix went further than the threshold.

Both tests now assert the widest observed gap, which no amount of load can produce without the widening, and the expected value is written out rather than read back from `NextPollTick`: deriving it there would let the widening be removed from both sides at once and the test would still pass.

They also moved onto `testing/synctest` in 125d724, so the gaps come off a fake clock and are exact instead of a threshold with a margin. That ruled out `httptest`, since a real network read is never durably blocked and the bubble's clock would not advance, hence the stub `http.RoundTripper` in `pkg/testutil`. Seen RED by making `NextPollTick` return its base tick unchanged: expected 100ms against an actual 10ms.

`api/event/sudolistener_test.go:224` got the same treatment, as you noted it would need.

api/mfa/mfa.go
```diff
@@ -164,20 +173,22 @@ func StepUpForSudo(ac *client.AlpaconClient, serverName string) error {
...
 		case <-deadline:
 			spinner.Stop()
-			return fmt.Errorf("MFA step-up timed out after %v", stepUpTimeout)
-		case <-ticker.C:
+			return fmt.Errorf("MFA step-up timed out after %v", timeout)
+		case <-poll.C:
 			completed, cerr := CheckMFACompletion(ac)
 			if cerr != nil || !completed {
 				// Non-fatal: the endpoint may lag the browser; keep polling.
+				poll.Reset(utils.NextPollTick(pollInterval, time.Since(started)))
```
junho226: Warning: this widening costs a window a human is watching, and the quota argument behind it does not reach this loop.

With `tick = 500 ms`, `NextPollTick` gives 500 ms for the first 5 s (`pollFastWindow` 10), then 2.5 s out to 30 s (`pollMediumWindow` 60, `pollMediumTick` 5), then 5 s—about 26 polls across the 60 s window where the fixed ticker did 120. Two consequences:

- Worst-case detection lag after the user finishes in the browser goes from 0.5 s to 5 s.
- A completion at t≈56–59 s can be missed entirely: the last poll lands at 55 s and the next races the 60 s deadline, where the fixed tick would have caught it within half a second. 55 s is slow but ordinary when someone has to find their phone.

The PR reasons carefully about *not* adding backoff here—"one failure would eat half a window somebody is watching a browser through"—and then applies widening, which eats up to 5 s of that same window at the tail. And the quota this PR exists to fix is `service_token: 1000/hour`, while by this function's own docstring a step-up is interactive-only ("scripts, CI, and AI agents never receive a presence denial in the first place"), so the ~94 requests saved per step-up are not the ones that were exhausting anything.

Simplest resolution is to leave the two MFA loops on their fixed tick, or cap them at `pollMediumTick` so the worst lag is 2.5 s rather than 5 s. `api/event/sudolistener.go:315` is the same change and the same reasoning.

api/event/sudolistener.go
```diff
@@ -288,24 +295,24 @@ func (sl *SudoListener) handleSudoMFA(event sudoMFAEvent) {
...
 			completed, err := mfa.CheckMFACompletion(sl.ac)
-			if err != nil {
-				continue
-			}
-			if completed {
+			if err == nil && completed {
 				return true
 			}
+			// The endpoint may lag the browser; a failure here is not an answer.
+			poll.Reset(utils.NextPollTick(sl.pollInterval, time.Since(started)))
```
junho226: Warning: same as `api/mfa/mfa.go:191`—see that comment for the numbers.

The listener's window is the same 60 s against a server-side pending-grant expiry, and the buffer over that expiry is what `defaultMFAPollTimeout`'s comment says the design depends on. Spending up to 5 s of it per poll at the tail narrows exactly that buffer, which is the opposite of what the constant is there for.

Moving the pacing to fields is a clear improvement and unaffected by this—only the `NextPollTick` call is in question.

cmd/exec/run.go
```diff
@@ -496,34 +539,38 @@ func RunExecWithApprovalWait(ac *client.AlpaconClient, serverName, command, user
...
+				// the cap below—the throttle budget and the deadline bound it
+				// instead, and counting it would end the wait before the budget it
+				// just spent could carry it anywhere.
+				if utils.HTTPStatusCode(err) == http.StatusTooManyRequests {
+					delay := utils.NextPollBackoff(approvalWaitPollInterval, throttles, utils.RetryAfter(err))
+					throttles++
+					if newDeadline, extended := budget.Extend(deadline, delay); extended {
+						deadline = newDeadline
+						// No drain before Reset: under this module's Go 1.23+ timer
+						// semantics Stop reports true for a timer nobody received from,
```
junho226: Warning: dropping the `if !timer.Stop() { <-timer.C }` drain is right, but the reason given is not the reason.

`Timer.Stop` returns **false** if the timer has already expired—whether or not anyone received the value. Reaching this line from the `poll.C` case with `timer` already fired (both channels ready, `select` picked `poll.C`) is exactly when it returns false, so "Stop reports true for a timer nobody received from" does not hold.

What makes the removal correct is only the second clause: since Go 1.23 the timer channel is unbuffered and `Stop` guarantees no stale value is delivered, so the drain is a receive that blocks forever. The old code was not dead—it was a hang waiting for the both-ready interleaving.

Worth fixing because this comment exists to justify a deliberate removal. Left as is, it tells the next reader "`Stop` cannot return false here," which is the premise under which someone re-adds the drain.

```suggestion
						// No drain before Reset: under this module's Go 1.23+ timer
						// semantics the channel is unbuffered and a receive after Stop
						// is guaranteed to block, so the drain could only hang.
```

cmd/worksession/worksession_create.go
```diff
@@ -304,14 +308,26 @@ so it is recorded and scoped accordingly.`,
 	},
 }
 
+// useDecision is what the post-create switch does about --use: nothing, attach
+// now, wait for the session to become attachable, or refuse.
+type useDecision int
```
junho226, line 314: Suggestion: the constants stayed at line 28 and the type came down here, ~280 lines away.

Go does not care about the order, but a reader hitting `useDecisionNoop useDecision = iota` at the top now has to go looking for the type. The doc comment the move was for reads just as well above the const block, where the declaration already was—so this trades placement for nothing.

It is also the only edit in the PR unrelated to polling, in a diff that is already 1382 lines.

The new `approvalOutcome` in `cmd/exec/run.go` has the same shape (constants at 60, type at 202). If the split is a deliberate convention—configuration constants near the top, types near their use—then one line saying so settles it for both and I will stop reading it as drift.

junho226 (dismissed):
## Summary

`exec --wait` polled for a sudo approval by re-submitting the whole command every 5 s, and one tick was six-plus requests, so a 30-minute wait outspent the `service_token: 1000/hour` quota and ended on the 429 rather than on an answer. Server 2.28.0 added `sudo_grant_status` to the command detail, so the wait is now one GET per tick reading a status. The other three poll loops move onto shared pacing in `utils/poll.go`, and the 429 deadline allowance becomes `utils.ThrottleBudget`.

I traced the parts that carry the behavior rather than taking them from the description:

- `RemoteCommandError` has exactly one construction site (`errorFromDetails`), and it now sets `CommandID`, so the `denialErr.CommandID == ""` fallback is a genuine backstop rather than the common path.
- `ThrottleCeiling(timeout, tick) = 2*timeout + 60*tick` holds as an upper bound: `Extend` refuses once `spent >= limit`, so cumulative spend is at most `limit` plus one grant, and that grant is itself capped at `PollMaxBackoffTick * tick`. The extension count cannot beat it either.
- `cmd/exec` never calls `budget.Reset()`, so the ceiling is absolute there. `api/event` resets only alongside a full deadline refresh on a running command, and `cmd/worksession` only on an actual status change—the doc comment's "a wait that refills its budget has no ceiling at all" is accurate about all three.
- `pollSleep` caps its step at the remaining deadline, so worksession's 10-minute capped backoff cannot overshoot the window.
- `streamApprovedCommand` is still live at `run.go:513` (the initial status-hold path); only the mid-wait resume went away with the re-submitting tick.

## Critical issues (0)

None.

## Warnings (2)

**1. [`api/mfa/mfa.go:191`, `api/event/sudolistener.go:315`] Widening the two MFA polls costs a human-facing window, and the quota argument that motivates it does not apply to these loops.** Worst-case detection lag after the user finishes in the browser goes from 0.5 s to 5 s, and a completion at t≈56–59 s can be missed entirely. See the inline comment—the PR's own argument against backoff here applies to the widening too.

**2. [`cmd/exec/run.go:566`] The comment justifying the drain removal states a property `Timer.Stop` does not have.** The removal is right; the stated reason is not, and it is the premise under which someone re-adds the drain.

## Suggestions (2)

**1. [`cmd/worksession/worksession_create.go:313`] `useDecision` moved ~280 lines from its own constants**—the only edit in the PR unrelated to polling. `approvalOutcome` in `cmd/exec/run.go` has the same shape (constants at 60, type at 202); if the split is a convention rather than an accident, one line saying so settles both.

**2. [`cmd/exec/run.go`] The exec wait never calls `budget.ShouldWarn()`**, so a throttled 30-minute wait says nothing while the other two waits warn once. Probably deliberate (a warning mid-spin garbles the spinner), but nothing says so.

## Highlights

- **The fix removes the mechanism rather than tuning it.** The old loop was side-effect-safe only by an argument about server behavior; the new one is a status read, so `isPollFailure` can afford to be liberal and its old three-paragraph justification collapses to three lines. That is the right direction for a comment that long.
- **Rejected and expired are kept apart at the type level** while sharing an exit code, because the sentence a user reads decides where they go next. `CommandRejectedError.Expired` is the smallest thing that carries it.
- **`ThrottleBudget.Reset` is tied to an observed status change, not to any successful poll**, with the reasoning written where it is decided (`lastStatus != "" && s.Status != lastStatus`). Refilling per poll would have left no ceiling at all, and the first-read exclusion is the subtle half.
- **The elapsed-time bug that extending the deadline exposed**—`timeout - remaining` shrinking as the wait grew—was fixed at all four waits rather than at the one that showed it.
- **The tests measure the schedule, not the poll count.** `PollRecorder.WidestGap` under `testing/synctest`, with the reason `httptest` cannot work here (a real network read is never durably blocked, so the bubble's clock will not advance) written down in `StubTransport`. The mutation results are quoted with actual values (`"1.33s" is not less than "260ms"`), which is what makes them checkable rather than claimed.
- **Three tests removed with the mechanism they tested**, and the spinner comment that promised a per-tick heartbeat corrected in the same diff rather than left describing behavior that no longer exists.
- `SudoListener` moving its pacing from package vars to fields, with the race it removes named: a listener outliving its own test would otherwise read the next test's assignment.

## Verdict: APPROVED

- [x] Tests included (CI is the pass gate)
- [x] Documentation updated—`CLAUDE.md` and `README.md` touched; one comment states a wrong reason (Warning 2)

api/mfa/mfa.go
jisung-02: You are right, and the numbers are what settled it. Reverted in 4683425: both step-up loops are back on a fixed 500ms ticker.

I had the widening backwards on the one axis that matters here. The quota `NextPollTick` exists for is the service token's 1000/hour, and by `StepUpForSudo`'s own contract no script, CI job or agent ever reaches a step-up—so the ~94 requests it saves are not requests anything was short of. What it does cost is the window a person is watching, exactly as you counted: 26 polls where the ticker did 120, worst-case detection lag from 0.5s to 5s, and a completion at t=56-59s missed outright because the last poll lands at 55s.

I did not take the `pollMediumTick` cap. It is still five times the lag for a saving this loop has no use for, and it would leave a third pacing rule to explain. The pacing-on-fields refactor is untouched—only the `NextPollTick` call went.

The two tests now assert `WidestGap() == tick`; seen RED by putting `NextPollTick` back, which reports an expected 10ms against an actual 100ms. `CLAUDE.md`'s poll-pacing entry records the carve-out and why.

api/event/sudolistener.go
jisung-02: Fixed in the same commit, 4683425. The listener is back on a fixed ticker too.

Your reading of `defaultMFAPollTimeout` is the part I had missed: the buffer it keeps over the server's pending-grant expiry is the whole reason the constant is 60s, and a tail that widens to 5s per poll spends that buffer against the person still in the browser. The comment above the loop now says that, so the next reader has the reason in front of the code rather than in this thread.

The move to fields stays as it was.

cmd/exec/run.go
jisung-02: Applied your suggestion verbatim in 484e74d.

You are right that the premise was false. `Stop` returns false for any timer that has already fired, received or not—and reaching that line from the `poll.C` case with the timer fired is exactly when it does. The removal stands on the second clause alone: unbuffered channel since Go 1.23, `Stop` guaranteeing no stale delivery, so the drain is a receive that can only block. The old code was a hang waiting for the both-ready interleaving, not dead code, and I have corrected that claim in the PR body as well.

cmd/worksession/worksession_create.go
jisung-02: Not moving it back, and here is the one line you asked for—it already exists, in `CLAUDE.md` under "Go declaration order": top-level declarations follow const then var then type then func, and private helper types belong in the type block, never between functions. So the split is the convention rather than drift, and `approvalOutcome` in `cmd/exec/run.go` has the same shape for the same reason.

The old position was the one that broke it: `type useDecision int` sat between two const blocks. Where it is now it sits with `terminalWaitError` and `pendingWaitError`, which were already there, after the last `var` and before the first `func`.

You did find something in that comment, though, just not the placement. It said "nothing, attach now, wait for the session to become attachable, or refuse", which swaps the last two against `decideUseAction`: `useDecisionErrorNeedsWait` is the default case and it is the one that refuses without `--wait`, while `useDecisionSkipScheduled` is the approved-but-not-active case that waits. Corrected in c15a7ce.

You are also right that this is the only edit here unrelated to polling. It went in when the doc comment was written; I have kept it rather than reverting it into a form the declaration-order rule rejects, but say the word and I will pull both the move and the comment out into their own PR.

junho226 (approved):
## Summary

`exec --wait` stops re-submitting the command and reads the denied command's own `sudo_grant_status` instead, and the CLI's other three poll loops move onto one set of pacing rules in `utils/poll.go`. The re-submitting poll was a real defect—one tick was six or more requests, so a 30-minute wait outspent the service-token quota and then ended on the 429 rather than on the approval—and reading a status at one GET per tick is the right shape for it. The supporting work (a throttle budget that is spent rather than re-earned, a 429 that no longer counts toward the failed-poll cap, elapsed time measured from the start) all follows from that.

The branch was rebased onto `8356098` mid-review, so most of the file count is #417's testifylint sweep rather than this work. I re-read the four commits that landed on top and they close two things I had written up: `aa6fe69`/`59df3fd` give the exec wait the throttle notice its two siblings already had, and `524bbc0` stops `next_actions` offering `work-session use` for a session that can never be active again.

## What I checked against the code

**The three decided statuses match the server exactly.** `SudoGrant.STATUS_CHOICES` in alpacon-server is `pending_mfa`, `pending_approval`, `authorized`, `rejected`, `expired`, `used`, and `events/api/serializers.py::get_sudo_grant_status` reports the latest grant's status verbatim. `approvalOutcomeOf` decides on `authorized` / `rejected` / `expired` and leaves the other three—plus a nil field—pending, which is the correct partition.

**Dropping the timer drain is sound in this module.** The Go 1.23 unbuffered-timer-channel semantics are gated on the main module's language version, and `go.mod` declares `go 1.25.13`, so `timer.Stop()` followed by `Reset` really cannot leave a stale value, and the removed receive really could only block.

**The new `CommandID == ""` early return cannot silently disable the wait.** `RemoteCommandError` has exactly one construction site (`api/event/event.go`) and it now carries `CommandID: d.ID`, which `EventDetails` decodes from the detail's `id`. `approvalDenialOutput` already required that error type, so the guard reduces to the id being empty, which the detail response does not produce.

**The MFA carve-out is argued correctly.** At a 500ms base, `NextPollTick` holds 500ms for 5s, widens to 2.5s out to 30s, then 5s—about 26 polls where the fixed ticker did 120, and a completion at t=56-59s falls in the gap. Keeping a fixed tick for a wait bounded by a person in a browser, and moving the pacing off package vars onto fields and arguments so a listener goroutine cannot race the next test, are both right.

## Critical Issues (0)

None.

## Warnings (1)

### 1. `--wait` degrades silently against alpacon-server < 2.28.0, and nothing tells the user

`sudo_grant_status` arrives only from a server that has alpacax/alpacon-server#2911. Against an older one the field is absent, `approvalOutcomeOf` reads nil as `outcomePending`, and the wait runs the whole window and exits 4 saying the command is still pending—even though a reviewer approved it minutes earlier. The old re-submitting poll picked that approval up on the next tick, so this is a behavior regression on that server range, and the failure mode is silence for up to `--wait-approval`.

The design decision itself is right: a status nobody can read must not end a wait, and guessing "approved" from an absent field would be worse. What is missing is any way for the user to tell the two apart. The CLI has no version negotiation at all (no `MinServerVersion`, no server-version read anywhere), and the README's exit-4 entry—rewritten in this PR—names no minimum server version either.

alpacon-server is at 2.31.0, so the exposed range is narrow and this is not worth building version negotiation for. But a sentence in the README next to `--wait`, or a one-line note when the wait times out having never once read a non-nil `sudo_grant_status`, would turn a silent 30-minute nothing into a diagnosable one. The wait already tracks enough state to know it never saw the field.

## Suggestions (1)

### 1. `exec --wait` lost its non-TTY heartbeat, and the sibling wait in this same PR still has one

`Spinner.Start` prints its message once in a non-TTY, and the old loop's `restartSpinner()` reprinted it every tick, which is what gave a redirected log a heartbeat. Nothing restarts it now, so an `exec --wait` in CI or under an agent prints one line and then nothing for up to 30 minutes. The spinner comment is correctly updated to say so, so this is a known trade rather than an oversight.

The asymmetry is what makes it worth raising: `pollForApproval` in this same PR prints `utils.CliInfo("%s (%s elapsed of %s)", ...)` on every successful poll, and the exec loop already computes `time.Since(started)` for `NextPollTick`. A progress line on the same terms would close the gap without reintroducing a per-tick spinner restart.

## Nit (1)

### 1. `MaxConsecutivePollFailures` is not measuring a consecutive run once 429s are exempt

A 429 neither increments `failures` nor resets it, in both `RunExecWithApprovalWait` and `pollForApproval`. So `fail, fail, 429, fail, fail, fail` gives up at five failures even though the server answered in the middle, which is the opposite of what the constant's own comment argues ("a run of them means the CLI cannot reach the server"). Both loops behave identically, so this is consistency, not a divergence—but "consecutive" is now doing work the code does not do.

## Highlights

- **The 429 exemption is the load-bearing fix, not the status read.** Counting a 429 toward the failure cap ended both waits at five throttled polls no matter how far the budget had just carried the deadline. Spotting that the budget and the cap were fighting each other, and that the budget loses, is the kind of thing that only shows up when someone actually reads both.
- **`Reset` as the progress signal, with only a status change counting.** The note that any non-429 response refilling the allowance leaves the wait with no ceiling at all is exactly right, and `ThrottleCeiling` turning that bound into something two tests assert against is better than a comment claiming it.
- **`524bbc0` is the best catch in the PR.** `next_actions` naming a command that cannot succeed is worse than naming none, because the consumer is an agent that will run it and then have nothing left to try. Noticing that the two tests would have passed either way, and rewriting them to assert the absence plus a whole-envelope `JSONEq`, is the right repair—the assertion was the actual defect.
- **Backing the `NextPollTick` change out of the MFA polls after review.** Measuring what the widened schedule actually does to a 60s window—26 polls where there were 120, a completion at t=56s missed—and then keeping the fixed ticker is the right answer, and the reasoning is recorded where the next person will look for it.
- **Two settled outcomes sharing an exit code but not a sentence.** Telling someone whose window lapsed that a reviewer refused them points at the wrong next move; `CommandRejectedError.Expired` costs one bool and fixes it.
- **`aa6fe69` moving the notice onto the budget rather than copying the literal a third time.** The rule of two had already been passed, and putting the sentence on the type that decides whether it prints is the version that cannot drift.
- **`testing/synctest` for the timing tests.** A fake clock giving exact gaps instead of thresholds with a margin for a loaded runner is the difference between a test that pins the pacing and one that goes flaky in CI—and the stub `RoundTripper` in `pkg/testutil` follows from it rather than being a preference.

## Verification

Worktree `/Users/junho/alpacax/workspaces/alpacon-cli/review-pr-416`, HEAD `f9240f4` (matches the PR head).

| Command | Result |
|---|---|
| `go build ./...` | ok |
| `go vet ./...` | no output |
| `go test -race ./...` | 42 packages ok, no failures |
| CI (`gh pr checks 416`) | all pass, including `golangci-lint` on linux and windows and `govulncheck` |

Cross-checked against alpacon-server at `e096fadc1`: `sudo/models.py::SudoGrant.STATUS_CHOICES` and `events/api/serializers.py::get_sudo_grant_status` / `get_sudo_approval_request_id`.

## Verdict: APPROVED

The warning is about a server range that is already three minor versions behind and it does not put anything at risk—it costs a user one wait window and a re-run. Worth a README line either here or as a follow-up, your call.

- [x] Tests included and passing (locally and in CI), every new test seen red first
- [x] Documentation updated (`README.md` exit codes 4 and 6, `CLAUDE.md` poll pacing)

Issue #415, opened 2026-09-02
[FEATURE] test: run timer-bound tests under synctest and make the streaming fallback delays injectable


jisung-02:
## Summary

Two follow-ups from the #320 sweep and the test-parallelization branch (PR #413): move the remaining timer-bound tests onto `testing/synctest`, and give the five slow streaming-fallback tests in `api/event` a way to shorten the product delays they currently wait out for real.

## Problem Statement

A time-dependence audit of the suite (148 tests read, 98 touch real time) found no test that fails under load, but two things worth fixing.

First, about a dozen tests still wait on real timers or sleeps for no reason other than that nobody moved them: `WaitConnected` timeouts, the `RunExecWithApprovalWait` family in `cmd/exec`, MFA polling in `utils`, and elapsed-time bounds such as `assert.Less(elapsed, 1*time.Second)` that only say "eventually". Under `testing/synctest` (stable since Go 1.25) the same tests run in a bubble with a fake clock, the wait is instant, and a bound like "between 100ms and 1s" becomes exactly 100ms. Thirteen such conversions already exist as a WIP commit on branch `synctest-waits` (`9cbf94c`, on top of `fix/parallel-tests` at `f951aaa`); each was watched go red against a mutated implementation and passes `-race -count=30`.

Second, the suite's wall time is set by five tests that burn 12s, 7s, 7s, 7s and 4s of real time: `TestRunCommandStreaming_TerminalStatusErrors`, `_FallbackQuietWhenChunksUnavailable`, `_FallbackOnSessionFailure`, `_FallbackDrainsChunks` and `_GapFillRaceDoesNotDropChunk`. They drive the public entry point `runCommandStreamingWithWriter`, and the delays come from three product constants the test cannot reach: `streamPollTick = time.Second` (`api/event/event.go:34`, handed to `streamSubscribed` at lines 411 and 432, which builds its own `pollSeams` inline), `commandOutputConnectTimeout = 5 * time.Second` (`api/event/commandoutput_listener.go:18`, waited on at `event.go:400`), and `wsReconnectBaseDelay = 1 * time.Second` (`api/event/wslistener.go:14`, baked into the listener that `event.go:398` constructs). `synctest` cannot help here either: these tests talk to an httptest server over a real websocket, and a goroutine blocked on a socket is not durably blocked, so the bubble's clock never advances.

Measured on a 10-core machine with `go test -race ./...`:

| Configuration | Suite | `api/event` |
|---|---|---|
[...]

PR #417, opened 2026-09-02
test: run independent tests in parallel and finish testifylint adoption (#320)

### LINKED ISSUE #320: test: run independent tests in parallel and finish testifylint adoption

jisung-02:
Closes #320. Supersedes #413.

## What this changes

Twelve commits, test files plus `.golangci.yml`, `CLAUDE.md`, the CI workflow and three `intrange` rewrites in `utils` production code that the linter also sees.

1. `t.Parallel()` as the first statement of every top-level test that owns no process-wide state: 806 of 1067. The 261 that stay serial touch `t.Setenv`/`t.Chdir`, swap `os.Stdout`/`os.Stderr` (including `testutil.CaptureOutput`), reassign a package-level seam, read a package-level Cobra command (`Commands()` sorts in place and `Find()` builds flag sets lazily, so a read-only test still races), re-exec themselves as a helper process, or watch a timing window a parallel neighbour would narrow.
2. The nine bare table loops in `cmd/approval` and `cmd/worksession` run their cases under `t.Run`, with the empty filter named `empty` (Go would call it `#00`) and a parenthesised case renamed so `-run 'TestX/case'` can select it.
3. The classification corrections a review of #413 found: four pure tests made parallel, five helper-process tests and one busy-spin test made serial again.
4. `TestStreamSubscribed_FinCancelsThePoll` failed once under CPU saturation and, on inspection, proved nothing about cancellation (its fake server answers terminal, so the poll ends on its own). Two direct tests on `pollCommandExecution` now carry that guarantee and were watched go red with the cancel branches removed; the one poll test on a real clock runs on `fakePollClock` like its neighbours.
5. Eleven `require` calls inside httptest handlers and goroutines (`t.FailNow` off the test goroutine is undefined) become `assert` plus an early return or a channel hand-off, and nine `t.Fatalf` of the same shape with them: three in `cmd/worksession`, and six in `api/server` and `api/auth` that the first review found still standing, which now report with `t.Errorf`, answer 500 and return. `go-require` is enabled, and since it inspects testify calls only, the `CLAUDE.md` handler rule names `t.Fatalf` beside `require`.
6. The rest of #320: 175 `require-error` sites, 29 `empty` sites, 13 bare `assert.True(strings.HasSuffix(...))` turned into `assert.Equal` on the full path, `t.Helper()` on 15 helpers, 29 counted loops as `range` over int, three sleeps under `testing/synctest`, one `os.Unsetenv` made a real unset behind a `t.Setenv` restore, one `interface{}` as `any`. `require-error`, `empty`, `intrange` and `thelper` are enabled and the issue caps that hid the size of the sweep are lifted. The tests `main` gained since the branch point (`pkg/selfupdate`, `api/rbac`, `cmd/iam`, `cmd/update`, and the `t.Chdir` case in `utils/unzip_test.go` from #418, which stays serial) are brought to the same conventions.

## Why

Go runs test packages concurrently but the tests inside one package one after another unless each calls `t.Parallel()`. Nothing here did, so `go test -race ./...` took the sum of every sleep in `api/event`.

| Configuration | Suite | `api/event` |
| --- | --- | --- |
| `main` (serial) | 76 s | 74.2 s |
| this branch (parallel) | 19 s | 17.7 s |

Local numbers, Apple Silicon, 10 cores; the `ubuntu-latest` runner has 4 vCPUs, so CI gains less than the table shows. The suite is now bounded by its single longest test; shortening that one needs product-code seams and is tracked in #415 together with the remaining `synctest` conversions.

## Review of the branch

Six independent reviewers read every changed test on both sides of the diff for assertions that got weaker. Two findings, both fixed: the `unset` case of `TestDebugEnabled` had stopped exercising the absent-variable branch of `os.LookupEnv` (it now unsets for real, and the test fails when that branch is mutated), and an ftp backoff upper bound widened for parallel load went back to its original value, then was removed in the first review round (below): `alignedPollDelay` has its own table test and the poll count is asserted exactly, so the ceiling duplicated both. Handler `require` removals, `HasSuffix` to full-path `assert.Equal`, `Empty` conversions (every operand is a `string`) and the subtest conversions were confirmed equal or stronger.

## Review round 1

The first human review raised three warnings and six suggestions; all nine are applied, one commit each, so every thread can be read against its commit.

| Item | Commit |
| --- | --- |
| Six `t.Fatalf` calls inside httptest handlers, five in tests now parallel | f463817 |
| `api/ftp` wall-clock ceiling on a parallel test; `TestStreamSubscribed_FinCancelsThePoll` watches a timing window while parallel | f65d22c |
| Ceilings inside `synctest` bubbles are vacuous; `TestWSListener_WaitConnected_Timeout` left on the real clock; sudo listener test leans on `mfaPollingInterval` unstated | 81d1e90 |
| CI runs without `-shuffle=on` | dda3691 |
| Four `captureStderr` copies | 6ff0e74 |
| `cmd/csr` message restates `Empty` | 716d1f0 |

Two decisions differ from the suggestion as written. The handler fix uses `t.Errorf` rather than `assert.Fail`, because `api/server` and `api/auth` use the `testing` package directly and `t.Errorf` gives the same non-unwinding failure without a new import. The sudo listener test pins its dependency on `mfaPollingInterval` with a `require.Greater` at the top of the test rather than a comment, so lowering the constant fails with a message instead of a nil dereference inside the bubble.

## Verification

- `go test -race -count=1 -shuffle=on ./...` passes on 716d1f0; three earlier runs with 16 CPU-saturating processes and `-parallel 32` passed with no data race. Every commit before the review round was extracted on its own and builds, lints with 0 issues and passes `-race`; the six review-round commits were verified together on the final tree, and the three new exact-duration assertions were each seen fail with the expected duration off by a millisecond.
- `golangci-lint run ./...` reports 0 issues with the five newly enabled checkers; `gofmt -l .` and `go vet ./...` are clean on 716d1f0.
- Total statement coverage is unchanged at 64.3% (measured on `main` and on 716d1f0).

https://claude.ai/code/session_01LuA3veD1VrwdWoQ9RqBKko


api/server/server_test.go
```diff
@@ -193,6 +197,7 @@ func TestListRegistrationTokens(t *testing.T) {
 }
 
 func TestCreateRegistrationToken_WithExpiresAt(t *testing.T) {
+	t.Parallel()
```
geunwoonoh, line 200: Warning: this test is now parallel, and its handler still calls `t.Fatalf` at line 209.

`t.Fatalf` off the test goroutine calls `runtime.Goexit`, so the handler dies without answering and the client fails on an unrelated EOF. That is the shape 46b976a removed from `cmd/worksession`, and its commit message names parallel execution as the reason. `go-require` only inspects testify calls, so the linter does not see this one.

```go
if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
    assert.Fail(t, "failed to decode request body", "%v", err)
    return
}
```

geunwoonoh (commented):
## Review summary

This moves 807 of 1067 tests onto `t.Parallel()` and turns on the three testifylint checkers #320 deferred (`require-error`, `empty`, `go-require`) plus `intrange` and `thelper`. The only production changes are three `intrange` rewrites in `utils`, and their semantics are identical: `s.Len()` is evaluated once either way, and `i` was already unused in `resolveWritePath`.

Static review, no suite run; CI is green on all nine jobs. I checked the parallel/serial classification with a script rather than by reading, and it holds up.

---

## Critical issues (0)

None.

---

## Warnings (3)

### 1. Six `t.Fatalf` calls remain inside `http.HandlerFunc` bodies, five of them in tests this PR made parallel

This is the exact shape 46b976a removed from `cmd/worksession`. `t.Fatalf` calls `runtime.Goexit`, so the handler goroutine dies without answering and the client fails on an unrelated EOF. The commit message's own words apply here: "under parallel execution that misattributed failure lands interleaved with other tests' output". `go-require` only knows testify calls, so the linter stays green.

| Site | Test | Parallel |
| --- | --- | --- |
| `api/server/server_test.go:209` | `TestCreateRegistrationToken_WithExpiresAt` | yes |
| `api/server/server_test.go:466` | `TestGetAnsibleRegistrationGuideJSON` | yes |
| `api/server/server_test.go:560` | `TestRequestServerAction` | yes |
| `api/auth/auth_test.go:322`, `:335` | `TestCreateAPIToken` | yes |
| `api/auth/auth_test.go:61` | `TestLoginAndSaveCredentialsPasswordPreservesTargetMetadata` | no |

The fix is the one already applied in `cmd/worksession`:

```go
if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
    assert.Fail(t, "failed to decode request body", "%v", err)
    return
}
```

### 2. `api/ftp/ftp_test.go:1393` keeps a wall-clock upper bound on a test this PR made parallel

```go
assert.Less(t, elapsed, 3*time.Second, "adaptive backoff should finish well under the old fixed 2s+2s")
```

This contradicts the rule this same PR added to `CLAUDE.md`: "A test whose subject is a timing window stays serial too". The PR body records rejecting a widening of this bound from 3s to 3.9s for parallel load, so keeping it and going parallel looks deliberate, but the nominal run is 750ms and `ubuntu-latest` has 4 vCPUs, so under `-parallel 4` with `-race` the headroom is thinner than 4x suggests.

More to the point, the bound is redundant. `alignedPollDelay` already has a pure table test at `api/ftp/ftp_test.go:1337-1356`, and `assert.Equal(t, int32(3), calls.Load())` two lines up pins the poll count. Dropping the upper bound costs no coverage and removes the flake surface. If you want to keep it, keep this one test serial.

### 3. `api/event/event_test.go:1013` has the same rule tension

`TestStreamSubscribed_FinCancelsThePoll` is parallel and its subject is a `time.Sleep(10 * tick)` observation window. The relaxed verdict (`LessOrEqual(..., 1)`) makes it robust in practice, so this is consistency rather than risk. Either mark it serial or add a line saying why this particular window survives parallel load, so the next reader does not have to re-derive it.

---

## Suggestions (6)

### 1. Inside a synctest bubble the upper bounds are close to vacuous

`api/event/wslistener_test.go:145` and `api/event/sudolistener_test.go:212` measure fake time now, which is deterministic. `assert.Equal(t, 50*time.Millisecond, elapsed)` (and `100*time.Millisecond` for the sudo listener) is exact and would catch a stop path that spends an extra tick. Determinism is why these went into a bubble, so this strengthening is free.

### 2. `api/event/wslistener_test.go:55` was left on the real clock

`TestWSListener_WaitConnected_Timeout` kept `assert.Less(t, elapsed, 1*time.Second)` over a 100ms wait and became parallel, while its two neighbors in the same file went into `synctest.Test`. `WaitConnected` uses only channels and timers, no I/O, so the same wrapper applies directly.

### 3. `api/event/sudolistener_test.go:197` leans on an unstated constant relationship

`TestSudoListener_PollMFACompletion_Timeout` never reaches `mfa.CheckMFACompletion(sl.ac)` only because `mfaPollingInterval` is 500ms (`api/event/sudolistener.go:23`) and the stop lands at fake-time 100ms. Lower that constant below 100ms and the test dereferences a nil client, inside a bubble that forbids real I/O. A one-line comment naming the dependency, or a non-nil fake client, removes the trap.

### 4. `cmd/csr/csr_test.go:119` carries a message that restates the assertion

`CLAUDE.md` says to keep only a reason the code cannot say, and `Empty` prints its operand itself.

### 5. CI does not run what the PR verified with

The PR body's evidence is `go test -race -count=1 -shuffle=on ./...`, but `.github/workflows/build-and-test.yaml:62` runs `go test -race -v ./...`. Adding `-shuffle=on` keeps the serial/parallel classification honest as new tests land; without it an ordering dependency introduced later stays invisible until it flakes.

Separately, 76s to 19s is a 10-core local measurement and the runner has 4 vCPUs, so the CI gain will be smaller. Not a defect, just worth a line in the body so nobody reads 4x off the table.

### 6. Duplication, pre-existing, mentioned not requested

Four `captureStderr` implementations exist: `utils/log_test.go:15`, `api/auth0/auth0_test.go:337`, `api/event/event_test.go:1771`, `cmd/exec/run_test.go:539`. The shared `pkg/testutil.CaptureOutput` already does this, and `cmd/server/server_create_test.go:43` shows the right pattern by wrapping it. None of these came from this PR, so it is out of scope, but consolidating would shrink the "swaps `os.Stderr`, therefore serial" surface to one place.

---

## Highlights

- **The classification is genuinely correct.** Scripted over the tree: no parallel test reaches `t.Setenv`, `t.Chdir`, an `os.Stdout`/`os.Stderr` swap, or `utils.OutputFormat`, directly or through a helper, and none of the 180 package-level Cobra commands is read from a parallel test. `parseTunnelCommandArgs` building a fresh `&cobra.Command{}` instead of reading `TunnelCmd` is the reason `cmd/tunnel` could go parallel at all.
- **The bare-table sweep is complete.** No `[]struct{...}` table is still asserted inside a bare `for` loop anywhere in the repo.
- **`filterCaseName` is a real fix, not cosmetic.** `t.Run("")` becomes `#00`, and the same commit had deleted the `input: %q` message that used to identify that row, so the empty filter would have ended up less identifiable than before.
- **Replacing `TestStreamSubscribed_FinCancelsThePoll`'s cancellation claim was the right call.** A test that stays green with the feature deleted is worse than no test, and the two direct `pollCommandExecution` tests were watched go red with the cancel branches removed.
- **Commit messages carry constraint, rejected, and confidence notes.** I disagree with the ftp bound decision, but the reasoning was recorded, which made this review much faster.

---

## Verdict: COMMENTED

No critical issues. Worth fixing before merge: warning 1 (the six handler `t.Fatalf` sites, since this PR declares that class eliminated) and warning 2 (the ftp wall-clock bound, which costs nothing to drop). The rest can follow later.

## Checklist

- [x] Code follows stack conventions (`gofmt`, `go vet` clean, declaration order, errcheck patterns)
- [x] Tests included and passing (CI 9/9 green)
- [x] No security vulnerabilities (no hardcoded secrets, test-only changes)
- [x] Documentation updated (three `CLAUDE.md` test-pattern entries)
- [ ] New rules and code are consistent (warnings 2 and 3)

api/server/server_test.go
```diff
@@ -424,6 +436,7 @@ func TestGetRegistrationTokenAttributes_EmptyList(t *testing.T) {
 }
 
 func TestGetAnsibleRegistrationGuideJSON(t *testing.T) {
+	t.Parallel()
```
geunwoonoh, line 441: Warning: same as `TestCreateRegistrationToken_WithExpiresAt` above. The handler for this now-parallel test calls `t.Fatalf` at line 466.

```go
if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
    assert.Fail(t, "failed to decode request body", "%v", err)
    return
}
```

api/server/server_test.go
```diff
@@ -511,6 +525,7 @@ func TestDeleteServer(t *testing.T) {
 }
 
 func TestRequestServerAction(t *testing.T) {
+	t.Parallel()
```
geunwoonoh, line 532: Warning: same pattern again. The handler for this now-parallel test calls `t.Fatalf` at line 560.

```go
if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
    assert.Fail(t, "failed to decode request body", "%v", err)
    return
}
```

api/auth/auth_test.go
```diff
@@ -288,6 +291,7 @@ func TestResolveTokenID(t *testing.T) {
 }
 
 func TestCreateAPIToken(t *testing.T) {
+	t.Parallel()
```
geunwoonoh, line 296: Warning: this test is now parallel and its handler calls `t.Fatalf` twice, at lines 322 and 335.

Same class as the `api/server/server_test.go` sites: the handler goroutine exits without writing a response, so the failure surfaces as an EOF from the client rather than the message you wrote. `TestLoginAndSaveCredentialsPasswordPreservesTargetMetadata` (line 61) has a third instance; it stays serial, but the goroutine problem is the same.

```go
if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
    assert.Fail(t, "failed to decode request body", "%v", err)
    return
}
```

api/ftp/ftp_test.go
```diff
@@ -1333,6 +1359,7 @@ func TestAlignedPollDelay(t *testing.T) {
 }
 
 func TestPollTransferStatus_BacksOffThenSucceeds(t *testing.T) {
+	t.Parallel()
```
geunwoonoh, line 1362: Warning: this test became parallel while keeping a wall-clock upper bound at line 1393 (`assert.Less(t, elapsed, 3*time.Second, ...)`), which contradicts the rule this PR adds to `CLAUDE.md`: "A test whose subject is a timing window stays serial too".

The nominal run is 750ms and `ubuntu-latest` has 4 vCPUs, so under `-parallel 4` with `-race` the 4x headroom is thinner than it looks. The PR body notes rejecting a widening from 3s to 3.9s, so I take it the bound is considered load-proof, but the bound is also redundant: `alignedPollDelay` has its own pure table test at lines 1337-1356, and `assert.Equal(t, int32(3), calls.Load())` already pins the poll count.

Dropping `assert.Less(t, elapsed, 3*time.Second, ...)` costs no coverage. Keeping it argues for keeping this one test serial.

api/event/event_test.go
```diff
@@ -915,6 +1011,7 @@ func TestStreamSubscribed_FinEndsRunWithoutThePoll(t *testing.T) {
 // on requesting a command that is already over—spending the throttle budget this
 // pacing exists to protect, and outliving the process's other streams.
 func TestStreamSubscribed_FinCancelsThePoll(t *testing.T) {
```
geunwoonoh, line 1016: Warning: same rule tension as `api/ftp/ftp_test.go` - the subject here is a `time.Sleep(10 * tick)` observation window, and the new `CLAUDE.md` line says such a test stays serial.

The relaxed verdict (`LessOrEqual(..., 1)`) makes it robust in practice, so this is consistency rather than flake risk. Either mark it serial, or add a line saying why this window survives parallel load so the next reader does not have to work it out.

api/event/wslistener_test.go
```diff
@@ -118,22 +128,26 @@ func TestWSListener_ListenLoop_DoesNotDialWhenAlreadyStopped(t *testing.T) {
...
-	result := w.WaitConnected(5 * time.Second)
-	elapsed := time.Since(start)
+		start := time.Now()
+		result := w.WaitConnected(5 * time.Second)
+		elapsed := time.Since(start)
 
-	assert.False(t, result, "should return false when done is closed")
-	assert.Less(t, elapsed, 1*time.Second, "should exit quickly on shutdown")
+		assert.False(t, result, "should return false when done is closed")
+		assert.Less(t, elapsed, 1*time.Second, "should exit quickly on shutdown")
```
geunwoonoh: Suggestion: inside the bubble this bound is close to vacuous, because `elapsed` is fake time and therefore deterministic. Exact equality is free here and catches a stop path that spends an extra tick.

```suggestion
		assert.Equal(t, 50*time.Millisecond, elapsed, "the stop must land on the close, not a tick later")
```

api/event/sudolistener_test.go
```diff
@@ -187,22 +195,26 @@ func TestSudoListener_VerifySudoGrant_ServerError(t *testing.T) {
...
 
-	result := sl.pollMFACompletion()
-	elapsed := time.Since(start)
+		result := sl.pollMFACompletion()
+		elapsed := time.Since(start)
 
-	assert.False(t, result, "should return false when stopped")
-	assert.Less(t, elapsed, 2*time.Second, "should exit quickly when stopped")
+		assert.False(t, result, "should return false when stopped")
+		assert.Less(t, elapsed, 2*time.Second, "should exit quickly when stopped")
```
geunwoonoh: Suggestion: same as `wslistener_test.go` - fake time makes `elapsed` exact, so an equality assertion is strictly stronger than a 2s ceiling that nothing in the bubble can reach.

```suggestion
		assert.Equal(t, 100*time.Millisecond, elapsed, "the stop must land on Stop(), not a poll tick later")
```

api/event/wslistener_test.go
```diff
@@ -34,19 +37,23 @@ func TestWSListener_StopIsIdempotent(t *testing.T) {
...
+			time.Sleep(50 * time.Millisecond)
+			close(w.connected)
+		}()
+
+		result := w.WaitConnected(2 * time.Second)
+		assert.True(t, result, "should return true when connected")
+	})
 }
 
 func TestWSListener_WaitConnected_Timeout(t *testing.T) {
```
geunwoonoh, line 55: Suggestion: this one was left on the real clock while its two neighbors in this file moved into `synctest.Test`, and it became parallel keeping `assert.Less(t, elapsed, 1*time.Second)` over a 100ms wait.

`WaitConnected` uses only channels and timers with no I/O, so the same wrapper applies directly, and the bound then becomes an exact equality instead of a ceiling chosen for slack.

api/event/sudolistener_test.go
```diff
@@ -187,22 +195,26 @@ func TestSudoListener_VerifySudoGrant_ServerError(t *testing.T) {
 }
 
 func TestSudoListener_PollMFACompletion_Timeout(t *testing.T) {
```
geunwoonoh, line 197: Suggestion: this test only avoids calling `mfa.CheckMFACompletion(sl.ac)` because `mfaPollingInterval` is 500ms (`api/event/sudolistener.go:23`) and the stop lands at fake-time 100ms. Lower that constant below 100ms and the test dereferences a nil client, inside a bubble that forbids real I/O.

A one-line comment naming the dependency, or passing a non-nil fake client, removes the trap.

cmd/csr/csr_test.go
```diff
@@ -110,17 +111,17 @@ func TestListSubcommandFlags(t *testing.T) {
...
 func TestDownloadSubcommandFlags(t *testing.T) {
 	cmd, _, err := CsrCmd.Find([]string{"download-crt"})
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	outFlag := cmd.Flags().Lookup("out")
 	assert.NotNil(t, outFlag, "--out flag should exist")
 	assert.Equal(t, "o", outFlag.Shorthand, "short flag should be -o")
-	assert.Equal(t, "", outFlag.DefValue, "default value should be empty")
+	assert.Empty(t, outFlag.DefValue, "default value should be empty")
```
geunwoonoh: Suggestion: the message restates the assertion, which `CLAUDE.md` asks us to drop - `Empty` prints the operand itself.

```suggestion
	assert.Empty(t, outFlag.DefValue)
```

api/server/server_test.go
jisung-02: Applied in f463817. You were right that go-require never sees this shape: it inspects testify calls only, so the six `t.Fatalf` handlers stayed green while doing exactly what 46b976a removed. The handler now reports with `t.Errorf`, answers 500 and returns. I used `t.Errorf` rather than `assert.Fail` because this file uses the `testing` package directly; it fails the test without unwinding the goroutine, so the client gets a response and the decode error is what lands in the log. The `CLAUDE.md` handler rule now names `t.Fatalf` beside `require` and says why the linter is silent on it.

api/server/server_test.go
jisung-02: Applied in f463817, same rewrite as the site above: `t.Errorf`, 500, return.

api/server/server_test.go
jisung-02: Applied in f463817, same rewrite: `t.Errorf`, 500, return.

api/auth/auth_test.go
jisung-02: Applied in f463817. Both sites in `TestCreateAPIToken` and the serial one at line 61 take the same rewrite: `t.Errorf`, 500, return. You were right to include the serial one; the goroutine problem does not depend on `t.Parallel()`, only on where `t.Fatalf` runs.

api/ftp/ftp_test.go
jisung-02: Applied in f65d22c: the ceiling is gone, the lower bound and the poll count stay, and the test stays parallel. You were right that it was redundant. `alignedPollDelay` has its own table test and `calls.Load()` is asserted exactly, so the ceiling could only add a flake surface on a 4 vCPU runner. The rejected 3.9s widening is noted in the commit body as rejected for that reason, not the earlier one, and the PR body's sentence about it is corrected.

api/event/event_test.go
jisung-02: Applied in f65d22c: the test is serial now, with a two-line comment saying that the ten-tick window is the subject and parallel load would stretch the ticks it counts. Consistency with the rule this branch wrote is the reason, as you said; the `LessOrEqual(..., 1)` verdict already made it robust in practice.

api/event/wslistener_test.go
jisung-02: Applied in 81d1e90 as suggested: `assert.Equal(t, 50*time.Millisecond, elapsed, ...)`. Seen red with the expected duration set to 51ms, so the equality does carry information the ceiling did not.

api/event/sudolistener_test.go
jisung-02: Applied in 81d1e90. The expected value is the same 100ms, spelled as the `stopAfter` constant the test now declares once (see the thread at line 197). Seen red with the expected duration off by a millisecond.

api/event/wslistener_test.go
jisung-02: Applied in 81d1e90: the test runs inside `synctest.Test` like its neighbors and asserts `assert.Equal(t, 100*time.Millisecond, elapsed, ...)` in place of the 1s ceiling. Seen red with the expected duration set to 101ms. You were right that `WaitConnected` is channels and timers only, so the bubble applies directly.

api/event/sudolistener_test.go
jisung-02: Applied in 81d1e90, slightly differently from either option. The test now declares `const stopAfter = 100 * time.Millisecond` and opens with `require.Greater(t, mfaPollingInterval, stopAfter, "a poll before the stop would dereference the nil client")`, so the dependency is executable: lowering the constant fails with that message instead of a nil dereference inside the bubble. I did not pass a fake client because the poll would then do real HTTP inside the bubble, and the test's subject is the stop, not the poll.

cmd/csr/csr_test.go
jisung-02: Applied in 716d1f0: the message is dropped.

geunwoonoh (approved):
## Second pass: review round verified

Re-reviewed the six review-round commits and re-ran the verification scripts from the first pass against `716d1f0`. All nine items are applied, one commit each, so every thread reads against its commit.

| Item | Commit | Verified |
| --- | --- | --- |
| W1 six handler `t.Fatalf` | `f463817` | all six now `t.Errorf` + 500 + return; scan finds zero `require`/`t.Fatalf`/`t.FailNow` inside handlers or goroutines |
| W2 ftp wall-clock ceiling | `f65d22c` | `assert.Less(elapsed, 3s)` gone, lower bound and poll count kept, test stays parallel |
| W3 `FinCancelsThePoll` | `f65d22c` | serial with a two-line reason |
| S1 vacuous bounds in bubbles | `81d1e90` | two exact-duration equalities |
| S2 `WaitConnected_Timeout` | `81d1e90` | moved into the bubble, exact equality |
| S3 unstated constant | `81d1e90` | `require.Greater` guard |
| S4 csr message | `716d1f0` | dropped |
| S5 CI shuffle | `dda3691` | workflow and `CLAUDE.md` together |
| S6 four `captureStderr` copies | `6ff0e74` | all four wrap `testutil.CaptureOutput` |

### Two decisions that improved on the suggestion

`t.Errorf` over `assert.Fail` is the better call: `api/server` and `api/auth` use the `testing` package directly, and `t.Errorf` gives the same non-unwinding failure without pulling in testify. My suggestion would have added an import for nothing.

The sudo listener guard is also better than what I proposed. Instead of a comment or a fake client, the dependency is now executable:

```go
const stopAfter = 100 * time.Millisecond
require.Greater(t, mfaPollingInterval, stopAfter, "a poll before the stop would dereference the nil client")
```

Lowering the constant fails with that message instead of a nil dereference inside the bubble, and declining the fake client is right, since the poll would then do real HTTP inside a bubble that forbids it.

### S6 turned out to be a latent bug, not tidying

I flagged the `captureStderr` duplication as pre-existing and out of scope. The `api/event` copy read the pipe only after `fn()` returned:

```go
fn()
_ = w.Close()
_, _ = io.Copy(&buf, r)   // fn already deadlocked if it wrote past the 64KB pipe buffer
```

`testutil.CaptureOutput` drains concurrently (`pkg/testutil/capture.go:41`), so wrapping it removed a deadlock that was waiting for a large enough stderr write.

### Independent verification on `716d1f0`

- `go vet ./...` and `gofmt -l .` clean
- zero `require`/`t.Fatalf`/`t.FailNow` inside httptest handlers or goroutines
- zero `[]struct{...}` tables still asserted in a bare `for` loop
- zero parallel tests reading any of the 180 package-level Cobra commands
- zero parallel tests reaching `t.Setenv`, `t.Chdir`, an `os.Stdout`/`os.Stderr` swap, or `utils.OutputFormat`, directly or through a helper
- 806 parallel of 1067, matching the PR body

The timing invariant is now clean tree-wide, which is what makes this safe to leave alone. Every remaining upper bound sits on a serial test; every duration assertion on a parallel test is a lower bound, and load can only strengthen those.

| Site | Parallel | Kind |
| --- | --- | --- |
| `api/ftp/ftp_test.go:1390` | yes | lower |
| `cmd/worksession/worksession_create_test.go:205` | yes | lower |
| `cmd/exec/purpose_demand_test.go:184` | no | upper |
| `cmd/exec/run_test.go:261` | no | lower |
| `utils/spinner_test.go:125` | no | upper |

CI is green on all nine jobs, now including `-shuffle=on`. The PR body is current, down to correcting the stale sentence about the rejected 3.9s widening.

Nothing outstanding.

## Verdict: APPROVED

## Checklist

- [x] Tests included and passing (CI 9/9, `-shuffle=on` included)
- [x] Documentation updated (`CLAUDE.md` handler rule and test command)
- [x] New rules and code are consistent (the item left open in the first pass)

Issue #412, opened 2026-09-01
Sign release checksums so alpacon update verifies who published a release, not just that it arrived intact


jisung-02:
Follow-up to #408, which added `alpacon update`. Raised in review there and deferred deliberately; the README now states the gap plainly rather than leaving a reader to infer more than the checksum provides.

## Background

`alpacon update` downloads a release archive and verifies it against `alpacon-<version>-checksums.sha256`, which it picks out of the same release's asset list (`pkg/selfupdate/download.go`). `.goreleaser.yaml` publishes no signature.

That comparison proves the bytes arrived intact and match what the release store holds. It proves nothing about who put them there, because the archive and the checksums file come from the same place and nothing vouches for either. Anyone who can write to a release — a leaked release token, a compromised maintainer account, a poisoned release workflow — uploads a trojaned archive together with a checksums file that matches it, verification passes, and the CLI reports a successful update.

This CLI is the terminal-side client for a PAM product, so the binary it replaces is the one an operator uses to reach production. A compromised release is code execution on every operator workstation that updates, with that operator's privileges.

## What would close it

Sign the checksums file at release time and verify that signature against a pinned identity before any hash is trusted:

- `.goreleaser.yaml` gains a `signs:` block. Cosign keyless signing via the release workflow's OIDC token avoids a long-lived key; minisign with a key in the release environment is the simpler alternative.
- `pkg/selfupdate` verifies the signature before `VerifyChecksum` runs, against an identity compiled into the binary rather than one named by the release. A release whose signature does not verify is refused, not downgraded to a checksum-only check.
- The verification order matters as much as the check: signature, then checksum, then extraction. The current code already puts checksum before extraction and has a test pinning that; the signature step needs the same treatment.

## Out of scope here

Signing the binaries themselves (Apple notarization, Authenticode). Those solve a different problem — the OS refusing to run an unsigned binary — and neither covers the deb/rpm paths.

## Until then

`README.md` says the update is checksum-verified but unsigned, explains what that does and does not prove, and gives the out-of-band check: compare `alpacon version` against the Releases page, or verify the archive by hand before installing it.

Refs #408

PR #419, opened 2026-09-03
feat(install): install and update alpacon on Windows with one command

### LINKED ISSUE #377: [FEATURE] Lower the barrier to installing and updating the CLI on Windows
## Summary

Provide a one-command install and update path for Windows—winget and/or scoop distribution, or at minimum a PowerShell install script.

## Problem Statement

macOS installs with a single `brew install` and Linux with apt/yum repositories, and both update through the same package manager. Windows is the only platform where installation is fully manual: download a zip from the Releases page, extract it, and edit the PATH environment variable by hand. Updating repeats the same manual process every time, and there is no way to know a new version exists. Users must also pick the right architecture (amd64/arm64/arm) themselves. This is a real barrier for the CLI's audience—engineers on managed Windows machines, CI pipelines, and AI agents that bootstrap their own tooling.

## Proposed Solution

- Publish to winget: GoReleaser supports generating winget manifests, so this attaches to the existing release pipeline. `winget install alpacax.alpacon` then covers both install and update.
- Add a scoop bucket: GoReleaser also supports scoop with a self-hosted bucket at low maintenance cost.
- Minimum viable option: ship a PowerShell install script (auto-detects architecture, downloads the latest release, registers PATH) and document it in the README and docs.alpacax.com.

Done when a Windows user can install with one command, update with the same tool, and the README and official docs reflect the new path.

## Priority

medium

jisung-02:
Closes #377. winget and scoop are not planned: the script covers the platform on its own, and each manifest would be another publishing surface to keep current on every release.

## What

`install.ps1` ships as a release asset and is run with `irm https://github.com/alpacax/alpacon-cli/releases/latest/download/install.ps1 | iex`. It installs into `%LOCALAPPDATA%\Alpacon\bin`, appends that directory to the user `PATH` and updates `$env:Path` in the calling shell, so no administrator rights are needed and the shell that ran it can use `alpacon` right away. Running it again updates in place. `-Version` pins a release and `-Force` reinstalls the current one, both through a script block since `irm | iex` cannot pass arguments. A `-Version` that arrives empty, which is what an unset CI variable expands to, is refused rather than read as unpinned.

The archive is verified against the release's SHA-256 checksums before anything is moved into place, and `install.ps1` itself is hashed into that same checksums file, so it can be verified before it is run. README says what that proves and what it does not. README also states what the machine must already have, which it did not before: PowerShell 5.0, where `Expand-Archive` arrives and which Windows 10 and 11 ship built in, and a session in full language mode. It says the execution policy needs no relaxing, since `iex` runs a string rather than a script file—the same property that makes a `#Requires -Version` useless here, as that governs a file. The by-hand suite commands in `CLAUDE.md` do run a file, so they now pass `-ExecutionPolicy Bypass`, which a Windows client's default policy requires.

`.github/workflows/install-script.yaml` runs Pester on both PowerShell editions, PSScriptAnalyzer on Windows PowerShell, and an end-to-end job that installs on a real runner, re-runs, reinstalls after the binary is deleted, pins a version, refuses a `-Version` that arrives empty, confirms a locked binary is refused, and confirms a run that installs nothing still puts the directory on the calling shell's `PATH`. `release.yaml` gates GoReleaser on that workflow. Its `actions/checkout` names the same commit SHA the repository's other four workflows already do, rather than the floating tag it was first written with.

`alpacon update` now keeps the installer's `installed-version.txt` in step, both after it replaces the binary and on a run that finds nothing to install.

## Why

Windows was the only platform without a one-command path: download a zip, unpack it, edit `PATH`, and repeat all of it to update.

The script decides what to do from `installed-version.txt`, because reading the version out of `alpacon version` would reach the network on every install. `alpacon update` replaced the binary and left that file alone, so using both left them disagreeing: `-Version <the marker's version>` matched the stale marker and skipped, leaving a different binary in place than the caller pinned, and a plain re-run reported an upgrade it had not performed. Rewriting the marker on the already-current path as well is what repairs a marker stranded by an update killed before it got there.

`-Version` was tested for truth rather than for having been passed, so `-Version "$env:PINNED_VERSION"` with the variable unset installed whatever release was newest and said nothing about it. Pinning a version is done to rule that out. The check now asks whether the parameter was passed at all, which works because the script splats its own bound parameters into `Invoke-AlpaconInstall`: an omitted `-Version` never reaches the function, while an empty one does and is refused by `ConvertFrom-VersionArgument`, whose `[AllowEmptyString()]` was there for exactly that.

The latest version comes from the `releases/latest` redirect rather than the API, which is rate limited to 60 calls an hour per IP and so runs out on a shared office network or a CI runner. The redirect is followed with a HEAD request and the tag read out of the URL it lands on, because Windows PowerShell 5.1 turns `-MaximumRedirection 0` into an exception that carries no response for the tag to be read from.

Two constraints shape the script and are enforced by tests. `irm | iex` runs it in the caller's own shell, so `Set-StrictMode`, `$ErrorActionPreference` and `$ProgressPreference` all live inside `Invoke-AlpaconInstall` and a test walks the syntax tree to keep them there. The file stays pure ASCII, because Windows PowerShell 5.1 reads a BOM-less file as ANSI while a BOM would reach `iex` as a stray U+FEFF.

## Testing

`go test -race -shuffle=on ./...` passes, 42 packages ok and no failures, and `golangci-lint run ./...` reports 0 issues. Each new or changed Go test was seen failing first: the marker-name test with `installedVersionMarker` set to another value, which printed the two-line diff it now exists to print rather than the whole of `install.ps1`.

`pkg/selfupdate/marker_test.go` stays serial and says why. `updateFixture` reaches `releaseServer`, which reassigns the package-level `allowedAssetOrigins`, so `t.Parallel()` on a test that calls `Run` races the parallel test that reads it: adding it lost under `-race` three runs out of three. The one test in the file that reaches neither does run in parallel.

On this head Pester reports 113 passed, 0 failed and 1 skipped on both Windows PowerShell 5.1 and PowerShell 7, PSScriptAnalyzer reports nothing, and every step of the end-to-end job passes on a real Windows runner, the empty `-Version` refusal among them. `install.ps1` was also checked here for non-ASCII bytes and a BOM, and `actionlint` reports nothing on `install-script.yaml`.

The suite and the install were also run on a real Windows 11 ARM64 machine, which is coverage CI cannot reach: `windows-latest` is x64 only, so the `ARM64` branch of `Get-AlpaconArch` and the `windows-arm64` archive had never been exercised. Pester reports the same 113 passed, 0 failed and 1 skipped there under Windows PowerShell 5.1.26100.7920, with the files in LF and again in CRLF, and PSScriptAnalyzer 1.24.0 reports nothing. `install.ps1` was confirmed on that machine to hold no byte above 127 and no BOM. That machine has no PowerShell 7, so the pwsh leg stays CI-only.

Piping the script into `iex` there installed 1.12.0: the binary's PE header reads machine ARM64, `alpacon version` answers 1.12.0, and the user `PATH` gained the directory while keeping its `REG_EXPAND_SZ` kind. Re-running skipped, `-Version 1.11.2` reported a replacement rather than an update, pinning it again skipped, `-Force` reinstalled, and `-Version ""` was refused. Deleting the binary with the marker left behind reinstalled rather than skipping, and an install attempted while a handle was held on `alpacon.exe` was refused with the marker left where it was.

The claim that no administrator rights are needed was checked rather than assumed, since an OpenSSH session for an administrator is elevated and so is a GitHub runner. With the install wiped and its `PATH` entry removed, a process launched through `runas /trustlevel:0x20000`—not in the Administrators role, and refused a write to `HKLM\SOFTWARE`—placed the binary, wrote the marker and added the directory to the user `PATH` with its `REG_EXPAND_SZ` kind intact. An install through `iex` also succeeded in a process whose execution policy was `Restricted`, and one in constrained language mode was refused by name before it reached its first .NET call.

`syncVersionMarker` was checked against the same install with the branch cross-compiled for `windows/arm64`, since no published release carries it yet. An update from 1.11.0 moved the marker to 1.12.0 with the binary; a marker reading 9.9.9 beside an already-current 1.12.0 binary was repaired to 1.12.0 on the run that found nothing to install; and an update run with no marker present created none.

Narrowing the marker-name assertion cost one round trip worth recording: a regexp ending in `.*$` keeps the carriage return that git adds when it checks `install.ps1` out on Windows, which `assert.Contains` had never been sensitive to, and the Windows leg failed on it. `[^\r\n]*` takes the line without either terminator. It was reproduced locally by converting the file to CRLF before the fix went in.


install.ps1
```diff
@@ -0,0 +1,531 @@
...
+$script:ReleaseBaseUrl = 'https://github.com/alpacax/alpacon-cli/releases/download'
+$script:LatestReleaseUrl = 'https://github.com/alpacax/alpacon-cli/releases/latest'
+function Get-DefaultInstallDir {
+    param([Parameter(Mandatory)][AllowEmptyString()][string]$LocalAppData)
+
+    if ($LocalAppData -cnotmatch "$script:DriveRootedPattern") { return $null }
+
+    # Joined as a string, not with Join-Path: a drive-qualified path is a
+    # PowerShell drive reference off Windows, and Join-Path fails on it.
+    return $LocalAppData.TrimEnd('\') + '\Alpacon\bin'
```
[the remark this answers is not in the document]
jisung-02: Fixed in b40f73a. `TrimEnd` now takes both separators, and the existing test became a two-case one covering a value that ends in a backslash and one that ends in a forward slash.

install.ps1
```diff
@@ -0,0 +1,531 @@
...
+    try {
+        $arch = Get-AlpaconArch
+        $target = if ($Version) { ConvertFrom-VersionArgument -Version $Version } else { Get-LatestAlpaconVersion }
+
+        $installed = Get-InstalledAlpaconVersion -InstallDir $InstallDir
+        $action = Resolve-InstallAction -InstalledVersion $installed -TargetVersion $target -Force:$Force
+        if ($action -eq 'skip') {
+            Write-Host "alpacon is already at $target. Pass -Force to install it again."
+            return
+        }
```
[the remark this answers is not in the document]
jisung-02: Fixed in 6fb2bce. Both endings now go through `Add-ToSessionPath`, so the skip path puts the directory on `$env:Path` before it returns. The end-to-end job strips the directory out of the step PATH before the second run, so the skip path has to put it back for the job to pass.

.github/scripts/run-installer-tests.ps1
```diff
@@ -0,0 +1,18 @@
...
+# step's `shell:`, so the edition has to be the host the step launches.
+Set-StrictMode -Version 3.0
+$ErrorActionPreference = 'Stop'
+
+Install-Module Pester -RequiredVersion 6.1.0 -Force -SkipPublisherCheck -Scope CurrentUser
+# The runner image ships its own Pester, and auto-loading takes the highest
+# version it finds, not the one installed above.
+Import-Module Pester -RequiredVersion 6.1.0 -Force
+
+$result = Invoke-Pester ./install.Tests.ps1 -Output Detailed -PassThru
```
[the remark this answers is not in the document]
jisung-02: Fixed in 195d89e. The suite is now located with `$PSScriptRoot`, so the script runs the same way from any working directory. Worth having for the reason you name: Pester raises nothing when discovery finds no file, so the only thing standing between that and a green check was the `TotalCount -eq 0` guard.

eunyoung14 (approved):
An `irm | iex` installer that writes the user PATH has two ways to do real damage — install the wrong bytes, or corrupt PATH — and both are handled carefully here.

## PATH

This is the part I looked hardest at, because a broken user PATH is the failure a user cannot easily undo and will not attribute to us.

```powershell
$current = $key.GetValue('Path', $null, [RegistryValueOptions]::DoNotExpandEnvironmentNames)
$kind    = ... $key.GetValueKind('Path')
if ($kind -ne String -and $kind -ne ExpandString) { throw }
$key.SetValue('Path', $updated, $kind)
```

Three separate hazards, each closed:

- **`SetEnvironmentVariable` rewrites PATH as `REG_SZ`**, freezing entries like `%USERPROFILE%\tools` into expanded strings that break when the profile moves. Going through the registry directly and preserving the original kind avoids it. This is a well-known Windows footgun that most installers walk straight into.
- **Reading with `DoNotExpandEnvironmentNames`** is the other half of the same problem: read it expanded and you write back the expanded form even if you preserve the kind.
- **Refusing a `REG_MULTI_SZ`/`REG_BINARY` PATH** rather than casting it to string and writing it back. The comment says exactly what that would do. Declining to act on a shape you did not expect is the right instinct in a path like this.

The `CreateSubKey` null check is a small thing done properly too — it answers null rather than throwing, so without the guard the `finally` would fail on null and report a method call instead of the cause.

`Add-PathEntry` handling quoted, trailing-separator and equivalent-but-differently-spelled entries matches how PATH is actually written in the wild, which is by hand as often as by installers.

## Verification

- The archive is checked against the release checksums **before** anything is moved into place.
- `$parts[1] -ceq $FileName` is case-sensitive. On a case-insensitive platform that is easy to get wrong, and a case-insensitive match is how you end up validating against the wrong row.
- **No checksum for the file is a `throw`**, not a skip. That is the direction that matters: an installer that verifies when it can and proceeds when it cannot has no verification at all.
- Hashing `install.ps1` into the same checksums file, and stating in the README what that proves and what it does not, is the honest framing. `irm | iex` does not verify the script it runs; the hash lets someone who wants to verify do so, and the README does not pretend the default path is verified.

## Details worth keeping

`-Version` arriving empty being refused rather than read as unpinned is the CI case (`-Version $env:UNSET`), and it fails toward "tell me" instead of "install latest".

Refusing constrained language mode with an instruction for what to do instead, rather than failing on the first .NET call, turns an obscure error into an actionable one. Same for the unsupported-architecture throw.

Stating the PowerShell 5.0 floor and why (`Expand-Archive`, shipped in Windows 10/11) is the kind of prerequisite that is otherwise discovered by a user hitting it.

## Scope

Declining winget and scoop with a reason — each manifest is another publishing surface to keep current every release — is the right call at this size, and worth having written down so it is a decision rather than an omission.

724 lines of `install.Tests.ps1` against 537 of installer, plus a CI workflow that runs them, is the right ratio for a script that runs on machines we cannot see.

junho226 (commented):
## Summary

Adds the Windows one-line installer (`install.ps1`), a 724-line Pester suite, a PSScriptAnalyzer configuration, a dedicated CI workflow that gates GoReleaser, and a Go-side `syncVersionMarker` that keeps `installed-version.txt` in step with `alpacon update`.

The script is careful in the places that decide whether an `irm | iex` surface is safe. The checksum is fetched and compared before anything is unpacked or moved into place, `-Version` is validated against a listed character set anchored with `\z` and matched with `-cmatch` so a Kelvin sign cannot case-fold into a URL, `Set-StrictMode` and both preferences live inside `Invoke-AlpaconInstall` with an AST test walking the tree to keep them there, and the archive and checksum names are pinned to `.goreleaser.yaml` from the PowerShell side and the marker name from the Go side.

I checked the release-side assumptions against the live releases rather than taking them from the comments:

- `alpacon-1.12.0-windows-amd64.zip` is flat, with `alpacon.exe` at its root beside `LICENSE` and `README.md`, so the `Move-Item` out of `$work` finds what it expects.
- `alpacon-1.12.0-checksums.sha256` is the two-column GNU form with two spaces between hash and name, which is exactly what `Get-ExpectedChecksum`'s `-split '\s+', 2` reads.
- Windows `amd64`, `arm` and `arm64` archives are all published and no `windows/386` build exists, so every branch of `Get-AlpaconArch` maps to a real asset and the refusal for 32-bit x86 is the truth rather than a guess.
- `release.Version` (v-stripped in `release.go`) and `utils.Version` (GoReleaser's `{{.Version}}`, also v-stripped) are the same shape the script writes, so the marker the Go side rewrites still compares equal under `Resolve-InstallAction`'s `-ceq`. This is the one thing that would have quietly turned every post-update run into a needless re-download, and it holds.

Three axes, as usual. **Security**: nothing found. The one property the design cannot have—a checksum from the same host as the archive proves transport integrity and not provenance—is stated in the script, in the README and in CLAUDE.md, and #412 is where it gets closed. **Correctness**: one gap, below. **YAGNI**: nothing to say. The suite-to-script ratio is high, but this is a script users pipe into their shell unread, and the AST tests and the cross-file pins are what stop it drifting away from the release it has to match.

All three Copilot findings are resolved in the branch, and I confirmed each fix in the code rather than from the replies: `Get-DefaultInstallDir` now trims both separators, the skip path reaches `Add-ToSessionPath`, and the runner script locates the suite from `$PSScriptRoot`.

## Critical Issues

None.

## Warnings

- **A run that installs nothing never repairs the permanent user PATH** (`install.ps1:423-429`). `Add-ToUserPath` is called at one place only, line 490, on the install/upgrade branch. The skip branch fixes `$env:Path` and returns, so a user whose `HKCU\Environment` entry went missing gets `alpacon is already at <version>` and no repair, and every terminal opened afterwards still cannot find the binary. The argument already made here for the session PATH applies to the user PATH as well, and `Add-ToUserPath` is idempotent.

## Suggestions

- `.github/scripts/run-installer-tests.ps1` is in neither `paths` filter and is not in the `analyze` job's file list, so the one file that pins Pester and asserts the skip count is the only piece of this feature CI neither triggers on nor lints.
- The end-to-end job hard-pins release `1.10.0`, and `release.yaml` now makes this workflow a `needs:` of `goreleaser`, so a pruned old release would block future tags rather than fail anything the tag changed.
- About half the functions the script leaves in the caller's session carry no `Alpacon` in the name. The leak is unavoidable; its blast radius is not.

## Highlights

- The `$PSBoundParameters.ContainsKey('Version')` test, carried through the splat at line 536, is the right fix for the empty-pin bug: an omitted `-Version` never reaches the function, an empty one does and is refused. The end-to-end job exercises the exact shape a pipeline hits.
- `Get-LatestAlpaconVersion` reading the `releases/latest` redirect instead of the API sidesteps a rate limit that a shared office network or a CI runner really does exhaust, and `Get-ResponseUri` handling both the 5.1 `HttpWebResponse` and the 7 `HttpResponseMessage` shapes is the sort of thing that is only found by running both.
- `Add-ToUserPath` going through the registry rather than `[Environment]::SetEnvironmentVariable`, and refusing a `Path` stored as a kind a string write would destroy, avoids the classic installer bug of freezing `%USERPROFILE%` into an expanded string. The end-to-end job records the value kind before installing rather than asserting one afterwards, which is the only way that check means anything on a runner.
- `syncVersionMarker` rewriting an existing marker and creating none is the correct asymmetry: the Go side has no way to know whether a directory is one the script owns, and the file's presence is the only evidence available.
- `Test-AlpaconFileLocked` asking for read access on a read-only file and letting the share mode answer, rather than conflating "read-only" with "held", is a distinction most installers get wrong.

## Verdict

COMMENT. Nothing blocking: the security-relevant paths hold up, the release-side assumptions check out against the published assets, and CI is green on both PowerShell editions. The one correctness gap and the three suggestions are all non-blocking and can land here or as a follow-up.

## Checklist

- [x] Security: download verified before use, version input cannot steer a URL or a file name, no credential or token handling, no elevation
- [x] Correctness: archive layout, checksum format, asset names and version string shapes verified against the live releases
- [x] Cross-file pins: `.goreleaser.yaml` templates and the marker name are asserted from both languages
- [x] Test coverage (static): 113 Pester cases, four new Go tests, AST guards for the two script-wide invariants
- [x] Go conventions: declaration order, `_ =` on the ignored write, serial tests with a stated reason, `assert.Equal` over a wrapped predicate
- [x] Workflow pins: every third-party `uses:` carries a 40-character SHA with a version comment
- [ ] CI trigger coverage: `.github/scripts/run-installer-tests.ps1` is outside both `paths` filters and the analyzer's file list

install.ps1
```diff
@@ -0,0 +1,537 @@
...
+        # a pipeline that meant to pin a version must hear about it rather than
+        # get the latest release.
+        $target = if ($PSBoundParameters.ContainsKey('Version')) { ConvertFrom-VersionArgument -Version $Version } else { Get-LatestAlpaconVersion }
+
+        $installed = Get-InstalledAlpaconVersion -InstallDir $InstallDir
+        $action = Resolve-InstallAction -InstalledVersion $installed -TargetVersion $target -Force:$Force
+        if ($action -eq 'skip') {
+            # This shell can predate the install that put alpacon here, so it
+            # still needs the directory on a run that installs nothing.
+            Add-ToSessionPath -Directory $InstallDir
```
junho226, line 426: Warning: this branch fixes the session PATH but never the permanent one. `Add-ToUserPath` is reached at line 490 only, on the install/upgrade branch, so a user whose `HKCU\Environment` entry went missing—another installer rewriting PATH, the 2047-character legacy dialog truncating it, a manual clean-up—gets `alpacon is already at <version>` and no repair, and every terminal opened afterwards still cannot find the binary. README says the script puts the directory on the user PATH, and re-running the one-liner is the first thing such a user tries.

The reasoning written two lines above for `$env:Path`—this shell can predate the install—holds for the user PATH just as well: the entry can predate a PATH edit that dropped it. `Add-ToUserPath` is idempotent and returns `$false` when the entry is already there, so calling it here costs one registry read on the common path. It would want the same try/catch the install branch uses at lines 489-510, since a locked-down HKCU has to stay a warning rather than turn a no-op run into a failure.

.github/workflows/install-script.yaml
```diff
@@ -0,0 +1,221 @@
...
+      # The suite cross-checks these two against the documented download URL.
+      - "README.md"
+      - ".goreleaser.yaml"
+  pull_request:
+    types: [opened, reopened, ready_for_review, synchronize]
+    paths:
+      - "install.ps1"
+      - "install.Tests.ps1"
+      - ".github/workflows/install-script.yaml"
+      - "PSScriptAnalyzerSettings.psd1"
```
junho226, line 20: Suggestion: `.github/scripts/run-installer-tests.ps1` is in neither `paths` filter, and the `analyze` job's file list below names only the three files at the repository root. That leaves the one file that pins Pester 6.1.0, locates the suite and asserts `SkippedCount -eq 1` as the only piece of this feature CI neither triggers on nor lints—a PR that bumps that pin, or relaxes the skip assertion because a new test was skipped, merges without the suite having run once.

Adding the path to both filters and the file to the analyzer list closes it, and the file is exactly the kind that earns the analyzer: it is the harness the rest of the checks run inside.

.github/workflows/install-script.yaml
```diff
@@ -0,0 +1,221 @@
...
+          $key.Close()
+          $hits = @($raw -split ';' | Where-Object { $_ -eq $dir }).Count
+          if ($hits -ne 1) { throw "install directory appears $hits times on PATH, expected 1" }
+
+      # 1.10.0 is simply an older release that still ships Windows archives.
+      # Any published version older than the latest one does the job.
+      - name: Pinning an older version installs that version
+        shell: powershell
+        run: |
+          ./install.ps1 -Version 1.10.0
```
junho226, line 171: Suggestion: this hard-pins release 1.10.0, and `release.yaml` now makes this workflow a `needs:` of `goreleaser`. So if 1.10.0 is ever deleted, or its Windows archives pruned in a release clean-up, every future tag blocks here rather than on anything that tag changed—and the failure names an unrelated version, which is the slow kind to diagnose.

The comment above already says any published version older than the latest one does the job. Reading the second-newest tag out of `gh release list` would put that in code instead of in prose, and would keep the step working as the pin ages past what the fleet still has.

install.ps1
```diff
@@ -0,0 +1,537 @@
...
+        # Folds '..' segments. Only for absolute paths: GetFullPath would resolve
+        # anything else against this process's working directory, which has
+        # nothing to do with where that entry points for anyone else.
+        try { return [System.IO.Path]::GetFullPath($trimmed).TrimEnd('\') } catch { return $trimmed }
+    }
+
+    return $trimmed
+}
+
+function Add-PathEntry {
```
junho226, line 284: Suggestion: `irm | iex` leaves every function in the caller's session, as CLAUDE.md says, and about half of these carry no `Alpacon` in the name—`Add-PathEntry`, `Get-ResponseUri`, `Assert-Checksum`, `Get-ExpectedChecksum`, `ConvertTo-ComparablePath`, `Resolve-InstallAction`, `Send-SettingChange`. Functions come before cmdlets in PowerShell's resolution order, so each of these silently shadows a same-named helper the user had, for the rest of that session and with nothing to say it happened.

The leak itself is unavoidable and the PR is right that it is; its blast radius is not. Prefixing them the way `Get-AlpaconArch` and `Test-AlpaconFileLocked` already are would bound it, at the cost of a rename across the Pester suite.

PR #420, opened 2026-09-03
ci: pin workflow actions to commit SHAs and attest release artifacts

jisung-02:
## What

CI configuration and docs only. No Go code changes.

1. Every `uses:` in `.github/workflows/` is pinned to a 40-character commit SHA, with the version kept in a trailing comment. `.github/dependabot.yml` checks the GitHub Actions and Go module ecosystems weekly, each grouped into one PR with a Conventional Commit prefix, and raises the SHA and its comment together. One action, `danielmundi/upload-packagecloud`, has had a single tag since 2021-08 and nothing to raise the pin to; the docs now say so, and the follow-up is replacing it with a direct PackageCloud API call.
2. The `goreleaser` job in `release.yaml` runs `actions/attest-build-provenance` v4.2.2 over every tar.gz, zip, deb and rpm in `dist/`, and a second run of the same action attests the Docker image GoReleaser pushes, keyed to the manifest digest read back from the registry since the image has no file in `dist/` to name. The job holds `id-token: write` and `attestations: write` for both. The workflow's top-level `permissions: {}` is unchanged.

README's "Updating" section says what the checksum proves and what the attestation adds, and shows how to verify an archive and the Docker image by hand with `gh attestation verify`. CLAUDE.md and `.github/copilot-instructions.md` record the pin rule, how to raise a pin, the one pin that cannot be raised, and why the two permissions must stay.

## Why

Both changes are about the supply chain of the release itself: what the build pulls in, and what ends up on the release page. They close two of the three ways in, and narrow the third.

A tag is a ref its own repository can move, `v6.1.0` no less than `v6`. A moved tag runs new code inside the release build with nothing in our diff to show it. A commit SHA cannot come to mean different content. The cost is that a pinned SHA never moves on its own either, so an upstream fix reaches this repo only as a raised pin; Dependabot is what raises it.

The checksums file says an archive is the one the release names, and nothing about who built it: the archive and the checksums come from the same release, so one stolen release token rewrites both and `alpacon update` installs the result. A build provenance attestation says which workflow, at which commit, built the artifact. It is signed with a certificate minted from the job's OIDC identity that lives for minutes, so there is no key to store or leak, and the signature is recorded in a public transparency log. A release published any other way cannot be made to verify.

Provenance answers only who built the artifact, never what went into it. A dependency or an action compromised upstream is built by the real workflow and gets a valid attestation. Pinning is what narrows that.

## Out of scope

`alpacon update` still compares checksums and nothing else. Teaching it to verify provenance is #412.

Replacing `danielmundi/upload-packagecloud` with a direct call to the PackageCloud upload API, since its pin can never be raised, is a separate change.

## Testing

- Each of the ten pinned SHAs was resolved through the GitHub API and compared against its tag. `golangci/golangci-lint-action` v9.3.0 is an annotated tag and was dereferenced to the commit it wraps.
- `actionlint` passes on all three workflows.
- The four `subject-path` globs were checked against the file names `.goreleaser.yaml` writes to `dist/`, the same paths the `upload-artifact` steps already name.
- `docker buildx imagetools inspect --format '{{json .Manifest}}' alpacax/alpacon-cli:latest | jq -r .digest` was run against the image already on Docker Hub to confirm the format string returns a `sha256:...` digest and not buildx's default text output, which a bare `{{.Manifest.Digest}}` field template silently falls back to.
- `.github/dependabot.yml` parses under a YAML loader.
- No attestation exists yet: the first one is created by the next tag. After it is published, `gh attestation verify alpacon-<version>-<os>-<arch>.tar.gz --repo alpacax/alpacon-cli --signer-workflow alpacax/alpacon-cli/.github/workflows/release.yaml` verifies the archive, and `gh attestation verify oci://index.docker.io/alpacax/alpacon-cli:<version> --repo alpacax/alpacon-cli --signer-workflow alpacax/alpacon-cli/.github/workflows/release.yaml` verifies the image.


README.md
```diff
@@ -68,9 +68,11 @@ go build && sudo mv alpacon-cli /usr/local/bin/alpacon
...
+`alpacon update` compares SHA-256 checksums and nothing else. That comparison proves the archive arrived intact and matches the checksums file the release publishes; it proves nothing about who published either, because both come from the same release. Anyone who could write to a release—a leaked token, a compromised account, a poisoned release workflow—could upload a trojaned archive together with a checksums file that matches it, and the update would install it. Teaching the CLI to verify the publisher is tracked in [#412](https://github.com/alpacax/alpacon-cli/issues/412).
+
+Every release newer than v1.11.2 does carry a [build provenance attestation](https://docs.github.com/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds), which the update does not read but you can. The release workflow signs each archive and package with a short-lived certificate minted from its own GitHub identity, and the signature is recorded in a public transparency log—so a release published any other way cannot be made to verify, and one published this way leaves a record nobody can quietly remove.
```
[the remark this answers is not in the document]
jisung-02: Fixed—the version boundary is gone. The sentence now reads "Every release published since provenance was added to the release workflow", so there is no number to keep current.

I did not take the suggested phrasing as written, though: "releases built by this repo's release workflow" would be false for v1.11.2 and earlier. Those were built by the same workflow and carry no attestation, so that wording would promise a record that is not there. The paragraph now also says what a reader hits on one of those—`gh attestation verify` answers `404`, which is an absent record rather than a bad signature.

README.md
```diff
@@ -79,8 +81,15 @@ To check out of band, compare `alpacon version` against the [Releases](https://g
 curl -LO https://github.com/alpacax/alpacon-cli/releases/download/v<version>/alpacon-<version>-<os>-<arch>.tar.gz
 curl -LO https://github.com/alpacax/alpacon-cli/releases/download/v<version>/alpacon-<version>-checksums.sha256
 sha256sum --check --ignore-missing alpacon-<version>-checksums.sha256
+
+# Who built it, from which commit. Needs the GitHub CLI.
+gh attestation verify alpacon-<version>-<os>-<arch>.tar.gz \
+  --repo alpacax/alpacon-cli \
+  --signer-workflow alpacax/alpacon-cli/.github/workflows/release.yaml
```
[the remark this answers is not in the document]
jisung-02: The value is in the format the flag documents. `gh attestation verify --help` (gh 2.99.0) describes `--signer-workflow` as `[host/]<owner>/<repo>/<path>/<to>/<workflow>`, so `alpacax/alpacon-cli/.github/workflows/release.yaml` is the full form, not a prefixed one. A bare `.github/workflows/release.yaml` would not match: the signer identity in the certificate is `https://github.com/alpacax/alpacon-cli/.github/workflows/release.yaml@refs/tags/vX.Y.Z`, and the flag anchors on the owner and repo.

- The README value matches the format printed by `gh attestation verify --help`.
- Dropping the owner and repo would let a release.yaml in any other repository satisfy the filter, which is the opposite of what the check is for.

.github/workflows/release.yaml
```diff
@@ -91,39 +95,51 @@ jobs:
...
+          subject-path: |
+            dist/*.tar.gz
+            dist/*.zip
+            dist/*.deb
+            dist/*.rpm
+
       - name: Strip "v" prefix from tag
         run: echo "TAG_NAME=${GITHUB_REF_NAME#v}" >> $GITHUB_ENV
         env:
           GITHUB_REF_NAME: ${{ github.ref_name }}
```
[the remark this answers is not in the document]
jisung-02: Fixed. Both lines, 121 and 156, lose the trailing space in the follow-up commit; `actionlint` still passes.

.github/workflows/release.yaml
```diff
@@ -91,39 +95,51 @@ jobs:
...
           version: "~> v1"
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ steps.release-publisher-token.outputs.token }}
 
+      # The checksums file says an archive is the one this release names; the
+      # attestation says which workflow, at which commit, built it. Stealing a
+      # release token is enough to rewrite the first and not the second.
+      - name: Attest release artifacts
+        uses: actions/attest-build-provenance@4d101475d8b20a2381f78447822ac1eab6504dd8 # v4.2.2
```
junho226, line 111: Suggestion: the Docker image is the one released artifact this leaves unattested.

`.goreleaser.yaml`'s `dockers:` block pushes `alpacax/alpacon-cli:latest` and `alpacax/alpacon-cli:{{ trimprefix .Tag "v" }}` to Docker Hub from this same job, and `subject-path` below covers archives and packages only:

```
dist/*.tar.gz
dist/*.zip
dist/*.deb
dist/*.rpm
```

The docs are technically precise about this — CLAUDE.md and copilot-instructions both say "every archive and package it builds", and an image is neither — so this is not a false claim. But for a PR titled "harden release supply chain", the remaining distribution channel is worth either closing or recording as out of scope.

Closing it means attesting by digest rather than by path, which needs the pushed digest: goreleaser v1 pushes the image itself, so the digest has to be read back (`docker buildx imagetools inspect`, or `docker inspect --format='{{index .RepoDigests 0}}'`) and handed to a second `attest-build-provenance` step as `subject-name` + `subject-digest`. That is not a one-liner, so recording it as a follow-up is a fine answer — the ask is that the choice be explicit, since the next reader of `subject-path` cannot tell whether the image was considered.

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  # SHA-pinned actions receive no upstream patch on their own—the pin is
+  # exact, so a fix upstream reaches this repo only as a raised pin.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
```
junho226: Suggestion: this diverges from the house pattern, and the file is new so it costs nothing to match now.

`alpacon-server/.github/dependabot.yml` carries two things this one does not:

```yaml
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    commit-message:
      prefix: "chore(actions)"
    groups:
      actions:
        patterns:
          - "*"
```

- **`commit-message.prefix`** matters because Dependabot's default subject is `Bump actions/checkout from A to B`, which is not a Conventional Commit — and the org convention applies to commit subjects in every repo.
- **`groups`** matters because this repo now pins 9 distinct actions. Weekly with no grouping is up to 9 PRs a week instead of 1, all of them mechanical SHA-plus-comment rewrites.

Separately, `gomod` is absent. That half is weaker: `govulncheck` already runs in `lint.yaml` on every push, so detection is covered and only automated bumping is missing, and `go.mod` is small. Your call whether to include it — but the README added in this PR names an upstream-compromised dependency as the residual that provenance cannot answer, which is an argument for having Dependabot watch that ecosystem too.

.github/workflows/release.yaml
```diff
@@ -137,30 +153,30 @@ jobs:
...
 
       - name: Download ARM64 RPM Artifacts
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
         with:
           name: alpacon_${{ env.TAG_NAME }}_linux_arm64.rpm
 
       - name: Upload AMD64 DEB to PackageCloud
-        uses: danielmundi/upload-packagecloud@v1
+        uses: danielmundi/upload-packagecloud@46cd0e61152bf952dbc0d1759e609d3d22649030 # v1
```
junho226, line 199: Suggestion: worth a note that this pin is now frozen with no upstream signal to raise it from.

Facts on the action, for the record:

- last upstream push **2021-08-11**, 11 stars, not archived
- exactly one tag ever: `v1` (hence `# v1` rather than a precise version)
- it is the `uses:` in this workflow that receives `PACKAGECLOUD_TOKEN`

Pinning it is the single biggest win in this PR — a four-year-dormant third-party action holding a publish credential is exactly what a moved tag would exploit, and a SHA closes that.

What does not follow is the maintenance path copilot-instructions documents for it. "Raising a pin means changing both the SHA and its comment together. Resolve the SHA rather than guessing it: `gh api repos/OWNER/REPO/commits/vX.Y.Z --jq .sha`" needs a newer semver tag to resolve, and there will not be one. Dependabot raises a pin from a newer version, so it has nothing to offer here either. The `# v1` comment also cannot distinguish revisions of a moving major.

So the follow-up for this one is not a version bump but replacing it — the PackageCloud push is a `curl` to their API, and vendoring it inline drops a credential-holding third-party action from the release path entirely. Fine as a separate change; the ask here is only that the copilot-instructions rule not read as if it covered this pin.

junho226 (dismissed):
## Review: #420

**Stack**: common (repo is Go; the diff touches CI and docs only) · **CI**: green

### Summary

Pins every third-party `uses:` to a 40-character SHA with the version in a trailing comment, adds `actions/attest-build-provenance` to the `goreleaser` job, creates `.github/dependabot.yml` for the actions ecosystem, and documents both in CLAUDE.md, copilot-instructions, and the README.

### Verified

Everything checkable, checked.

| Check | Result |
|---|---|
| All 10 SHA → tag pairs | **10/10 match** (`gh api repos/OWNER/REPO/commits/vX.Y.Z`) |
| Unpinned `uses:` at head | **0** — only `uses: ./.github/workflows/build-and-test.yaml`, exactly as documented |
| Each pin vs. the major it replaced | tip of that major in all four cases (checkout v6.1.0, setup-go v6.5.0, upload-artifact v6.0.0, download-artifact v7.0.0), so nothing changes behavior and the four available major bumps become reviewable Dependabot PRs |
| `permissions: {}` top-level claim | true (`release.yaml:8`); the two new permissions sit on the one job that uses them |
| `subject-path` globs vs. goreleaser output | all four match — `.tar.gz`, `.zip` (windows `format_overrides`), `.deb`/`.rpm` (`nfpms.formats`), ~18 subjects, so no empty-glob failure |
| `gh attestation verify` invocation in the README | correct flag form (`--repo`, `--signer-workflow <owner>/<repo>/<path>`) |

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (3)

1. **[.github/workflows/release.yaml:110]** The Docker image is the one released artifact with no attestation. Inline.
2. **[.github/dependabot.yml:8]** Diverges from the house pattern in `alpacon-server/.github/dependabot.yml`. Inline.
3. **[.github/workflows/release.yaml:179]** `danielmundi/upload-packagecloud` has no upstream signal to raise the pin from. Inline.

### Highlights

- Comments say `# v6.1.0`, not `# v6` — an exact release, so a raised pin is a reviewable diff rather than an unexplained hash swap.
- The docs state what provenance does **not** prove, in both directions: `alpacon update` does not read it, and an action compromised upstream still gets a valid attestation. The pinning-provenance complementarity is written on both sides rather than on one.
- `id-token: write` / `attestations: write` on the one job that needs them, with the short-lived-certificate reason inline, under a `permissions: {}` default.
- The stale-release case is named: an older release carries no record rather than a bad one, and `gh attestation verify` reports it as a `404`.
- Pinning is a faithful pin, not a silent major bump — every pin is the tip of the major the workflow already tracked.

---

**Verdict**: APPROVED

eunyoung14 (dismissed):
"Every `uses:` is pinned" is a completeness claim, so I checked all of them. It holds.

## Pin completeness

Across the four workflows, **26 of 27 `uses:` entries carry a 40-character SHA**. The one that does not is

```
uses: ./.github/workflows/build-and-test.yaml
```

a local reusable-workflow reference, resolved at the same commit and therefore not something a SHA can pin. It is out of scope rather than missed. No gap.

Keeping the version in a trailing comment is the right call too. A bare SHA tells a reader nothing about what is pinned, and having Dependabot raise the SHA and its comment together is what keeps that comment true. A comment that stops being updated becomes a lie shortly afterwards, and this closes that path.

## Permission scope

The two permissions the attestation needs sit on the **`goreleaser` job only**, not on the workflow.

```
permissions: {}            <- top level
  create-release:      contents: write
  build-and-test:      contents: read
  goreleaser:          contents/packages: write, id-token: write, attestations: write
  packagecloud-deploy: contents: read
  publish-release:     contents: write
```

`id-token: write` is the ability to mint an OIDC token, so granting it at the top level would hand that identity to every job. Scoping it to the one job that needs it is correct, and keeping the top-level `permissions: {}` matters just as much.

## The reasoning is right

The distinction between the checksums and the attestation is the most important sentence in this PR:

> the archive and the checksums come from the same release, so one stolen release token rewrites both

Because the checksums file ships from the same release as the archive, one token rewrites both. The checksums therefore say only "this is the file the release names" and nothing about who built it. Carrying that through to `alpacon update` installing the result is what makes the consequence concrete.

The same goes for the tag argument. It takes on the common assumption that `v6.1.0` is safer than `v6` and rejects it directly: both are refs the upstream repository can move.

## One thing left open (outside this PR)

`danielmundi/upload-packagecloud` is used in four places and is a third-party action under a personal account. A SHA pin fixes the code as it was at that moment, which is exactly what this PR sets out to do. Whether to trust that code is a different axis, and pinning cannot answer it.

That sits alongside the description's own framing that this closes two of the three ways in and narrows the third, so it is a note rather than a finding. Packagecloud upload handles release artifacts, so it looks worth a separate look at some point.

## Verification instructions

Good call adding `gh attestation verify` to the README. An attestation guarantees nothing if nobody verifies it, and nobody verifies it if the instructions are not near the release page.

.github/dependabot.yml
jisung-02: Fixed in 7b1ae43. `.github/dependabot.yml` now matches the alpacon-server pattern—a `commit-message.prefix` and a `groups` block per ecosystem—and adds a `gomod` entry alongside `github-actions`.

Kept `gomod` in, since it costs nothing and the README's new provenance section names an upstream-compromised dependency as exactly the residual govulncheck detects but nothing here was auto-bumping.

.github/workflows/release.yaml
jisung-02: Fixed in ba77e01. Added the exception to both `.github/copilot-instructions.md` and CLAUDE.md, and a short comment above the first `upload-packagecloud` step: this pin has had one tag, `v1`, since 2021-08, so there is no newer version for the documented "raise a pin" procedure to resolve.

Agreed the SHA pin itself is the right fix here regardless—it is what closes the moved-tag risk on the action holding `PACKAGECLOUD_TOKEN`. Replacing the action with a direct PackageCloud API call is the real follow-up, and it is out of scope for this PR.

.github/workflows/release.yaml
jisung-02: Fixed in 0baaa4d. Added a step that reads the pushed image's manifest digest back from the registry (`docker buildx imagetools inspect`, since GoReleaser v1 pushes the image itself rather than leaving a file under `dist/`) and a second `attest-build-provenance` step keyed to that digest via `subject-name`/`subject-digest`. Both tags share one manifest, so one digest covers `latest` too.

One implementation note: a bare-field `--format '{{.Manifest.Digest}}'` silently no-ops on this image and prints buildx's default text instead of erroring, so the digest is read via `{{json .Manifest}}` piped through `jq -r .digest`, with a `test -n` guard so an empty digest fails the step instead of reaching the attest action.

Updated CLAUDE.md, copilot-instructions, and the README (new `gh attestation verify oci://...` example) to say the image is attested too, and noted that a release published before this change carries no image record.

eunyoung14 (approved):
Re-reviewing after the three commits that landed since my earlier pass. The pin completeness and permission scoping I checked before still hold; this covers what is new.

## Docker image attestation (`0baaa4d`)

The gap this closes is real: `dist/` never held the pushed image, so the archives were attested and the image was not.

Reading the digest back from the registry rather than from the build is the right shape — it attests what was actually pushed, not what the build believed it produced. Two details make it hold:

- **`test -n "${digest}"`** fails the step on an empty read. Without it an attestation would be made against an empty subject, which is worse than none: it looks like coverage.
- **Ordering.** Login (`:69`) → GoReleaser push (`:97`) → digest read (`:121`) → attest (`:130`). The read comes after the push, so `imagetools inspect` cannot answer about a stale or absent image.

`subject-name: index.docker.io/alpacax/alpacon-cli` is fully qualified, which is what `gh attestation verify` needs to match against.

The comment explaining why `{{json .Manifest}}` plus `jq` rather than a bare field template is worth having — a template that silently no-ops is exactly the failure that produces an empty digest, and the guard above is what catches it.

## Dependabot (`7b1ae43`)

Grouping each ecosystem into one PR is the right call for pinned actions specifically: a raise touches the SHA and its trailing comment together, and ungrouped that is one PR per action every week, which is how a pin queue stops being read. Adding `gomod` closes the gap where the workflows were watched and the module graph was not.

## The packagecloud pin (`ba77e01`)

This answers the note I left, and with more than I had. Verified:

```
danielmundi/upload-packagecloud   tags: v1 only
                                  latest commit: 2021-08-11 (46cd0e611)
```

So the pin is not merely pinned, it is **frozen**: Dependabot has nothing to raise it to, and the usual "the pin gets raised" mitigation does not apply here at all. Recording that in the docs rather than leaving it to be rediscovered is right.

The detail I did not have is the one that matters most: it is **the only action that receives `PACKAGECLOUD_TOKEN`**. So the action with no upstream maintenance since 2021 is also the one holding a publishing credential. That makes the recorded follow-up — replacing it with a direct call to the PackageCloud upload API — the substantive one of the two, not a tidy-up.

Nothing to do in this PR. Pinning is what was available here and it was applied.

## Standing verification

Unchanged from the earlier pass, restated because the review was dismissed:

- **26 of 27 `uses:` carry a 40-character SHA.** The remaining one is `./.github/workflows/build-and-test.yaml`, a local reusable workflow resolved at the calling commit — out of scope rather than missed.
- **`id-token: write` and `attestations: write` sit on the `goreleaser` job only**, with `permissions: {}` at the top level. Granting OIDC minting workflow-wide would hand that identity to every job.
- The checksums-versus-attestation reasoning in the description is the load-bearing argument, and it is correct: both files ship from the same release, so one stolen token rewrites both, and `alpacon update` installs the result.

PR #425, opened 2026-09-04
chore(deps): bump testify to 1.12.1 and ignore smux major bumps

jisung-02:
## Summary

Dependabot's weekly `gomod` group carries two bumps: testify 1.11.1 to 1.12.1, which is routine, and `github.com/xtaci/smux` v1.5.57 to v2.0.1+incompatible, which does not compile. This PR takes the first and tells Dependabot to stop offering the second.

## The smux bump

smux tagged v2 in September 2019 without moving to a `/v2` module path. Go therefore reads the tag as `v2.0.1+incompatible` and ranks it above `v1.5.57` (released March 2026) on the version number alone, even though the tree behind it is six and a half years older. That tree has no `MaxStreamBuffer` field, which `config.GetSmuxConfig` sets in `config/config.go`, so the "upgrade" fails to build. Upstream has shipped from v1 ever since.

An `ignore` entry scoped to `version-update:semver-major` keeps the group from stalling on that downgrade every week. It is a standing exception, so `CLAUDE.md` records it beside the one action pin nothing raises, and the YAML comment names `config.GetSmuxConfig` so a future reader can confirm the reason still holds before deleting the entry.

## Changes

- `go.mod` / `go.sum`: testify 1.11.1 to 1.12.1. The new release swaps `gopkg.in/yaml.v3` for `go.yaml.in/yaml/v3` and drops the `go-spew` and `go-difflib` indirects, which is the rest of the diff.
- `.github/dependabot.yml`: ignore major bumps for `github.com/xtaci/smux`, with the reason in a comment.
- `CLAUDE.md`: document the ignore as a standing exception to the weekly Go module updates.

## Testing

`go build ./...` and `go test -race -shuffle=on ./...` both pass; every package reports `ok` or has no test files.


junho226 (approved):
Verified at `49b062f16`. The testify bump is safe, the ignore rule is scoped to what it claims, and the reason behind it is recorded in three places and holds up when checked against upstream. All four files accounted for. Approving.

### The ignore rule silences majors only, and that scoping is load-bearing

`.github/dependabot.yml:29-31` sits under the `gomod` entry as a sibling of `groups:`, and reads:

```yaml
ignore:
  - dependency-name: "github.com/xtaci/smux"
    update-types: ["version-update:semver-major"]
```

So it silences exactly one thing: a semver-major bump of `github.com/xtaci/smux`. Patch and minor bumps of smux still get raised, and no other module is touched.

That scoping matters more than it looks. GitHub's docs are explicit that `ignore` suppresses pull requests for "version updates and security updates", so a `dependency-name` entry without `update-types` would have stopped smux security PRs too. This one does not: a security fix released in the v1 line still raises a PR. The only case it suppresses is a fix existing solely in a v2 major, and that tree does not compile here anyway. Even then the Dependabot *alert* still fires, since `dependabot.yml` settings do not interact with alerts.

### The reason is recorded, and it checks out

The comment at `dependabot.yml:25-28` names `config.GetSmuxConfig` as the caller that breaks, `CLAUDE.md` records it as a standing exception, and both carry the exit condition ("drop this once a real `/v2` module exists"). I confirmed the claim rather than taking it: `github.com/xtaci/smux@v2.0.1+incompatible` declares `Config` with four fields (`KeepAliveInterval`, `KeepAliveTimeout`, `MaxFrameSize`, `MaxReceiveBuffer`) and no `MaxStreamBuffer`, which `config/config.go:259` sets. The bump genuinely does not build.

The pin is current rather than stranded: `proxy.golang.org` resolves `@latest` to v1.5.57 (2026-03-13), so ignoring majors leaves the repo on the newest release upstream actually ships. No known advisories either—the GitHub advisory database returns nothing for `github.com/xtaci/smux` in the Go ecosystem, and `govulncheck ./...` reports none. Both sources were reachable and answered, so this is a real negative, not an unavailable check.

### The testify bump does not change behavior here

v1.12.1 is only the `go.yaml.in/yaml/v3` swap. v1.12.0 does carry functional changes, but none reach this repo: it imports only `testify/assert` and `testify/require` across 127 test files, with no `mock`, no `suite`, and no `*AssertionFunc` types, and zero uses of `IsIncreasing`, `PanicsWithError`, `NotSubset`, or `YAMLEq`. The one touched API is `assert.Same` (4 call sites), where v1.12.0 changed the failure message text only, not the pass/fail decision.

The dropped indirects follow from the same release: testify vendored `go-spew` and `go-difflib` in v1.12.0, which is why they leave `go.mod`. `go.sum` is consistent—`go mod tidy` produces no diff, which is the repo's own CI gate.

### All four files accounted for

`go.mod` and `go.sum` are the bump, `.github/dependabot.yml` is the ignore, `CLAUDE.md` is the note. No Go source changes.

One optional nit on the `CLAUDE.md` line: the Go-module sentence lands inside the "Workflow actions are pinned by commit SHA" bullet, between the PackageCloud follow-up and "Pinning is also what narrows...", so it interrupts the pinning argument to talk about `gomod`. That bullet does already own `dependabot.yml`, so this is a placement preference, not a correctness problem. Fine to leave as is.

### Gates run

Local, on `49b062f16`, Go 1.25.13:

| Gate | Result |
| --- | --- |
| `go mod tidy` + `git diff --exit-code go.mod go.sum` | clean, no diff |
| `go build -v .` | pass |
| `go vet ./...` | pass |
| `golangci-lint run` v2.9.0, `GOOS=linux` | `0 issues.` |
| `golangci-lint run` v2.9.0, `GOOS=windows` | `0 issues.` |
| `go test -race -shuffle=on ./...` | pass, 42 ok / 16 no test files / 0 fail |
| `govulncheck ./...` | `No vulnerabilities found.` |

CI on this head is green across all 12 checks.

Issue #427, opened 2026-09-04
install-script CI misses run-installer-tests.ps1 and hard-pins release 1.10.0


jisung-02:
## Background

`.github/workflows/install-script.yaml` is the CI for the Windows installer script. `release.yaml` wires this workflow into the `goreleaser` job through `needs:`, so a block here blocks the release tag itself. The review raised two separate spots in this workflow.

## 1. `run-installer-tests.ps1` is missing from both the CI triggers and the lint list

`.github/scripts/run-installer-tests.ps1` pins Pester 6.1.0, locates the suite, and asserts `SkippedCount -eq 1`. It is the runner every other check executes inside.

That path appears in neither the `push` nor the `pull_request` `paths` filter (lines 7-22), and the file list the `analyze` job checks (line 71) names only `install.ps1`, `install.Tests.ps1`, and `PSScriptAnalyzerSettings.psd1`.

So a PR that touches only this file merges without the suite running once. Raising the Pester pin, or loosening the skip-count assertion because a newly skipped test appeared, can land unverified.

The fix is to add `.github/scripts/run-installer-tests.ps1` to both `paths` filters and to the `analyze` job's file list. If anything, it is the file that gains the most from the analyzer.

## 2. The end-to-end job hard-codes release `1.10.0`

Lines 166-184 pin release `1.10.0` to check a version-pinned install, and the comment above them says any release older than the latest would do.

The problem is that `release.yaml` wires this workflow into `goreleaser`'s `needs:`. If 1.10.0 is deleted, or its Windows archive is dropped during a cleanup, every tag after that blocks here. The failure message then names a version unrelated to the tag being cut, which makes the cause slow to find.

Move the condition the comment states in prose into code: read the second-newest tag from `gh release list` and use that, so an aging pin keeps working.

## Source

Two suggestion items from the review on PR #419 (posted after the merge). https://github.com/alpacax/alpacon-cli/pull/419#pullrequestreview-5111414632

PR #430, opened 2026-09-07
fix(install): repair the user PATH when the installer skips the download

### LINKED ISSUE #426: Windows installer skip branch never repairs the permanent user PATH
## 배경

`install.ps1`은 설치할 것이 없으면 `Resolve-InstallAction`이 `skip`을 돌려주고, 그 분기에서 `Add-ToSessionPath`만 부른 뒤 바로 반환합니다. 영구 PATH를 쓰는 `Add-ToUserPath`는 설치/업그레이드 분기 한 곳(`install.ps1:490`)에서만 불립니다.

## 문제

`HKCU\Environment`의 PATH 항목이 사라진 사용자는 스크립트를 다시 돌려도 복구되지 않습니다. 항목이 사라지는 경로는 여러 가지입니다. 다른 설치 프로그램이 PATH를 통째로 다시 쓰거나, 옛 환경 변수 편집 창이 2047자에서 잘라내거나, 사용자가 직접 정리하는 경우입니다.

이때 실제로 벌어지는 일은 이렇습니다.

1. `alpacon.exe`는 제자리에 있고 버전도 최신이라 `Resolve-InstallAction`이 `skip`을 돌려줍니다.
2. 스크립트는 `alpacon is already at <version>.`만 출력하고 끝납니다.
3. 현재 셸에서는 `$env:Path`가 고쳐졌으니 동작하지만, 새로 여는 터미널은 전부 여전히 바이너리를 못 찾습니다.

README는 스크립트가 설치 디렉토리를 사용자 PATH에 넣는다고 적어 두었고, 이런 상황에서 사용자가 가장 먼저 하는 일이 원라이너를 다시 돌리는 것입니다. 그런데 그 재실행이 아무것도 고치지 않습니다.

`skip` 분기 바로 위에 이미 같은 논리의 주석이 붙어 있습니다. 이 셸이 설치보다 먼저 떠 있었을 수 있으니 아무것도 설치하지 않는 실행에서도 세션 PATH는 손봐야 한다는 것입니다. 사용자 PATH도 마찬가지입니다. 레지스트리 항목이 그것을 지운 PATH 편집보다 먼저 있었을 수 있습니다.

## 고칠 방향

`skip` 분기에서도 `Add-ToUserPath`를 부릅니다. 이 함수는 멱등이고 항목이 이미 있으면 `$false`를 돌려주므로, 정상적인 경우에 드는 비용은 레지스트리 읽기 한 번입니다.

설치 분기(`install.ps1:489-510`)와 똑같이 try/catch로 감싸야 합니다. 그룹 정책으로 `HKCU\Environment`가 잠긴 환경에서 아무것도 설치하지 않는 실행이 실패로 끝나면 안 되고, 지금처럼 경고로 남아야 합니다. `Add-ToUserPath`가 `$true`를 돌려준 경우에만 `Send-SettingChange`와 안내 문구를 내보내는 것도 설치 분기와 같습니다.

## 테스트

`install.Tests.ps1`에 이런 경우를 추가합니다. 버전이 이미 최신이고 사용자 PATH에 설치 디렉토리가 없을 때, `Add-ToUserPath`가 불리고 안내 문구가 나오는지 확인합니다. 항목이 이미 있으면 아무 문구도 나오지 않아야 합니다.

## 출처

PR #419 리뷰(머지 이후에 달린 리뷰)의 Warning 항목입니다. https://github.com/alpacax/alpacon-cli/pull/419#pullrequestreview-5111414632

jisung-02:
## Background

`install.ps1` installs into `%LOCALAPPDATA%\Alpacon\bin` and puts that directory on the user `PATH`, and the README tells people to run the same one-liner again to update. When the version already on disk matches the one it would install, the script takes a skip branch: it downloads nothing and returns.

That branch only put the directory on the session `PATH`. The user `PATH` entry lives in `HKCU\Environment` and is what a newly opened terminal reads, and it can go missing for reasons that have nothing to do with alpacon: another installer rewriting `PATH`, the 2047-character truncation of the old environment variable editor, a hand edit. The user who notices `alpacon` is no longer found and re-runs the one-liner to get the entry back is the one user the skip branch did not help.

Closes #426

## Changes

Four commits: the extraction, the fix, and two on the workflow. Each builds and passes the suite on its own.

### `refactor(install)` — the PATH work becomes a function

`Set-AlpaconPath` is the block the install path ran inline at its end: the user `PATH` write, the setting-change broadcast, the session `PATH`, and the warning about a copy of `alpacon` that comes earlier on `PATH`. It is called where it used to sit, with the same arguments in the same order, so nothing changes here—it just becomes reachable from a second caller.

`Get-ShadowingAlpacon` is the shadow lookup pulled out of that same block, with the resolved command reachable as a parameter so the suite can supply it instead of answering from whatever is on the test machine's own `PATH`. `$script:ExeName` replaces the three places that spelled out `alpacon.exe`; they have to agree for the install to work at all.

The lint suppression comment named `Set-InstalledAlpaconVersion`. There are two `Set-` functions now, so it names what they have in common instead.

### `fix(install)` — the skip branch repairs both PATHs

Both exits from `Invoke-AlpaconInstall`—the one that installs and the one that finds the binary already current—now hand every `PATH` decision to `Set-AlpaconPath`. The skip branch gains the user `PATH` write, the broadcast, and the shadow warning; it made none of those before. The `install.ps1` diff here is eight lines.

A user `PATH` that this process may not write stays a warning rather than a failed run. `alpacon.exe` is in place on both paths that reach `Set-AlpaconPath`, so the install is usable by full path even when group policy refuses the registry write, and the warning says what to do instead.

`README.md` said running the command again updates. It now also says it puts the directory back on `PATH`.

### `ci(install)` — the workflow is named for what it does

No product behavior changes here. `install-script.yaml` read as something that publishes the installer; it only checks it, and holds `contents: read`. It is now `install-script-tests.yaml`, and `release.yaml` and the `paths` filters follow. The job ids are unchanged, so the check-run names are unchanged too, and `build-and-test` is the only required status check on this repo.

Both `paths` filters also gain `.github/scripts/run-installer-tests.ps1`, the script the `unit` job runs. It was never listed, so a change to the test harness alone started none of these jobs. Raised by Copilot on this PR; it predates the branch.

Inside the workflow, `analyze` and `end-to-end` set `shell: powershell` once under `defaults` rather than on each of eleven steps; the install directory and exe path are computed once and carried on through `GITHUB_ENV`; and reading `HKCU\Environment` goes through `Get-Item HKCU:\Environment` instead of `Microsoft.Win32.Registry`, which drops the explicit `Close` and the fully qualified `RegistryValueOptions`. Behavior is unchanged there: the steps run under `$ErrorActionPreference = 'Stop'`, so a missing key still stops the step.

The end-to-end re-run step now empties the user `PATH` as well as the session `PATH` before running the installer again, and requires exactly one entry back in each. It only ever emptied the session `PATH`, which is the half the old skip branch did repair, so it passed against the bug it stood closest to.

### `ci(install)` — the analyzer reads the test runner too

Listing `.github/scripts/run-installer-tests.ps1` in the `paths` filters was half of what that file needed. The `analyze` job's own file list still named only `install.ps1`, `install.Tests.ps1` and `PSScriptAnalyzerSettings.psd1`, so the runner started these jobs without ever being read by PSScriptAnalyzer: a break in it was caught only if it happened to take the `unit` job down at runtime. It is now the fourth entry in that list, and in the by-hand command `CLAUDE.md` carries beside it. Raised by junho226 in review.

## Testing

Run on a Windows 11 VM, under both editions CI uses.

```
# Windows PowerShell 5.1
SUMMARY result=Passed total=123 passed=122 failed=0 skipped=1
# PowerShell 7
SUMMARY result=Passed total=123 passed=122 failed=0 skipped=1

PSScriptAnalyzer: no findings
```

The one skip is the suite's single deliberately Unix-only test.

The `refactor` commit was run on its own as well, since a commit that does not pass is not a commit anyone can bisect to. It carries the four `Get-ShadowingAlpacon` cases and not yet the five skip-branch ones:

```
# both editions, at the refactor commit
SUMMARY result=Passed total=118 passed=117 failed=0 skipped=1

PSScriptAnalyzer: no findings
```

New coverage, in `Describe 'A run that installs nothing'`, which mocks `Get-InstalledAlpaconVersion` equal to `Get-LatestAlpaconVersion` so every case runs the skip branch:

- `puts the install directory back on the user PATH` — `Add-ToUserPath`, `Add-ToSessionPath`, the broadcast, and the message all fire, with the install directory as the argument.
- `says nothing about the PATH when the entry is already there` — no message and no broadcast when nothing changed.
- `still reports the version it found` — the existing report survives the added work.
- `warns instead of failing when the user PATH cannot be written` — a throw from `Add-ToUserPath` warns, still reaches the session `PATH`, and still reports the version.
- `warns about a copy of alpacon that comes earlier on the PATH`.

And `Describe 'Get-ShadowingAlpacon'`, four cases: a different copy is named, the installed copy is not, `bin\..\bin` spelling is not, and nothing resolving at all is not.

Both mutants were run. The first is the `refactor` commit's own tree, which is the bug with the extraction already in place:

```
# skip branch on Add-ToSessionPath alone (the #426 bug)
SUMMARY result=Failed total=123 passed=118 failed=4 skipped=1
FAILED: A run that installs nothing.puts the install directory back on the user PATH
FAILED: A run that installs nothing.says nothing about the PATH when the entry is already there
FAILED: A run that installs nothing.warns instead of failing when the user PATH cannot be written
FAILED: A run that installs nothing.warns about a copy of alpacon that comes earlier on the PATH

# Add-ToSessionPath dropped from Set-AlpaconPath
SUMMARY result=Failed total=123 passed=120 failed=2 skipped=1
FAILED: A run that installs nothing.puts the install directory back on the user PATH
FAILED: A run that installs nothing.warns instead of failing when the user PATH cannot be written
```

The end-to-end job's steps were also run by hand on that VM through a harness that stands in for `GITHUB_ENV`; all steps passed, and the user `PATH` value kind came back `ExpandString` with one entry.

The VM runs above were made before the analyzer's file list gained the test runner, so their `no findings` line covers three files, not four. PSScriptAnalyzer needs Windows PowerShell or pwsh and the machine this commit was written on has neither, so the `analyze` job is the first run over `.github/scripts/run-installer-tests.ps1`. It found nothing, and it did read the file: the step runs under `$ErrorActionPreference = 'Stop'`, so a path `Invoke-ScriptAnalyzer` could not open would have failed the job rather than passing quietly.

Actions is green on `fd99cca`, the head as it stands:

```
Install script tests   unit (powershell)  success
Install script tests   unit (pwsh)        success
Install script tests   analyze            success
Install script tests   end-to-end         success
Build and Test                            success
Lint                                      success
```


.github/workflows/install-script-tests.yaml
```diff
@@ -1,12 +1,15 @@
...
 
 on:
   push:
     branches: [main]
     paths:
       - "install.ps1"
       - "install.Tests.ps1"
-      - ".github/workflows/install-script.yaml"
+      - ".github/workflows/install-script-tests.yaml"
+      - ".github/scripts/run-installer-tests.ps1"
```
junho226, line 12: Suggestion: this line is half of what the harness script needed. The other half is the `analyze` job's file list further down, which still names only `install.ps1`, `install.Tests.ps1` and `PSScriptAnalyzerSettings.psd1`.

`run-installer-tests.ps1` is the file that pins the Pester version and turns the suite's result into the job's exit code, and it stays the one PowerShell file in this repository PSScriptAnalyzer never reads. Adding it here means a change to it now starts these jobs, which is the improvement. It is still only caught if it happens to break the `unit` job at runtime.

Adding `'.github/scripts/run-installer-tests.ps1'` to that array is the whole change. I could not attach a suggestion block to it because the line is outside this diff.

junho226 (dismissed):
## Summary

This closes the Warning I left on #419: the skip branch put the install directory on the session `PATH` and nothing else, so the one user who re-runs the one-liner because `alpacon` stopped resolving got no repair. Both exits of `Invoke-AlpaconInstall` now go through `Set-AlpaconPath`, which does the user `PATH` write, the broadcast, the session `PATH` and the shadow warning. The `install.ps1` change is eight lines; the rest is the extraction that made them reachable, the suite, and the workflow rename.

## Verified

**The fix is at the right layer.** `Set-AlpaconPath` is the block the install branch already ran, moved whole, and both call sites pass the same two arguments. The skip branch calls it before its "already at" message, so a run that downloads nothing still repairs both `PATH`s.

**Repair is idempotent.** `Add-PathEntry` compares each existing entry through `ConvertTo-ComparablePath` and returns `$null` when the directory is already there, so `Add-ToUserPath` answers `$false`, the broadcast does not fire, and nothing is appended. A user who re-runs the one-liner ten times gets one entry.

**The claim that the binary is always in place holds.** `Set-AlpaconPath` warns about a shadowing copy by comparing against `$ExePath`, which would be misleading if the skip branch could be reached with no binary there. It cannot: `Get-InstalledAlpaconVersion` tests for the exe before it reads the marker and returns `$null` without it, so a quarantined binary with a surviving marker does not resolve to `skip`. The e2e covers that case directly.

**The user `PATH` write still avoids the trap it was written to avoid.** `Add-ToUserPath` reads with `DoNotExpandEnvironmentNames`, preserves the existing `RegistryValueKind`, and refuses outright when the value is neither `String` nor `ExpandString`. Widening the number of runs that reach this code makes that handling matter more than it did, and it is unchanged.

**The e2e reproduces the bug rather than standing next to it.** The re-run step now strips the directory out of `HKCU\Environment` as well as the session `PATH` and requires exactly one entry back in each. Against `main` that assertion fails. The step also restores the value kind before the run, so a failure there points at the fix and not at the setup. I confirmed the job ran all thirteen steps on this head rather than being satisfied by a path filter.

**The rename keeps the check names.** The job ids `unit`, `analyze` and `end-to-end` are unchanged, and `release.yaml` follows the file to its new name, so the required `build-and-test` status is untouched.

**The two new functions carry the `Alpacon` prefix**, which is the shape my #419 note asked for on new code. Fourteen of the twenty-two existing ones still do not, and leaving those alone is the right call rather than a mass rename in a fix PR.

## Suggestions (1)

One, inline, and it is the half of my #419 suggestion this PR did not reach: both `paths` filters gained `.github/scripts/run-installer-tests.ps1`, but the `analyze` job's file list beside them still names only the three files at the repository root.

Separately, and only as a note rather than a request: the `1.10.0` pin further down now carries a comment saying any older published release does the job. That answers why this version was chosen, which was not quite the question. `goreleaser` still has `needs: install-script`, so if `1.10.0` or its Windows archives are ever pruned, this step fails and every subsequent tag stops there. It is a slow-moving risk and it is not this PR's to fix, so I am not asking for anything here, only leaving the reason written down somewhere other than a merged PR's review.

## Highlights

- The suite pins the behavior rather than the shape. It asserts the broadcast fires, that a second run with the entry already present says nothing about `PATH` and does not broadcast, and that an unwritable user `PATH` warns instead of throwing. Those are the three ways this fix could be wrong.
- `Get-ShadowingAlpacon` takes the resolved command as a parameter so the suite supplies it. Without that the test would answer from whatever is on the test machine's own `PATH`, which is how a check like this ends up passing everywhere and meaning nothing.
- The description says plainly that the old e2e step "passed against the bug it stood closest to." That is the useful sentence in this PR.

## Verdict

**APPROVED**

The fix is correct, minimal, and covered by a test that fails without it.

## Checklist

- [x] Both exits verified to reach `Set-AlpaconPath` with the same arguments
- [x] Idempotence traced through `Add-PathEntry` and `ConvertTo-ComparablePath`
- [x] The "binary is always in place" claim checked against `Get-InstalledAlpaconVersion`
- [x] Registry value-kind handling re-read, unchanged
- [x] The e2e job confirmed to have run all thirteen steps on this head
- [x] Job ids and `release.yaml` reference checked for the rename
- [x] CI green; suite not run locally

.github/workflows/install-script-tests.yaml
jisung-02: Applied in `fd99cca`. You were right that the `paths` entry only starts the jobs and never gets the file read: `run-installer-tests.ps1` was the one PowerShell file here PSScriptAnalyzer never opened, so a break in it surfaced only if it happened to take the `unit` job down at runtime.

`.github/scripts/run-installer-tests.ps1` is now the fourth entry in the `analyze` job's array, and in the by-hand command `CLAUDE.md` carries beside it, which had the same three-file list.

I could not run the analyzer locally—it needs Windows PowerShell or pwsh and this machine has neither—so the `analyze` job on that push is the first run over the file. It passed with no findings, and it did read the file rather than skipping it: the step runs under `$ErrorActionPreference = 'Stop', so a path `Invoke-ScriptAnalyzer` could not open would have failed the job.

junho226 (approved):
## Summary

Delta review of `df9cb2a..fd99cca`. One commit, `ci(install): lint the installer test runner too`, four insertions and two deletions across `.github/workflows/install-script-tests.yaml` and `CLAUDE.md`. `install.ps1`, `install.Tests.ps1` and the end-to-end job are untouched, so everything I verified at `df9cb2a` still holds and I did not redo it.

## Suggestions

Resolved. The `analyze` job's file list now carries `.github/scripts/run-installer-tests.ps1` as its fourth entry, and the by-hand command `CLAUDE.md` keeps beside it gained the same entry, so the two lists that have to agree still do.

Three things I checked rather than took on trust:

- **The fourth path is analyzed, not silently dropped.** The entry lands with a trailing comma on the first line and the pipe on the second, so the expression is `'install.ps1', 'install.Tests.ps1', 'PSScriptAnalyzerSettings.psd1', '.github/scripts/run-installer-tests.ps1' | ForEach-Object { ... }`. The comma operator binds tighter than `|`, so all four elements are enumerated into `ForEach-Object` rather than only the last one—which is the way an edit of this shape goes wrong. Both copies are written the same way.
- **The list is complete at this head.** `git ls-files` finds exactly four PowerShell files in the repository (`install.ps1`, `install.Tests.ps1`, `PSScriptAnalyzerSettings.psd1`, `.github/scripts/run-installer-tests.ps1`), and all four are now analyzed. All four also sit in both `paths` filters, so the job that reads a file is started by a change to that file.
- **The `analyze` run on this head is the evidence.** I could not run PSScriptAnalyzer locally either—this machine has neither `pwsh` nor `powershell`—so I am leaning on the same run you are. `analyze` is green on `fd99cca`, alongside `unit` on both editions, `end-to-end`, `build-and-test`, `golangci-lint` and CodeQL.

The description now says the Windows VM numbers predate this commit and cover three files rather than four. That was the one line that would otherwise have gone quietly stale.

## Highlights

- Section 1 of #427 is closed by this PR in both of its halves: the `paths` filters in `df9cb2a`, the `analyze` list in `fd99cca`. The `1.10.0` pin is section 2 and stays open there, which is the right split. I was not asking for the pin here, and folding it in would have left this PR closing an issue it only half fixes.
- `CLAUDE.md` moved in the same commit as the workflow. That pairing is what keeps the documented by-hand command from drifting into a different check from the one CI runs—the same class of gap this PR just closed one level up.

Not a request, only written down: the file list is still spelled out by hand in two places, so a fifth PowerShell file added later gets analyzed only if someone remembers both. At four files in one repository that is cheaper than a recursive glob, and I would leave it as it is.

## Verdict

**APPROVED**

The suggestion from my last review is resolved, correctly, in the smallest change that closes it. Nothing new in the delta.

## Checklist

- [x] Delta is one commit; `install.ps1` and the Pester suite untouched, so the `df9cb2a` verification carries over
- [x] All four entries reach `Invoke-ScriptAnalyzer`: the comma binds tighter than the pipe, in the workflow and in `CLAUDE.md` alike
- [x] Every PowerShell file in the repository is in the analyzer's list and in both `paths` filters
- [x] `analyze` green on `fd99cca`, and every other check on the head is green
- [x] Job ids unchanged, so the required check contexts are untouched
- [x] The description states that its "no findings" line covers three files, not four

PR #434, opened 2026-09-07
fix(client): keep the access token behind a locked accessor pair

### LINKED ISSUE #397: AlpaconClient.AccessToken is read and written from different goroutines without synchronization
Raised by @junho226 in review on #393.

## Background

`client.AlpaconClient.AccessToken` is a plain string field. `RefreshToken` writes it at `client/client.go:406` and `createRequest` reads it at `client/client.go:204` to build the `Authorization` header. Nothing guards either side.

Until #393 this was contained inside `SudoListener`: every writer and every reader sat under `mfaMu`, because `handleSudoMFA` holds that lock across both of its `RefreshToken` calls, across `pollMFACompletion`, and across `verifySudoGrant`, and the listen goroutine dialed a URL provisioned before `Start`, so it issued no HTTP at all.

After #393 every dial runs `CreateEventSession` in `provisionSession` and `SubscribeEvent` in `subscribe`, both on the listen goroutine, and neither takes `mfaMu`. An MFA flow holds that lock for up to `mfaPollingTimeout` (60s) while the reconnect loop redials every 1 to 30 seconds, so a drop during an MFA flow puts an unsynchronized read next to `handleSudoMFA`'s write.

## Consequence

A torn string header yields a truncated or overlong `Bearer` value. The truncated case is a 401, which `SudoListener.refreshExpiredToken` now recovers from, so it mostly self-heals. The overlong case reads past the allocation.

`-race` does not catch it today: the tests substitute `sl.refreshToken`, so nothing drives a real refresh concurrently with a reconnect.

## Why not fix it in api/event

Taking `mfaMu` around the provisioning call would park the reconnect loop behind a 60-second MFA poll, which is worse than the race. The field belongs to `client`, so that is where the fix belongs.

## Proposed change

A `sync.RWMutex`, or an `atomic.Pointer[string]`, behind an `ac.accessToken()` accessor in `client`. That closes it for every caller at once rather than per call site.

## The listen goroutine writes the field too

Added after @junho226 re-anchored the note on `6c85ff25`. The description above covers the read side, which is what the original report saw. `01ec2cf` widened it: `SudoListener.refreshExpiredToken` calls `sl.refreshToken` (`= ac.RefreshToken`), on the listen goroutine, so that goroutine now both reads `AccessToken` through `setHTTPHeader` and writes it through `RefreshToken`, while `handleSudoMFA` may be holding `mfaMu` across its own two writes on another.

`mfaMu`'s `TryLock` in `refreshExpiredToken` keeps the two refreshes from overlapping, which is all its comment claims. It does nothing for the reads in `setHTTPHeader`, so the write side needs the same accessor rather than a second lock in `api/event`.

jisung-02:
Closes #397

## Background

`AlpaconClient.AccessToken` was an exported string field. `sendRequest` replaces it mid-flight when the server reports the token stale, while every other request reads it to build the `Authorization` header. A Go string is a pointer and a length, so an unsynchronized read landing between those two writes produces a truncated or overlong header.

The locking itself already landed on main in `2180b1a` and `85eea7a`, neither written for this issue. What was missing is what keeps it closed: nothing stopped the next edit from reading the field directly again, and the race had no test behind it.

## Changes

The field is now `accessToken`, unexported, reachable only through `AccessToken()` and `SetAccessToken()`. Both take `tokenMu` across their whole body and call nothing while holding it. The setter is exported so tests in other packages can build an authenticated client without writing the field directly and skipping the lock; there are four such call sites, three in `cmd/iam/user_role_test.go` and one in `cmd/worksession/worksession_wait_token_test.go`, and all four are tests. No production code outside `client` writes the token—`api/event/sudolistener.go`, `api/mfa/mfa.go` and `cmd/tunnel/tunnel.go` all go through `RefreshToken()`.

`client/token_boundary_test.go` walks the AST of every `.go` file in the package, test files included, and fails on any reference to the field outside those two bodies, as a selector or as a composite-literal key. Unexporting alone would not have caught `setHTTPHeader`, the read that started this issue, since it lives in the same package. Matching on the AST rather than on text means a comment naming the field is not an offense and renaming the receiver variable does not evade the rule. An accessor is matched by receiver as well as by name, so a plain function called `AccessToken` inherits none of the pass. `NewAlpaconAPIClient` plants both of its token values through the setter, so the rule has no exceptions.

`client/client_test.go` gains the regression test the issue asked for. Eight goroutines send requests while another refreshes the token, and nothing below the client is stubbed: the fake workspace serves the auth env and the token endpoint, so `RefreshToken` runs the real grant against the real header path. The endpoint the readers hit checks the `Authorization` header it receives and answers 500 on one that is neither token, so a torn header fails the test rather than waiting on `-race` to notice.

`api/config_boundary_test.go` gets the same protection against passing vacuously that the token guard already had. It matched `configReaders` against selector names as strings with nothing tying those strings to the symbols they name, so renaming `config.LoadConfig` would have left the walk nothing to find and the guard green on an empty offender list. The four names are now referenced at compile time from the test file, which the walk skips, so a rename fails to build instead.

No behavior changes. The legacy `Token` field and the `api/event` token handling are untouched.

## Verification

Every new test and check here was watched failing before it passed.

Stripping `tokenMu` from the two accessors:

```
WARNING: DATA RACE
Write at 0x00c000138a88 by goroutine 18:
  client.(*AlpaconClient).SetAccessToken()  client/client.go:616
  client.(*AlpaconClient).refreshLocked()   client/client.go:634
Previous read at 0x00c000138a88 by goroutine 14:
  client.(*AlpaconClient).AccessToken()     client/client.go:602
  client.(*AlpaconClient).setHTTPHeader()   client/client.go:266
```

Making `setHTTPHeader` read the field directly:

```
Error: Should be empty, but was [client.go:266:23 in setHTTPHeader]
```

Adding a plain function that borrows the accessor name, `func AccessToken(ac *AlpaconClient) string { return ac.accessToken }`:

```
Error: Should be empty, but was [zz_scratch_red_test.go:3:56 in AccessToken]
```

Making `SetAccessToken` append to the token it installs, so the header arrives corrupted:

```
--- FAIL: TestRefreshTokenRacesConcurrentRequests (0.20s)
    client_test.go:1200: Authorization header was torn: "Bearer staleCORRUPT"
```

Renaming one of the compile-time references in the config guard:

```
vet: api/config_boundary_test.go:33:13: undefined: config.LoadConfigRenamed
```

On the branch as it stands, `go test -race -shuffle=on ./...` exits 0 with 42 packages reporting `ok`, and `golangci-lint run ./...` reports `0 issues.`


client/client_test.go
```diff
@@ -1174,3 +1175,63 @@ func TestSetWebsocketHeaderWithCapabilities(t *testing.T) {
...
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{}`))
+	})
+
+	require.NoError(t, config.CreateConfig(
+		ts.URL, "ws", "", "", "stale", "r1", "alpacon.io", 0, false,
+	))
+
+	ac := &AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+	ac.SetAccessToken("stale")
```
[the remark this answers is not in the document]
jisung-02: Fixed in 28132ac. `ts.Client()` carries no timeout, so a stalled round trip would have hung until the go test deadline instead of failing the `require.NoError` the loop already runs; the client now gets a 10s timeout.

One limit worth stating: this bounds the round trip only. A lock held across the refresh grant blocks in `setHTTPHeader` before `Do` is ever called, so that case still surfaces as the test binary's own timeout dump rather than as a returned error.

CLAUDE.md
```diff
@@ -109,8 +109,8 @@ PSScriptAnalyzerSettings.psd1  # Lint rules for the installer
...
+- **Dual auth tokens**: the access token (Auth0 Bearer JWT) takes priority; `Token` (legacy API key) is fallback. Set in `client.setHTTPHeader()`, which reads the token through `ac.AccessToken()`—the field itself is unexported and `SetAccessToken()` is the only write path, both under `tokenMu`. `client/token_boundary_test.go` enforces that over the AST: a reference to the field anywhere but those two accessor bodies fails, whether it is a selector or a composite-literal key, which is why `NewAlpaconAPIClient` plants the token through the setter rather than in its own struct literal
```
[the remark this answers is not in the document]
jisung-02: Fixed in 28132ac, on the test side rather than the doc. The walk no longer skips `_test.go`, so the sentence in `CLAUDE.md` is true as written. A test file compiles into package `client` and can touch the field exactly as `setHTTPHeader` did, so it belongs in scope; `api/config_boundary_test.go` keeps its own skip because its subject is production reads of `config`, which tests may legitimately call.

Verified RED: adding `_ = ac.accessToken` to `client_test.go` fails with `Should be empty, but was [client_test.go:731:9 in newBearerTestClient]`.

client/token_boundary_test.go
```diff
@@ -0,0 +1,109 @@
...
+			if field == accessTokenField {
+				declared = true
+			}
+		}
+		offenders = append(offenders, fieldRefsOutside(fset, file, accessTokenField, accessTokenAccessors)...)
+	}
+
+	// A renamed field would leave the walk nothing to find and pass vacuously.
+	assert.True(t, declared, "AlpaconClient declares no unexported field named %q", accessTokenField)
+	assert.Empty(t, offenders, "every access token reference must sit inside AccessToken() or SetAccessToken()")
```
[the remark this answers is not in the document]
jisung-02: Not changing this one. testify does apply printf formatting once more than one message argument is passed. From `assert/assertions.go` in v1.12.1, the version this module pins:

```go
if len(msgAndArgs) > 1 {
    return fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)
}
```

So `%q` is expanded and the field name reaches the failure message. `Truef` would read the same; testifylint's formatter check does not flag the non-f form here, and `golangci-lint run ./...` reports `0 issues.`

client/client_test.go
```diff
@@ -747,7 +748,7 @@ func stubTokenRenewal(t *testing.T, newToken string) *int {
 	calls := 0
 	swapTokenRenewal(t, func(ac *AlpaconClient) error {
 		calls++
-		ac.setAccessToken(newToken)
+		ac.SetAccessToken(newToken)
```
geunwoonoh, line 751: Warning: the doc comment on `stubTokenRenewal` (lines 743-745, just above this hunk) still names `setAccessToken`, which this PR deleted. A grep for that identifier across the tree now finds only that line.

The boundary test parses with mode `0`, so comments are invisible to it by design—which is exactly what lets a stale identifier reference survive here.

```go
// stubTokenRenewal swaps the refresh seam for one that installs newToken and
// counts its runs. The seam runs with refreshMu held, so it installs the token
// through SetAccessToken the way refreshLocked does.
```

client/client.go
```diff
@@ -596,26 +596,25 @@ func (ac *AlpaconClient) RefreshToken() error {
...
 // legacy API key. Some endpoints refuse the API key outright, and their refusal arrives
 // with no error code, so a caller rewriting it has to know which credential it sent.
 func (ac *AlpaconClient) IsBearerAuth() bool {
-	return ac.accessToken() != ""
+	return ac.AccessToken() != ""
 }
 
-// setAccessToken installs the token every later request carries.
-func (ac *AlpaconClient) setAccessToken(token string) {
+// SetAccessToken is exported so callers outside this package cannot skip tokenMu.
```
geunwoonoh: Warning: this comment states a rationale that is not true today, and never says what the method does.

Every caller outside `client` is a test: `cmd/iam/user_role_test.go` (3) and `cmd/worksession/worksession_wait_token_test.go` (1). No production code outside the package writes the token—`api/event/sudolistener.go`, `api/mfa/mfa.go` and `cmd/tunnel/tunnel.go` all go through `RefreshToken()`. So the export exists to let cross-package tests build an authenticated client, and "callers outside this package cannot skip tokenMu" reads as if such callers exist.

Go convention also wants an exported method's doc to open with what it does.

```suggestion
// SetAccessToken installs the token every later request carries. It is exported
// so tests in other packages can build an authenticated client without writing
// the field directly and skipping tokenMu.
```

client/client.go
```diff
@@ -596,26 +596,25 @@ func (ac *AlpaconClient) RefreshToken() error {
 	return ac.refreshLocked()
 }
 
-// accessToken reads the token a request should carry. Every read goes through
-// here because sendRequest can replace it between two requests.
-func (ac *AlpaconClient) accessToken() string {
+// AccessToken locks because sendRequest can swap the token mid-flight.
```
geunwoonoh: Warning: the comment gives the reason for the lock but never says the method returns the token a request should carry. The old unexported version said it; the exported one, which now shows up in godoc, says less.

```suggestion
// AccessToken returns the token a request should carry. It locks because
// sendRequest can swap the token mid-flight.
```

client/token_boundary_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const accessTokenField = "accessToken"
+
+// Only these two take tokenMu, so the rule rejects every other reference rather
+// than try to prove from the AST that a lock is held (issue #397).
+var accessTokenAccessors = map[string]bool{"AccessToken": true, "SetAccessToken": true}
```
geunwoonoh, line 21: Suggestion: this allowlist keys on the function name alone, so the rule is bypassable by anything that happens to carry that name. The following compiles, touches the field with no lock held, and the boundary test passes:

```go
func AccessToken(ac *AlpaconClient) string { return ac.accessToken }
```

Same for a method named `AccessToken` on some other type in the package. Checking that the declaration is a method on `*AlpaconClient` closes it:

```go
// Only these two take tokenMu, so the rule rejects every other reference rather
// than try to prove from the AST that a lock is held (issue #397). Matching the
// receiver too keeps a plain function of the same name from inheriting the pass.
var accessTokenAccessors = map[string]bool{"AccessToken": true, "SetAccessToken": true}

func isAccessor(fn *ast.FuncDecl, allowed map[string]bool) bool {
	if !allowed[fn.Name.Name] || fn.Recv == nil || len(fn.Recv.List) != 1 {
		return false
	}
	star, isPtr := fn.Recv.List[0].Type.(*ast.StarExpr)
	if !isPtr {
		return false
	}
	ident, isIdent := star.X.(*ast.Ident)
	return isIdent && ident.Name == "AlpaconClient"
}
```

with `fieldRefsOutside` calling `isAccessor(fn, allowed)` where it currently reads `allowed[fn.Name.Name]`.

client/client_test.go
```diff
@@ -1174,3 +1175,67 @@ func TestSetWebsocketHeaderWithCapabilities(t *testing.T) {
...
+		_, _ = w.Write([]byte(`{"auth0":{"method":"auth0","client_id":"cli","domain":"` + strings.TrimPrefix(ts.URL, "https://") + `"}}`))
+	})
+	mux.HandleFunc("/oauth/token/", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"access_token":"fresh","expires_in":3600,"token_type":"Bearer"}`))
+	})
+	mux.HandleFunc("/api/test/", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{}`))
+	})
```
geunwoonoh, line 1206: Suggestion: this handler answers `{}` to anything, so the test asserts only that no request errored. A truncated or overlong `Authorization` header—the exact damage issue #397 describes—still gets a 200, which means the test detects the bug only through `-race`.

Asserting the header here makes it fail on its own merits. Per `CLAUDE.md`, a handler reports with `t.Errorf` rather than `require`, answers 500, and returns.

```suggestion
	mux.HandleFunc("/api/test/", func(w http.ResponseWriter, r *http.Request) {
		// A torn read of the token yields a truncated or overlong header, which
		// a handler that answers everything would let through.
		if got := r.Header.Get("Authorization"); got != "Bearer stale" && got != "Bearer fresh" {
			t.Errorf("Authorization header was torn: %q", got)
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
		w.Header().Set("Content-Type", "application/json")
		_, _ = w.Write([]byte(`{}`))
	})
```

geunwoonoh (commented):
### Summary

The access token moves behind an accessor pair guarded by `tokenMu`, and two tests keep it there: an AST walk that rejects any direct reference to the field inside `client/`, and a concurrency test that drives the real refresh grant against eight goroutines building `Authorization` headers. The lock ordering is sound—`renewAccessToken` and `RefreshToken` take `refreshMu` and then `tokenMu` through the setter, and neither accessor calls anything while holding `tokenMu`, so the reverse order cannot arise. No behavior change, and CI is green on `28132ac`.

### Critical issues (0)

None.

### Warnings (3)

1. **[client/client_test.go:743-745]** The comment on `stubTokenRenewal` still names `setAccessToken`, which this PR deleted. A grep for that identifier now finds only this line. The boundary test parses with mode `0`, so comments are invisible to it by design—which is what lets a stale identifier reference survive here.

2. **[client/client.go:613]** `SetAccessToken`'s doc comment states a rationale that is not true today, and never says what the method does. Grepping the tree, every caller outside `client` is a test: `cmd/iam/user_role_test.go` (3) and `cmd/worksession/worksession_wait_token_test.go` (1). No production code outside the package writes the token—`api/event/sudolistener.go`, `api/mfa/mfa.go` and `cmd/tunnel/tunnel.go` all go through `RefreshToken()`. So the export exists to let cross-package tests build an authenticated client, and the comment should say that; "callers outside this package cannot skip tokenMu" reads as if such callers exist. Go convention also wants an exported method's doc to open with what it does.

3. **[client/client.go:599]** Same convention point on the read side: the comment gives the reason for the lock but never says the method returns the token a request should carry. The old unexported version said it; the exported one, which now shows up in godoc, says less.

### Suggestions (4)

1. **[client/token_boundary_test.go:20]** `accessTokenAccessors` keys on the function name alone, so the rule is bypassable by anything that happens to carry that name. A plain function named `AccessToken` compiles, touches the field with no lock held, and the boundary test passes. Same for a method named `AccessToken` on some other type in the package. Checking that the declaration is a method on `*AlpaconClient` closes it.

2. **[client/client_test.go:1197]** The `/api/test/` handler answers `{}` to anything, so the new test asserts only that no request errored. A truncated or overlong `Authorization` header—the exact damage issue #397 describes—still gets a 200, which means the test detects the bug only through `-race`. Asserting the header in the handler makes it fail on its own merits.

3. **[client/client_test.go:1213]** `readers, reads, refreshes = 8, 50, 20` sends 400 real TLS round trips plus 20 grants, each grant carrying a `config.LoadConfig()` read and a config file rewrite. The test is serial (correctly—`t.Setenv`), so it runs before any parallel test in the package starts and its cost is wall clock the whole package waits on. It passes comfortably today; worth remembering as the first dial to turn if the `client` package shows up in CI time.

4. **[client/client.go:103]** `NewAlpaconAPIClient` now calls the setter twice on the expired path, taking `tokenMu` once to install a token it is about to replace. Harmless—the client is not shared yet—and resolving the token first would need a local variable that the current shape avoids. Noting it only because the redundancy is new with this PR; leaving it as is, is defensible.

### Dead, duplicate, and unused code

Nothing to flag. The rename left no orphan: `setAccessToken` and `accessToken()` survive only in the stale comment at Warning 1. `RefreshToken()`, `IsBearerAuth()` and the `newBearerTestClient` helper all still have callers. The two accessors are not duplicates of each other, and no third read path was left behind—the boundary test proves that mechanically, which is the point of it.

### Performance

No regression, and no change worth making.

- `AccessToken()` adds one uncontended `sync.Mutex` acquire per outbound request. That is tens of nanoseconds against an HTTPS round trip; it is not measurable.
- **Do not switch `tokenMu` to `sync.RWMutex`.** The critical section is a single string copy, and `RWMutex.RLock` carries more overhead than `Mutex.Lock` at this level of contention. The change would very likely be slower, not faster.
- The AST boundary test parses four files once per run.
- Out of scope but worth knowing: `refreshLocked` does a `config.LoadConfig()` disk read on every renewal, and `auth0.RefreshAccessToken` writes the config file. Serialized by `refreshMu`, so it is correct, just I/O-heavy per renewal. Pre-existing, not introduced here.

### Highlights

- Matching on the AST rather than on text is the right call, and the PR body explains why: a comment naming the field is not an offense, and renaming a receiver does not evade the rule.
- The `declared` flag is the detail that makes the boundary test worth having. Without it a rename of the field would leave the walk nothing to find and the test would pass vacuously—a failure mode most guard tests of this shape ship with.
- The regression test stubs nothing below the client. The fake workspace serves the auth env and the token endpoint, so `RefreshToken` runs the real grant through the real header path rather than a seam.
- Both new tests were watched failing before they passed, and the PR body records both failure outputs. That is the evidence the tests test what they claim.
- `CLAUDE.md` is updated in the same PR, so the next reader learns the rule from the place they already look.

---

**Verdict**: COMMENTED—no blockers; three comment/doc fixes and one real hole in the boundary rule (Suggestion 1) worth closing before merge.

**Checklist**:
- [x] Code follows stack conventions (one exception: Go doc-comment form on the two newly exported methods, Warnings 2 and 3)
- [x] Tests included and passing (CI green on `28132ac`, all 10 checks)
- [x] No security vulnerabilities—the change narrows the token's write surface rather than widening it; no token value is logged
- [x] Documentation updated (`CLAUDE.md`)

client/client_test.go
```diff
@@ -1174,3 +1175,67 @@ func TestSetWebsocketHeaderWithCapabilities(t *testing.T) {
...
+	))
+
+	// A stalled round trip must fail the test, not hang it until the go test deadline.
+	httpClient := ts.Client()
+	httpClient.Timeout = 10 * time.Second
+
+	ac := &AlpaconClient{HTTPClient: httpClient, BaseURL: ts.URL}
+	ac.SetAccessToken("stale")
+
+	const readers, reads, refreshes = 8, 50, 20
```
geunwoonoh, line 1219: Suggestion: these counts send 400 real TLS round trips plus 20 grants, each grant carrying a `config.LoadConfig()` read and a config file rewrite.

The test is serial (correctly—`t.Setenv`), so it runs before any parallel test in the package starts and its cost is wall clock the whole package waits on. It passes comfortably today; worth remembering as the first dial to turn if the `client` package shows up in CI time.

client/client.go
```diff
@@ -98,9 +98,9 @@ func NewAlpaconAPIClient() (*AlpaconClient, error) {
 		BaseURL:       validConfig.WorkspaceURL,
 		WorkspaceName: validConfig.WorkspaceName,
 		Token:         validConfig.Token,
-		AccessToken:   validConfig.AccessToken,
 		UserAgent:     utils.GetUserAgent(),
 	}
+	client.SetAccessToken(validConfig.AccessToken)
```
geunwoonoh, line 103: Suggestion: on the expired path the setter now runs twice, taking `tokenMu` once to install a token it is about to replace three lines later.

Harmless—the client is not shared yet—and resolving the token first would need a local variable that the current shape avoids. Noting it only because the redundancy is new with this PR; leaving it as is, is defensible.

client/client.go
jisung-02: Applied in b6ff99e. You are right that the exported doc said less than the unexported one it replaced—the reason for the lock is not what a godoc reader needs first. It now opens with what the method returns.

client/client.go
jisung-02: Applied in af8ff2a. The claim was false as you show: all four callers outside the package are tests, and the production paths go through RefreshToken(). The doc now opens with what the method does and states the real reason for the export.

client/client_test.go
jisung-02: Applied in ad2d122. The boundary test parses with mode 0 by design, so it was never going to catch this one—the stale name survived the rename precisely because nothing mechanical looks at comments.

client/token_boundary_test.go
jisung-02: Applied in c24c461, with your isAccessor as written. This was a real hole: the allowlist keyed on the name alone, so anything carrying it inherited the pass. Seen RED before the fix landed—adding your example function gives:

    Error: Should be empty, but was [zz_scratch_red_test.go:3:56 in AccessToken]

client/client_test.go
jisung-02: Applied in d649371. Agreed—the handler answering everything left -race as the only thing that could see the damage. Seen RED by making SetAccessToken append to the token it installs, so the header arrives corrupted:

    --- FAIL: TestRefreshTokenRacesConcurrentRequests (0.20s)
        client_test.go:1200: Authorization header was torn: "Bearer staleCORRUPT"

client/client_test.go
jisung-02: Left as is. Measured on the pushed head, the whole client package takes 1.456s under go test -race -shuffle=on, so the 400 round trips are not what would show up in CI time today. Noted as the first dial to turn if it ever does; no issue filed for it.

client/client.go
jisung-02: Left as is, for the reason you gave: the client is not shared yet, and resolving the token first would need a local variable the current shape avoids. The redundant lock acquire is one uncontended Mutex on a client nobody else holds.

client/token_boundary_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+const accessTokenField = "accessToken"
+
+// Only these two take tokenMu, so the rule rejects every other reference rather
+// than try to prove from the AST that a lock is held (issue #397). The receiver
+// match keeps a plain function of the same name from inheriting the pass.
+var accessTokenAccessors = map[string]bool{"AccessToken": true, "SetAccessToken": true}
+
+// Unexporting stops other packages; setHTTPHeader and the package's own tests
+// live here. The AST match ignores comments and any receiver name.
```
geunwoonoh: Suggestion (non-blocking, the only thing left): now that `isAccessor` checks `ident.Name == "AlpaconClient"`, "ignores comments and any receiver name" can be read as saying the receiver's *type* is ignored, which is the opposite of what `c24c461` just established.

The sentence was written when only the receiver variable's name was irrelevant, and that is still true—`fieldRefsOutside` matches `SelectorExpr.Sel.Name` and never looks at `X`. Naming the variable removes the ambiguity.

```suggestion
// live here. The AST match ignores comments and the receiver's variable name.
```

geunwoonoh (dismissed):
### Summary

Every finding from the previous review that asked for a change has landed, one commit each, and CI is green on `5baf0e6` (all 10 checks). The two findings marked as needing no action—the race test's iteration counts and the double setter call in `NewAlpaconAPIClient`—were correctly left alone. The one addition beyond the review, pinning `api/config_boundary_test.go`'s barred reader names at compile time, is sound and self-consistent.

### Critical issues (0)

None.

### Warnings (0)

None. All three from the previous round are fixed:

| Previous finding | Commit | State |
|---|---|---|
| `AccessToken` doc says why it locks, not what it returns | `b6ff99e` | Applied verbatim |
| `SetAccessToken` doc states a rationale untrue of production callers | `af8ff2a` | Applied verbatim |
| `stubTokenRenewal` comment names the deleted `setAccessToken` | `ad2d122` | Applied |

### Suggestions (1)

1. **[client/token_boundary_test.go:24]** Non-blocking, and the only thing left. Now that `isAccessor` checks `ident.Name == "AlpaconClient"`, "ignores comments and any receiver name" can be read as saying the receiver's *type* is ignored, which is the opposite of what `c24c461` just established. The sentence was written when only the receiver variable's name was irrelevant, and that is still true—`fieldRefsOutside` matches `SelectorExpr.Sel.Name` and never looks at `X`. Naming the variable removes the ambiguity.

### Verification of the previous round

- **`c24c461`, the receiver match.** `isAccessor` requires a single pointer receiver whose type identifier is `AlpaconClient`, and `fieldRefsOutside` calls it where it previously read `allowed[fn.Name.Name]`. The bypass raised last round—a plain `func AccessToken(ac *AlpaconClient) string { return ac.accessToken }`—now fails on `fn.Recv == nil`, and a method of that name on any other type fails on the identifier check. `CLAUDE.md` records the rule.
- **`d649371`, the torn-header assertion.** Applied with `t.Errorf`, a 500 and a return, per the repo's rule against `require` inside a handler, and the comment is one line rather than the three proposed. It cannot flake: the client carries no legacy `Token`, so the header is never the `token="…"` form; `SetAccessToken` runs only from the initial plant and from `refreshLocked`, whose grant endpoint always answers `fresh`, so `Bearer stale` and `Bearer fresh` are the complete set. The grant and auth-env endpoints on the same mux are unasserted, which is right—neither request carries a bearer.
- **Correctly not actioned.** The race test's `8, 50, 20` and the double `SetAccessToken` in `NewAlpaconAPIClient` were both flagged as observations rather than requests, and both were left as they are.

### The change outside the previous review

`5baf0e6` adds four compile-time references to `api/config_boundary_test.go`, so renaming any barred reader breaks the build instead of leaving the name-keyed walk nothing to find. This generalizes the `declared` vacuity guard from the token boundary test to its sibling, and it is the right shape for the problem.

The failure mode worth checking was whether the new block flags itself—that file exists to bar `config` reads under `api/`, and it now references `config.LoadConfig` four times. It does not: `TestAPILayerNeverReadsConfig`'s `WalkDir` filter returns early on any path ending in `_test.go`, so the guard only ever walks production files. `package api` importing `config` from a test file introduces no cycle, and the build passing is itself the proof that all four symbols exist as package-level functions.

The one cost is that the four names now live in two lists that must be kept in step. Adding a fifth reader to `configReaders` without a matching `_ =` line silently loses the compile-time pin for that one. There is no way around it—a name cannot be recovered from a function value at compile time—so this is a note, not a request.

The change does put `api/` inside a PR titled for the client's token field. It is nine lines, it closes the same class of hole, and the `CLAUDE.md` bullet is updated alongside it, so it reads as deliberate rather than as drift.

### Highlights

- One finding per commit, each subject naming the change rather than the review round. `docs(client): say what AccessToken returns, not only why it locks` says what changed without the reader opening the diff.
- The torn-header assertion was tightened, not just pasted: the three-line comment proposed in review became one line that says the thing that matters, which is the repo's own comment bar applied to a suggestion rather than accepted whole.
- Carrying the vacuity guard across to `api/config_boundary_test.go` was not asked for. Both guard tests now fail loudly on a rename rather than passing with nothing checked.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (CI green on `5baf0e6`, all 10 checks)
- [x] No security vulnerabilities—the token's write surface is narrower than before; no token value is logged
- [x] Documentation updated (`CLAUDE.md`, both bullets)

client/token_boundary_test.go
jisung-02: Applied in 00f1796, with your wording. You are right that the sentence now reads against c24c461—the receiver type is exactly what the rule started checking, and only the variable name is still ignored.

The same ambiguity was in the PR body ("renaming a receiver does not evade the rule"), so that is reworded too.

On the two lists having to be kept in step: agreed there is no way to recover a name from a function value, so I have left it. If a fifth reader is ever added, the pin it loses is silent, and the place that would catch it is the CLAUDE.md bullet naming the barred set.

geunwoonoh (approved):
### Summary

`00f1796` applies the previous review's only open item verbatim. The delta from `5baf0e6` is one comment line in `client/token_boundary_test.go`; no code, no test, and no behavior changed, which the unchanged PR totals confirm. CI is green on `00f1796` (all 10 checks).

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (0)

None. Nothing is left open on this PR.

### Verification of the previous round

The wording is now correct against both matchers, and they pull in opposite directions, which is why the old sentence had to change. `fieldRefsOutside` matches `SelectorExpr.Sel.Name` and never reads `X`, so the receiver's variable name is genuinely ignored—renaming `ac` evades nothing. `isAccessor` reads the receiver's *type* and requires it to be `AlpaconClient`, so that half is very much not ignored. "The receiver's variable name" names the half that is, and a reader no longer has to guess which one the sentence meant.

### Highlights

- The commit subject, `docs(client): say which part of the receiver the match ignores`, states the distinction rather than pointing at a review round. Someone running `git log` on this file gets the reason without opening the diff.
- A one-line comment fix landed as its own commit rather than being folded into a rebase, so the history keeps saying what changed and why.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (CI green on `00f1796`, all 10 checks)
- [x] Documentation updated (`CLAUDE.md`, both bullets)

PR #435, opened 2026-09-08
fix: fail instead of reporting success when a command's output cannot be read

jisung-02:
## Background

Under the streaming contract a command's output is stored as chunks and the `result` field on the command detail stays empty, so every non-streaming path rebuilds the output through `api/event`'s `GetCommandOutput` and keeps `result` only as the fallback for a command that predates the chunk endpoint.

Two paths did that, and both discarded the fetch error.

`cmd/exec/logs.go`, behind `alpacon exec logs JOB_ID`:

```go
if output, oerr := event.GetCommandOutput(alpaconClient, jobID); oerr == nil && output != "" {
	details.Result = output
}
```

`api/event/event.go`'s `runCommandFallbackFromID`, the path `exec` and `websh` take when the output stream cannot be set up and the CLI falls back to polling:

```go
output := details.Result
if reconstructed, oerr := GetCommandOutput(ac, cmdID); oerr == nil && reconstructed != "" {
	output = reconstructed
}
```

On either path a 500 or a network failure on `/api/events/commands/<id>/chunks/` took the `oerr == nil` guard out and left the output at the empty string it already held. The command then printed nothing and reported success—exit `0` from `exec logs`, a nil error from the fallback. Under the streaming contract that is indistinguishable from a command that ran fine and produced no output, so a script or an AI agent branching on the exit code read an unreadable result as a clean one, which is the case the exit codes exist to keep apart.

The fallback path was the more misleading of the two, because it looks covered. It warns first, and its comment said the warning was enough:

```
real-time output unavailable (%v); falling back to polling
```

That names a lost stream. It says nothing about a lost output, and it is printed before the fetch that failed.

### One copy of the decision, in `api/event`

`api/event/chunks.go` gains the two functions that replace both copies:

- `ResolveCommandOutput(ac, cmdID, result)` pairs the fetch with the fallback. Both call sites now call this and nothing else.
- `pickCommandOutput(result, chunked, chunkErr)` is the fallback with the fetch lifted out, so a case table can read it without a server.

`cmd/exec` already imports `api/event`, so this is the only direction the shared version could go.

A chunk fetch failure is reported exactly when `result` carries nothing to fall back on:

| chunk fetch | `result` | printed as the output | what the caller ends on |
|---|---|---|---|
| ok, non-empty | anything | the chunks | the command's own outcome |
| ok, empty | non-empty | `result`, the legacy field | the command's own outcome |
| ok, empty | empty | nothing | the command's own outcome |
| failed with 404 | anything | `result`, the legacy field | the command's own outcome |
| failed otherwise | non-empty | `result`, the legacy field | the command's own outcome |
| failed otherwise | empty | nothing | the read failure, unless the command failed on its own |

A server predating the chunk endpoint answers only the legacy field, which is why a failed fetch with a non-empty `result` still prints and still ends on the command's own outcome. There is nothing a caller could do with an error about a store that server does not have.

The 404 row is that same server saying it has no chunk store at all, and there an empty `result` is the whole output rather than an output nobody could read—a command that printed nothing must not become `failed to read command output` on those deployments. `utils.HTTPStatusCode` reads the status the client tags onto the error, and `api.FetchAllPages` wraps with `%w`, so the tag reaches the decision.

The last row is the only one where the read failure has anything to say, and even there it yields to a command that failed on its own—that failure carries the remote exit code, and both paths keep it. Both sections below spell that out.

### `alpacon exec logs`

A command that otherwise succeeded now reports `failed to read command output: ...` and exits `1`. The message names the fetch and not the command, because the command itself may well have succeeded and it is the CLI that could not read what it wrote. It goes through `utils.CliErrorWithExit`, which sanitizes the server's text (#364). Exit `1` needs no new README row: the table there already covers a server error under the general failure code.

The read failure is held until the outcome is known and surfaces only where the outcome would otherwise be exit `0`, so a command that failed on its own keeps the exit code it carries—the same order the polling fallback keeps, below. Without that, an unreachable chunk store turned a command that exited `7` into a generic CLI failure, which is the exit code a script branches on.

The fetch itself is skipped on the statuses whose outcome prints no output, and the two halves of that condition are skipped for different reasons.

`stuck`/`error`/`cancelled` are answered with a status line and exit `1`, and `logsCommandOutcome` returns an empty `stdoutLine` there, so the round trip buys nothing and a chunk store that hangs rather than errors costs the wait. An earlier revision of this description credited that half with keeping an unreachable chunk store from burying the status line. It does not, and neither does the comment that used to say so in `logs.go` (`0e2c8d2`). The ordering above does it alone: dropping `!event.IsStatusOnlyFailure` leaves both the status line and the exit code standing, and only the request counter in `TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure` goes red. What that half saves is the round trip.

A running command is the other half, and there the skip is what produces the answer rather than what saves a request. `logsCommandOutcome` returns exit `0` for a running status, so a read error takes the `exitCode == 0` gate instead of yielding to the outcome. Dropping `!event.IsRunningStatus` alone, against the same chunk store answering 500, replaces the "still running" line with a read failure:

```
exit=1 chunkReads=1 stdout=""
stderr="Error: failed to read command output: fetching page 1 from
/api/events/commands/a1b2c3d4-5678-abcd-ef01-234567890abc/chunks/:
server returned an empty error response"
```

`go test ./cmd/exec/` stays green with that half dropped, for the reason the last section of this description gives, so the comment above the guard is what carries it (`3e2466f`).

### The polling fallback

`runCommandFallbackFromID` returns `failed to read command output: %w` wrapping the fetch error, which reaches the caller as the command's failure.

The command's own failure keeps precedence over an unreadable output, because it is what carries the remote exit code `exec` propagates as its own; replacing it with a generic read error would change the exit code scripts branch on. Only a command that otherwise reported success now ends on the read failure. `exec logs` applies the same rule, so this is one decision written once and honored on both paths.

### One spelling of the status-only failures

`"stuck"`, `"error"` and `"cancelled"` were matched in two places: `errorFromDetails` carried them as a switch case on the polling path, and `cmd/exec` had a private `isStatusOnlyFailure` for the detach path. Since the point of the change above is that the two paths decide alike on an unreadable output, a fourth terminal status reaching one spelling and missing the other would split them again. `event.IsStatusOnlyFailure` in `api/event/types.go` now holds the triple, beside `IsRunningStatus` and `IsRejectedStatus`, and both callers read it (`e4dde03`).

A switch case cannot call a predicate, so `errorFromDetails` matches the three in an `if` ahead of its switch—the shape the awaiting and rejected statuses already had there. The messages it builds are unchanged.

`api/ftp/ftp.go:696` and `:751` are deliberately left alone. They test `"stuck" || "error"` without `"cancelled"`, so they are a different judgment, and folding them in would change what `alpacon cp` reports.

## Testing

The full suite, as CI runs it, plus the linters, re-run on the pushed head:

```
$ gofmt -l .
(no output)
$ go vet ./...
(no output)
$ golangci-lint run ./...
0 issues.
exit=0    ok packages: 42    FAIL lines: 0
```

Every new test was seen RED. Mutating the shared decision back into swallowing the error (`if chunkErr != nil { return result, nil }`) reddens both packages at once, which is the check that the duplication is actually gone:

```
--- FAIL: TestPickCommandOutput/fetch_failure_with_nothing_to_fall_back_on
--- FAIL: TestResolveCommandOutputSurfacesFetchFailure/nothing_to_fall_back_on_is_an_error
--- FAIL: TestRunCommandFallbackFromID_ChunkFetchFailure/a_successful_command_with_nothing_to_print
--- FAIL: TestExecLogsChunkFetchFailureExitsNonZero
    Messages: helper exited 0; stderr:
```

Restoring `runCommandFallbackFromID`'s old body alone, with the shared decision left correct, reddens the fallback test with the lie itself:

```
--- FAIL: TestRunCommandFallbackFromID_ChunkFetchFailure/a_successful_command_with_nothing_to_print
```

New coverage:

- `TestPickCommandOutput` (`api/event/chunks_test.go`), a case table over the `(fetch, result)` combinations in the table above, the 404 included.
- `TestResolveCommandOutputSurfacesFetchFailure` (`api/event/chunks_test.go`), three cases through an `httptest` server answering the chunks endpoint with a 500 or a 404. It pins the wiring the pure table cannot: that the fetch error reaches the decision, that `result` and the chunks are not passed the wrong way round, and that the 404 tag survives the wrapping `FetchAllPages` puts around it. Mutating the constant to `http.StatusGone` reddens the 404 case in both tests.
- `TestRunCommandFallbackFromID_ChunkFetchFailure` (`api/event/event_test.go`), three cases driving the real fallback against a server whose command detail is terminal and whose chunks endpoint is a 500: a successful command with nothing to print returns the read error, a legacy `result` stands in for the chunks, and a failed command still returns its own `RemoteCommandError`.
- `TestExecLogsChunkFetchFailureExitsNonZero` (`cmd/exec/logs_test.go`), which drives the real `exec logs` through the package's helper process against the same shape of server. It asserts exit `1`, the message on stderr, an empty stdout, and that the message carries no bad-verb marker—the reason the read failed is the only thing the caller has to go on, and a mis-rendered verb would replace it with `%!d(...)` while the prefix assertion stayed green.
- `TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure` (`cmd/exec/logs_test.go`), two cases against the same shape of server: a remote failure carrying `exit_code: 2` still exits `2` and prints no read error, and a `stuck` command still prints `command failed with status: stuck` while a counter in the handler pins that the chunks endpoint was never called. Reverting the ordering reddens both, the second on all three counts:

```
--- FAIL: .../a_remote_failure_keeps_its_own_exit_code
    Error: Not equal: expected: 2, actual: 1
--- FAIL: .../a_stuck_command_keeps_its_status_line
    Error: "...failed to read command output..." does not contain "command failed with status: stuck"
    Error: Should be zero, but was 1
    Messages: the outcome prints no output, so nothing is fetched
```

- `TestIsStatusOnlyFailure` (`cmd/exec/logs_test.go`), beside the existing `TestIsRunningStatus`, over the extracted predicate: the three statuses it holds, and the ones that must stay out of it—`"failed"` above all, since that one carries the remote exit code and a result, so its output is fetched and printed. Mutating the predicate to `return false` reddens both packages, which is what says the extraction preserved behavior:

```
--- FAIL: TestRunCommandStreaming_TerminalStatusErrors
--- FAIL: TestIsStatusOnlyFailure
--- FAIL: TestLogsCommandOutcome
--- FAIL: TestLogsCommandOutcomeSanitizesServerText
--- FAIL: TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure
```

Dropping only `"cancelled"` from the triple reddens the new test alone.

One thing is deliberately not covered: that a running command reaches its own "still running" line and exit `0` without touching the chunks endpoint. The package's helper-process harness (`runHelperProcess`, `cmd/exec/pending_approval_test.go:107`) requires a non-zero exit and a running command exits `0`, so an exit-`0` path cannot ride it, and widening that contract would touch every helper-process test in the package. `TestIsRunningStatus` and `TestLogsCommandOutcome` cover the predicate and the outcome on both sides, but neither reaches the guard that calls it, which is why dropping the running half leaves the package green and why the comment above the guard states what that half holds.

### DIFF
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -621,9 +621,9 @@ func drainRemainingChunks(ac *client.AlpaconClient, cmdID string, lastSeq int, o
 // errorFromDetails maps a terminal command status to an error so unrecognized
 // statuses are not masked as success.
 func errorFromDetails(d EventDetails) error {
-	// A switch case cannot call a predicate, so every status matched by one is
-	// handled ahead of the switch—a server-side rename then lands in types.go
-	// alone.
+	// A switch case cannot call a predicate, so every status this function answers
+	// specially is handled ahead of the switch—a server-side rename then lands in
+	// types.go alone.
 	if IsAwaitingPurposeStatus(d.Status) {
 		return &AwaitingPurposeError{CommandID: d.ID, ExpiresAt: d.PurposeExpiresAt}
 	}
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -621,8 +621,9 @@ func drainRemainingChunks(ac *client.AlpaconClient, cmdID string, lastSeq int, o
 // errorFromDetails maps a terminal command status to an error so unrecognized
 // statuses are not masked as success.
 func errorFromDetails(d EventDetails) error {
-	// A switch case cannot call a predicate, so the two approval statuses are
-	// matched ahead of it—a server-side rename then lands in types.go alone.
+	// A switch case cannot call a predicate, so every status matched by one is
+	// handled ahead of the switch—a server-side rename then lands in types.go
+	// alone.
 	if IsAwaitingPurposeStatus(d.Status) {
 		return &AwaitingPurposeError{CommandID: d.ID, ExpiresAt: d.PurposeExpiresAt}
 	}
@@ -631,6 +632,16 @@ func errorFromDetails(d EventDetails) error {
 	}
 	if IsRejectedStatus(d.Status) {
 		return &CommandRejectedError{CommandID: d.ID}
+	}
+	if IsStatusOnlyFailure(d.Status) {
+		phase := ""
+		if d.ErrorPhase != nil {
+			phase = *d.ErrorPhase
+		}
+		if phase == "" {
+			return fmt.Errorf("command failed with status: %s", d.Status)
+		}
+		return fmt.Errorf("command failed: [%s] %s (status=%s)", phase, DescribePhase(phase), d.Status)
 	}
 	switch d.Status {
 	case "completed", "success", "failed":
@@ -646,15 +657,6 @@ func errorFromDetails(d EventDetails) error {
 			return &RemoteCommandError{Output: d.Result, ExitCode: exitCode, ErrorPhase: phase, CommandID: d.ID}
 		}
 		return nil
-	case "stuck", "error", "cancelled":
-		phase := ""
-		if d.ErrorPhase != nil {
-			phase = *d.ErrorPhase
-		}
-		if phase == "" {
-			return fmt.Errorf("command failed with status: %s", d.Status)
-		}
-		return fmt.Errorf("command failed: [%s] %s (status=%s)", phase, DescribePhase(phase), d.Status)
 	default:
 		return fmt.Errorf("unexpected command status: %s (command may still be running)", d.Status)
 	}
--- a/api/event/event.go
+++ b/api/event/event.go
@@ -680,15 +680,18 @@ func runCommandFallbackFromID(ac *client.AlpaconClient, cmdID string, out io.Wri
 	if err != nil {
 		return err
 	}
-	// Command has finished: reconstruct output from chunks best-effort, falling
-	// back to Result when chunks are empty or unavailable. No warning on failure—
-	// the polling-fallback warning above already covers it.
-	output := details.Result
-	if reconstructed, oerr := GetCommandOutput(ac, cmdID); oerr == nil && reconstructed != "" {
-		output = reconstructed
-	}
+	output, oerr := ResolveCommandOutput(ac, cmdID, details.Result)
 	if output != "" {
 		_, _ = fmt.Fprint(out, output)
 	}
-	return errorFromDetails(details)
+	// The command's own failure carries the remote exit code the caller
+	// propagates, so it outranks an unreadable output. The warning above names a
+	// lost stream, not a lost output, so it cannot stand in for this error.
+	if err := errorFromDetails(details); err != nil {
+		return err
+	}
+	if oerr != nil {
+		return fmt.Errorf("failed to read command output: %w", oerr)
+	}
+	return nil
 }
--- a/api/event/event_test.go
+++ b/api/event/event_test.go
@@ -2272,3 +2272,73 @@ func TestGetCommandByID_MissingSudoFieldsStayNil(t *testing.T) {
 	assert.Nil(t, details.SudoGrantStatus)
 	assert.Nil(t, details.SudoApprovalRequestID)
 }
+
+// TestRunCommandFallbackFromID_ChunkFetchFailure pins what the polling fallback
+// owes the caller when the chunk store is the thing that failed. The warning
+// this path prints first says the stream was lost; a successful command that
+// then prints nothing and returns nil says the command produced no output.
+func TestRunCommandFallbackFromID_ChunkFetchFailure(t *testing.T) {
+	t.Parallel()
+	const cmdID = "a1b2c3d4-1234-5678-abcd-000000000000"
+	tests := []struct {
+		name       string
+		result     string
+		success    bool
+		wantOut    string
+		wantErrMsg string
+		wantRemote bool
+	}{
+		{
+			name:       "a successful command with nothing to print",
+			success:    true,
+			wantErrMsg: "failed to read command output",
+		},
+		{
+			name:    "a legacy result stands in for the chunks",
+			result:  "legacy output\n",
+			success: true,
+			wantOut: "legacy output\n",
+		},
+		{
+			name:       "the command's own failure outranks the read failure",
+			wantRemote: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				switch {
+				case strings.Contains(r.URL.Path, "/chunks/"):
+					w.WriteHeader(http.StatusInternalServerError)
+				case r.URL.Path == "/api/events/commands/"+cmdID+"/":
+					w.Header().Set("Content-Type", "application/json")
+					_ = json.NewEncoder(w).Encode(map[string]any{
+						"id": cmdID, "status": "completed", "success": tt.success,
+						"exit_code": 0, "result": tt.result,
+					})
+				default:
+					http.NotFound(w, r)
+				}
+			}))
+			defer ts.Close()
+
+			ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+
+			var out bytes.Buffer
+			err := runCommandFallbackFromID(ac, cmdID, &out, false, fmt.Errorf("websocket dial failed"))
+
+			assert.Equal(t, tt.wantOut, out.String())
+			switch {
+			case tt.wantRemote:
+				var remoteErr *RemoteCommandError
+				require.ErrorAs(t, err, &remoteErr)
+			case tt.wantErrMsg != "":
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.wantErrMsg)
+			default:
+				require.NoError(t, err)
+			}
+		})
+	}
+}
--- a/cmd/exec/logs.go
+++ b/cmd/exec/logs.go
@@ -74,9 +74,9 @@ Run the command again later to check for completion.`,
 			return
 		}
 
-		// Only the statuses whose outcome prints the output are worth a fetch. A
-		// running command and a status-only failure both answer with a line of
-		// their own, which an unreachable chunk store would only bury.
+		// Only the statuses whose outcome prints the output are worth a fetch: the
+		// rest answer with a line of their own and an empty stdoutLine, so the
+		// round trip buys nothing and a chunk store that hangs costs the wait.
 		var readErr error
 		if !event.IsRunningStatus(details.Status) && !isStatusOnlyFailure(details.Status) {
 			output, oerr := event.ResolveCommandOutput(alpaconClient, jobID, details.Result)
--- a/cmd/exec/logs.go
+++ b/cmd/exec/logs.go
@@ -78,7 +78,7 @@ Run the command again later to check for completion.`,
 		// rest answer with a line of their own and an empty stdoutLine, so the
 		// round trip buys nothing and a chunk store that hangs costs the wait.
 		var readErr error
-		if !event.IsRunningStatus(details.Status) && !isStatusOnlyFailure(details.Status) {
+		if !event.IsRunningStatus(details.Status) && !event.IsStatusOnlyFailure(details.Status) {
 			output, oerr := event.ResolveCommandOutput(alpaconClient, jobID, details.Result)
 			if oerr != nil {
 				readErr = oerr
@@ -120,12 +120,6 @@ func init() {
 	ExecCmd.AddCommand(logsCmd)
 }
 
-// isStatusOnlyFailure reports the statuses whose outcome is a line about the
-// status alone: the command never reached a result worth printing.
-func isStatusOnlyFailure(status string) bool {
-	return status == "stuck" || status == "error" || status == "cancelled"
-}
-
 // logsCommandOutcome guarantees a non-empty stderrLine ends with \n. Neither the
 // awaiting_purpose demand, the awaiting_approval hold, nor a rejection reaches
 // here: the caller answers all three on their own contracts first.
@@ -142,7 +136,7 @@ func logsCommandOutcome(details event.EventDetails) (stdoutLine, stderrLine stri
 		return "", stderrLine, 0
 	}
 
-	if isStatusOnlyFailure(details.Status) {
+	if event.IsStatusOnlyFailure(details.Status) {
 		if details.ErrorPhase != nil && *details.ErrorPhase != "" {
 			phase, desc := sanitizedPhaseParts(*details.ErrorPhase)
 			stderrLine = fmt.Sprintf("%s: [%s] %s (status=%s)\n",
--- a/cmd/exec/logs.go
+++ b/cmd/exec/logs.go
@@ -74,9 +74,12 @@ Run the command again later to check for completion.`,
 			return
 		}
 
-		// Only the statuses whose outcome prints the output are worth a fetch: the
-		// rest answer with a line of their own and an empty stdoutLine, so the
-		// round trip buys nothing and a chunk store that hangs costs the wait.
+		// Only the statuses whose outcome prints the output are worth a fetch, for
+		// two different reasons. A status-only failure answers with a line of its
+		// own and an empty stdoutLine, so the round trip buys nothing and a chunk
+		// store that hangs costs the wait. A running status needs the skip to stay
+		// correct: its outcome exits 0, so a read error would take the
+		// exitCode == 0 gate below and bury the "still running" line.
 		var readErr error
 		if !event.IsRunningStatus(details.Status) && !event.IsStatusOnlyFailure(details.Status) {
 			output, oerr := event.ResolveCommandOutput(alpaconClient, jobID, details.Result)
--- a/cmd/exec/logs.go
+++ b/cmd/exec/logs.go
@@ -74,18 +74,15 @@ Run the command again later to check for completion.`,
 			return
 		}
 
-		// Output is stored as chunks (Result is empty under the streaming
-		// contract); reconstruct it, falling back to Result for legacy commands.
 		// A running command's output is not printed here, so skip the fetch and
 		// keep an unreachable chunk store from burying the "still running" line.
 		if !event.IsRunningStatus(details.Status) {
-			output, oerr := event.GetCommandOutput(alpaconClient, jobID)
-			result, rerr := resolveCommandOutput(details.Result, output, oerr)
-			if rerr != nil {
-				utils.CliErrorWithExit("failed to fetch command output: %s", rerr)
+			output, oerr := event.ResolveCommandOutput(alpaconClient, jobID, details.Result)
+			if oerr != nil {
+				utils.CliErrorWithExit("failed to read command output: %s", oerr)
 				return
 			}
-			details.Result = result
+			details.Result = output
 		}
 
 		// What the requester said the command was for, ahead of the outcome and
@@ -112,24 +109,6 @@ Run the command again later to check for completion.`,
 
 func init() {
 	ExecCmd.AddCommand(logsCmd)
-}
-
-// resolveCommandOutput picks what `exec logs` prints as the command's output.
-// A chunk fetch failure is fatal only when Result carries nothing: a server
-// predating the chunk endpoint still answers the legacy field, while an empty
-// Result leaves nothing to print, and exiting 0 there would report a command
-// that finished having produced no output.
-func resolveCommandOutput(result, chunked string, chunkErr error) (string, error) {
-	if chunkErr != nil {
-		if result != "" {
-			return result, nil
-		}
-		return "", chunkErr
-	}
-	if chunked != "" {
-		return chunked, nil
-	}
-	return result, nil
 }
 
 // logsCommandOutcome guarantees a non-empty stderrLine ends with \n. Neither the
--- a/cmd/exec/logs.go
+++ b/cmd/exec/logs.go
@@ -74,15 +74,17 @@ Run the command again later to check for completion.`,
 			return
 		}
 
-		// A running command's output is not printed here, so skip the fetch and
-		// keep an unreachable chunk store from burying the "still running" line.
-		if !event.IsRunningStatus(details.Status) {
+		// Only the statuses whose outcome prints the output are worth a fetch. A
+		// running command and a status-only failure both answer with a line of
+		// their own, which an unreachable chunk store would only bury.
+		var readErr error
+		if !event.IsRunningStatus(details.Status) && !isStatusOnlyFailure(details.Status) {
 			output, oerr := event.ResolveCommandOutput(alpaconClient, jobID, details.Result)
 			if oerr != nil {
-				utils.CliErrorWithExit("failed to read command output: %s", oerr)
-				return
+				readErr = oerr
+			} else {
+				details.Result = output
 			}
-			details.Result = output
 		}
 
 		// What the requester said the command was for, ahead of the outcome and
@@ -95,6 +97,13 @@ Run the command again later to check for completion.`,
 		}
 
 		stdoutLine, stderrLine, exitCode := logsCommandOutcome(details)
+		// The command's own failure carries the remote exit code this command
+		// propagates, so it outranks an unreadable output—the order
+		// runCommandFallbackFromID keeps on the polling path.
+		if exitCode == 0 && readErr != nil {
+			utils.CliErrorWithExit("failed to read command output: %s", readErr)
+			return
+		}
 		if stdoutLine != "" {
 			fmt.Println(stdoutLine)
 		}
@@ -109,6 +118,12 @@ Run the command again later to check for completion.`,
 
 func init() {
 	ExecCmd.AddCommand(logsCmd)
+}
+
+// isStatusOnlyFailure reports the statuses whose outcome is a line about the
+// status alone: the command never reached a result worth printing.
+func isStatusOnlyFailure(status string) bool {
+	return status == "stuck" || status == "error" || status == "cancelled"
 }
 
 // logsCommandOutcome guarantees a non-empty stderrLine ends with \n. Neither the
@@ -127,7 +142,7 @@ func logsCommandOutcome(details event.EventDetails) (stdoutLine, stderrLine stri
 		return "", stderrLine, 0
 	}
 
-	if details.Status == "stuck" || details.Status == "error" || details.Status == "cancelled" {
+	if isStatusOnlyFailure(details.Status) {
 		if details.ErrorPhase != nil && *details.ErrorPhase != "" {
 			phase, desc := sanitizedPhaseParts(*details.ErrorPhase)
 			stderrLine = fmt.Sprintf("%s: [%s] %s (status=%s)\n",
--- a/cmd/exec/logs.go
+++ b/cmd/exec/logs.go
@@ -76,8 +76,16 @@ Run the command again later to check for completion.`,
 
 		// Output is stored as chunks (Result is empty under the streaming
 		// contract); reconstruct it, falling back to Result for legacy commands.
-		if output, oerr := event.GetCommandOutput(alpaconClient, jobID); oerr == nil && output != "" {
-			details.Result = output
+		// A running command's output is not printed here, so skip the fetch and
+		// keep an unreachable chunk store from burying the "still running" line.
+		if !event.IsRunningStatus(details.Status) {
+			output, oerr := event.GetCommandOutput(alpaconClient, jobID)
+			result, rerr := resolveCommandOutput(details.Result, output, oerr)
+			if rerr != nil {
+				utils.CliErrorWithExit("failed to fetch command output: %s", rerr)
+				return
+			}
+			details.Result = result
 		}
 
 		// What the requester said the command was for, ahead of the outcome and
@@ -104,6 +112,24 @@ Run the command again later to check for completion.`,
 
 func init() {
 	ExecCmd.AddCommand(logsCmd)
+}
+
+// resolveCommandOutput picks what `exec logs` prints as the command's output.
+// A chunk fetch failure is fatal only when Result carries nothing: a server
+// predating the chunk endpoint still answers the legacy field, while an empty
+// Result leaves nothing to print, and exiting 0 there would report a command
+// that finished having produced no output.
+func resolveCommandOutput(result, chunked string, chunkErr error) (string, error) {
+	if chunkErr != nil {
+		if result != "" {
+			return result, nil
+		}
+		return "", chunkErr
+	}
+	if chunked != "" {
+		return chunked, nil
+	}
+	return result, nil
 }
 
 // logsCommandOutcome guarantees a non-empty stderrLine ends with \n. Neither the
--- a/api/event/chunks.go
+++ b/api/event/chunks.go
@@ -1,6 +1,7 @@
 package event
 
 import (
+	"net/http"
 	"net/url"
 	"sort"
 	"strconv"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/alpacax/alpacon-cli/api"
 	"github.com/alpacax/alpacon-cli/client"
+	"github.com/alpacax/alpacon-cli/utils"
 )
 
 // noSeqBound is the toSeq sentinel that leaves getCommandChunks unbounded (seq
@@ -68,10 +70,12 @@ func ResolveCommandOutput(ac *client.AlpaconClient, cmdID, result string) (strin
 // pickCommandOutput decides between the two. A chunk fetch failure is fatal only
 // when result carries nothing: a server predating the chunk endpoint answers the
 // legacy field alone, while an empty result leaves nothing to print, and
-// reporting success there describes a command that produced no output.
+// reporting success there describes a command that produced no output. A 404 is
+// that older server saying it has no chunk store at all, so its empty result is
+// the whole output rather than a result nobody could read.
 func pickCommandOutput(result, chunked string, chunkErr error) (string, error) {
 	if chunkErr != nil {
-		if result != "" {
+		if result != "" || utils.HTTPStatusCode(chunkErr) == http.StatusNotFound {
 			return result, nil
 		}
 		return "", chunkErr
--- a/api/event/chunks.go
+++ b/api/event/chunks.go
@@ -57,3 +57,27 @@ func GetCommandOutput(ac *client.AlpaconClient, cmdID string) (string, error) {
 	}
 	return b.String(), nil
 }
+
+// ResolveCommandOutput reconstructs a finished command's output from its chunks,
+// falling back to the result field the command detail carries.
+func ResolveCommandOutput(ac *client.AlpaconClient, cmdID, result string) (string, error) {
+	chunked, err := GetCommandOutput(ac, cmdID)
+	return pickCommandOutput(result, chunked, err)
+}
+
+// pickCommandOutput decides between the two. A chunk fetch failure is fatal only
+// when result carries nothing: a server predating the chunk endpoint answers the
+// legacy field alone, while an empty result leaves nothing to print, and
+// reporting success there describes a command that produced no output.
+func pickCommandOutput(result, chunked string, chunkErr error) (string, error) {
+	if chunkErr != nil {
+		if result != "" {
+			return result, nil
+		}
+		return "", chunkErr
+	}
+	if chunked != "" {
+		return chunked, nil
+	}
+	return result, nil
+}
--- a/api/event/chunks_test.go
+++ b/api/event/chunks_test.go
@@ -15,6 +15,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// statusErr carries an HTTP status the way the API client's own errors do,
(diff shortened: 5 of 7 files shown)


cmd/exec/logs.go
```diff
@@ -74,9 +74,14 @@ Run the command again later to check for completion.`,
...
-		// contract); reconstruct it, falling back to Result for legacy commands.
-		if output, oerr := event.GetCommandOutput(alpaconClient, jobID); oerr == nil && output != "" {
+		// A running command's output is not printed here, so skip the fetch and
+		// keep an unreachable chunk store from burying the "still running" line.
+		if !event.IsRunningStatus(details.Status) {
+			output, oerr := event.ResolveCommandOutput(alpaconClient, jobID, details.Result)
+			if oerr != nil {
+				utils.CliErrorWithExit("failed to read command output: %s", oerr)
+				return
+			}
```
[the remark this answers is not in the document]
jisung-02: `%s` is correct here, so this is not changed. `fmt` treats `%s` and `%v` identically for any value implementing `error`: `handleMethods` checks for the `error` interface before falling through to the default formatting, calls `Error()`, and prints the result. A `%!s(...)` marker only appears for a verb `fmt` cannot apply to the operand, such as `%d` on a string. `%s` on an error is also what every other message in this file uses (`logs.go:35`, `logs.go:42`).

Nothing pinned that rendering, though, so `TestExecLogsChunkFetchFailureExitsNonZero` now asserts the message carries no bad-verb marker (5ec0af7). Mutating the verb to `%d` reddens it and shows what a real mistake would cost—the reason the read failed is replaced by the marker:

```
"failed to read command output: &{%!d(string=fetching page 1 from /api/events/commands/.../chunks/: server returned an empty error response) ...}"
    should not contain "%!"
```

With `%s` the same path prints the reason in full: `failed to read command output: fetching page 1 from /api/events/commands/<id>/chunks/: server returned an empty error response`.

api/event/event_test.go
```diff
@@ -2272,3 +2272,73 @@ func TestGetCommandByID_MissingSudoFieldsStayNil(t *testing.T) {
 	assert.Nil(t, details.SudoGrantStatus)
 	assert.Nil(t, details.SudoApprovalRequestID)
 }
+
+// TestRunCommandFallbackFromID_ChunkFetchFailure pins what the polling fallback
+// owes the caller when the chunk store is the thing that failed. The warning
+// this path prints first says the stream was lost; a successful command that
+// then prints nothing and returns nil says the command produced no output.
+func TestRunCommandFallbackFromID_ChunkFetchFailure(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
jisung-02: Kept parallel—the race this describes needs a swapper running at the same time, and this package has none.

Go pauses a test at `t.Parallel()` and resumes it only after the sequential pass over the binary's tests has finished, so a parallel test never overlaps a serial one. Every test here that swaps `os.Stderr` is serial: the ten `captureStderr` callers in `api/event/event_test.go` (lines 1929-2199, all `TestApplyChunk_*`, `TestRecoverSkippedChunks_*` and `TestDrainRemainingChunks_*`) and the eight `testutil.CaptureOutput` callers in `api/event/sudolistener_test.go`. Not one of them calls `t.Parallel()`. The swap happens in the sequential phase and this test's write happens in the parallel phase, so there is no reader/writer pair on the variable and nothing for a capture pipe to pick up.

That is the repo's stated rule rather than an accident—`CLAUDE.md`, under "Test patterns": a test that swaps `os.Stdout`/`os.Stderr`, `testutil.CaptureOutput` included, does not call `t.Parallel()`. What keeps this safe is the convention on the swapping side, so a guard belongs there if anywhere, not on a test that merely writes.

Verified rather than argued:

```
go test -race -shuffle=on -count=5 ./api/event/
ok  	github.com/alpacax/alpacon-cli/api/event	72.726s
```

junho226 (commented):
The core fix is right and the tests are real, not decorative. One open finding: on `exec logs` the read failure now outranks the command's own failure, which is the opposite of the ordering this same PR chose for `runCommandFallbackFromID`, and it changes an exit code the PR body itself argues must not change. Everything else checks out, including both of the author's rebuttals to Copilot.

## What I verified

Environment: worktree off `origin/main`, `gh pr checkout 435`, HEAD `5ec0af7d040a0fa126255e85fb887e71ff67a86b`, tree clean throughout.

```
$ gofmt -l .
(no output)
$ go vet ./...
(no output)
exit=0    FAIL lines: 0    ok packages: 42
```

Line split, recomputed from `git diff origin/main...HEAD --numstat`: +216/-11 over 6 files is +173 test and +43/-11 non-test (`chunks.go` +24, `event.go` +11/-8, `logs.go` +8/-3). Four fifths of the diff is test code, so the size is not a YAGNI problem. `pickCommandOutput` is not a speculative seam either — it is the only way the five-case table runs without a server, and `ResolveCommandOutput` has both call sites the PR names.

## The RED claims reproduce, exactly

Mutating the shared decision back into swallowing (`if chunkErr != nil { return result, nil }`) reddens both packages:

```
--- FAIL: TestPickCommandOutput/fetch_failure_with_nothing_to_fall_back_on
--- FAIL: TestResolveCommandOutputSurfacesFetchFailure/nothing_to_fall_back_on_is_an_error
--- FAIL: TestRunCommandFallbackFromID_ChunkFetchFailure/a_successful_command_with_nothing_to_print
--- FAIL: TestExecLogsChunkFetchFailureExitsNonZero
    Messages:   helper exited 0; stderr: 
```

Restoring only `runCommandFallbackFromID`'s old body reddens the fallback test with the lie itself:

```
--- FAIL: TestRunCommandFallbackFromID_ChunkFetchFailure/a_successful_command_with_nothing_to_print
    Error:      An error is expected but got nil.
```

Both match the PR body verbatim.

## Both Copilot rebuttals hold

The `%s` claim was wrong and the new assertion pins it. Mutating the verb to `%d`:

```
Error: "Error: failed to read command output: &{%!d(string=fetching page 1 from
/api/events/commands/.../chunks/: server returned an empty error response) 824635564224}..."
should not contain "%!"
--- FAIL: TestExecLogsChunkFetchFailureExitsNonZero
```

The `t.Parallel()` race claim was also wrong. I walked every function in `api/event/event_test.go`, `sudolistener_test.go` and `chunks_test.go`, stripped comment lines, and looked for a body holding both a `captureStderr(`/`CaptureOutput(` call and `t.Parallel()`. There is none, so no stderr swapper ever overlaps the new parallel test. `-race -shuffle=on` ran clean three times.

## Finding: `exec logs` orders the two failures the other way round

The PR body's rule for the fallback path is that the command's own failure outranks an unreadable output, "because it is what carries the remote exit code `exec` propagates as its own; replacing it with a generic read error would change the exit code scripts branch on." `cmd/exec/logs.go` propagates the same remote exit code (`logsCommandOutcome` → `os.Exit(*details.ExitCode)`), but there the read failure wins, and the PR does not say so.

I measured it by driving the real `exec logs` through `runExecLogsHelper` against a server whose chunk endpoint answers 500, once on this branch and once with the three non-test files reverted to `origin/main`:

| command detail | base exit / stderr | head exit / stderr |
|---|---|---|
| `status=completed success=false exit_code=2 result=""` | `2`, empty stderr | `1`, `failed to read command output: ...` |
| `status=stuck result=""` | `1`, `command failed with status: stuck` | `1`, `failed to read command output: ...` |

Two separate consequences:

1. **A remote exit code becomes 1.** A script branching on `exec logs` reads a command that exited 7 as a generic CLI failure whenever the chunk store is unreachable. That is the outcome the fallback path's ordering was written to protect.
2. **`stuck`/`error`/`cancelled` lose their diagnostic for nothing.** `logsCommandOutcome` returns `""` as `stdoutLine` on that branch (`cmd/exec/logs.go:139`), so the output is never printed there — the same reason the PR gives for skipping the fetch on a running command. The fetch buys nothing on those three statuses either, and now it replaces the specific status line with a read error the caller cannot act on.

Neither is a correctness hole in the sense the PR set out to close: both exits are non-zero, so no caller reads an unreadable result as a clean one. That is why this is a comment and not a block. But the ordering should be one decision applied twice, not two, and (2) extends the guard the PR already wrote rather than adding a new one.

## Security

Not the main axis here, and I found nothing. Checked anyway, because the new failure path prints a server-derived string: the message resolves to `fetching page 1 from /api/events/commands/<uuid>/chunks/: <server text>` (`api/pagination.go:48` formats the endpoint path, not the built URL), so no host, no query parameters, no token. It goes to stderr through `utils.CliErrorWithExit` → `cliMessage` → `SanitizeTerminalBlock`, so the server's text is stripped of control characters (#364), and stdout stays empty — the test asserts that. On the fallback path the same error reaches `HandleCommandResult`'s `utils.CliErrorWithExit("%s", err)`, sanitized the same way.

## Other axes, checked and clean

- **No `io.EOF` conflation.** Nothing in the diff touches EOF semantics; `FetchPagesUpTo` terminates on `response.Next == 0 || len(response.Results) == 0`, not on a short read, and returns `nil` on any page error, so `GetCommandOutput` never hands back a silently truncated string.
- **Wrapping.** `%w` on the fallback path keeps the cause reachable; `%s` in `logs.go` is a terminal message nothing unwraps.
- **Partial output.** In `runCommandFallbackFromID` the only case that writes to `out` is the one that returns no read error, so there is no "printed half of it and then failed" state. On `exec logs` the fetch fails before anything is printed.
- **Exit 1 needs no README row.** The table at `README.md:397` already covers a server error under the general failure code.

## Could not verify

`golangci-lint` is not installed on this machine (`which golangci-lint` → not found), so the "0 issues" claim is unchecked. `gofmt -l` and `go vet ./...` are both clean.

cmd/exec/logs.go
```diff
@@ -74,9 +74,14 @@ Run the command again later to check for completion.`,
...
 		}
 
-		// Output is stored as chunks (Result is empty under the streaming
-		// contract); reconstruct it, falling back to Result for legacy commands.
-		if output, oerr := event.GetCommandOutput(alpaconClient, jobID); oerr == nil && output != "" {
+		// A running command's output is not printed here, so skip the fetch and
+		// keep an unreachable chunk store from burying the "still running" line.
+		if !event.IsRunningStatus(details.Status) {
+			output, oerr := event.ResolveCommandOutput(alpaconClient, jobID, details.Result)
+			if oerr != nil {
```
junho226, line 86: This orders the two failures the opposite way from `runCommandFallbackFromID`, which the same PR gave the rule: the command's own failure outranks an unreadable output because it carries the remote exit code. `exec logs` propagates that code too (`logsCommandOutcome` → `os.Exit(*details.ExitCode)`), so exiting here loses it.

Measured against the same 500-on-chunks server, base vs this branch:

- `status=completed success=false exit_code=2 result=""` — base exits `2`; head exits `1` with `failed to read command output`.
- `status=stuck result=""` — base prints `command failed with status: stuck`; head prints the read error instead. Exit stays `1`, but note `logsCommandOutcome` returns `""` as `stdoutLine` on the stuck/error/cancelled branch (line 139), so the output is never printed there — the same argument the comment two lines up makes for skipping the fetch on a running command applies to those three statuses.

Nothing here reports an unreadable result as a clean one, so it is not a blocker. But computing the outcome first and surfacing the read error only when the outcome would otherwise be exit `0` would make one decision serve both paths, and would fold the stuck/error/cancelled case into the guard you already wrote.

jisung-02: You are right on both counts, and both are fixed in e175647.

The read failure now waits for the outcome and speaks only where the outcome would otherwise be exit `0`, so a command that failed on its own keeps the code it carries. That is the same order `runCommandFallbackFromID` takes, so it is one decision honored on both paths rather than two.

The fetch is skipped on `stuck`, `error` and `cancelled` as well. They answer with a status line and an empty `stdoutLine`, exactly as a running command does, so `isStatusOnlyFailure` now gives the guard and the branch in `logsCommandOutcome` a single spelling of the three statuses.

`TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure` pins both against the same 500-on-chunks server you measured with. Reverting the ordering reddens both cases, the stuck one on all three counts:

```
--- FAIL: .../a_remote_failure_keeps_its_own_exit_code
    Error: Not equal: expected: 2, actual: 1
--- FAIL: .../a_stuck_command_keeps_its_status_line
    Error: "...failed to read command output..." does not contain "command failed with status: stuck"
    Error: Should be zero, but was 1
    Messages: the outcome prints no output, so nothing is fetched
```

One thing I did not add: a warning on stderr naming the unread output when the command failed on its own. The polling path stays silent there, and keeping them identical is the point of the change.

On the unverified claim—`golangci-lint run ./...` reports `0 issues.` here, and `gofmt -l .` and `go vet ./...` are clean. The PR body now describes the ordering and the widened guard.

cmd/exec/logs.go
```diff
@@ -74,10 +74,17 @@ Run the command again later to check for completion.`,
 			return
 		}
 
-		// Output is stored as chunks (Result is empty under the streaming
-		// contract); reconstruct it, falling back to Result for legacy commands.
-		if output, oerr := event.GetCommandOutput(alpaconClient, jobID); oerr == nil && output != "" {
-			details.Result = output
+		// Only the statuses whose outcome prints the output are worth a fetch. A
+		// running command and a status-only failure both answer with a line of
+		// their own, which an unreachable chunk store would only bury.
```
junho226: Accuracy note on the comment, not on the code. Since the ordering gate at line 103 landed in the same commit, this guard no longer keeps an unreachable chunk store from burying the status line — the gate does that on its own.

Measured: dropping `!isStatusOnlyFailure(details.Status)` from line 81 and re-running the new test reddens only the request counter, while the status line and the exit code both survive.

```
--- FAIL: TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure/a_stuck_command_keeps_its_status_line
    Error:    Should be zero, but was 1
    Messages: the outcome prints no output, so nothing is fetched
```

So what the guard buys now is the round trip itself, and the latency of a chunk store that hangs rather than errors — still worth keeping, just less than "an unreachable chunk store would only bury." Reads as if it were written before the gate existed.

cmd/exec/logs.go
```diff
@@ -106,6 +120,12 @@ func init() {
 	ExecCmd.AddCommand(logsCmd)
 }
 
+// isStatusOnlyFailure reports the statuses whose outcome is a line about the
+// status alone: the command never reached a result worth printing.
+func isStatusOnlyFailure(status string) bool {
+	return status == "stuck" || status == "error" || status == "cancelled"
```
junho226: Optional, and not for this PR if you would rather leave `event.go` alone.

This is the second spelling of the triple rather than the first: `api/event/event.go:649` carries `case "stuck", "error", "cancelled":` in `errorFromDetails`, which is the polling path's counterpart to `logsCommandOutcome`. Since the whole point of the change is that the two paths decide alike, `api/event/types.go` — where `IsRunningStatus`, `IsAwaitingApprovalStatus` and `IsRejectedStatus` already live — is where a shared predicate would keep them from drifting when a fourth terminal status shows up.

Raising it mainly so the PR body's "the single spelling of those three statuses" is read as scoped to `logs.go`, which it is: two copies collapsed to one there.

junho226 (dismissed):
Delta review over `5ec0af7d..a51443f8`, read in the context of the whole PR. My round at `5ec0af7d` left one open finding and one verification gap:

1. `exec logs` ordered the two failures the opposite way from `runCommandFallbackFromID` — the read failure outranked the command's own failure, so a command that exited `7` became a generic exit `1` whenever the chunk store was unreachable, and a `stuck`/`error`/`cancelled` command lost its status line to a fetch whose output that branch never prints.
2. `golangci-lint` was not on the machine I reviewed from, so the "0 issues" claim was unchecked.

Copilot then raised a third point against the unchanged `pickCommandOutput`: a blanket "any fetch error is fatal" would fail a legacy server that answers 404 on the chunks endpoint for a command that genuinely printed nothing.

All three are closed. Approving. Two small notes inline, neither blocking.

## Both halves of the ordering finding are fixed, and the mutations reproduce

`cmd/exec/logs.go` now holds the fetch error in `readErr`, computes the outcome first, and surfaces the read error only where the outcome would otherwise be exit `0` (line 103) — the order `runCommandFallbackFromID` keeps at `api/event/event.go:690-695`, where `errorFromDetails` is returned before `oerr`. Reverting `logs.go` to its `5ec0af7d` shape reddens both new cases, the stuck one on all three counts:

```
--- FAIL: TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure/a_stuck_command_keeps_its_status_line
    Error: "...failed to read command output..." should not contain "failed to read command output"
    Error: "..." does not contain "command failed with status: stuck"
    Error: Should be zero, but was 1
--- FAIL: .../a_remote_failure_keeps_its_own_exit_code
    Error: Not equal: expected: 2, actual: 1
```

I walked the gate against every branch of `logsCommandOutcome`, not only the two the tests drive. Running and status-only never reach the fetch. The `success=false` remote failure and the unrecognised-status fallback both return a non-zero code, so they keep it and drop the read error — and in every case that drops it, `pickCommandOutput` returned an error only because `result` was empty, so nothing printable goes with it. Neither path has a half-printed state: `runCommandFallbackFromID` writes `output` before it decides, but `pickCommandOutput` returns `""` on the error path, so it never prints part of the output and then fails.

## The 404 narrowing is wired end to end, and it is narrow

The tag survives the whole chain, which I read rather than assumed: `roundTrip` calls `withStatus(parseAPIError(...), resp.StatusCode)` (`client/client.go:439`), `withStatus` attaches the code to `*apiError` or wraps in `*statusError` (line 679-689), `FetchPagesUpTo` wraps with `%w` (`api/pagination.go:48`), and `utils.HTTPStatusCode` unwraps until it finds a `statusCoder` (`utils/error.go:134-141`). `TestResolveCommandOutputSurfacesFetchFailure` pins that wiring through an `httptest` server rather than only the pure table, which is the right split. Dropping the clause reddens both:

```
--- FAIL: TestPickCommandOutput/a_404_leaves_an_empty_legacy_result_standing
--- FAIL: TestResolveCommandOutputSurfacesFetchFailure/a_server_without_the_chunk_endpoint_is_not_a_failure
```

The part worth checking is whether a *current* server can 404 that path, because a 404 that does not mean "no such endpoint" would re-open the exact swallow this PR exists to close. On alpacon-server the chunks reader is an `@action(detail=True)` on the same `CommandViewSet` as the detail read (`events/api/views.py:381`) and resolves through `self.get_object()`; that viewset's `get_queryset` (line 163) varies only `select_related` by action, never the filter. Both CLI call sites read the command detail through that viewset first and fetch chunks only after it succeeded, so the sole 404 left on a current server is the row disappearing between the two requests — and there the detail snapshot already in hand is the honest answer. The endpoint's own refusals are not 404s: the requester/admin gate raises `ValidationError(code=PERMISSION_DENIED)`, a 400, so a principal who can see the command but not its chunks still gets a read failure and a non-zero exit rather than a clean empty one. A transport failure carries no status at all, so `HTTPStatusCode` answers `0` and it stays fatal. The narrowing holds.

## Consistency across the readers of command output

`ResolveCommandOutput` has exactly two callers — `cmd/exec/logs.go:82` and `api/event/event.go:683` — and both now take the same order. The remaining chunk readers are the streaming path's (`api/event/event.go:142`, `438`, `524`, `582`), which this PR does not touch and which stay warn-and-continue: `drainRemainingChunks` prints `failed to fetch trailing chunks (from seq N): ...; output may be incomplete` and returns, so a stream whose tail cannot be fetched still ends on the command's own outcome. That is a different shape from the bug fixed here — the chunks that did arrive are already on stdout and the gap is named on stderr, rather than an empty stdout reported as success — so I am not asking for it in this PR. Worth recording that it is the one reader the new rule does not reach.

## Verification

Worktree off `origin/main`, `gh pr checkout 435`, HEAD `a51443f848ba0b98b46b078fa9366ff99c9e63b7`, tree clean before and after every mutation.

```
$ gofmt -l .
(no output)
$ go build ./...
(no output)
$ go vet ./...
(no output)
$ golangci-lint run ./...
0 issues.
exit=0    ok packages: 42    FAIL lines: 0
```

That closes the gap I left last round: `golangci-lint` was installed under `~/go/bin` and not on `PATH`, which is why I reported it missing. `0 issues.` and the 42-package count both match the PR body.

The `t.Parallel()` thread Copilot re-raised at this head is answered by the repo's own rule rather than by argument — `CLAUDE.md`, *Test patterns*: the constraint sits on the side that swaps `os.Stdout`/`os.Stderr`, and serial tests run to completion before the parallel ones start, so a parallel test that merely writes to stderr has no reader to race. The suite is green under `-race -shuffle=on`.

## Security

Not an axis this delta moves, and nothing new. The one new decision reads a status code off an error the client had already tagged; it prints nothing it did not print before, and the 404 branch prints strictly less — a `result` the CLI already held. The failure message's sanitizing path is unchanged from what I checked last round.

## YAGNI

The delta is +112/-12 over four files, +28/-9 of it non-test. The one new production predicate, `isStatusOnlyFailure`, replaces a literal triple `logs.go` spelled twice. Nothing speculative.

cmd/exec/logs.go
jisung-02: Right, and the measurement is the part that settles it. The guard was written before the ordering gate existed, so its stated reason was true at the time and stopped being true in the same commit that made the gate.

`0e2c8d2` rewrites the comment to what the guard actually saves—the round trip, and the wait of a chunk store that hangs rather than errors:

```go
// Only the statuses whose outcome prints the output are worth a fetch: the
// rest answer with a line of their own and an empty stdoutLine, so the
// round trip buys nothing and a chunk store that hangs costs the wait.
```

I dropped the claim rather than restating it in weaker form, since the gate already carries its own comment two lines below the fetch. The PR body made the same claim and now says the same thing the comment does.

cmd/exec/logs.go
jisung-02: Taken, and done in `e4dde03` rather than deferred—by the rule of two it is the second copy of a judgment expression, and the reason you give is the reason it should not wait: the whole point of the change is that the two paths decide alike, so a fourth terminal status reaching one spelling and missing the other would split them again.

`event.IsStatusOnlyFailure` now lives in `api/event/types.go` beside `IsRunningStatus` and `IsRejectedStatus`, and both callers read it. A switch case cannot call a predicate, so `errorFromDetails` matches the three in an `if` ahead of its switch—the shape the awaiting and rejected statuses already had there. The messages it builds are unchanged.

`TestIsStatusOnlyFailure` sits beside `TestIsRunningStatus` and pins the triple plus what must stay out of it, `"failed"` above all: it carries the remote exit code and a result, so its output is still fetched and printed. Mutating the predicate to `return false` reddens both packages, which is what says the extraction preserved behavior:

```
--- FAIL: TestRunCommandStreaming_TerminalStatusErrors
--- FAIL: TestIsStatusOnlyFailure
--- FAIL: TestLogsCommandOutcome
--- FAIL: TestLogsCommandOutcomeSanitizesServerText
--- FAIL: TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure
```

Dropping only `"cancelled"` reddens the new test alone.

One thing I left alone: `api/ftp/ftp.go:696` and `:751` test `"stuck" || "error"` without `"cancelled"`. That is a different judgment, and folding them in would change what `alpacon cp` reports, so it is not in this PR and no issue is filed for it.

The PR body no longer scopes that sentence to `logs.go`—it now describes the shared predicate.

cmd/exec/logs.go
```diff
@@ -74,10 +74,17 @@ Run the command again later to check for completion.`,
 			return
 		}
 
-		// Output is stored as chunks (Result is empty under the streaming
-		// contract); reconstruct it, falling back to Result for legacy commands.
-		if output, oerr := event.GetCommandOutput(alpaconClient, jobID); oerr == nil && output != "" {
-			details.Result = output
+		// Only the statuses whose outcome prints the output are worth a fetch: the
+		// rest answer with a line of their own and an empty stdoutLine, so the
+		// round trip buys nothing and a chunk store that hangs costs the wait.
```
junho226: The rewrite is right about the half it was written for, and now undersells the other half of the same condition.

Measured on this head, against a server whose `/chunks/` answers 500:

- Condition as written, `status: "running"`: exit `0`, zero requests to `/chunks/`.
- Only `!event.IsRunningStatus(details.Status)` dropped, same server: exit `1`, one request, stderr `failed to read command output: ...`, and the `command is still running (status: running)` line never prints. `logsCommandOutcome` returns exit `0` for a running status, so the `exitCode == 0 && readErr != nil` gate fires there rather than yielding to the outcome.

So the running half buys the answer, not the round trip. `go test ./cmd/exec/` stays green with it dropped, which is the gap the body already names for the helper-process harness — the comment is the only thing standing in for it. As written, "the round trip buys nothing and a chunk store that hangs costs the wait" reads as true of both halves when it is true of the status-only one, which is the shape of the claim the last round asked about, pointed the other way. A clause naming the running half as load-bearing would keep a later reader from folding the condition on the strength of this comment. Non-blocking.

api/event/event.go
```diff
@@ -621,8 +621,9 @@ func drainRemainingChunks(ac *client.AlpaconClient, cmdID string, lastSeq int, o
 // errorFromDetails maps a terminal command status to an error so unrecognized
 // statuses are not masked as success.
 func errorFromDetails(d EventDetails) error {
-	// A switch case cannot call a predicate, so the two approval statuses are
-	// matched ahead of it—a server-side rename then lands in types.go alone.
+	// A switch case cannot call a predicate, so every status matched by one is
+	// handled ahead of the switch—a server-side rename then lands in types.go
+	// alone.
```
junho226: "every status matched by one is handled ahead of the switch" claims more than holds. `IsRunningStatus` is a predicate in the same `types.go`, and the statuses it matches are deliberately *not* handled ahead of the switch — they fall to `default`, which is what produces `command may still be running`. Naming the four statuses, or "every status this function answers specially", says the same thing without inviting that inference. Non-blocking.

junho226 (dismissed):
Delta review of `a51443f8..e4dde037` — the two commits since the dismissed approval, read against the whole PR.

Where the previous round's two notes stand:

| Previous note | Commit | Status |
|---|---|---|
| `logs.go`, the guard comment credited the guard with keeping an unreachable chunk store from burying the status line | `0e2c8d2` | Rewritten, and the measurement in the body matches mine. The new wording undersells the other half of the same condition — inline below, non-blocking |
| `logs.go`, `isStatusOnlyFailure` was a second spelling of the triple `errorFromDetails` already carried | `e4dde03` | Settled. `event.IsStatusOnlyFailure` in `api/event/types.go`, both callers read it, no private copy left (`grep -rn isStatusOnlyFailure` finds only the exported name) |

## Verification

Run on `e4dde037` in a clean worktree, not taken from the body:

```
$ gofmt -l .
(no output)
$ go build ./...
(no output)
$ go vet ./...
(no output)
$ golangci-lint run ./...        # 2.9.0
0 issues.
exit=0    ok packages: 42    FAIL lines: 0
```

Mutating `IsStatusOnlyFailure` to `return false` reddens both packages, and exactly the tests the body names: `TestRunCommandStreaming_TerminalStatusErrors` (four subtests) in `api/event`, and `TestLogsCommandOutcome`, `TestLogsCommandOutcomeSanitizesServerText`, `TestIsStatusOnlyFailure` and `TestExecLogsChunkFetchFailureYieldsToTheCommandsOwnFailure/a_stuck_command_keeps_its_status_line` in `cmd/exec`. That is what says the extraction preserved behavior rather than only compiling.

## Correctness

The move out of the switch is behavior-preserving. The statuses matched ahead of it are disjoint from the `completed`/`success`/`failed` case, both `fmt.Errorf` strings are copied verbatim, and `default` still claims every status none of the predicates does. `api/ftp/ftp.go:696` and `:751` are correctly left out: they test `stuck || error` without `cancelled`, and a grep for `"stuck"` shows they are the only other sites, so nothing else was left holding a fourth spelling.

## Security

Not an axis this delta touches, with one invariant worth naming because the delta moved a comparison behind a call. `logsCommandOutcome`'s comment skips sanitizing `details.Status` on the strength of "a Status that reached one of these branches equals the literal it was compared against". That still holds — `IsStatusOnlyFailure` is exact equality — but the premise now lives one package away from the comment relying on it. Nothing to change here.

## YAGNI

Rule of two, two callers in two packages, so the extraction is not speculative; the commit message's directive (a new terminal status carrying no printable result goes in the predicate, not in either caller) is the part that makes it pay. `TestIsStatusOnlyFailure` sits in `cmd/exec/logs_test.go` rather than beside the predicate, which matches `TestIsRunningStatus` and the `IsAwaitingPurposeStatus` asserts in `purpose_demand_test.go` — `api/event` has no `types_test.go` to hold it.

Also checked, since it is the one bot thread still open at this head: no test in `api/event` both swaps a std stream and calls `t.Parallel()`. The ten `captureStderr` callers in `event_test.go` and the eight `testutil.CaptureOutput` callers in `sudolistener_test.go` are all serial, so the reply on that thread holds under the rule `CLAUDE.md` states for parallel tests.

Approving. Both inline notes are comment accuracy on delta lines, nothing that blocks.

cmd/exec/logs.go
jisung-02: Right, and the measurement is what settles it. The rewrite explained the status-only half and left the running half reading as the same kind of saving, which it is not.

I reproduced your numbers on this head before touching the comment, driving the real `exec logs` through `runExecLogsHelper` against a chunk store answering 500, with only `!event.IsRunningStatus(details.Status)` dropped:

```
exit=1 chunkReads=1 stdout=""
stderr="Error: failed to read command output: fetching page 1 from
/api/events/commands/a1b2c3d4-5678-abcd-ef01-234567890abc/chunks/:
server returned an empty error response"
```

The `command is still running (status: running)` line never prints, so there the skip produces the answer rather than saving a round trip. `go test ./cmd/exec/` stays green with the half dropped, which is the gap the description already names for the helper-process harness.

`3e2466f` splits the comment into the two reasons:

```go
// Only the statuses whose outcome prints the output are worth a fetch, for
// two different reasons. A status-only failure answers with a line of its
// own and an empty stdoutLine, so the round trip buys nothing and a chunk
// store that hangs costs the wait. A running status needs the skip to stay
// correct: its outcome exits 0, so a read error would take the
// exitCode == 0 gate below and bury the "still running" line.
```

The commit body carries the measurement, and a `Directive` trailer says neither half may be dropped on the strength of the other half's reason. The description now separates the two halves the same way, and its "not covered" section says the running half is what the green package leaves unpinned.

api/event/event.go
jisung-02: Taken. `IsRunningStatus` is exactly the counterexample the wording invited, and it is deliberate that its statuses fall to `default`—that is what produces `command may still be running`, so "every status matched by one is handled ahead of the switch" claimed a rule the function does not keep.

`3f6e136` narrows it to what the function actually covers:

```go
// A switch case cannot call a predicate, so every status this function answers
// specially is handled ahead of the switch—a server-side rename then lands in
// types.go alone.
```

I took your second phrasing rather than the list of four: the list would have to be re-read against the predicates every time one is added, while "answers specially" stays true on its own. The reason the comment exists—that a server-side rename lands in `types.go` alone—is unchanged.

junho226 (approved):
Delta review over `e4dde037..3f6e136e`. Comment-only, 9 insertions and 6 deletions across two files, and it closes both notes the previous round left open. Nothing new.

## Both notes settled

**`cmd/exec/logs.go`** now names the running half as load-bearing rather than as another saved round trip, and the mechanism it states is the one in the file. `logsCommandOutcome` answers a running status with `return "", stderrLine, 0`, and the gate ten lines below the guard is `if exitCode == 0 && readErr != nil`. So a read error on a running command does take that gate and does replace the "still running" line with `failed to read command output` — which is the measurement from last round, now written down where the next reader meets it. The two reasons are separated in the comment, so neither half reads as the other's kind of saving.

**`api/event/event.go`** narrows the claim to what the function keeps. `types.go` declares five `Is*` predicates and `errorFromDetails` calls four of them ahead of the switch (`IsAwaitingPurposeStatus`, `IsAwaitingApprovalStatus`, `IsRejectedStatus`, `IsStatusOnlyFailure`). `IsRunningStatus` is the fifth and is deliberately not among them: its statuses reach `default`, which is what emits `command may still be running`. "every status matched by one" therefore named a rule with a standing counterexample, and "every status this function answers specially" is true as written — the four specially-answered classes are all pre-switch, and a running status gets the generic answer rather than a special one. The rename argument the sentence exists for survives the narrowing.

## Verified at this head

- `gofmt -l .` empty, `go build ./...` and `go vet ./...` clean, `golangci-lint run ./...` → `0 issues.`
- `go test -race -shuffle=on ./...` → 42 packages ok, no failures; `go test -count=1 ./...` re-run separately, exit 0 with no non-ok line
- `api/event` and `cmd/exec` re-run on their own → both ok
- Read `logsCommandOutcome`'s running branch and the `exitCode == 0` gate directly rather than taking the comment's word for either, and enumerated the five predicates against the four call sites

No code changed, so the axes the previous rounds covered need no re-run: the ordering rule, the 404 narrowing and the single-spelling triple are all untouched here.

PR #440, opened 2026-09-09
fix(utils): confine zip extraction to its destination

### LINKED ISSUE #438: Unzip writes through a pre-existing symlink instead of the path it validated
## Description

`Unzip` (`utils/utils.go:386`) validates each archive entry against the destination directory as a string, then writes to that path with a call that resolves symlinks. The two do not agree, so the file the check approved is not always the file that gets written.

The check is lexical:

```go
fpath := filepath.Join(destDir, f.Name)

if !strings.HasPrefix(fpath, destPrefix) {
    return fmt.Errorf("invalid file path: %s", f.Name)
}
```

The write is not:

```go
outFile, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
```

`O_WRONLY|O_CREATE|O_TRUNC` follows a symlink at `fpath`. If a path component under `destDir` is already a symlink pointing elsewhere on disk, the string comparison still passes (the path text is under the destination) while the write lands wherever the link resolves to. `Unzip` returns `nil`, so a caller has no way to tell that an entry went somewhere other than the directory it asked for.

`Unzip` never creates symlinks itself, so the link has to be on disk before extraction starts. That makes this narrow, not unreachable: `alpacon cp -r` extracts into a destination the user names, and re-extracting into a directory that already holds one is an ordinary thing to do.

Callers: `api/ftp/ftp.go:650` and `api/ftp/ftp.go:768`.

## Steps to reproduce

```go
root := t.TempDir()
outside := filepath.Join(root, "outside.txt")
os.WriteFile(outside, []byte("original"), 0600)

dest := filepath.Join(root, "dest")
os.MkdirAll(dest, 0755)
os.Symlink(outside, filepath.Join(dest, "link"))

// a.zip contains one entry named "link" holding "replacement"
err := utils.Unzip(zipPath, dest)
```

## Expected behavior

Either the entry is rejected, or it is written to `dest/link` as a regular file. Nothing outside `dest` changes, and a rejection comes back as an error rather than silently.

## Actual behavior

`Unzip` returns `nil`, `dest/link` is still a symlink, and `root/outside.txt` now holds the archive entry's content:

```
Unzip err = <nil>
expected: "original"
actual  : "replacement"
```

## Suggested fix

Two options, either of which closes the gap:

- Open with `O_NOFOLLOW` in `extractFile` so a symlink at the final component fails the open. `pkg/selfupdate/replace_unix.go:47-50` already discusses this trade-off for the update path and is the nearest precedent in the tree.
- Resolve each path component and confirm the real path is under the destination root before creating or writing.

The first is smaller and covers the final component; the second also covers a link at an intermediate directory. Keep a regression test that extracts over an existing symlink.

---

Found by a Codex audit pass and reproduced against `main` at 694e9e9.

jisung-02:
## Background

Unzip validated the joined path with a string-prefix check and then opened it with plain `os.OpenFile`. A symlink already present under the destination redirected the write outside it, and the prefix check could not see that because it only looks at the name.

## Changes

`Unzip` pins the destination with `os.OpenRoot` and does every mkdir and file write through that root, so the final open is confined whatever the name says.

The new `utils/unzip_path.go` resolves each path component before the write. `resolveZipPath` walks the components with `Lstat` and `Readlink`, follows relative links inside the root, and rejects cycles and escapes. It exists because `os.Root` refuses every absolute symlink outright, including one that points back inside the destination; `zipRelativeTarget` rewrites such a target relative to the root, matching against `destDir` and its `filepath.EvalSymlinks` form so a destination reached through a symlink still resolves. A target that does not match either form is rejected. Link expansion happens before `..` is consumed, because cleaning first can select a different file. The function carries a doc comment stating that contract and recording why it permits an absolute internal symlink at all, since failing closed instead would be about five lines and the choice is otherwise invisible.

A component that does not exist is recorded as such, and a following `..` is refused rather than popped. Without that, a target of `<dest>/missing/../victim` resolved to `<dest>/victim` and the entry overwrote it, where the kernel answers ENOENT. Nothing escaped the destination in that case, but the entry landed on a file the archive never named.

`extractFile` passes `f.Mode().Perm()` rather than `f.Mode()`. `os.Root.OpenFile` rejects a mode carrying non-permission bits outright — `openat raw: unsupported file mode` — so the change is forced by the move to `os.Root`, and it also stops an archive from setting setuid, setgid or sticky on an extracted file, which the old `os.OpenFile` call passed straight through to `open(2)`.

Directories an archive asks for are created 0755 through the new `zipDirPerm` constant, not `os.ModePerm`. 0777 predates this branch, but the branch hands it to `root.MkdirAll` on both extraction paths as well, and a umask is the only thing that was narrowing it. `TestUnzip_DirectoriesUseFixedPerm` pins that: it lowers the umask to 0, so the requested bits survive to disk, and checks the destination, a directory entry the archive asks for at 0777, and the parent of a nested file. `syscall.Umask` is process-wide, so the test is serial and restores what it found, and the file carries a `unix` build tag because Windows has no umask and the Windows leg now runs `./utils`. Its expected mode is the literal 0755 rather than the constant, so widening `zipDirPerm` fails the test instead of moving it.

Entries declared as symlinks in the archive are written as regular files holding their target text, so extraction never creates a link of its own.

The Windows leg of `build-and-test.yaml` now runs `./utils`. `filepath.VolumeName` always returns `""` on Unix, and a target starting with the separator has already taken the `filepath.IsAbs` branch above, so that check in `resolveZipPath` and `zipPathEqual`'s `strings.EqualFold` branch are unreachable there. They guard a security boundary and no CI leg was executing them. One existing case moved with the leg: `TestUnzip_PathTraversalAttack` expected `/etc/passwd` to be rejected, but on Windows an absolute path needs a volume name, so `filepath.IsAbs` is false for a leading slash and the entry is joined onto the destination instead, where it stays confined. That expectation is now Unix-only.

Three more cases moved with it, after the leg's first run said so. `TestUnzip_ResolvesLinkBeforeParentComponent` and both cases of `TestUnzip_RejectsParentAfterNonDirectory` build an absolute symlink target containing `..` and exist to pin what `resolveZipPath` does when it consumes one. Windows normalizes an absolute target when the link is created, so `Readlink` hands back a path with the `..` already resolved and the precondition is simply not there. `skipUnlessParentInLink` reads the link back and skips on that, rather than on `runtime.GOOS`, so the cases stay live wherever the target survives intact.

## Safety

The destination-prefix check in `Unzip` is not redundant. An earlier version of this description said it was, and that was wrong.

It is redundant for confinement only: `os.Root` holds the boundary whatever the check does. It is load-bearing for rejection. `strings.TrimPrefix` is a no-op when the prefix does not match, so without the check `resolveZipPath` receives an absolute path, splits it on the separator, skips the leading empty component, and silently re-roots the entry under the destination. `../evil.txt` would land inside the destination instead of being refused. Removing only that check:

```
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_traversal
--- FAIL: TestUnzip_PathTraversalAttack/multiple_parent_directory_traversal
--- FAIL: TestUnzip_PathTraversalAttack/mixed_path_traversal
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_only
--- FAIL: TestUnzip_PathTraversalAttack/sibling_directory_sharing_the_destination_prefix
--- FAIL: TestUnzip_DirectoryTraversal
```

The check stays exactly as it is, and it is also the shape CodeQL's `go/zipslip` recognizes. Enforcing the invariant locally instead of at a caller check three lines up—`filepath.Rel`, which cannot silently produce an absolute path, or rejecting an absolute `name` on entry to `resolveZipPath`—will be a follow-up, tracked in #444.

Two behaviors are deliberately conservative and fail closed rather than resolve. An absolute symlink target that contains `..` but lands back inside the destination is rejected, because cleaning it before expansion would reintroduce the wrong-file selection described above. `zipPathEqual` folds case only on Windows, so on a case-insensitive macOS volume a target spelled with different case is rejected rather than followed.

Writing through a pre-existing dangling symlink inside the destination creates the missing parents and lands there. This stays inside the destination, and an archive cannot set up that precondition, since its own symlink entries become regular files.

## Testing

```
go test -race -shuffle=on ./...
ok  	github.com/alpacax/alpacon-cli/utils	2.619s
```

```
golangci-lint run ./...
0 issues.
```

The Windows leg cannot be run from this machine. Its first run was on this PR and it failed on the three cases described above; the run after `skipUnlessParentInLink` is green.

`utils/unzip_symlink_test.go` covers the symlink axis. The existing `TestUnzip_ValidFiles`, `TestUnzip_PathTraversalAttack` and `TestUnzip_DirectoryTraversal` cover classic zip slip and still pass; they now build their fixtures through the same `writeUnzipArchive` helper rather than hand-rolling `zip.NewWriter` four times.

`TestUnzip_PathTraversalAttack` now runs its containment check on every case, not only the ones expecting an error. The check was gated on `wantErr`, which left the case this branch turns into a success on Windows writing a file and asserting nothing about where it landed. No entry may leave `extractDir` whether it was rejected or accepted, so the gate came off. Seen red by writing a file into the parent directory inside the subtest: all eight cases fail, `valid_file_with_dots_in_name` and `valid_subdirectory` among them, which is exactly the pair the gate had been skipping.

Every test was seen red against a mutated target, and the table below was re-run against the final diff:

| Mutation | Cases that failed |
|---|---|
| `resolveZipPath` dropped, mkdir and write done with `os.MkdirAll` and `os.OpenFile` on the absolute path | `TestUnzip_ConfinesSymlinks` (9 of 10), `TestUnzip_AllowsAbsoluteInternalSymlinks/dangling_directory` |
| `resolveZipPath` returns its argument unchanged | `TestUnzip_AllowsAbsoluteInternalSymlinks` (5), `TestUnzip_AbsoluteLinksWithDestinationAlias` (2), `TestUnzip_ResolvesLinkBeforeParentComponent` |
| `zipRelativeTarget` allows instead of rejecting | `TestUnzip_ConfinesSymlinks` (4) |
| Parent-escape rejection replaced by `continue` | `TestUnzip_ConfinesSymlinks` (5: the four relative cases and `parent_escape`) |
| Archive symlink entries created as real links | `TestUnzip_MaterializesArchiveSymlinkAsFile` |
| Every symlink rejected | `TestUnzip_AllowsInternalSymlink`, `TestUnzip_ResolvesLinkBeforeParentComponent`, `TestUnzip_AllowsAbsoluteInternalSymlinks` (5), `TestUnzip_AbsoluteLinksWithDestinationAlias` (2) |
| Not-a-directory check dropped | `TestUnzip_RejectsParentAfterNonDirectory/regular_file` |
| Missing-component guard disabled | `TestUnzip_RejectsParentAfterNonDirectory/missing_component` |
| Symlink hop cap dropped | `TestUnzip_ConfinesSymlinks/cycle` hangs until the test timeout |
| `zipDirPerm` set back to 0777 | `TestUnzip_DirectoriesUseFixedPerm`, on all three directories |

Two rows need a note. Under the first mutation `TestUnzip_ConfinesSymlinks/cycle` is the one case that still passes: with `resolveZipPath` gone the kernel walks the loop itself and `os.OpenFile` returns ELOOP, so the test gets its error from a different source. And the last row is the one that does not fail cleanly without its guard: removing the hop cap makes the walk loop forever, so the case ends the run on `panic: test timed out` rather than on an assertion.

Closes #438

### DIFF
--- a/utils/unzip_path.go
+++ b/utils/unzip_path.go
@@ -14,6 +14,7 @@ const maxZipSymlinkHops = 255
 func resolveZipPath(root *os.Root, rootNames []string, name string) (string, error) {
 	pending := strings.Split(name, string(os.PathSeparator))
 	var resolved []string
+	missing := false
 	links := 0
 	for len(pending) > 0 {
 		part := pending[0]
@@ -25,12 +26,18 @@ func resolveZipPath(root *os.Root, rootNames []string, name string) (string, err
 			if len(resolved) == 0 {
 				return "", fmt.Errorf("zip path escapes destination: %s", name)
 			}
+			// Everything after the first absent component is absent too, so the
+			// kernel would stop here rather than let '..' select a sibling.
+			if missing {
+				return "", fmt.Errorf("zip path traverses a missing directory: %s", name)
+			}
 			resolved = resolved[:len(resolved)-1]
 			continue
 		}
 		candidate := filepath.Join(filepath.Join(resolved...), part)
 		info, err := root.Lstat(candidate)
 		if os.IsNotExist(err) {
+			missing = true
 			resolved = append(resolved, part)
 			continue
 		}
@@ -59,6 +66,7 @@ func resolveZipPath(root *os.Root, rootNames []string, name string) (string, err
 				return "", err
 			}
 			resolved = nil
+			missing = false
 		} else if filepath.VolumeName(target) != "" || strings.HasPrefix(target, string(os.PathSeparator)) {
 			return "", fmt.Errorf("zip symlink escapes destination: %s", candidate)
 		}
--- a/utils/unzip_path.go
+++ b/utils/unzip_path.go
@@ -1,0 +1,89 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+func resolveZipPath(root *os.Root, rootNames []string, name string) (string, error) {
+	pending := strings.Split(name, string(os.PathSeparator))
+	var resolved []string
+	links := 0
+	for len(pending) > 0 {
+		part := pending[0]
+		pending = pending[1:]
+		switch part {
+		case "", ".":
+			continue
+		case "..":
+			if len(resolved) == 0 {
+				return "", fmt.Errorf("zip path escapes destination: %s", name)
+			}
+			resolved = resolved[:len(resolved)-1]
+			continue
+		}
+		candidate := filepath.Join(filepath.Join(resolved...), part)
+		info, err := root.Lstat(candidate)
+		if os.IsNotExist(err) {
+			resolved = append(resolved, part)
+			continue
+		}
+		if err != nil {
+			return "", err
+		}
+		if info.Mode()&os.ModeSymlink == 0 {
+			if len(pending) > 0 && !info.IsDir() {
+				return "", fmt.Errorf("zip path component is not a directory: %s", candidate)
+			}
+			resolved = append(resolved, part)
+			continue
+		}
+		links++
+		if links > 255 {
+			return "", fmt.Errorf("too many symlinks in zip path: %s", name)
+		}
+		target, err := root.Readlink(candidate)
+		if err != nil {
+			return "", err
+		}
+		target = filepath.FromSlash(target)
+		if filepath.IsAbs(target) {
+			target, err = zipRelativeTarget(rootNames, target)
+			if err != nil {
+				return "", err
+			}
+			resolved = nil
+		} else if filepath.VolumeName(target) != "" || strings.HasPrefix(target, string(os.PathSeparator)) {
+			return "", fmt.Errorf("zip symlink escapes destination: %s", candidate)
+		}
+		// Expand links before consuming '..'; cleaning first can select a different file.
+		pending = append(strings.Split(target, string(os.PathSeparator)), pending...)
+	}
+	if len(resolved) == 0 {
+		return ".", nil
+	}
+	return filepath.Join(resolved...), nil
+}
+
+func zipRelativeTarget(rootNames []string, target string) (string, error) {
+	for _, name := range rootNames {
+		prefix := strings.TrimRight(name, string(os.PathSeparator)) + string(os.PathSeparator)
+		if zipPathEqual(target, name) {
+			return ".", nil
+		}
+		if len(target) >= len(prefix) && zipPathEqual(target[:len(prefix)], prefix) {
+			return target[len(prefix):], nil
+		}
+	}
+	return "", fmt.Errorf("zip symlink escapes destination: %s", target)
+}
+
+func zipPathEqual(a, b string) bool {
+	if runtime.GOOS == "windows" {
+		return strings.EqualFold(a, b)
+	}
+	return a == b
+}
--- a/utils/unzip_path.go
+++ b/utils/unzip_path.go
@@ -34,7 +34,9 @@ func resolveZipPath(root *os.Root, rootNames []string, name string) (string, err
 			resolved = resolved[:len(resolved)-1]
 			continue
 		}
-		candidate := filepath.Join(filepath.Join(resolved...), part)
+		// resolved holds clean single components and part carries no separator,
+		// so joining the strings is what filepath.Join would return anyway.
+		candidate := strings.Join(append(resolved, part), string(os.PathSeparator))
 		info, err := root.Lstat(candidate)
 		if os.IsNotExist(err) {
 			missing = true
--- a/utils/unzip_path.go
+++ b/utils/unzip_path.go
@@ -8,6 +8,9 @@ import (
 	"strings"
 )
 
+// Same cap filepath.walkSymlinks applies, so a zip path gives up where a plain one does.
+const maxZipSymlinkHops = 255
+
 func resolveZipPath(root *os.Root, rootNames []string, name string) (string, error) {
 	pending := strings.Split(name, string(os.PathSeparator))
 	var resolved []string
@@ -42,7 +45,7 @@ func resolveZipPath(root *os.Root, rootNames []string, name string) (string, err
 			continue
 		}
 		links++
-		if links > 255 {
+		if links > maxZipSymlinkHops {
 			return "", fmt.Errorf("too many symlinks in zip path: %s", name)
 		}
 		target, err := root.Readlink(candidate)
--- a/utils/unzip_path.go
+++ b/utils/unzip_path.go
@@ -11,7 +11,25 @@ import (
 // Same cap filepath.walkSymlinks applies, so a zip path gives up where a plain one does.
 const maxZipSymlinkHops = 255
 
-func resolveZipPath(root *os.Root, rootNames []string, name string) (string, error) {
+// zipRoot is one spelling of the destination, paired with the prefix an absolute
+// symlink target must carry to be inside it. Both are fixed for a whole Unzip call.
+type zipRoot struct {
+	name   string
+	prefix string
+}
+
+func newZipRoots(names ...string) []zipRoot {
+	roots := make([]zipRoot, 0, len(names))
+	for _, name := range names {
+		roots = append(roots, zipRoot{
+			name:   name,
+			prefix: strings.TrimRight(name, string(os.PathSeparator)) + string(os.PathSeparator),
+		})
+	}
+	return roots
+}
+
+func resolveZipPath(root *os.Root, roots []zipRoot, name string) (string, error) {
 	pending := strings.Split(name, string(os.PathSeparator))
 	var resolved []string
 	missing := false
@@ -63,7 +81,7 @@ func resolveZipPath(root *os.Root, rootNames []string, name string) (string, err
 		}
 		target = filepath.FromSlash(target)
 		if filepath.IsAbs(target) {
-			target, err = zipRelativeTarget(rootNames, target)
+			target, err = zipRelativeTarget(roots, target)
 			if err != nil {
 				return "", err
 			}
@@ -81,14 +99,13 @@ func resolveZipPath(root *os.Root, rootNames []string, name string) (string, err
 	return filepath.Join(resolved...), nil
 }
 
-func zipRelativeTarget(rootNames []string, target string) (string, error) {
-	for _, name := range rootNames {
-		prefix := strings.TrimRight(name, string(os.PathSeparator)) + string(os.PathSeparator)
-		if zipPathEqual(target, name) {
+func zipRelativeTarget(roots []zipRoot, target string) (string, error) {
+	for _, r := range roots {
+		if zipPathEqual(target, r.name) {
 			return ".", nil
 		}
-		if len(target) >= len(prefix) && zipPathEqual(target[:len(prefix)], prefix) {
-			return target[len(prefix):], nil
+		if len(target) >= len(r.prefix) && zipPathEqual(target[:len(r.prefix)], r.prefix) {
+			return target[len(r.prefix):], nil
 		}
 	}
 	return "", fmt.Errorf("zip symlink escapes destination: %s", target)
--- a/utils/unzip_path.go
+++ b/utils/unzip_path.go
@@ -29,6 +29,16 @@ func newZipRoots(names ...string) []zipRoot {
 	return roots
 }
 
+// resolveZipPath maps an archive entry name to a path relative to root, with every
+// symlink expanded, and rejects anything that leaves the destination. Confinement
+// does not rest on it: os.Root is the boundary, so a bug here can misplace a file
+// inside the destination but never outside it.
+//
+// os.Root refuses an absolute symlink even when its target is inside the root, so
+// permitting one is what these hops cost; failing closed instead would be about
+// five lines. We resolve and follow the link because the destination is a folder
+// the user chose and may already have populated, and because an archive naming
+// "link" asks for what "link" points at.
 func resolveZipPath(root *os.Root, roots []zipRoot, name string) (string, error) {
 	pending := strings.Split(name, string(os.PathSeparator))
 	var resolved []string
--- a/utils/unzip_symlink_test.go
+++ b/utils/unzip_symlink_test.go
@@ -1,0 +1,110 @@
+package utils
+
+import (
+	"archive/zip"
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUnzip_ConfinesSymlinks(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name   string
+		target string
+		entry  string
+	}{
+		{"file", "../outside/file", "link"},
+		{"directory", "../outside", "link/file"},
+		{"new directory", "../outside", "link/new/"},
+		{"new file", "../outside", "link/new/file"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			dir := t.TempDir()
+			dest := filepath.Join(dir, "dest")
+			outside := filepath.Join(dir, "outside")
+			require.NoError(t, os.Mkdir(dest, 0755))
+			require.NoError(t, os.Mkdir(outside, 0755))
+			require.NoError(t, os.WriteFile(filepath.Join(outside, "file"), []byte("original"), 0600))
+			makeUnzipSymlink(t, filepath.FromSlash(tt.target), filepath.Join(dest, "link"))
+			archive := writeUnzipTestArchive(t, dir, tt.entry)
+
+			require.Error(t, Unzip(archive, dest))
+			content, err := os.ReadFile(filepath.Join(outside, "file"))
+			require.NoError(t, err)
+			assert.Equal(t, "original", string(content))
+			entries, err := os.ReadDir(outside)
+			require.NoError(t, err)
+			assert.Len(t, entries, 1)
+		})
+	}
+}
+
+func TestUnzip_AllowsInternalSymlink(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	dest := filepath.Join(dir, "dest")
+	require.NoError(t, os.MkdirAll(filepath.Join(dest, "real"), 0755))
+	makeUnzipSymlink(t, "real", filepath.Join(dest, "link"))
+	require.NoError(t, Unzip(writeUnzipTestArchive(t, dir, "link/file"), dest))
+	content, err := os.ReadFile(filepath.Join(dest, "real", "file"))
+	require.NoError(t, err)
+	assert.Equal(t, "replacement", string(content))
+}
+
+func TestUnzip_MaterializesArchiveSymlinkAsFile(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	archive := filepath.Join(dir, "test.zip")
+	f, err := os.Create(archive)
+	require.NoError(t, err)
+	zw := zip.NewWriter(f)
+	header := &zip.FileHeader{Name: "link"}
+	header.SetMode(os.ModeSymlink | 0777)
+	w, err := zw.CreateHeader(header)
+	require.NoError(t, err)
+	_, err = w.Write([]byte("../outside"))
+	require.NoError(t, err)
+	require.NoError(t, zw.Close())
+	require.NoError(t, f.Close())
+	dest := filepath.Join(dir, "dest")
+	require.NoError(t, Unzip(archive, dest))
+	info, err := os.Lstat(filepath.Join(dest, "link"))
+	require.NoError(t, err)
+	assert.True(t, info.Mode().IsRegular(), "extracted mode: %v", info.Mode())
+	content, err := os.ReadFile(filepath.Join(dest, "link"))
+	require.NoError(t, err)
+	assert.Equal(t, "../outside", string(content))
+}
+
+func makeUnzipSymlink(t *testing.T, target, link string) {
+	t.Helper()
+	err := os.Symlink(target, link)
+	if err != nil && runtime.GOOS == "windows" {
+		t.Skipf("symlinks unavailable: %v", err)
+	}
+	require.NoError(t, err)
+}
+
+func writeUnzipTestArchive(t *testing.T, dir, entry string) string {
+	t.Helper()
+	archive := filepath.Join(dir, "test.zip")
+	f, err := os.Create(archive)
+	require.NoError(t, err)
+	zw := zip.NewWriter(f)
+	w, err := zw.Create(entry)
+	require.NoError(t, err)
+	if entry[len(entry)-1] != '/' {
+		_, err = w.Write([]byte("replacement"))
+		require.NoError(t, err)
+	}
+	require.NoError(t, zw.Close())
+	require.NoError(t, f.Close())
+	return archive
+}
--- a/utils/unzip_symlink_test.go
+++ b/utils/unzip_symlink_test.go
@@ -84,7 +84,7 @@ func TestUnzip_ResolvesLinkBeforeParentComponent(t *testing.T) {
 	makeUnzipSymlink(t, filepath.Join("real", "subdir"), filepath.Join(dest, "alias"))
 	target := dest + string(os.PathSeparator) + "alias" + string(os.PathSeparator) + ".." + string(os.PathSeparator) + "file"
 	makeUnzipSymlink(t, target, filepath.Join(dest, "link"))
-	requireParentInLink(t, filepath.Join(dest, "link"))
+	skipUnlessParentInLink(t, filepath.Join(dest, "link"))
 	require.NoError(t, Unzip(writeUnzipTestArchive(t, dir, "link"), dest))
 	data, err := os.ReadFile(filepath.Join(dest, "real", "file"))
 	require.NoError(t, err)
@@ -196,7 +196,7 @@ func TestUnzip_RejectsParentAfterNonDirectory(t *testing.T) {
 			}
 			sep := string(os.PathSeparator)
 			makeUnzipSymlink(t, dest+sep+tc.middle+sep+".."+sep+"victim", filepath.Join(dest, "link"))
-			requireParentInLink(t, filepath.Join(dest, "link"))
+			skipUnlessParentInLink(t, filepath.Join(dest, "link"))
 			require.Error(t, Unzip(writeUnzipTestArchive(t, dir, "link"), dest))
 			data, err := os.ReadFile(filepath.Join(dest, "victim"))
 			require.NoError(t, err)
@@ -214,10 +214,10 @@ func makeUnzipSymlink(t *testing.T, target, link string) {
 	require.NoError(t, err)
 }
 
-// requireParentInLink skips a case whose subject is a ".." left for resolveZipPath
+// skipUnlessParentInLink skips a case whose subject is a ".." left for resolveZipPath
 // to consume. Windows normalizes an absolute target when the link is created, so
 // the ".." is already gone by the time Readlink sees it and the case cannot be built.
-func requireParentInLink(t *testing.T, link string) {
+func skipUnlessParentInLink(t *testing.T, link string) {
 	t.Helper()
 	target, err := os.Readlink(link)
 	require.NoError(t, err)
--- a/utils/unzip_symlink_test.go
+++ b/utils/unzip_symlink_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -239,7 +240,7 @@ func writeUnzipTestArchive(t *testing.T, dir, entry string) string {
 	return writeUnzipArchive(t, dir, func(zw *zip.Writer) {
 		w, err := zw.Create(entry)
 		require.NoError(t, err)
-		if entry[len(entry)-1] != '/' {
+		if !strings.HasSuffix(entry, "/") {
 			_, err = w.Write([]byte("replacement"))
 			require.NoError(t, err)
 		}
--- a/utils/unzip_symlink_test.go
+++ b/utils/unzip_symlink_test.go
@@ -108,3 +108,142 @@ func writeUnzipTestArchive(t *testing.T, dir, entry string) string {
 	require.NoError(t, f.Close())
 	return archive
 }
+
+func TestUnzip_AllowsAbsoluteInternalSymlinks(t *testing.T) {
+	t.Parallel()
+	for _, tc := range []struct {
+		name, target, entry, want string
+		directory                 bool
+	}{
+		{"directory", "real", "link/new/file", "real/new/file", false},
+		{"directory entry", "real", "link/new/", "real/new", true},
+		{"existing file", "real/file", "link", "real/file", false},
+		{"dangling file", "real/missing", "link", "real/missing", false},
+		{"dangling directory", "missing", "link/new/file", "missing/new/file", false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			dir := t.TempDir()
+			dest := filepath.Join(dir, "dest")
+			require.NoError(t, os.MkdirAll(filepath.Join(dest, "real"), 0755))
+			require.NoError(t, os.WriteFile(filepath.Join(dest, "real", "file"), []byte("original"), 0600))
+			target := filepath.Join(dest, filepath.FromSlash(tc.target))
+			makeUnzipSymlink(t, target, filepath.Join(dest, "link"))
+			require.NoError(t, Unzip(writeUnzipTestArchive(t, dir, tc.entry), dest))
+			path := filepath.Join(dest, filepath.FromSlash(tc.want))
+			if tc.directory {
+				info, err := os.Stat(path)
+				require.NoError(t, err)
+				assert.True(t, info.IsDir(), "mode: %v", info.Mode())
+			} else {
+				data, err := os.ReadFile(path)
+				require.NoError(t, err)
+				assert.Equal(t, "replacement", string(data))
+			}
+			got, err := os.Readlink(filepath.Join(dest, "link"))
+			require.NoError(t, err)
+			assert.Equal(t, target, got)
+		})
+	}
+}
+
+func TestUnzip_ResolvesLinkBeforeParentComponent(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	dest := filepath.Join(dir, "dest")
+	require.NoError(t, os.MkdirAll(filepath.Join(dest, "real", "subdir"), 0755))
+	makeUnzipSymlink(t, filepath.Join("real", "subdir"), filepath.Join(dest, "alias"))
+	target := dest + string(os.PathSeparator) + "alias" + string(os.PathSeparator) + ".." + string(os.PathSeparator) + "file"
+	makeUnzipSymlink(t, target, filepath.Join(dest, "link"))
+	require.NoError(t, Unzip(writeUnzipTestArchive(t, dir, "link"), dest))
+	data, err := os.ReadFile(filepath.Join(dest, "real", "file"))
+	require.NoError(t, err)
+	assert.Equal(t, "replacement", string(data))
+	_, err = os.Stat(filepath.Join(dest, "file"))
+	assert.ErrorIs(t, err, os.ErrNotExist)
+}
+
+func TestUnzip_AbsoluteLinksWithDestinationAlias(t *testing.T) {
+	t.Parallel()
+	for _, canonical := range []bool{false, true} {
+		name := "original spelling"
+		if canonical {
+			name = "canonical spelling"
+		}
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			dir := t.TempDir()
+			real := filepath.Join(dir, "real")
+			dest := filepath.Join(dir, "dest")
+			require.NoError(t, os.MkdirAll(filepath.Join(real, "files"), 0755))
+			makeUnzipSymlink(t, real, dest)
+			target := filepath.Join(dest, "files")
+			if canonical {
+				var err error
+				target, err = filepath.EvalSymlinks(target)
+				require.NoError(t, err)
+			}
+			makeUnzipSymlink(t, target, filepath.Join(real, "link"))
+			require.NoError(t, Unzip(writeUnzipTestArchive(t, dir, "link/file"), dest))
+			data, err := os.ReadFile(filepath.Join(real, "files", "file"))
+			require.NoError(t, err)
+			assert.Equal(t, "replacement", string(data))
+		})
+	}
+}
+
+func TestUnzip_RejectsAbsoluteEscapesAndCycles(t *testing.T) {
+	t.Parallel()
+	for _, name := range []string{"external file", "external directory", "sibling prefix", "nested escape", "cycle", "parent escape"} {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			dir := t.TempDir()
+			dest := filepath.Join(dir, "dest")
+			outside := filepath.Join(dir, "dest-other")
+			require.NoError(t, os.Mkdir(dest, 0755))
+			require.NoError(t, os.Mkdir(outside, 0755))
+			require.NoError(t, os.WriteFile(filepath.Join(outside, "file"), []byte("original"), 0600))
+			entry := "link/file"
+			target := outside
+			switch name {
+			case "external file":
+				target = filepath.Join(outside, "file")
+				entry = "link"
+			case "external directory":
+				target = dir
+				entry = "link/dest-other/file"
+			case "nested escape":
+				makeUnzipSymlink(t, outside, filepath.Join(dest, "second"))
+				target = filepath.Join(dest, "second")
+			case "cycle":
+				target = filepath.Join(dest, "link")
+			case "parent escape":
+				target = dest + string(os.PathSeparator) + ".." + string(os.PathSeparator) + "dest-other"
+			}
+			makeUnzipSymlink(t, target, filepath.Join(dest, "link"))
+			require.Error(t, Unzip(writeUnzipTestArchive(t, dir, entry), dest))
+			data, err := os.ReadFile(filepath.Join(outside, "file"))
+			require.NoError(t, err)
+			assert.Equal(t, "original", string(data))
+			entries, err := os.ReadDir(outside)
+			require.NoError(t, err)
+			assert.Len(t, entries, 1)
+		})
+	}
+}
+
+func TestUnzip_RejectsParentAfterRegularFile(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	dest := filepath.Join(dir, "dest")
+	require.NoError(t, os.Mkdir(dest, 0755))
+	for _, name := range []string{"plain", "victim"} {
+		require.NoError(t, os.WriteFile(filepath.Join(dest, name), []byte("original"), 0600))
+	}
+	target := dest + string(os.PathSeparator) + "plain" + string(os.PathSeparator) + ".." + string(os.PathSeparator) + "victim"
+	makeUnzipSymlink(t, target, filepath.Join(dest, "link"))
+	require.Error(t, Unzip(writeUnzipTestArchive(t, dir, "link"), dest))
+	data, err := os.ReadFile(filepath.Join(dest, "victim"))
+	require.NoError(t, err)
+	assert.Equal(t, "original", string(data))
+}
--- a/utils/unzip_symlink_test.go
+++ b/utils/unzip_symlink_test.go
@@ -12,41 +12,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestUnzip_ConfinesSymlinks(t *testing.T) {
-	t.Parallel()
-	tests := []struct {
-		name   string
-		target string
-		entry  string
-	}{
-		{"file", "../outside/file", "link"},
-		{"directory", "../outside", "link/file"},
-		{"new directory", "../outside", "link/new/"},
-		{"new file", "../outside", "link/new/file"},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			t.Parallel()
-			dir := t.TempDir()
-			dest := filepath.Join(dir, "dest")
-			outside := filepath.Join(dir, "outside")
-			require.NoError(t, os.Mkdir(dest, 0755))
-			require.NoError(t, os.Mkdir(outside, 0755))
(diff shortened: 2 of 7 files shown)


geunwoonoh (commented):
## Review

Real fix for a real bug, and the layering is right: `os.Root` is the hard boundary, so even a bug in `resolveZipPath` can only misplace a file *inside* the destination, never outside it. I verified the three standard-library behaviors the design depends on rather than taking them from the description:

```
OpenFile(abslink/a.txt) -> ERR  openat abslink/a.txt: path escapes from parent   # absolute symlink, target inside root
OpenFile(rellink/b.txt) -> OK                                                    # relative symlink, inside root
OpenFile with ModeSetuid -> openat mode.txt: unsupported file mode
MkdirAll(".") -> <nil>
```

So the justification for `unzip_path.go` holds, and `.Perm()` is genuinely forced. Findings below are one factual correction to the PR's own safety reasoning, a coverage gap, and cleanup.

### Critical issues (0)

None. Confinement holds on every path I traced.

### Warnings (4)

**1. `utils/utils.go:432` — the destination-prefix check is *not* redundant, and the PR body says it is**

The Safety section states the check "is now redundant with both `resolveZipPath` and `os.Root`, and is kept because CodeQL's `go/zipslip` recognizes only that shape." That is not accurate, and the inaccuracy is the risk: it invites a future cleanup that removes the check.

`strings.TrimPrefix(fpath, destPrefix)` is a no-op when the prefix does not match, so it hands `resolveZipPath` an *absolute* path. `resolveZipPath` splits on the separator and skips the leading empty component, silently re-rooting it under the destination. I removed only that check in a throwaway copy:

```
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_traversal
        Expected error for malicious path "../evil.txt", but got none
--- FAIL: TestUnzip_PathTraversalAttack/multiple_parent_directory_traversal
--- FAIL: TestUnzip_PathTraversalAttack/mixed_path_traversal
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_only
--- FAIL: TestUnzip_PathTraversalAttack/sibling_directory_sharing_the_destination_prefix
--- FAIL: TestUnzip_DirectoryTraversal
```

And the entry lands here instead of being rejected:

```
LANDED AT: dest/var/folders/cf/q8w.../T/TestProbeRelocation.../001/evil.txt
```

Nothing escapes—`os.Root` holds—but a rejected entry becomes a silently relocated one. The check is load-bearing for *rejection*, redundant only for *confinement*. Two things worth doing: correct that paragraph, and enforce the invariant locally instead of relying on a caller check twenty lines up—`filepath.Rel(destDir, fpath)` returns an error or a `..`-prefixed path on mismatch, or `resolveZipPath` could reject an absolute `name` on entry.

**2. `utils/unzip_path.go:70` — Windows-only branches with zero coverage anywhere**

```go
} else if filepath.VolumeName(target) != "" || strings.HasPrefix(target, string(os.PathSeparator)) {
```

On Unix `filepath.VolumeName` always returns `""`, and any target starting with the separator already took the `IsAbs` branch above—so this `else if` is unreachable on Unix. Same for `zipPathEqual`'s `strings.EqualFold` branch. They only fire on Windows, and `.github/workflows/build-and-test.yaml:33` runs the Windows leg against `./pkg/selfupdate/... ./cmd` only, so `utils` never executes there. New security-boundary logic with no test that ever runs it. Adding `./utils` to the Windows leg is the fix; if the older `utils` tests assert Unix file modes and would fail, that is worth knowing now rather than at the next Windows regression.

**3. `utils/utils.go:400` — the empty-archive guard carries no reason**

```go
if len(r.File) == 0 {
    return nil
}
```

This preserves the old behavior that an empty archive creates no destination directory—without it, the new upfront `os.MkdirAll(destDir)` would start creating one. Nothing in the code says that, so it reads as a pointless micro-optimization and the next reader deletes it. One line of comment. (It also sits after `destPrefix` is computed; moving it above is free.)

**4. `utils/utils.go:456` — `extractFile`'s parameter is still named `fpath`**

```go
func extractFile(root *os.Root, fpath string, f *zip.File) (err error) {
```

It is no longer a full path—the caller passes `relativePath`, root-relative. The old name now says the opposite of what the value is, on the one function that performs the write. `name` or `relPath`.

### Suggestions (6)

**5. `utils/unzip_path.go` — is permitting absolute internal symlinks worth this?**

The 100 lines exist to allow what the standard library deliberately refuses (confirmed in the probe above). Failing closed instead is roughly five lines: `os.Root` alone, absolute symlinks rejected. What is lost is extracting into a destination that already contains an absolute symlink the archive also names—for `alpacon cp` that destination is a user-chosen download folder, so the case is narrow. Related policy question: extraction currently *follows* a pre-existing symlink and overwrites its target (`TestUnzip_AllowsAbsoluteInternalSymlinks/existing file`); refusing to write through any pre-existing symlink is the more conservative reading of "the archive named `link`, so write `link`." Not asking for a rewrite—asking that the trade-off be recorded in the code, since the current comments explain the mechanism but not the choice.

**6. `utils/unzip_path.go:37` — quadratic path building in the component loop**

```go
candidate := filepath.Join(filepath.Join(resolved...), part)
```

Two `Join`s (each with a `Clean`) per component, and the inner one re-joins the whole prefix every iteration—O(depth^2) in path length. `resolved` holds only clean single components and `part` has no separator, so `strings.Join` over the components is equivalent with a much smaller constant. Bounded and not a hot path, but it is free.

**7. `utils/unzip_path.go:84` — `prefix` rebuilt on every hop**

`strings.TrimRight(name, sep) + sep` allocates per rootName per absolute-symlink resolution, and `rootNames` is fixed for the whole call. Precompute the prefixes once in `Unzip` and pass those.

**8. Per-entry resolution cost**

`resolveZipPath` adds an `Lstat` per component per entry, and `root.MkdirAll(filepath.Dir(...))` still runs for every file entry (pre-existing). For an archive with many files under one deep directory that is a few syscalls per file that resolve to the same answer. A one-entry memo of the last resolved directory would cover the common case; worth it only if `cp -r` on large trees shows up, so treat as a note.

**9. `resolveZipPath` has no doc comment**

The tidy bar deletes comments the code already says—this is the opposite case. A security-critical function whose contract is non-obvious ("returns a root-relative, symlink-free path; rejects anything leaving the root") should state it. The inline comments present are good and should stay.

**10. Test hygiene—target file `utils/unzip_test.go`**

Two things. `writeUnzipArchive`/`writeUnzipTestArchive` are exactly what `unzip_test.go` hand-rolls three separate times with `zip.NewWriter`; the new helpers are the better version and the old tests should use them. And `TestUnzip_ConfinesSymlinks` and `TestUnzip_RejectsAbsoluteEscapesAndCycles` share a verbatim setup-and-assert block (mkdir dest + outside, write `outside/file`, then assert error, content unchanged, `Len(entries, 1)`), differing only in relative versus absolute targets—one table with a `target func(...)` covers both. Neither is a duplicate case, so no coverage is lost either way.

**11. `utils/unzip_symlink_test.go:242` — `entry[len(entry)-1]` panics on an empty entry**

Only literals reach it today. `strings.HasSuffix(entry, "/")` costs nothing and does not depend on that staying true.

**12. `os.ModePerm` for extracted directories**

0777 (pre-existing, now on `root.MkdirAll` too). Umask usually saves it, but this is the PR that hardens extraction of untrusted archives, so 0755 is the consistent choice.

### Highlights

- Layering `os.Root` under `resolveZipPath` means a resolution bug can misplace a file but cannot escape. That is the right shape for a security fix, and it is what makes the remaining findings non-critical.
- `f.Mode().Perm()` closes a real hole: the old `os.OpenFile(fpath, ..., f.Mode())` passed setuid/setgid/sticky through to `open(2)`. Framed in the description as forced by `os.Root`, but it stands on its own.
- Expanding symlinks before consuming `..` matches kernel semantics, and `TestUnzip_ResolvesLinkBeforeParentComponent` pins exactly the case where cleaning first picks the wrong file.
- The mutation table in the PR body—every new test seen red against a specific mutation, including the note that the cycle case fails by timeout rather than assertion—is the standard other security PRs should be held to.
- Tests follow the repo conventions: `t.Parallel()` first, `t.Run` case tables, testify helpers over predicate wrappers, and `assert.True` messages carrying the operand.

---

**Verdict**: COMMENTED

**Checklist**:
- [x] Tests included and passing (CI green; Windows leg does not run `utils`—Warning 2)
- [x] No security vulnerabilities (confinement verified; Warning 1 is a rejection-vs-relocation issue inside the destination)
- [ ] Documentation updated if needed (PR body's Safety section needs the Warning 1 correction)

utils/utils.go
```diff
@@ -410,28 +429,32 @@ func Unzip(src string, dest string) error {
 		if !strings.HasPrefix(fpath, destPrefix) {
 			return fmt.Errorf("invalid file path: %s", f.Name)
 		}
+		relativePath, err := resolveZipPath(root, rootNames, strings.TrimPrefix(fpath, destPrefix))
```
geunwoonoh: Warning: this line depends on the `HasPrefix` check three lines up having passed, and the PR body says that check is now redundant. It is not, and that mismatch is the risk — it invites a future cleanup that removes it.

`strings.TrimPrefix` is a no-op when the prefix does not match, so it hands `resolveZipPath` an **absolute** path. `resolveZipPath` then splits on the separator and skips the leading empty component, silently re-rooting the entry under the destination. Removing only the `HasPrefix` check in a throwaway copy:

```
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_traversal
        Expected error for malicious path "../evil.txt", but got none
--- FAIL: TestUnzip_PathTraversalAttack/multiple_parent_directory_traversal
--- FAIL: TestUnzip_PathTraversalAttack/mixed_path_traversal
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_only
--- FAIL: TestUnzip_PathTraversalAttack/sibling_directory_sharing_the_destination_prefix
--- FAIL: TestUnzip_DirectoryTraversal
```

and the entry lands at `dest/var/folders/.../T/.../evil.txt` instead of being rejected. Nothing escapes (`os.Root` holds), but a rejected entry becomes a silently relocated one — the check is load-bearing for *rejection*, redundant only for *confinement*.

Worth enforcing the invariant here rather than at a distant caller check: `filepath.Rel` returns an error or a `..`-prefixed path on mismatch, so it cannot silently produce an absolute path. (Alternatively, have `resolveZipPath` reject an absolute `name` on entry.)

```suggestion
		relativePath, err := resolveZipPath(root, rootNames, strings.TrimPrefix(fpath, destPrefix))
```

utils/unzip_path.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		target = filepath.FromSlash(target)
+		if filepath.IsAbs(target) {
+			target, err = zipRelativeTarget(rootNames, target)
+			if err != nil {
+				return "", err
+			}
+			resolved = nil
+			missing = false
+		} else if filepath.VolumeName(target) != "" || strings.HasPrefix(target, string(os.PathSeparator)) {
```
geunwoonoh, line 100: Warning: this branch is unreachable on Unix and never executed by CI anywhere.

On Unix `filepath.VolumeName` always returns `""`, and any target starting with the separator already took the `filepath.IsAbs` branch above — so the whole `else if` only fires on Windows. Same for `zipPathEqual`'s `strings.EqualFold` branch below.

`.github/workflows/build-and-test.yaml:33` runs the Windows leg against `./pkg/selfupdate/... ./cmd` only, so `utils` never executes there. That leaves new security-boundary logic with no test that ever runs it, on either platform.

Adding `./utils` to the Windows matrix leg is the fix. If some of the older `utils` tests assert Unix file modes and would fail there, that is worth finding out now rather than at the next Windows regression.

utils/utils.go
```diff
@@ -397,6 +397,25 @@ func Unzip(src string, dest string) error {
 		return err
 	}
 	destPrefix := strings.TrimSuffix(destDir, string(os.PathSeparator)) + string(os.PathSeparator)
+	if len(r.File) == 0 {
```
geunwoonoh: Warning: this guard encodes a deliberate behavior decision with nothing saying so.

It preserves the old behavior that an empty archive creates no destination directory — without it, the new upfront `os.MkdirAll(destDir)` on the next line would start creating one. As written it reads as a pointless micro-optimization, and the next reader deletes it.

It also sits after `destPrefix` is computed; moving it above that is free.

```suggestion
	if len(r.File) == 0 {
		// An empty archive creates nothing, as before the destination was pinned:
		// the MkdirAll below would otherwise leave an empty directory behind.
		return nil
	}
```

utils/utils.go
```diff
@@ -410,28 +429,32 @@ func Unzip(src string, dest string) error {
...
+		if err := extractFile(root, relativePath, f); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func extractFile(fpath string, f *zip.File) (err error) {
-	outFile, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
+func extractFile(root *os.Root, fpath string, f *zip.File) (err error) {
```
geunwoonoh: Warning: the parameter is no longer a full path — the caller passes `relativePath`, which is root-relative. The name now says the opposite of what the value is, on the one function that performs the write.

```suggestion
func extractFile(root *os.Root, name string, f *zip.File) (err error) {
	outFile, err := root.OpenFile(name, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode().Perm())
```

utils/unzip_path.go
```diff
@@ -0,0 +1,100 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+// Same cap filepath.walkSymlinks applies, so a zip path gives up where a plain one does.
+const maxZipSymlinkHops = 255
+
+func resolveZipPath(root *os.Root, rootNames []string, name string) (string, error) {
```
geunwoonoh: Suggestion: two things on this function.

**Contract.** No doc comment. The tidy bar deletes comments the code already says, but this is the opposite case — a security-critical function whose contract is not obvious from the signature. Worth stating: returns a root-relative, symlink-free path; rejects anything that leaves the root. (The inline comments inside the body are good and should stay.)

**Design.** These 100 lines exist to permit what the standard library deliberately refuses. I confirmed the premise rather than assuming it:

```
OpenFile(abslink/a.txt) -> ERR  openat abslink/a.txt: path escapes from parent   # absolute symlink, target inside root
OpenFile(rellink/b.txt) -> OK                                                    # relative symlink, inside root
```

So the justification is real. But failing closed instead is roughly five lines — `os.Root` alone, absolute symlinks rejected — and what is lost is extracting into a destination that already contains an absolute symlink the archive also names. For `alpacon cp` that destination is a user-chosen download folder, so the case is narrow.

Related policy question: extraction currently *follows* a pre-existing symlink and overwrites its target (`TestUnzip_AllowsAbsoluteInternalSymlinks/existing file`). Refusing to write through any pre-existing symlink is the more conservative reading of "the archive named `link`, so write `link`."

Not asking for a rewrite — the confinement is sound either way. Asking that the trade-off be recorded in the code, since the comments explain the mechanism but not the choice.

utils/unzip_path.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+			// Everything after the first absent component is absent too, so the
+			// kernel would stop here rather than let '..' select a sibling.
+			if missing {
+				return "", fmt.Errorf("zip path traverses a missing directory: %s", name)
+			}
+			resolved = resolved[:len(resolved)-1]
+			continue
+		}
+		candidate := filepath.Join(filepath.Join(resolved...), part)
```
geunwoonoh: Suggestion: two `filepath.Join`s (each with a `Clean`) per component, and the inner one re-joins the whole prefix every iteration — O(depth^2) in path length.

`resolved` holds only clean single components (symlinks are never appended, and `.`/`..`/`""` are consumed by the switch above), and `part` came from a `Split` so it contains no separator. `strings.Join` is therefore equivalent with a much smaller constant.

Bounded and not a hot path, but it is free.

```suggestion
		candidate := strings.Join(append(resolved, part), string(os.PathSeparator))
```

utils/unzip_path.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if len(resolved) == 0 {
+		return ".", nil
+	}
+	return filepath.Join(resolved...), nil
+}
+
+func zipRelativeTarget(rootNames []string, target string) (string, error) {
+	for _, name := range rootNames {
+		prefix := strings.TrimRight(name, string(os.PathSeparator)) + string(os.PathSeparator)
```
geunwoonoh: Suggestion: `prefix` is rebuilt on every call, for every rootName, on every absolute-symlink hop — but `rootNames` is fixed for the whole `Unzip` call.

Precomputing the prefixes once in `Unzip` and passing those instead of `rootNames` removes the allocation entirely.

utils/utils.go
```diff
@@ -410,28 +429,32 @@ func Unzip(src string, dest string) error {
...
-			err := os.MkdirAll(fpath, os.ModePerm)
+			err := root.MkdirAll(relativePath, os.ModePerm)
 			if err != nil {
 				return err
 			}
 			continue
 		}
 
-		if err := os.MkdirAll(filepath.Dir(fpath), os.ModePerm); err != nil {
+		if err := root.MkdirAll(filepath.Dir(relativePath), os.ModePerm); err != nil {
```
geunwoonoh: Suggestion (note, not a change request): `resolveZipPath` now adds an `Lstat` per path component per entry, and this `MkdirAll` still runs for every file entry (pre-existing). For an archive with many files under one deep directory, that is several syscalls per file all resolving to the same answer.

A one-entry memo of the last resolved directory would cover the common case. Worth doing only if `cp -r` on large trees actually shows up in practice — flagging it so the cost is known, not asking for it now.

utils/utils.go
```diff
@@ -397,6 +397,25 @@ func Unzip(src string, dest string) error {
 		return err
 	}
 	destPrefix := strings.TrimSuffix(destDir, string(os.PathSeparator)) + string(os.PathSeparator)
+	if len(r.File) == 0 {
+		return nil
+	}
+	if err := os.MkdirAll(destDir, os.ModePerm); err != nil {
```
geunwoonoh: Suggestion: `os.ModePerm` is 0777 (pre-existing, and now on `root.MkdirAll` at lines 438 and 445 too). Umask usually saves it, but this is the PR that hardens extraction of untrusted archives, so 0755 is the consistent choice — especially next to the `f.Mode().Perm()` change, which exists precisely to stop an archive dictating permission bits.

utils/unzip_symlink_test.go
```diff
@@ -0,0 +1,271 @@
...
+		header := &zip.FileHeader{Name: entry}
+		header.SetMode(os.ModeSymlink | 0777)
+		w, err := zw.CreateHeader(header)
+		require.NoError(t, err)
+		_, err = w.Write([]byte(target))
+		require.NoError(t, err)
+	})
+}
+
+func writeUnzipArchive(t *testing.T, dir string, addEntry func(*zip.Writer)) string {
```
geunwoonoh, line 254: Suggestion: these helpers are the better version of what `utils/unzip_test.go` hand-rolls three separate times with a raw `zip.NewWriter` (`TestUnzip_ValidFiles`, `TestUnzip_PathTraversalAttack`, `TestUnzip_RelativeDestination`). Worth pointing the existing tests at `writeUnzipArchive` so there is one way to build a fixture archive in this package.

Not blocking — the duplication is in the old file, not this one.

utils/unzip_symlink_test.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	require.NoError(t, err)
+}
+
+func writeUnzipTestArchive(t *testing.T, dir, entry string) string {
+	t.Helper()
+	return writeUnzipArchive(t, dir, func(zw *zip.Writer) {
+		w, err := zw.Create(entry)
+		require.NoError(t, err)
+		if entry[len(entry)-1] != '/' {
```
geunwoonoh: Suggestion: `entry[len(entry)-1]` panics on an empty `entry`. Only literals reach it today, and `strings.HasSuffix` costs nothing and does not depend on that staying true.

```suggestion
		if !strings.HasSuffix(entry, "/") {
```

utils/unzip_symlink_test.go
```diff
@@ -0,0 +1,271 @@
...
+			makeUnzipSymlink(t, target, filepath.Join(real, "link"))
+			require.NoError(t, Unzip(writeUnzipTestArchive(t, dir, "link/file"), dest))
+			data, err := os.ReadFile(filepath.Join(real, "files", "file"))
+			require.NoError(t, err)
+			assert.Equal(t, "replacement", string(data))
+		})
+	}
+}
+
+func TestUnzip_RejectsAbsoluteEscapesAndCycles(t *testing.T) {
```
geunwoonoh: Suggestion: this test and `TestUnzip_ConfinesSymlinks` share a verbatim setup-and-assert block — mkdir dest and outside, write `outside/file`, then assert the error, the unchanged content, and `Len(entries, 1)`. The only real difference is relative versus absolute link targets, which the `target func(...)` field here already expresses.

One table covers both. No coverage is lost: none of the cases are duplicates of each other, only the scaffolding is.

utils/utils.go
jisung-02: Not applied in code, by decision; the description is what changed. You are right on the substance, and I reproduced it rather than taking it from the comment: removing only the `HasPrefix` check gives the same six failures you quoted, with the entry relocated under the destination instead of refused.

```
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_traversal
--- FAIL: TestUnzip_PathTraversalAttack/multiple_parent_directory_traversal
--- FAIL: TestUnzip_PathTraversalAttack/mixed_path_traversal
--- FAIL: TestUnzip_PathTraversalAttack/parent_directory_only
--- FAIL: TestUnzip_PathTraversalAttack/sibling_directory_sharing_the_destination_prefix
--- FAIL: TestUnzip_DirectoryTraversal
```

The Safety section claimed the check was redundant. It now says it is redundant for confinement only and load-bearing for rejection, quotes that output, and states that the check stays as it is. So the paragraph no longer invites the cleanup that would break it, which was the actual risk you named.

Enforcing the invariant next to the use rather than at a caller check twenty lines up—`filepath.Rel`, or rejecting an absolute `name` on entry to `resolveZipPath`—will be a follow-up rather than part of this PR. No issue is filed for it yet.

utils/unzip_path.go
jisung-02: Fixed in 57c184d. You are right that the branch is unreachable on Unix and that no leg was executing it, which is the part that matters for a security boundary.

The Windows leg now runs `./utils`. The older tests turned out to be ready for it: `requireUnixModes` already skips the file-mode assertions and `makeUnzipSymlink` skips when symlink creation is refused. One case was not. `TestUnzip_PathTraversalAttack` expected `/etc/passwd` to be rejected, but on Windows an absolute path needs a volume name, so `filepath.IsAbs` is false for a leading slash and the entry is joined onto the destination, where it stays confined. That case is now `wantErr: runtime.GOOS != "windows"` with the reason inline.

I cannot run the Windows leg from this machine, so its first real run is this PR's CI.

utils/utils.go
jisung-02: Fixed in 2cae821. You are right that the guard encodes a decision and said nothing about it, which is exactly how it gets deleted.

```go
if len(r.File) == 0 {
	// An empty archive creates nothing, as it did before the destination was
	// pinned: the MkdirAll below would otherwise leave a directory behind.
	return nil
}
```

Also moved above `destPrefix`, which it never needed.

utils/utils.go
jisung-02: Fixed in cad0270. The parameter now reads `name`. You are right that it was the worst place for a misleading name, since it is the value the write goes to.

utils/unzip_path.go
jisung-02: Both parts addressed in af14331, though the design is unchanged.

The doc comment states the contract you asked for and adds the bound that makes it readable: `os.Root` is the boundary, so a bug in this function can misplace a file inside the destination but never outside it.

On the design, I kept the resolution and recorded why, since that was the ask. The comment says the hops exist only to permit an absolute symlink that `os.Root` refuses outright, that failing closed instead would be about five lines, and why we follow the link anyway: the destination is a folder the user chose and may already have populated, and an archive naming `link` asks for what `link` points at. The commit body also carries `Rejected:` and a `Directive:` saying that refusing to write through a pre-existing symlink is a policy change rather than a cleanup, so whoever reopens it finds the reasoning rather than re-deriving it.

Your probe output matches mine; I ran the same three checks before touching anything.

utils/unzip_path.go
jisung-02: Fixed in 14d7de7, exactly as suggested. Your reasoning about why the components are already clean is what made it safe to swap, so the comment now says it in the code:

```go
// resolved holds clean single components and part carries no separator,
// so joining the strings is what filepath.Join would return anyway.
candidate := strings.Join(append(resolved, part), string(os.PathSeparator))
```

utils/unzip_path.go
jisung-02: Fixed in 3d36b3c, slightly differently from the suggestion. Passing the prefixes alone would lose the `name` that `zipPathEqual(target, name)` needs, so each spelling is paired with its prefix instead:

```go
type zipRoot struct {
	name   string
	prefix string
}
```

`newZipRoots` builds the pairs once in `Unzip`, and `zipRelativeTarget` takes `[]zipRoot`. The per-hop allocation is gone and the comparison is unmoved.

utils/utils.go
jisung-02: Not applied, following your own framing of it as a note. Recording the measurement so the cost is on the record rather than only in this thread: `resolveZipPath` does one `Lstat` per component per entry, and `root.MkdirAll(filepath.Dir(...))` still runs for every file entry.

For what it is worth, a memo would be sound here rather than merely fast: archive-declared symlinks are materialized as regular files, so no new link appears mid-extraction and a resolved parent cannot go stale within one call. I left it out because nothing has measured `cp -r` on a large tree yet, and caching on a security path for an unmeasured gain is the wrong trade to make blind. Happy to add it the moment there is a number.

utils/utils.go
jisung-02: Fixed in a7536ce. You are right that hardening extraction while leaving directories world-writable is inconsistent, and that a umask was the only thing narrowing it.

0755 now, through a named constant since three sites needed the same answer:

```go
// zipDirPerm is what a directory an archive asks for is created as. An archive
// dictates no permission bits here, the same reason extractFile takes only Perm().
zipDirPerm = 0o755
```

utils/unzip_symlink_test.go
jisung-02: Fixed in cec02de. Four rather than three: `TestUnzip_DirectoryTraversal` hand-rolled the same fixture, so it moved too. The package now has one way to build a zip fixture.

utils/unzip_symlink_test.go
jisung-02: Fixed in a4243c1. `strings.HasSuffix(entry, "/")` now, for the reason you gave: it does not depend on only literals reaching the helper.

utils/unzip_symlink_test.go
jisung-02: Fixed in b8e0944. Merged into `TestUnzip_ConfinesSymlinks`, since that name covers relative and absolute alike; the four relative cases became `target func(...)` entries and the outside directory took the single name `dest-other`.

Seen red rather than assumed: replacing the parent-escape rejection in `resolveZipPath` with a `continue` fails all four merged cases plus `parent_escape`.

```
--- FAIL: TestUnzip_ConfinesSymlinks/relative_file
--- FAIL: TestUnzip_ConfinesSymlinks/relative_directory
--- FAIL: TestUnzip_ConfinesSymlinks/relative_new_directory
--- FAIL: TestUnzip_ConfinesSymlinks/relative_new_file
--- FAIL: TestUnzip_ConfinesSymlinks/parent_escape
```

utils/unzip_path.go
jisung-02: Following up on my own reply, since the leg has now run: it failed, and your "worth finding out now" was the whole point.

Three cases broke, all of them ones that build an absolute symlink target containing `..`: `TestUnzip_ResolvesLinkBeforeParentComponent`, and both cases of `TestUnzip_RejectsParentAfterNonDirectory`. Windows resolves an absolute target when the link is created, so the `..` is gone before `Readlink` ever sees it and the precondition each case exists to exercise is not there. The two rejection cases got `nil` where they expect an error; the link-before-parent case found nothing at `dest/real/file`.

f531e12 adds `requireParentInLink`, which reads the link back and skips on what is actually missing rather than on `runtime.GOOS`, so the cases stay live wherever the target survives. The runner then printed exactly what the guard claims:

```
unzip_symlink_test.go:87: platform resolved ".." out of the link target: "C:\Users\RUNNER~1\...\001\dest\file"
--- SKIP: TestUnzip_ResolvesLinkBeforeParentComponent (0.01s)
unzip_symlink_test.go:199: platform resolved ".." out of the link target: "C:\Users\RUNNER~1\...\001\dest\victim"
    --- SKIP: TestUnzip_RejectsParentAfterNonDirectory/missing_component (0.00s)
    --- SKIP: TestUnzip_RejectsParentAfterNonDirectory/regular_file (0.00s)
```

`dest\file` and `dest\victim` with no `..` left in them, which turns the diagnosis from my guess into the runner's own output. Everything else in `utils` ran on Windows, symlink confinement included: `TestUnzip_ConfinesSymlinks` passed all ten cases there.

`test (windows-latest, ./pkg/selfupdate/... ./cmd ./utils)` is green, and so is the rest of the run.

utils/unzip_test.go
```diff
@@ -87,7 +66,9 @@ func TestUnzip_PathTraversalAttack(t *testing.T) {
 		{
 			name:     "absolute path unix",
 			filename: "/etc/passwd",
-			wantErr:  true,
+			// Windows reads a leading slash as rooted but not absolute, so the entry
+			// is joined onto the destination rather than rejected. It stays confined.
+			wantErr: runtime.GOOS != "windows",
```
geunwoonoh, line 71: Warning: the diagnosis here is right and the behavior is fine — it is the same relocation-not-rejection trade-off the Safety section already accepts for the prefix check, and `os.Root` still holds the boundary. The problem is the assertion.

The "no files created outside extract directory" block below runs only `if tt.wantErr`. On Windows this case now writes a file and checks nothing about where it went, so the comment's own claim — "It stays confined" — is the part left unverified. Before this PR the case was `wantErr: true` on every platform and the check ran; the change narrows an existing assertion at exactly the point where new behavior was introduced.

The containment check holds for every case in this table, not just the failing ones — no valid entry should land outside `extractDir` either. Ungating it covers this for free:

```suggestion
			wantErr: runtime.GOOS != "windows",
```

(the edit is on the block below, not this line: drop the `if tt.wantErr {` wrapper around the `os.ReadDir(parentDir)` loop so it runs for every case.)

utils/utils.go
```diff
@@ -25,6 +25,10 @@ import (
 const (
 	// DefaultApprovalWaitTimeout is the shared --wait default so exec, work-session create, and event wait match; not a ceiling (--wait-approval exceeds it), preserving the old 30 × 10s window.
 	DefaultApprovalWaitTimeout = 5 * time.Minute
+
+	// zipDirPerm is what a directory an archive asks for is created as. An archive
+	// dictates no permission bits here, the same reason extractFile takes only Perm().
+	zipDirPerm = 0o755
```
geunwoonoh, line 31: Suggestion: nothing asserts this. 0777 to 0755 is a deliberate security change and a user-visible one — anyone running `umask 0` loses group-write on extracted directories — so it is worth a test that fails if someone widens it back.

`requireUnixModes` already exists in `utils_test.go` for exactly this kind of assertion, so the test is a few lines: extract an archive with a directory entry, `os.Stat` it, assert `Perm() == 0o755`.

The comment itself is good — it ties the choice back to `extractFile` taking only `Perm()`, which is the reason a reader would want.

utils/unzip_symlink_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil && runtime.GOOS == "windows" {
+		t.Skipf("symlinks unavailable: %v", err)
+	}
+	require.NoError(t, err)
+}
+
+// requireParentInLink skips a case whose subject is a ".." left for resolveZipPath
+// to consume. Windows normalizes an absolute target when the link is created, so
+// the ".." is already gone by the time Readlink sees it and the case cannot be built.
+func requireParentInLink(t *testing.T, link string) {
```
geunwoonoh: Suggestion: the guard itself is the right design — keying off the observed precondition rather than `runtime.GOOS` means it stays silent wherever the case is buildable, and it reports what it actually saw. That is what makes it a precondition check instead of a platform opt-out.

Only the name is off: `require*` means "fail if not" everywhere else in the suite, and this skips. `skipUnlessParentInLink` says what it does.

```suggestion
// skipUnlessParentInLink skips a case whose subject is a ".." left for resolveZipPath
```

geunwoonoh (dismissed):
## Follow-up review

Verified all 12 findings from the previous review against the file contents at `f531e12`, not the commit messages. Ten are applied, two are deliberately deferred with reasoning I accept. Three new items below, none blocking.

| # | Finding | Status |
|---|---|---|
| W1 | prefix check is not redundant | **Code unchanged, by decision** — description corrected, follow-up declared |
| W2 | Windows CI gap | Applied — `./utils` on the Windows leg, green |
| W3 | empty-archive guard uncommented | Applied — commented and moved above `destPrefix` |
| W4 | `extractFile`'s `fpath` param | Applied — now `name` |
| S5 | design decision unrecorded | Applied — `resolveZipPath` doc comment |
| S6 | quadratic path join | Applied — `strings.Join`, with the reasoning inline |
| S7 | `prefix` rebuilt per hop | Applied — `zipRoot` pairs each spelling with its prefix |
| S8 | per-entry resolution memo | Deferred — no measurement yet |
| S9 | no doc comment | Applied |
| S10 | test hygiene | Applied — one fixture helper, escape tables merged |
| S11 | `entry[len(entry)-1]` | Applied — `strings.HasSuffix` |
| S12 | 0777 directories | Applied — `zipDirPerm = 0o755` |

Two things I checked rather than took on trust.

**The test merge lost nothing.** `TestUnzip_ConfinesSymlinks` now carries all four original relative cases plus all six absolute ones, ten in one table, with the same three assertions per case. The merge also strengthened the first four: the sibling directory is now named `dest-other`, so those cases additionally exercise the prefix-confusion axis that only `sibling prefix` covered before.

**The Windows skip is not a neutered test.** `requireParentInLink` keys off the observed precondition — it reads the link back and skips only when the `..` is genuinely gone from the target — rather than off `runtime.GOOS`. So it stays silent on every platform where the case is buildable, and reports what it saw when it is not. The commit body records that it was confirmed both ways on darwin. That is the right shape; a `GOOS` check would have been the wrong one.

I also re-derived the `append` in `strings.Join(append(resolved, part), sep)`, since that change came from my suggestion. It is safe: nothing else aliases `resolved`'s backing array, every read is bounded by `len`, and the scratch slot past `len` is overwritten by the next real append. Equivalent to the old double `filepath.Join` because the switch above guarantees no component is `""`, `.`, `..`, or separator-bearing.

### Critical issues (0)

None.

### Warnings (1)

**1. `utils/unzip_test.go:71` — the Windows concession is the one case whose containment nothing asserts**

```go
name:     "absolute path unix",
filename: "/etc/passwd",
// Windows reads a leading slash as rooted but not absolute, so the entry
// is joined onto the destination rather than rejected. It stays confined.
wantErr: runtime.GOOS != "windows",
```

The diagnosis is right and the behavior is fine — it is the same relocation-not-rejection trade-off W1 already accepts, and `os.Root` still holds the boundary. The problem is the assertion. The "no files created outside extract directory" block at line 118 runs only `if tt.wantErr`, so on Windows this case now writes a file and checks nothing about where it went. Before this PR it was `wantErr: true` everywhere and the check ran; the change narrows an existing assertion at precisely the point where new behavior was introduced, and the comment's own claim — "It stays confined" — is the part left unverified.

The containment check holds for every case in this table, not just the failing ones. Ungating it covers this for free.

### Suggestions (2)

**2. `utils/utils.go:31` — `zipDirPerm` has no test**

0777 to 0755 is a deliberate security change and a user-visible one (anyone running umask 0 loses group-write on extracted directories). Nothing asserts it. `requireUnixModes` already exists in `utils_test.go` for exactly this kind of assertion.

**3. `utils/unzip_symlink_test.go:220` — `requireParentInLink` skips, but is named like a testify `require*`**

`require*` means "fail if not". This skips. `skipUnlessParentInLink` says what it does.

### On the one finding not applied

W1 is the right call. What I flagged was not the `TrimPrefix` line itself — confinement never rested on it — but that the description called a load-bearing check redundant, which is how such a check gets deleted. The Safety section now says the opposite, quotes the six failures, and states the check stays. The risk is closed; the local-invariant refactor (`filepath.Rel`, or rejecting an absolute `name` on entry) is genuinely follow-up-shaped.

One ask attached to that: the description says "No issue is filed for it yet." A follow-up with no issue behind it is a follow-up that does not happen. Worth filing before merge so it survives this PR.

### Highlights

- Every response reproduced the finding before acting on it rather than taking it from the comment, including re-running my probe independently.
- Both deferrals argue their case instead of going quiet, and S8's reasoning is better than my own framing was — it points out a memo would be *sound* here, not merely fast, because archive symlinks are materialized as regular files so no resolved parent can go stale mid-call, and then declines it anyway for want of a measurement. Declining to cache on a security path for an unmeasured gain is the correct instinct.
- `af14331` records what was rejected and why, so the "should we just fail closed?" question is answerable from the tree instead of re-derived.
- Turning the Windows leg on surfaced a real platform difference within one CI run, and it was handled by narrowing the precondition rather than the assertion.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (10/10 checks green, including the new `windows-latest ./utils` leg)
- [x] No security vulnerabilities (confinement verified; Warning 1 is an unasserted test claim, not a defect)
- [x] Documentation updated if needed (Safety section corrected)

utils/unzip_test.go
jisung-02: Applied in aec93bf. You were right that the gate is what left the claim unverified: the comment says the entry stays confined and the one check that could show it was the one not running.

The `if tt.wantErr {` wrapper is gone, so the `os.ReadDir(parentDir)` loop runs for every case. Seen red by writing a file into the parent directory inside the subtest -- all eight cases fail, including `valid_file_with_dots_in_name` and `valid_subdirectory`, which is the pair the gate had been skipping.

`wantErr: runtime.GOOS != "windows"` on the line itself is unchanged, since the behavior it describes is the one W1 already accepts.

utils/utils.go
jisung-02: Applied in 22558e9, as `TestUnzip_DirectoriesUseFixedPerm` in the new `utils/unzip_perm_unix_test.go`.

One thing the test had to handle beyond what you sketched: under the usual 022 umask, 0777 and 0755 both land as 0755 on disk, so a test that does not control the umask passes either way. It lowers the umask to 0 first, which makes `syscall.Umask` process-wide state -- so the test is serial and restores what it found, and the file carries a `unix` build tag rather than calling `skipUnlessUnixModes`, since Windows has no umask at all and the Windows leg now runs `./utils`.

It checks all three directories `Unzip` creates: the destination, a directory entry the archive asks for at 0777, and the parent of a nested file. The expected mode is the literal `0o755`, not `zipDirPerm` -- reading the constant would make the test follow a regression instead of catching it. Seen red by setting `zipDirPerm` back to 0777: all three come back 0x1ff against the expected 0x1ed.

utils/unzip_symlink_test.go
jisung-02: Applied in 66dd834, with one addition: `requireUnixModes` in `utils_test.go` skips as well and got the same rename, so both are now `skipUnless*`.

Renaming only `requireParentInLink` would have left the package spelling one behavior two ways, and `requireUnixModes` is the older of the two -- it is where the misleading name came from rather than a precedent for keeping it. Six call sites moved with it.

geunwoonoh (approved):
## Follow-up review

Three commits, three findings, all applied and all verified against the patch contents.

| # | Finding | Status |
|---|---|---|
| W1 | containment gated on `tt.wantErr` | `aec93bf` — gate removed, runs for all 8 cases |
| S2 | `zipDirPerm` untested | `22558e9` — `TestUnzip_DirectoriesUseFixedPerm` |
| S3 | `require*` helpers that skip | `66dd834` — both renamed `skipUnless*` |

The one piece of genuinely new logic is the perm test, so I read it rather than trusting the commit message:

- **`//go:build unix` is the right constraint and it is required.** A `_unix.go` filename suffix carries no implicit build constraint — only GOOS names do — so without the explicit tag this would compile on Windows and fail there. The tag is present.
- **Serial, correctly.** `syscall.Umask` is process-wide, so `t.Parallel()` is rightly absent and the comment above the test says why. Go runs all serial tests to completion before any parallel body resumes within a package binary, so the lowered umask cannot leak into a neighbor.
- **It covers all three `zipDirPerm` call sites**: the destination itself, an archive directory entry that asks for 0777, and the parent of a nested file.
- **It asserts the literal `0o755`, not `zipDirPerm`.** Reading the constant would make the test track a regression instead of catching it. This is the detail that decides whether the test is worth having.

The umask handling is a correction to my own suggestion, not just an implementation of it: under the usual 022 umask both 0777 and 0755 land as 0755 on disk, so the test I sketched would have passed against the un-hardened code. Lowering the umask to 0 is what gives it teeth, and it was found by writing the test rather than by reasoning about it.

On W1's containment ungating: the check now holds for every case because no entry — rejected or accepted — may land outside `extractDir`, and `os.MkdirAll(destDir, …)` runs before the loop so `extract` exists even on the cases that error out. Seen red by planting a file in the parent directory: all eight cases fail, including the two valid ones the gate had been skipping.

On S3: the rename reached `requireUnixModes` in `utils/utils_test.go`, which predates this branch. That is scope the finding did not ask for and it is the right call — renaming only the newer helper would have left the package spelling one behavior two ways. Verified no stale references remain: seven `skipUnlessUnixModes` call sites, three `skipUnlessParentInLink`, zero of either old name.

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (0)

None.

### Highlights

- Turning the Windows leg on paid for itself inside one CI run, and the follow-up reply quoted the runner's own output — `"C:\Users\RUNNER~1\...\dest\file"` with no `..` left in it — which converts the platform diagnosis from a hypothesis into an observation. `TestUnzip_ConfinesSymlinks` passed all ten cases on Windows, which is the assurance the whole `unzip_path.go` branch was missing two rounds ago.
- Every one of the three commits records the mutation it was seen red against, with the actual values (`0x1ff` against `0x1ed`), not just a claim that it was checked.
- The perm test found a hole in my suggestion instead of implementing it literally.
- Issue #444 is filed for the `resolveZipPath` invariant and the description now links it, so the one finding deliberately left unapplied is tracked rather than trusted to memory.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (10/10 green, including `windows-latest ./utils`)
- [x] Documentation updated if needed (Safety section corrected, #444 linked)

PR #441, opened 2026-09-09
fix(iam): handle missing membership on delete

### LINKED ISSUE #437: DeleteMember indexes the first membership without checking the list is non-empty
## Description

`DeleteMember` (`api/iam/iam.go:190-208`) fetches the membership list for a (user, group) pair and then indexes the first element without checking that the list has one:

```go
var memberDetails []MemberDetailResponse
err = json.Unmarshal(responseBody, &memberDetails)
if err != nil {
    return err
}

_, err = ac.SendDeleteRequest(utils.BuildURL(membershipURL, memberDetails[0].ID, nil))
```

When the membership does not exist the server answers with an empty list, `memberDetails` has length 0, and the process panics on `memberDetails[0]`. The function returns `error`, so a caller is written to handle the failure, but a panic never reaches it: `cmd/iam/member_delete.go` gets no error to report and the CLI dies with a Go stack trace instead of a message.

This is the only unguarded first-element access in `api/`. Every sibling checks the list first:

| Site | Guard |
|---|---|
| `api/auth/auth.go:191` | `response.Count == 0` |
| `api/rbac/rbac.go:73` | `len(roles) == 0` |
| `api/packages/packages.go:85,95` | `response.Count == 0` |
| `api/server/server.go:120` | `response.Count == 0` |
| `api/iam/iam.go:236,273` | `response.Count == 0` |
| `api/event/event.go:246` | `len(cmdResponse) == 0` |
| `api/iam/iam.go:208` | none |

`GetUserIDByName`, four lines further down the same file, is the model to follow: it returns `errors.New("no user found with the given name")` rather than indexing into an empty list.

## Steps to reproduce

Point the client at a server that answers the membership list with `[]`:

```go
ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    w.Header().Set("Content-Type", "application/json")
    switch r.URL.Path {
    case "/api/iam/groups/":
        _, _ = w.Write([]byte(`{"count":1,"results":[{"id":"g-1","name":"devs"}]}`))
    case "/api/iam/users/":
        _, _ = w.Write([]byte(`{"count":1,"results":[{"id":"u-1","username":"alice"}]}`))
    default:
        _, _ = w.Write([]byte(`[]`))
    }
}))
defer ts.Close()

ac := &client.AlpaconClient{BaseURL: ts.URL, HTTPClient: ts.Client()}
err := DeleteMember(ac, MemberDeleteRequest{Group: "devs", User: "alice"})
assert.Error(t, err)
```

Through the CLI this is `alpacon group member delete` for a user who is not in the group.

## Expected behavior

`DeleteMember` returns a domain error such as `no membership found for the given user and group`, and the command reports it through `utils.CliErrorWithExit`.

## Actual behavior

```
panic: runtime error: index out of range [0] with length 0

github.com/alpacax/alpacon-cli/api/iam.DeleteMember(...)
	api/iam/iam.go:208
```

## Suggested fix

Check `len(memberDetails) == 0` and return an error before the index. Keep an empty-list regression test alongside it.

---

Found by a Codex audit pass and reproduced against `main` at 694e9e9.

jisung-02:
## Summary

`DeleteMember` indexed `memberDetails[0]` without checking whether the list had any elements, so deleting a user who has no membership in the group caused `panic: runtime error: index out of range [0] with length 0`. It now checks the list length first and returns the error `no membership found for the given user and group`.

## Changes

- Added a length guard in `api/iam/iam.go` before the `memberDetails[0]` index in `DeleteMember`.
- Added a `TestDeleteMember` table in `api/iam/iam_test.go` covering an existing membership (deletes successfully), an empty list (`[]`), and a null response, all against `httptest`, without issuing a DELETE request for the missing-membership cases.

## Testing

```
ok  	github.com/alpacax/alpacon-cli/api/iam	2.809s
```

```
0 issues.
```

The new test was confirmed to panic against the pre-fix code (reverting the guard reproduces `panic: runtime error: index out of range [0] with length 0`).

Closes #437

### DIFF
--- a/api/iam/iam.go
+++ b/api/iam/iam.go
@@ -204,6 +204,9 @@ func DeleteMember(ac *client.AlpaconClient, memberDeleteRequest MemberDeleteRequ
 	if err != nil {
 		return err
 	}
+	if len(memberDetails) == 0 {
+		return errors.New("no membership found for the given user and group")
+	}
 
 	_, err = ac.SendDeleteRequest(utils.BuildURL(membershipURL, memberDetails[0].ID, nil))
 	if err != nil {
--- a/api/iam/iam_test.go
+++ b/api/iam/iam_test.go
@@ -397,3 +397,56 @@ func TestAddMember(t *testing.T) {
 		t.Error("membership POST was not called")
 	}
 }
+
+func TestDeleteMember(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name       string
+		response   string
+		wantErr    string
+		wantDelete int32
+	}{
+		{"existing membership", `[{"id":"membership-1"}]`, "", 1},
+		{"empty list", `[]`, "no membership found for the given user and group", 0},
+		{"null list", `null`, "no membership found for the given user and group", 0},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			var deleteCount atomic.Int32
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				switch {
+				case r.Method == http.MethodGet && r.URL.Path == groupURL:
+					assert.Equal(t, "admins", r.URL.Query().Get("name"))
+					_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"group-1","name":"admins"}]}`))
+				case r.Method == http.MethodGet && r.URL.Path == userURL:
+					assert.Equal(t, "alice", r.URL.Query().Get("username"))
+					_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"user-1","username":"alice"}]}`))
+				case r.Method == http.MethodGet && r.URL.Path == membershipURL:
+					assert.Equal(t, "group-1", r.URL.Query().Get("group"))
+					assert.Equal(t, "user-1", r.URL.Query().Get("user"))
+					_, _ = w.Write([]byte(tt.response))
+				case r.Method == http.MethodDelete:
+					deleteCount.Add(1)
+					assert.Equal(t, membershipURL+"membership-1/", r.URL.Path)
+					w.WriteHeader(http.StatusNoContent)
+				default:
+					t.Errorf("unexpected request: %s %s", r.Method, r.URL.Path)
+					w.WriteHeader(http.StatusNotFound)
+				}
+			}))
+			defer ts.Close()
+
+			ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+			err := DeleteMember(ac, MemberDeleteRequest{Group: "admins", User: "alice"})
+			if tt.wantErr != "" {
+				assert.EqualError(t, err, tt.wantErr)
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, tt.wantDelete, deleteCount.Load())
+		})
+	}
+}
--- a/api/iam/iam_test.go
+++ b/api/iam/iam_test.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"sync/atomic"
 	"testing"
 
@@ -13,6 +12,13 @@ import (
 	"github.com/alpacax/alpacon-cli/client"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+)
+
+const (
+	stubGroupName = "admins"
+	stubGroupID   = "group-uuid-111"
+	stubUserName  = "alice"
+	stubUserID    = "user-uuid-222"
 )
 
 func TestGetUserList_Pagination(t *testing.T) {
@@ -354,49 +360,51 @@ func TestInviteUser(t *testing.T) {
 	}
 }
 
+// newIAMStubServer answers the group and user lookups every membership test makes,
+// and hands every other request to membershipHandler.
+func newIAMStubServer(t *testing.T, membershipHandler http.HandlerFunc) *httptest.Server {
+	t.Helper()
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		switch {
+		case r.Method == http.MethodGet && r.URL.Path == groupURL:
+			assert.Equal(t, stubGroupName, r.URL.Query().Get("name"))
+			resp := api.ListResponse[GroupResponse]{Count: 1, Results: []GroupResponse{{ID: stubGroupID, Name: stubGroupName}}}
+			_ = json.NewEncoder(w).Encode(resp)
+		case r.Method == http.MethodGet && r.URL.Path == userURL:
+			assert.Equal(t, stubUserName, r.URL.Query().Get("username"))
+			resp := api.ListResponse[UserResponse]{Count: 1, Results: []UserResponse{{ID: stubUserID, Username: stubUserName}}}
+			_ = json.NewEncoder(w).Encode(resp)
+		default:
+			membershipHandler(w, r)
+		}
+	}))
+}
+
 func TestAddMember(t *testing.T) {
 	t.Parallel()
-	const (
-		groupID = "group-uuid-111"
-		userID  = "user-uuid-222"
-	)
-	var membershipPostCalled bool
+	var membershipPostCalled atomic.Bool
 
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	ts := newIAMStubServer(t, func(w http.ResponseWriter, r *http.Request) {
 		switch {
-		case r.Method == http.MethodGet && strings.Contains(r.URL.Path, "/groups/"):
-			resp := api.ListResponse[GroupResponse]{Count: 1, Results: []GroupResponse{{ID: groupID, Name: "admins"}}}
-			w.Header().Set("Content-Type", "application/json")
-			_ = json.NewEncoder(w).Encode(resp)
-		case r.Method == http.MethodGet && strings.Contains(r.URL.Path, "/users/"):
-			resp := api.ListResponse[UserResponse]{Count: 1, Results: []UserResponse{{ID: userID, Username: "alice"}}}
-			w.Header().Set("Content-Type", "application/json")
-			_ = json.NewEncoder(w).Encode(resp)
-		case r.Method == http.MethodPost && strings.Contains(r.URL.Path, "/memberships/"):
-			membershipPostCalled = true
+		case r.Method == http.MethodPost && r.URL.Path == membershipURL:
+			membershipPostCalled.Store(true)
 			var req MemberAddRequest
 			_ = json.NewDecoder(r.Body).Decode(&req)
-			if req.Group != groupID {
-				t.Errorf("expected group id %q, got %q", groupID, req.Group)
-			}
-			if req.User != userID {
-				t.Errorf("expected user id %q, got %q", userID, req.User)
-			}
+			assert.Equal(t, stubGroupID, req.Group)
+			assert.Equal(t, stubUserID, req.User)
 			w.WriteHeader(http.StatusCreated)
 		default:
 			t.Errorf("unexpected request: %s %s", r.Method, r.URL.Path)
+			w.WriteHeader(http.StatusInternalServerError)
 		}
-	}))
+	})
 	defer ts.Close()
 
 	ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
-	err := AddMember(ac, MemberAddRequest{Group: "admins", User: "alice", Role: "member"})
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	if !membershipPostCalled {
-		t.Error("membership POST was not called")
-	}
+	err := AddMember(ac, MemberAddRequest{Group: stubGroupName, User: stubUserName, Role: "member"})
+	assert.True(t, membershipPostCalled.Load(), "membership POST was not called")
+	require.NoError(t, err)
 }
 
 func TestDeleteMember(t *testing.T) {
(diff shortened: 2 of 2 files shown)


api/iam/iam_test.go
```diff
@@ -397,3 +398,56 @@ func TestAddMember(t *testing.T) {
...
+			defer ts.Close()
+
+			ac := &client.AlpaconClient{HTTPClient: ts.Client(), BaseURL: ts.URL}
+			err := DeleteMember(ac, MemberDeleteRequest{Group: "admins", User: "alice"})
+			if tt.wantErr != "" {
+				require.EqualError(t, err, tt.wantErr)
+			} else {
+				require.NoError(t, err)
+			}
+			assert.Equal(t, tt.wantDelete, deleteCount.Load())
```
geunwoonoh: Warning: `require` here where the two assertions are independent.

`CLAUDE.md` states: "`require` when later lines depend on the assertion, `assert` when checks are independent." Line 450 reads `deleteCount.Load()`, which is safe no matter what `err` was—nothing later dereferences a result. As written, a wrong error message aborts the subtest before the delete-count check reports, so a run that both returns the wrong error and issues a spurious DELETE only shows the first half.

Switching to `assert` reports both, and it drops `github.com/stretchr/testify/require` from the file's imports entirely.

```suggestion
			if tt.wantErr != "" {
				assert.EqualError(t, err, tt.wantErr)
			} else {
				assert.NoError(t, err)
			}
			assert.Equal(t, tt.wantDelete, deleteCount.Load())
```

api/iam/iam_test.go
```diff
@@ -397,3 +398,56 @@ func TestAddMember(t *testing.T) {
...
+					assert.Equal(t, "group-1", r.URL.Query().Get("group"))
+					assert.Equal(t, "user-1", r.URL.Query().Get("user"))
+					_, _ = w.Write([]byte(tt.response))
+				case r.Method == http.MethodDelete:
+					deleteCount.Add(1)
+					assert.Equal(t, membershipURL+"membership-1/", r.URL.Path)
+					w.WriteHeader(http.StatusNoContent)
+				default:
+					t.Errorf("unexpected request: %s %s", r.Method, r.URL.Path)
+					w.WriteHeader(http.StatusNotFound)
```
geunwoonoh: Suggestion: answer 500 rather than 404 for an unstubbed request.

`CLAUDE.md`: "A handler reports with `t.Errorf` or `assert`, answers 500, and returns." A 404 is a plausible real API answer, so an unstubbed request can be absorbed by client-side handling and read as an ordinary failure. 500 keeps "the test stubbed nothing for this" visually distinct from "the server said no".

```suggestion
					w.WriteHeader(http.StatusInternalServerError)
```

api/iam/iam.go
```diff
@@ -204,6 +204,9 @@ func DeleteMember(ac *client.AlpaconClient, memberDeleteRequest MemberDeleteRequ
 	if err != nil {
 		return err
 	}
+	if len(memberDetails) == 0 {
+		return errors.New("no membership found for the given user and group")
+	}
```
geunwoonoh, line 209: Suggestion: name the user and group in the error.

Neither this message nor the caller's format string in `cmd/iam/member_delete.go` carries the identifiers, so the final CLI line and any CI log read `Failed to remove the member from group: no membership found for the given user and group.` with no way to tell which pair failed. `api/rbac/rbac.go:70` sets the newer in-repo precedent: `no role named %q is visible to you; the name is matched exactly and is case-sensitive`. It also removes the message literal that is currently duplicated across `iam_test.go:411` and `:412`.

The suggestion adds the blank line too—every other guard in this file separates the unmarshal error check from the emptiness guard that way.

Non-blocking: the current wording matches the older sibling `no user found with the given name`, so keeping it is defensible if you would rather that family stay uniform. If you do take this, the two test cases at `iam_test.go:411-412` need their `wantErr` updated in the same commit—`fmt` is already imported here.

```suggestion

	if len(memberDetails) == 0 {
		return fmt.Errorf("user %q is not a member of group %q", memberDeleteRequest.User, memberDeleteRequest.Group)
	}
```

api/iam/iam_test.go
```diff
@@ -397,3 +398,56 @@ func TestAddMember(t *testing.T) {
...
+			var deleteCount atomic.Int32
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				switch {
+				case r.Method == http.MethodGet && r.URL.Path == groupURL:
+					assert.Equal(t, "admins", r.URL.Query().Get("name"))
+					_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"group-1","name":"admins"}]}`))
+				case r.Method == http.MethodGet && r.URL.Path == userURL:
+					assert.Equal(t, "alice", r.URL.Query().Get("username"))
+					_, _ = w.Write([]byte(`{"count":1,"results":[{"id":"user-1","username":"alice"}]}`))
```
geunwoonoh: Suggestion: these two stubs duplicate `TestAddMember`'s (`:365-373`)—same endpoints, same `ListResponse` shape, differing only in `strings.Contains` vs. the exact path match used here.

A small `newIAMStubServer(t, membershipHandler)` helper would leave each test holding only the endpoint it actually exercises. Worth doing once a third test needs the same two stubs; not on its own.

geunwoonoh (dismissed):
## Review summary

`DeleteMember` unmarshaled the membership list into a slice and then indexed `memberDetails[0]` unconditionally. When the (user, group) pair has no membership the server answers `[]`, the slice is empty, and the CLI died with a Go stack trace instead of the error its caller was already written to report. This PR adds a `len(memberDetails) == 0` guard returning `no membership found for the given user and group`, covered by a table-driven `httptest` regression test.

The fix is placed at the cause, not at the crash surface: the guard sits where the empty result first becomes observable, and the same check absorbs a `null` body because `json.Unmarshal` leaves the slice `nil`.

I checked every other first-element access under `api/` (21 sites). `webhook.go:114`, `rbac.go:73`, `auth.go:191`, `server.go:120`, `packages.go:85,95`, `event.go:142,246`, and `iam.go:239,276` all guard on `Count == 0` or `len() == 0` already. This was the last unguarded one, so no sibling follow-up is needed.

The caller renders the error as `Failed to remove the member from group: no membership found for the given user and group.` and exits 1 through `utils.CliErrorWithExit`, which is the behavior #437 asked for.

### Critical issues (0)

None. No security surface is touched: no credentials, no injection vector, no new logging of sensitive data. The change strictly narrows behavior from "panic" to "return error".

### Warnings (1)

1. **[api/iam/iam_test.go:445-450]** `require` is used where the two assertions are independent, and it pulls in the only new import in the file. `CLAUDE.md` states: "`require` when later lines depend on the assertion, `assert` when checks are independent." Line 450 reads `deleteCount.Load()`, which is safe regardless of what `err` was. As written, a wrong error message aborts the subtest before the delete-count check reports, so a run that both returns the wrong error and issues a spurious DELETE only shows the first half. Switching to `assert` reports both and drops `github.com/stretchr/testify/require` from the imports entirely.

### Suggestions (3)

1. **[api/iam/iam_test.go:438]** The `default` branch answers `404`; `CLAUDE.md` specifies `500` for this shape ("A handler reports with `t.Errorf` or `assert`, answers 500, and returns"). A 404 is a plausible real API answer, so an unstubbed request can be absorbed by client-side handling and read as a normal failure. 500 keeps "the test stubbed nothing for this" visually distinct from "the server said no".

2. **[api/iam/iam.go:207-209]** The error names neither the user nor the group, and the caller's format string in `cmd/iam/member_delete.go` does not add them either, so the final CLI line and any CI log carry no identifiers. `api/rbac/rbac.go:70` sets the newer in-repo precedent: `no role named %q is visible to you; the name is matched exactly and is case-sensitive`. Naming the pair also removes the message literal duplicated across `iam_test.go:411` and `:412`. Non-blocking: the current wording matches the older sibling `no user found with the given name`, so keeping it is defensible if the team prefers that family to stay uniform. Cosmetic, same block: every other guard in the file puts a blank line between the unmarshal error check and the emptiness guard; this one runs them together.

3. **[api/iam/iam_test.go:420-432]** The group and user lookup stubs duplicate `TestAddMember`'s (`:365-373`)—same two endpoints, same `ListResponse` shape, differing only in `strings.Contains` vs. exact path match. A small `newIAMStubServer(t, membershipHandler)` helper would leave each test holding only the endpoint it actually exercises. Worth doing when a third test needs the same two stubs; not on its own.

### Highlights

- The regression test is the real thing, not a symptom test: it asserts the value the missing guard produced (`deleteCount == 0`, no DELETE issued for a membership that does not exist), not merely "no panic". Wrapping the index in a recover would not make it pass.
- Covering `null` alongside `[]` is the detail most fixes miss. Both land on a `nil` slice, but only one of them is what the issue reported, and the test pins that a single guard handles both.
- `atomic.Int32` for the call counter, where the neighboring `TestAddMember` uses a plain `bool` written from the handler goroutine. Under the `-race -shuffle=on` CI run, the new test is the better-behaved of the two.
- It is also the only subtest in `api/iam/iam_test.go` that calls `t.Parallel()` inside `t.Run`, which is what `CLAUDE.md` asks for. Go 1.25 per-iteration loop variables make the `tt` capture safe.
- Exact path matching against the `groupURL`/`userURL`/`membershipURL` constants, plus asserting the DELETE path is `membershipURL + "membership-1/"`, pins that `BuildURL`'s trailing-slash behavior reaches the right resource. That is stricter than the `strings.Contains` used elsewhere in the file.

### Additional checks requested

**Python conventions**: not applicable. This PR touches two Go files, and the repository has no Python source (the PowerShell installer suite is separate). Reviewed against `stacks/golang.md` and the Go conventions in `CLAUDE.md` instead; the warning above and suggestions 1 and 2 are what came out of that.

**Stray inline imports**: Go has no function-scoped import syntax, so the concept does not apply. The equivalent check is unused imports: the only import this PR adds is `testify/require`, and it is used—though the warning above would remove it entirely. Production code adds no import, since `errors` was already at the top of the file.

**Duplicate or dead code**: nothing dead (golangci-lint's `unused` check passes). Two duplications, both in tests: the error string is a literal in one production site plus two test cases (suggestion 2 removes that), and the group/user stub handlers overlap with `TestAddMember` (suggestion 3).

**Performance**: no regression, a marginal improvement. The guard is an O(1) length check, and it saves one round trip against an implementation that would fire a DELETE at a nonexistent ID and wait for the server to refuse. One observation outside this PR's scope: `DeleteMember` makes three sequential round trips (group lookup, user lookup, membership lookup), and the first two are independent and could run concurrently. Not introduced here, and the codebase is sequential throughout, so it is a separate issue at most.

---

**Verdict**: APPROVED

The panic is fixed at its cause, the guard closes the last gap of its kind in `api/`, and the test would fail against the pre-fix code. All findings are optional polish.

**Checklist**:
- [x] Tests included and passing (all 10 checks green on b7d5209)
- [x] Documentation updated if needed (no user-facing surface change; `README.md:270` still accurate)

api/iam/iam_test.go
jisung-02: Taken, with one deviation. `assert` for the error assertions fails CI: testifylint's `require-error` checker is on by default and `.golangci.yml` enables testifylint with no settings block, so `assert.EqualError` / `assert.NoError` are reported as `require-error: for error assertions use require`.

The reordering carries the fix instead. The delete count is asserted first, so a run that both returns the wrong error and issues a spurious DELETE reports both halves, and the error assertion keeps `require` and stays lint-clean. `testify/require` therefore stays in the imports.

Fixed in 5b414f8.

api/iam/iam_test.go
jisung-02: Taken. The unstubbed branch answers `http.StatusInternalServerError` now, in `TestAddMember` as well since it grew a `default` branch of the same shape.

Fixed in 5b414f8.

api/iam/iam.go
jisung-02: Keeping the current wording. `no user found with the given name` and `no group found with the given name` are the two siblings in this file, and a third message in that family reading differently is the worse outcome for now. Naming the pair is worth doing as its own change across all three, not on one of them.

Not taking this one.

api/iam/iam_test.go
jisung-02: Taken now rather than on the third test. `newIAMStubServer(t, membershipHandler)` answers the group and user lookups against the `groupURL` / `userURL` constants and hands everything else to the handler the test passes in, so each test holds only the endpoint it exercises.

Two things followed from it. The exact path match is the stricter of the two forms, so `TestAddMember` picks up the check that its lookups reach the right path, replacing `strings.Contains` — `strings` is out of the imports. And both tests now name the same stub identifiers (`stubGroupID`, `stubUserID`, `stubGroupName`, `stubUserName`), which is what let the two stubs merge at all.

While in there, `TestAddMember`'s `membershipPostCalled` became an `atomic.Bool` — it is written from the server goroutine and read from the test goroutine, which the `-race` job can flag.

Fixed in 5b414f8.

geunwoonoh (approved):
## Follow-up review (commit 5b414f8)

Re-review of `5b414f8` against my previous review on `b7d5209`. The production fix in `api/iam/iam.go` is unchanged and still correct—the `len(memberDetails) == 0` guard, verified earlier as the last unguarded first-element access under `api/`. This commit is test-side only (+48 -45 in `api/iam/iam_test.go`).

Three of four findings are resolved, and the fourth is declined with reasoning I accept. No new issues in the added code.

### Disposition of the previous review

| Finding | Status |
|---|---|
| Warning: `require` masks the independent delete-count check | **Resolved, by a different and better route** |
| Suggestion 1: unstubbed request should answer 500, not 404 | **Applied**, and extended to `TestAddMember` |
| Suggestion 2: name the user and group in the error | **Declined**, with sound reasoning |
| Suggestion 3: shared stub server | **Applied**, and carried further |

**On the warning—my suggested fix was wrong.** I proposed switching `require.EqualError`/`require.NoError` to `assert`. That would have failed the lint gate: `.golangci.yml` enables `testifylint` with no settings block, so its default checkers run, and `require-error` is among them—it reports `assert.NoError`/`assert.EqualError` as "for error assertions use require". I verified this against the config at the head SHA rather than taking the reply at face value; the reply is correct.

The reordering chosen instead reaches the goal I actually stated: `assert.Equal(t, tt.wantDelete, deleteCount.Load())` now runs before the error assertions, so a run that both returns the wrong error and issues a spurious DELETE reports both halves. `require` stays, and stays lint-clean.

**On the declined suggestion.** `no user found with the given name` (`iam.go:236`) and `no group found with the given name` (`iam.go:273`) are both present, exactly as the reply describes. Making one of three messages in that family read differently is the worse outcome, and changing all three is its own commit. Reasoning accepted—worth a follow-up issue, not a blocker here.

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (0)

None. The one open item is the declined error-message wording, which belongs in a follow-up across all three messages rather than in this PR.

### Review of the new code

`newIAMStubServer(t, membershipHandler)` answers the group and user lookups and delegates everything else. Checked and clean:

- **No coverage lost, some gained.** `TestAddMember` still asserts the resolved IDs reach the POST body (`assert.Equal(t, stubGroupID, req.Group)`), and it now also picks up the lookup path and query-param assertions it never had, because the shared helper matches `groupURL`/`userURL` exactly where `TestAddMember` previously used `strings.Contains`.
- **Delegation is correct.** `/api/iam/memberships/` and `/api/iam/memberships/membership-1/` both fall through to the `default` branch and reach the test's own handler; an unexpected request still lands on its `t.Errorf` plus 500.
- **`membershipPostCalled` moved from a plain `bool` to `atomic.Bool`.** That closes a latent race in the pre-existing test—it was written from the handler goroutine and read from the test goroutine—under the `-race -shuffle=on` CI run.
- **Declaration order holds.** The new `const` block sits above every `func`, per `CLAUDE.md`'s `const → var → type → func`.
- **`t.Helper()` is present** in the helper, which the enabled `thelper` linter requires, and no `require` or `t.Fatalf` appears inside any handler, which `go-require` requires.
- **Parallel-safe.** Each subtest builds its own server, the stub constants are read-only, and the helper is called inside `t.Run`, so `assert` failures attribute to the right subtest.
- **The one added comment earns its place.** "The delete count is asserted first so a run that both returns the wrong error and issues a spurious DELETE reports both, not just the error" explains an ordering that otherwise looks arbitrary and that a future reader would plausibly tidy back.

### Highlights

- The reply to the warning is the right way to receive review feedback: the suggestion was checked against the actual lint configuration, found to break it, and replaced with something that meets the stated goal rather than the literal instruction.
- Suggestion 1 was applied to the neighboring `TestAddMember` as well, which had no `default` write at all before.
- Suggestion 3 was taken further than asked—shared stub identifiers, exact path matching, `strings` dropped from the imports, and the `atomic.Bool` race fix—leaving `TestAddMember` in better shape than the refactor strictly required.
- Every one of the four threads got a reasoned reply, including the declined one, which cites the specific sibling messages and proposes the right scope for the change.

---

**Verdict**: APPROVED

The panic is fixed at its cause, the regression test would fail against the pre-fix code, the test refactor loses no coverage and fixes a latent race in a neighboring test, and all 10 CI checks pass on `5b414f8`.

**Checklist**:
- [x] Tests included and passing (10/10 green on 5b414f8)
- [x] Documentation updated if needed (no user-facing surface change)

PR #442, opened 2026-09-09
fix(config): restrict config.json permissions while it holds the refresh token

### LINKED ISSUE #388: config.json keeps a wide mode while holding the refresh token
## 증상

`config.json`은 액세스 토큰과 리프레시 토큰을 담는데, 기존 파일의 모드를 아무도 좁히지 않습니다.

`config/config.go:69`입니다.

```go
os.OpenFile(configFile, O_CREATE|O_WRONLY|O_TRUNC, 0600)
```

`os.OpenFile`의 모드 인자는 **그 호출이 파일을 새로 만들 때만** 적용됩니다. 이미 0644로 존재하는 `config.json`은 계속 0644로 남습니다. 백업에서 복원했거나, 다른 머신에서 복사했거나, 프로비저닝 스크립트가 썼거나, 옛 CLI가 남긴 파일이 그렇습니다.

리포 전체에 `Chmod`를 걸면 `config/device.go`, `cmd/edit/edit.go:256`, `utils/file_save.go:104`뿐이고 `restrictConfigFileMode`에 해당하는 것이 없습니다.

## 왜 지금 올리나

alpacax/alpacon-cli#379가 `device_id`의 모드를 좁히는 작업을 하면서 이 비대칭이 드러났습니다.

`config/device.go:56`의 주석이 식별자를 두고 "비밀번호나 토큰이라는 의미의 비밀은 아니다"라고 적고 있습니다. 그런데 지금 상태는 CLI가 **비밀이 아닌 쪽의 모드는 매 실행마다 복구하고, 바로 옆에 있는 진짜 비밀은 그대로 둔다**는 것입니다.

`device_id`는 `readDeviceID:102`가 성공적인 읽기마다 `restrictDeviceIDFileMode`를 부르기 때문에 잘못된 모드가 다음 실행에 고쳐집니다. `config.json`에는 그 경로가 없습니다.

## 제안

`device.go`가 #379에서 확립한 형태를 그대로 씁니다.

- 경로가 아니라 **열린 핸들**에 `Stat`과 `Chmod`를 걸어 심링크 바꿔치기를 막습니다
- 요청한 모드를 가정하지 말고 실제 모드를 되읽습니다. umask가 끼면 요청값과 결과값이 다릅니다
- **좁히기만** 합니다. 이미 더 엄격한 파일을 넓히면 안 됩니다
- Windows에서는 조기 반환하고, 그 사실을 주장하는 주석 옆에 단서를 답니다

읽기 경로에 걸어 두면 다음 실행에 자동 복구됩니다.

## 함께 볼 것

디렉터리 모드도 강제되지 않습니다. `config/device.go:77`의 `os.MkdirAll(dir, 0700)`은 **이미 존재하는** 디렉터리의 모드를 건드리지 않아서, 0777 `~/.alpacon`에 0600 파일이 들어 있는 상태가 나옵니다. 파일 내용은 못 읽어도 파일을 지우고 다시 만들 수는 있습니다. `device_id`의 경우 `createDeviceID:121-130`이 발견한 값을 그대로 채택하므로, 공격자가 CLI가 Auth0에 보낼 `device:<id>`를 고르게 됩니다. ADR 0054의 MFA 존재 증명이 그 값에 묶입니다.

이 두 가지는 같은 변경으로 처리하는 편이 자연스럽습니다.

## 참고

- alpacax/alpacon-cli#379 — 이 비대칭이 드러난 PR이자 참고할 형태

jisung-02:
## Background

`config.json` holds the refresh token and could keep a wide mode indefinitely, because only the write path set `0600`: a file or directory that was already wider — created before this code existed, copied from a backup, or left behind by another account — was never narrowed on read. The device id file had the same gap.

Closing it turned out to need more than a `chmod`. A path-based `chmod` can be pointed at another file by anyone able to drop a symlink in the config directory, and the two files answer a refusal differently: the config file is the CLI's own target and credentials, while the device id is what an MFA presence proof binds to.

## Changes

Every mode decision and every `chmod` is bound to the handle it was taken on. `restrictConfigDirectoryMode` opens the directory and stats that handle rather than stat-ing the path and opening it afterwards. Two path-based reads remain and neither decides a mode on its own: the pre-check on the permission-denied branch, where an ordinary handle does not exist and the platform fallback re-derives the mode from its own handle before ANDing with `0700`; and `narrowToPublishedFileMode` in `config/device.go`, which reads the mode a rename is about to stand in for and intersects it, so it can only remove bits.

`config/permissions_unix.go` holds the guards the chmod and the read are gated on:

| Guard | Refuses | Why the mode alone does not say it |
|---|---|---|
| `O_NOFOLLOW` via `openToNarrow` | a symbolic link where a file or the directory belongs | the target could be a file the user does not own, and the same run under `sudo` would take access away from whatever the link named |
| `refuseHardLinked` | a file that answers to more than one name | `O_NOFOLLOW` says nothing about hard links, and macOS lets any account link a file it can merely read |
| `refuseForeignOwner` | a handle on a file this process does not own | root can read anything, and a `sudo` that keeps `HOME` would otherwise take an unprivileged user's file as root's device id |
| `refuseUnsafeConfigFile` | anything that is not a regular file | `O_NONBLOCK` keeps the open off a named pipe, but `os.NewFile` hands the descriptor to the poller and the read then waits for a writer that never arrives |

Narrowing stays best effort for a mode this process cannot change — a read-only mount, a directory another account owns, a filesystem without permission bits — and `warnUnrestricted` reports it once per subject and cause rather than ending the command. A symbolic link is the one failure that reads what it names before it speaks: the link is refused a `chmod` rather than the file behind it, so nothing is exposed unless the target's own mode grants another account access. `openCheckedConfigFile` reads that mode off the handle its fallback already opened, and `restrictConfigDirectoryMode` reaches the resolved directory with `O_NOFOLLOW` through `symlinkTargetGrantsOtherAccounts`. Keeping one `config.json` in a dotfiles tree at `0600` drew the warning on every command until now, for an exposure that was not there. `grantsOtherAccounts` is the judgment all of these share, the two read-backs inside `restrictOpenFileMode` included. A `chmod` that fails outright is held to the same read-back the successful path already used, rather than treated as an exposure by default — a Linux `vfat` mount answers `EPERM` for any mode its `fmask` disallows even when that mode already grants no other account access, and the device id must not be lost there for no reason. `errUnsafeConfigFile` is what separates that from a file this CLI did not write: a non-regular file and a foreign-owned one are refused outright, by `LoadConfig` as well as by `readDeviceID` — and `LoadConfig` now reads the very handle the narrowing opened rather than reopening the path, so the file cannot be swapped between the check and the decode, and the same refusal reaches what a symbolic link resolves to instead of stopping at whether the link itself is a regular file.

`resolveConfigWritePath` and `readlinkTarget` wrap their open errors with `%w`, so `errSymlink` and the errno underneath survive the write path. Nothing sorts on either today.

`saveConfig` and `DeleteConfig` resolve a symlinked `config.json` before acting, for the same reason the read follows the link: `os.Rename` and `os.Remove` act on the last path component, which would otherwise replace or unlink the dotfiles-tree link itself rather than the file it names. The resolved target is checked against the same two refusals a read applies.

`openConfigDirectory` follows a symbolic link on `~/.alpacon`, where `openToNarrow` refuses one. Pointing the config directory at a dotfiles tree is a supported setup; refusing it left that installation with no device id at all while the same link kept handing `LoadConfig` the refresh token. Only the directory's own `chmod` still refuses the link, which is the attack the refusal exists for. The file inside is reached with `openat` on the resolved handle, so every guard above still runs on it.

`symlinkErrnos` is declared per OS. FreeBSD and DragonFly report `EMLINK` for an `O_NOFOLLOW` open of a link and NetBSD reports `EFTYPE`, where Linux and macOS report `ELOOP`, so a `!windows` build tag no longer stands in for a platform.

Per-OS opens for a directory that permits traversal but not reads: `permissions_darwin.go` (`O_SEARCH`, and it now reads `EINVAL` to say macOS 13 is what it needs), `permissions_linux.go` (`O_PATH` and `Fchmodat`), `permissions_other.go` (a named missing-capability error instead of a bare `os.ErrPermission`, which read like a real permission problem on a platform that has no such open).

A `chmod` that reports success is checked by reading the mode back, since a filesystem without permission bits accepts the call and keeps the mode. Only bits that hand another account access count as a failure — macOS reports every file on a FAT volume as `0700`.

## Safety

Two hangs and one silent downgrade are closed by this PR:

- A named pipe at `~/.alpacon/config.json` or `~/.alpacon/device_id` parked the command with nothing printed and no way out from the CLI itself. Both paths now refuse a non-regular file, and `LoadConfig` re-checks what a symlinked path resolves to, so a link aimed at a pipe reports an error instead of blocking.
- A `config.json` owned by another account was read after a warning. That file names the host the CLI talks to and carries `Insecure`, which becomes `InsecureSkipVerify`, and re-login offers the stored target as the default — so an attacker who could write in `~/.alpacon` once kept a foothold that survived the narrowing meant to close it, since a file another account owns can never be chmod-ed back. It is now refused, and refused the same way when it is a symbolic link — that path skipped the owner check entirely and reached the decoder through nothing but an `IsRegular` test, so the refusal above held for a hard link and not for a symbolic one until now.
- Keeping one `config.json` in a dotfiles tree read correctly but wrote incorrectly: `saveConfig`'s rename replaced the link itself with a regular file on the first token refresh, and `alpacon logout` removed only the link, leaving the refresh token behind in the dotfiles tree either way. Both now resolve the link first and act on what it names.
- A `~/.alpacon` symlinked into a dotfiles tree lost its device id on every run, dropping MFA presence checks to a network fingerprint. It keeps its device id now.

On Windows nothing is repaired: `chmod` there changes the read-only attribute rather than access permissions, so the two modes are not enforced and the symlink, hard-link and owner refusals are stubs. The device id is not protected on Windows either.

## Testing

`go vet ./...` — no output.

```
42 packages ok, 0 failed
```

```
$ GOOS=linux golangci-lint run ./...
0 issues.
$ GOOS=darwin golangci-lint run ./...
0 issues.
$ GOOS=windows golangci-lint run ./...
0 issues.
```

`go build ./config/...` succeeds for `GOOS=linux`, `darwin`, `windows`, `freebsd`, `netbsd`, `openbsd`, and `dragonfly` (the last on `amd64`; Go has no `dragonfly/arm64`).

Every test added here was seen RED once, by mutating the guard it covers and confirming it fails: removing the regular-file check, removing the owner check, restoring the directory-link refusal, making `LoadConfig` warn instead of refuse, disabling its regular-file re-check, moving the hard-link guard ahead of the chmod, making `warnOnce` unconditional, and emptying `isSymlinkErrno` each broke exactly the tests naming them.

Five more were added answering the review, each seen RED the same way: `TestLoadConfigRefusesASymlinkToAFileOwnedByAnotherAccount` failed with the owner refusal removed, decoding `/etc/hosts` as JSON instead of refusing it (`"failed to decode config file: json: cannot unmarshal number into Go value of type config.Config"`); `TestSaveConfigWritesThroughASymlinkedConfigFile`, `TestDeleteConfigRemovesWhatASymlinkedConfigFileNames`, and `TestSaveConfigRefusesASymlinkToAFileOwnedByAnotherAccount` all failed with the write-path resolution removed, the first two by acting on the link instead of its target and the third by writing the refresh token straight into `/etc/hosts`'s owner; and `TestGetActiveWorkSessionReportsNoSessionWhenNoConfigExists` failed with `os.IsNotExist` restored in place of `errors.Is(err, fs.ErrNotExist)`, surfacing `"config file does not exist"` from a function whose doc comment promises `""` and no error.

Three tests cover the symlink warning and the refused-`chmod` branch. `TestLoadConfigWarnsOnASymlinkedConfigFileOnlyWhenItsTargetIsWide` and `TestRestrictConfigDirectoryModeReportsASymlinkOnlyWhenItsTargetIsWide` were each RED before `f8c6956`, on the case whose target is already narrow—the file one produced the exact sentence the review quoted (`"...config.json is a symbolic link, so its target is left alone: ... Another local account may be able to read your credentials while that stands.\n" should not contain "could not restrict"`) and the directory one returned `errSymlink` for a `0700` target. Both also assert the link's target keeps the mode it had, since it is never chmod'ed either way. `TestRestrictOpenFileModeJudgesAModeAChmodKept` grew a `chmodErr` column in `56f1771`, which is what finally runs the refused-`chmod` branch: returning `err` unconditionally failed the `0700` case with `"Received unexpected error: operation not permitted"`, and returning `nil` unconditionally failed the `0644` case with `"An error is expected but got nil."`

The regression guards are `TestLoadConfigNarrowsWhatIsWideAndKeepsWhatIsStricter`, `TestConfigOperationsNarrowAWideOpenDirectory`, `TestSearchableConfigDirectoryOperations`, `TestRestrictOpenFileModeJudgesAModeAChmodKept`, `TestWarnUnrestrictedSpeaksOncePerSubjectAndCause`, `TestGetActiveWorkSessionReportsNoSessionWhenNoConfigExists`, and in `config/permissions_unix_test.go`: `TestRestrictConfigDirectoryModeRejectsANonDirectoryWithoutBlocking`, `TestNarrowingRefusesASymbolicLink`, `TestRestrictConfigFileModeNarrowsThroughASymlinkedConfigDirectory`, `TestGetOrCreateDeviceIDSurvivesASymlinkedConfigDirectory`, `TestGetOrCreateDeviceIDRefusesAnIdentifierItCouldNotNarrow`, `TestRestrictConfigFileModeRefusesAHardLink`, `TestRestrictConfigFileModeAcceptsAHardLinkedFileAlreadyWithinTheMask`, `TestRefuseUnsafeConfigFileNamesWhatMustNotBeRead`, `TestConfigReadsRefuseANamedPipeWithoutBlocking`, `TestLoadConfigRefusesAConfigFileOwnedByAnotherAccount`, `TestLoadConfigRefusesASymlinkToAFileOwnedByAnotherAccount`, `TestLoadConfigRefusesASymlinkToANamedPipe`, `TestLoadConfigStillReadsThroughASymlinkedConfigFile`, `TestSaveConfigWritesThroughASymlinkedConfigFile`, `TestDeleteConfigRemovesWhatASymlinkedConfigFileNames`, `TestSaveConfigRefusesASymlinkToAFileOwnedByAnotherAccount`, `TestLoadConfigWarnsOnASymlinkedConfigFileOnlyWhenItsTargetIsWide`, `TestRestrictConfigDirectoryModeReportsASymlinkOnlyWhenItsTargetIsWide`, `TestOpenNoFollowInStaysWithTheDirectoryItHolds`, `TestRefuseForeignOwner`, and `TestIsSymlinkErrnoReadsThePlatformList`.

Two coverage limits are worth naming rather than leaving to be discovered. The owner refusal has no local shape on Linux — it will not let a test link a root-owned file into place — so `TestLoadConfigRefusesAConfigFileOwnedByAnotherAccount` and `TestSaveConfigRefusesASymlinkToAFileOwnedByAnotherAccount` skip there and run on the macOS leg only; `TestLoadConfigRefusesASymlinkToAFileOwnedByAnotherAccount` needs no hard link and runs on both. The `EINVAL` branch in `permissions_darwin.go` cannot be reached on a macOS 13 or later runner, so it has no test; it will stay uncovered until someone runs the CLI on an older release.

The BSD errno list is checked by `TestIsSymlinkErrnoReadsThePlatformList`, which pins the wiring, not the errno each BSD actually returns. Nothing in CI runs on a BSD, and the release build ships linux, darwin and windows only.

Closes #388


config/permissions.go
```diff
@@ -0,0 +1,62 @@
...
+	if os.IsPermission(err) {
+		if err = restrictSearchableDirectory(path); err != nil {
+			return fmt.Errorf("failed to restrict config directory permissions: %w", err)
+		}
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to open config directory: %w", err)
+	}
+	defer func() { _ = dir.Close() }()
```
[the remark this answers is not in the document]
jisung-02: Right about the window: the stat and the open resolved the path twice, so the directory could be swapped between them and the chmod would land on whatever took its place.

d6d220a opens the directory first and reads its type and mode from `dir.Stat()`, so the inspection and the chmod share one handle. The path-based read that remains is on the permission-denied branch only, where an ordinary handle does not exist; it decides nothing by itself, because `restrictSearchableDirectory` re-derives the mode from its own handle and only ANDs with `0700`.

Reaching a file inside the directory now goes through `openat` on that handle rather than by path, since `O_NOFOLLOW` guards the last component alone. `TestOpenNoFollowInStaysWithTheDirectoryItHolds` swaps the directory out after the handle exists and checks the read still lands on the original.

config/permissions_darwin.go
```diff
@@ -0,0 +1,19 @@
...
+const directorySearch = 0x40000000 | unix.O_DIRECTORY
+
+func restrictSearchableDirectory(path string) error {
+	file, err := os.OpenFile(path, directorySearch, 0)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = file.Close() }()
+	return restrictFileMode(file, 0700)
+}
```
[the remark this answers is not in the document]
jisung-02: Agreed — an `EINVAL` from `O_SEARCH` reached the user as an unexplained permissions failure with nothing pointing at the OS version.

d6d220a wraps it in `openSearchableDirectory`: the error now says that narrowing a directory without owner read permission needs macOS 13 or later. Both callers go through that function, so `restrictSearchableDirectory` and the openat base get the same message.

One limit worth stating: the branch cannot be reached on a macOS 13 or later runner, so it has no test and the new macOS leg does not cover it.

config/permissions_other.go
```diff
@@ -0,0 +1,9 @@
+//go:build !linux && !darwin
+
+package config
+
+import "os"
+
+func restrictSearchableDirectory(path string) error {
+	return &os.PathError{Op: "open", Path: path, Err: os.ErrPermission}
+}
```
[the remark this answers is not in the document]
jisung-02: Correct — `os.ErrPermission` on its own read as a real permission problem, when what it actually meant was that the platform has no search-only open at all.

d6d220a replaces it with `unsupportedSearchableDirectory`, which names the path, says opening a directory for search alone is unsupported on that `GOOS`, and still wraps `os.ErrPermission` so `os.IsPermission` keeps working for callers that sort on it.

The file stays rather than being deleted as unreachable on shipped platforms: it is what keeps the package compiling for the BSDs, which `go build` covers for `freebsd`, `netbsd`, `openbsd` and `dragonfly`.

config/config.go
```diff
@@ -159,6 +174,16 @@ func LoadConfig() (Config, error) {
...
+	// Reached only when the path is a symbolic link, which the narrowing above
+	// warns about rather than following—so this is the first look at what it
+	// resolves to, and a pipe there would park the decode below forever.
+	info, err := file.Stat()
+	if err != nil {
+		return Config{}, fmt.Errorf("failed to inspect config file: %v", err)
+	}
+	if !info.Mode().IsRegular() {
+		return Config{}, fmt.Errorf("%s is not a regular file: %w", configFile, errUnsafeConfigFile)
+	}
```
geunwoonoh: **Critical:** a symbolic link at `config.json` skips the owner refusal entirely.

`openNoFollowIn` returns `ELOOP` before `refuseUnsafeConfigFile` ever runs (`permissions.go:157-160`), so the only ownership check on this path never executes. `openConfigForRead` above then follows the link, and `IsRegular()` is all that stands between the decode and a file another account owns—the file that sets `WorkspaceURL` and `Insecure`, and that re-login offers as the default target. The description and the README say such a file is refused outright; that holds for a hard link, not for a symbolic one.

Calling `refuseUnsafeConfigFile` here closes the hole and removes the copy of its first half at the same time.

```suggestion
	// Reached only when the path is a symbolic link, which the narrowing above
	// warns about rather than following—so this is the first look at what it
	// resolves to: a pipe there would park the decode below forever, and a file
	// another account owns is the foothold this narrowing exists to close.
	if err = refuseUnsafeConfigFile(file); err != nil {
		return Config{}, err
	}
```

`TestLoadConfigRefusesASymlinkToANamedPipe` still gets `errUnsafeConfigFile` and `"is not a regular file"`, and `TestLoadConfigStillReadsThroughASymlinkedConfigFile` resolves to a file the same user owns. A guard for "user-owned symlink pointing at a foreign-owned file" would be worth adding.

config/permissions_unix.go
```diff
@@ -0,0 +1,79 @@
...
+	info, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	sys, ok := info.Sys().(*syscall.Stat_t)
+	if !ok {
+		return nil
+	}
+	if int(sys.Uid) != os.Geteuid() {
+		return fmt.Errorf("%s is owned by uid %d, not by uid %d", file.Name(), sys.Uid, os.Geteuid())
```
geunwoonoh: **Warning:** this refusal also fires in two ordinary situations, and the message sends the user the wrong way.

`sudo -E alpacon ...` (or any sudoers that keeps `HOME`) and `docker run -v ~/.alpacon:/root/.alpacon alpacax/alpacon-cli ...` both arrive here with mismatched uids. Both worked before; now every command ends in an error. CLAUDE.md names CI/CD pipelines as a primary user of this CLI, so the container case is worth weighing.

The security reasoning holds, so this is not an argument for dropping the refusal. But the message reads "is owned by uid 501, not by uid 0" and the README's remedy is "delete it and log in again"—which, for someone who got here through sudo or a container, deletes working credentials for no reason. Naming the identity the process is running as makes the real cause visible:

```suggestion
		return fmt.Errorf("%s is owned by uid %d, but this process runs as uid %d", file.Name(), sys.Uid, os.Geteuid())
```

A line in the README covering sudo and bind-mounted config directories would help too.

config/device.go
```diff
@@ -87,10 +94,23 @@ func GetOrCreateDeviceID() (string, error) {
 // rather than reported as absent, which would hide a permissions problem behind
 // an identifier that changes on every invocation.
 func readDeviceID(path string) (string, error) {
-	data, err := os.ReadFile(path)
+	// Fails closed where the config only warns, and reads the handle it narrowed
+	// rather than reopening the path: an MFA presence proof binds to whatever
+	// identifier this returns, so a mode this process could not narrow means no
+	// identifier rather than one another account may have chosen.
+	file, err := openNarrowedConfigFile(path)
```
geunwoonoh, line 101: **Warning:** failing closed here may make the device id permanently unavailable on Linux vfat/exFAT.

Every failure of `openNarrowedConfigFile` now means "no device id". FAT on macOS passes, because `restrictOpenFileMode` counts only bits that hand another account access (`&0077`) as a failure. Linux vfat looks different: `fat_setattr` refuses a mode change that does not match the mount's fmask with `EPERM`, so the `chmod` fails, and `createDeviceID`'s closing `readDeviceID` (line 170) fails with it.

That puts `createDeviceIDFileWithoutLink` in an odd position—its comment names "FAT, exFAT, some container mounts" as the reason it exists, yet on those filesystems no identifier is ever obtained, so the fallback has nothing left to do. If this is the intended trade-off, say so in that comment and in the README. If not, consider applying to a failed `chmod` the same standard the read-back already applies to a kept mode. Worth confirming the vfat behavior before acting on it.

config/config.go
```diff
@@ -77,6 +77,8 @@ func saveConfig(config *Config) error {
 		return fmt.Errorf("failed to create config directory: %v", err)
 	}
 
+	warnUnrestricted("config directory", restrictConfigDirectoryMode(configDir))
```
geunwoonoh, line 81: **Warning:** the dotfiles guarantee this PR documents survives reads but not the first write.

CLAUDE.md, the README, and the description all state that a single config file kept in a dotfiles tree keeps working. Reads do. But `os.Rename(tempFile, configFile)` below replaces the link itself with a regular file, so the first access-token refresh removes the dotfiles link and leaves the original stale. `DeleteConfig`'s `os.Remove` likewise removes only the link, so `alpacon logout` leaves the refresh token behind in the dotfiles tree.

The rename predates this PR (2bfd908); the claim written on top of it does not. Either narrow the claim to "reads follow the link, writes replace it", or resolve the link before the rename. There is no test covering what `saveConfig` does to a symlinked `config.json` either.

config/config.go
```diff
@@ -149,7 +151,20 @@ func LoadConfig() (Config, error) {
...
+	// talks to and decides whether its certificate is checked. Reading it would
+	// hand the next login's token to whoever chose those values.
+	if err = restrictConfigFileMode(configFile); err != nil {
+		if errors.Is(err, errUnsafeConfigFile) {
+			return Config{}, err
+		}
+		warnUnrestricted("config file", err)
+	}
+
+	file, err := openConfigForRead(configFile)
```
geunwoonoh: **Suggestion:** narrow and read open the same file twice.

`restrictConfigFileMode` closes and discards the handle it just narrowed, and this line resolves the path again. The file can change in between, and every call pays an extra open—`LoadConfig` runs two to five times per command across 21 call sites.

Reading the handle `openNarrowedConfigFile` already returned, and falling back to `openConfigForRead` only when the failure was `errSymlink`, closes the window and drops the syscalls together. Low priority once the critical issue above is fixed, since that is what actually closes the security gap.

config/permissions.go
```diff
@@ -0,0 +1,246 @@
...
+	errUnsafeConfigFile = errors.New("unsafe config file")
+	warnedUnrestricted  sync.Map
+)
+
+func restrictFileMode(file *os.File, allowed os.FileMode) error {
+	return restrictOpenFileMode(file, allowed, file.Chmod)
+}
+
+func restrictOpenFileMode(file *os.File, allowed os.FileMode, chmod func(os.FileMode) error) error {
+	if runtime.GOOS == "windows" {
```
geunwoonoh, line 173: **Suggestion:** this branch is unreachable.

Every caller of `restrictOpenFileMode` returns on `GOOS` first: `restrictConfigDirectoryMode` at line 183, `openNarrowedConfigFile` at line 134, and the `permissions_linux.go`/`permissions_darwin.go` callers only exist on those platforms. If it stays as defense in depth for a future caller, a comment saying so would keep the next reader from looking for the Windows path that reaches it.

config/permissions.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	return err
+}
+
+// refuseUnsafeConfigFile rejects a handle nothing should read. A named pipe at
+// the path is the reason for the first check: O_NONBLOCK keeps the open from
+// parking, but says nothing about the read behind it—os.NewFile hands a
+// non-blocking descriptor to the poller, and io.ReadAll then waits there for a
+// writer that never arrives.
+func refuseUnsafeConfigFile(file *os.File) error {
```
geunwoonoh: **Suggestion:** `fstat` runs three times on one handle before anything changes.

`refuseUnsafeConfigFile` stats, `refuseForeignOwner` stats again, and `restrictOpenFileMode` stats a third time. All three happen before any `chmod`, so all three see the same inode state. Passing the `os.FileInfo` down collapses them into one—the stats inside the `chmod` callback and the read-back genuinely need their own.

config/permissions.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	dir, err := openToNarrow(path)
+	if os.IsNotExist(err) {
+		return nil
+	}
+	// A directory that allows traversal but not reads has no ordinary handle, so
+	// its mode has to be read from the path. Granting no other account anything
+	// already satisfies the mask, and the platform fallbacks refuse a directory
+	// whose owner holds no execute bit.
+	if os.IsPermission(err) {
```
geunwoonoh: **Suggestion:** error classification is inconsistent within this file.

The comment at line 114 gives the reason to prefer `errors.Is(err, fs.ErrNotExist)`—`os.IsNotExist` unwraps only the error types the os package defines—and lines 145, 187, and 194 then use `os.IsNotExist`/`os.IsPermission`. It works today only because those errors reach here unwrapped; the moment someone wraps `openNoFollow` or `openConfigDirectory`, these stop matching and the failure is silent. `errors.Is(err, fs.ErrNotExist)` and `errors.Is(err, fs.ErrPermission)` would hold either way.

config/permissions_unix.go
```diff
@@ -0,0 +1,79 @@
...
+// says nothing about hard links, so a link planted in a directory this process
+// could not narrow would otherwise aim the chmod at an inode that also lives
+// outside the config directory—and macOS lets any account link a file it can
+// merely read.
+func refuseHardLinked(file *os.File) error {
+	info, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	sys, ok := info.Sys().(*syscall.Stat_t)
```
geunwoonoh: **Suggestion:** both guards fail open when the type assertion fails.

`refuseHardLinked` and `refuseForeignOwner` return `nil` when `info.Sys()` is not a `*syscall.Stat_t`, which skips the check. The path is unreachable on Unix, so this is not a live bug—but a function whose job is to refuse defaults to allowing, and that deserves a sentence saying why, or a refusal instead.

config/permissions.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+	}
+	return false
+}
+
+// asSymlink names the refusal behind that errno so a caller can tell it from a
+// filesystem failure. O_NOFOLLOW is what produces it: following the link would
+// let anyone who can write in the config directory aim the chmod at a file they
+// do not own, and the mask only ever takes access away.
+func asSymlink(path string, err error) error {
```
geunwoonoh: **Suggestion:** the name reads like the `errors.As` family.

In Go an `As` prefix suggests a type assertion that reports whether it matched. This function wraps an error with context and returns it unchanged otherwise, so something like `wrapSymlinkErr` describes what it does.

.github/workflows/build-and-test.yaml
```diff
@@ -20,16 +20,23 @@ jobs:
...
-    # Windows is not a portability formality: the files that replace a running
-    # executable are compiled out everywhere else and would ship untested without
-    # this runner. It runs only the packages that carry them—the rest of the
-    # suite asserts Unix file modes in tests that predate this job.
+    # Windows and macOS each carry files every other platform compiles out, and
+    # without a runner of their own those files ship untested: Windows holds the
+    # ones that replace a running executable, macOS the O_SEARCH open that
+    # narrows a config directory it cannot read. macOS runs the whole suite,
+    # which passes there unchanged, and is the only leg that exercises the
+    # owner-refusal path—Linux will not let a test link a root-owned file into
```
geunwoonoh: **Suggestion:** this comment claims more than holds.

`TestRefuseForeignOwner` (which borrows `/`) and `TestRefuseUnsafeConfigFileNamesWhatMustNotBeRead` (which borrows `/etc/hosts`) both run on Linux, so macOS is not the only leg exercising the owner-refusal path. What is macOS-only is `TestLoadConfigRefusesAConfigFileOwnedByAnotherAccount`—the end-to-end case through `LoadConfig`. Worth narrowing the sentence to that, so the next person reading it does not assume Linux covers none of it.

config/permissions_unix_test.go
```diff
@@ -0,0 +1,355 @@
...
+	dir := filepath.Join(os.Getenv("HOME"), ConfigFileDir)
+	require.NoError(t, os.MkdirAll(dir, 0700))
+	configFile := filepath.Join(dir, ConfigFileName)
+	// A file another account owns is one another account can rewrite, and this
+	// one names the host the CLI talks to and whether its certificate is checked.
+	// Linking a root-owned file is how macOS lets an attacker leave one behind;
+	// Linux refuses the link, and there the case has no local shape to build.
+	if err := os.Link("/etc/hosts", configFile); err != nil || os.Geteuid() == 0 {
+		t.Skip("cannot plant a file owned by another account here")
+	}
```
geunwoonoh: **Suggestion:** `os.Link` is evaluated before the root check, so as root this creates the link and then skips.

Harmless—the link lands in a temp `HOME` that gets cleaned up—but reversing the order keeps the skip from leaving anything behind:

```suggestion
	if os.Geteuid() == 0 {
		t.Skip("cannot plant a file owned by another account here")
	}
	if err := os.Link("/etc/hosts", configFile); err != nil {
		t.Skip("cannot plant a file owned by another account here")
	}
```

CLAUDE.md
```diff
@@ -221,7 +221,7 @@ _ = json.NewEncoder(w).Encode(resp)
...
...acOS reports every file on a FAT volume as `0700`. Everything else is best effort—a read-only mount, a directory another account owns, and a modeless filesystem all refuse the chmod—so `warnUnrestricted` warns once per subject and cause and the command carries on rather than dying over a mode it cannot change. `LoadConfig` therefore still reads a `config.json` that is itself a symbolic link, which is what keeps a single config file kept in a dotfiles tree working; it re-opens the path with `openConfigForRead` and refuses what that resolves to unless it is a regular file, so a link aimed at a pipe reports an error instead of parking the command. `~/.alpacon/device_id` fails closed throughout: an MFA presence proof binds to whatever identifier `readDeviceID` returns, so it reads the handle it narrowed and treats one it could not protect as no identifier at all, leaving the server to fall back to an IP fingerprint. All of it is Unix only. Windows chmod changes the read-only attribute rather than access permissions, so nothing is repaired there, the two modes above are not enforced, and the symlink, hard-link and owner refusals are stubs—the device id is not protected on Windows either
```
geunwoonoh: **Suggestion:** this bullet is 795 words—17% of a 4,716-word file in one item.

CLAUDE.md is loaded into context every session, so the length is a real cost paid on every run. The detail belongs in a package comment at the top of `config/permissions.go`, where it is read exactly when someone opens the file it describes. Three or four sentences here saying where the pieces live, what is narrowed, and what is refused outright would carry the rest.

geunwoonoh (changes requested):
## Summary

This PR narrows `~/.alpacon`, `config.json`, and `device_id` in place whenever they are read or written, instead of setting `0600` only on the write path. It does more than add a `chmod`: every mode decision and every `chmod` is bound to an open handle, the errno that reports a refused symbolic link is declared per OS, and a `chmod` that reports success is checked by reading the mode back, so a filesystem without permission bits cannot claim a narrowing it never performed. The reasoning behind each choice is recorded in the code, and every new test was seen RED once by mutating the guard it covers.

The one critical issue below is a hole in the guarantee this PR states for itself, so it is worth closing before merge. Everything else can follow.

---

### Critical issues (1)

**1. `config/config.go:180-186`—a symbolic link bypasses the owner refusal**

On the `LoadConfig` path the only ownership check is `refuseUnsafeConfigFile`, inside `openNarrowedConfigFile`. When `config.json` is a symbolic link, `openNoFollowIn` returns `ELOOP` first (`permissions.go:157-160`), so `refuseUnsafeConfigFile` never runs. `openConfigForRead` then follows the link, and the only re-check on what it resolves to is `IsRegular()`.

The attack is the one this PR names. On an installation whose `~/.alpacon` was already wide open, another account leaves a symbolic link called `config.json` pointing at a file it owns. One warning is printed and that file is read. It sets `WorkspaceURL` and `Insecure` (which becomes `InsecureSkipVerify`), and re-login offers it as the default target. The PR description and the README say such a file is "refused outright"; that holds for a hard link and not for a symbolic one.

The fix is to replace the inline check with `refuseUnsafeConfigFile`, which also removes the copy of its first half that now lives in `config.go`. The existing tests still pass: `TestLoadConfigRefusesASymlinkToANamedPipe` gets the same `errUnsafeConfigFile` and `"is not a regular file"`, and `TestLoadConfigStillReadsThroughASymlinkedConfigFile` resolves to a file the same user owns. A regression guard for "user-owned symlink pointing at a foreign-owned file" is worth adding.

---

### Warnings (3)

**2. `config/config.go:160`—the hard owner refusal breaks sudo and Docker use**

`refuseForeignOwner` compares `sys.Uid` against `os.Geteuid()` and nothing else. Two ordinary situations have mismatched uids:

- `sudo -E alpacon ...`, or any sudoers that keeps `HOME`. This worked before and now every command ends in an error.
- `docker run -v ~/.alpacon:/root/.alpacon alpacax/alpacon-cli ...`. The image runs as root while the host file is owned by uid 501 or 1000. CLAUDE.md names CI/CD pipelines as a primary user of this CLI, so this is worth stating.

The security reasoning holds, so this is not an argument for dropping the refusal. But the message reads `"is owned by uid 501, not by uid 0"` and the remedy the README gives is "delete it and log in again". For someone who arrived here through sudo or a container that is the wrong remedy, and following it deletes working credentials. At minimum, name the identity the process is running as, and add a line to the README.

**3. `config/device.go:101`—the device id can become permanently unavailable on Linux vfat/exFAT**

Now that `readDeviceID` fails closed, every failure of `openNarrowedConfigFile` means "no device id". FAT on macOS passes, because `restrictOpenFileMode` only treats bits that hand another account access (`&0077`) as a failure. Linux vfat looks different: `fat_setattr` refuses a mode change that does not match the mount's fmask with `EPERM`, so the `chmod` fails, and `createDeviceID`'s closing `readDeviceID` (`device.go:170`) fails with it.

That leaves `createDeviceIDFileWithoutLink` in an odd position: its comment names "FAT, exFAT, some container mounts" as the reason it exists, yet on those filesystems no identifier is ever obtained, so the fallback has nothing left to do. If this is the intended trade-off, say so in that comment and in the README. If it is not, consider applying the same standard to a failed `chmod` that the read-back already applies to a kept mode. The exact vfat behavior is worth confirming before acting on this.

**4. `config/config.go:80` / `saveConfig`—a symlinked `config.json` survives reads but not the first write**

CLAUDE.md, the README, and the PR description all state that a single config file kept in a dotfiles tree keeps working. Reads do. But `saveConfig`'s `os.Rename(tempFile, configFile)` replaces the link itself with a regular file, so the first access-token refresh removes the dotfiles link and leaves the original stale. `DeleteConfig`'s `os.Remove` likewise removes only the link, so `alpacon logout` leaves the refresh token behind in the dotfiles tree.

The rename predates this PR (2bfd908), but the guarantee written on top of it does not. Either narrow the claim to "reads follow the link, writes replace it", or resolve the link in `saveConfig` before the rename. There is also no test covering what `saveConfig` does to a symlinked `config.json`.

---

### Suggestions (7)

**5. `config/config.go:160-167`—narrow and read open the file twice.** `restrictConfigFileMode` closes and discards the handle it narrowed, and `openConfigForRead` then resolves the path again. The file can change in between, and every call pays for an extra open. Reading the handle `openNarrowedConfigFile` returned, and falling back to `openConfigForRead` only when the failure was `errSymlink`, closes the window and drops the syscalls together. Low priority once critical 1 is fixed.

**6. `config/permissions.go:29-32`—the Windows branch is unreachable.** The callers of `restrictOpenFileMode` are `restrictConfigDirectoryMode` (which returns at line 183), `openNarrowedConfigFile` (line 134), `permissions_linux.go`, and `permissions_darwin.go`. It never runs on Windows. If it stays as defense in depth, say so in a comment.

**7. `config/permissions.go:87-99`—`fstat` is called three times on one handle.** `refuseUnsafeConfigFile` stats, `refuseForeignOwner` stats again, and `restrictOpenFileMode` stats a third time. All three are before any `chmod`, so all three see the same thing. Passing the `os.FileInfo` down collapses them into one.

**8. `config/permissions.go:145,187,194`—error classification is inconsistent within the file.** The comment at line 114 gives the reason to use `errors.Is(err, fs.ErrNotExist)` (`os.IsNotExist` unwraps only the error types the os package defines), and lines 145, 187, and 194 then use `os.IsNotExist`/`os.IsPermission`. It works today because those errors are not wrapped, and it breaks silently the moment someone wraps `openNoFollow`.

**9. `config/permissions_unix.go:50-53,71-74`—the guards fail open.** When the `info.Sys().(*syscall.Stat_t)` assertion fails, both functions return `nil` and the check is skipped. That path is unreachable on Unix, but a function whose job is to refuse should say out loud why its default is to allow.

**10. `config/permissions.go:72`—`asSymlink` reads like the `errors.As` family.** In Go the `As` prefix suggests a type assertion. This function wraps, so something like `wrapSymlinkErr` matches what it does.

**11. `.github/workflows/build-and-test.yaml:23-31`—the comment claims more than holds.** `TestRefuseForeignOwner` (which uses `/`) and `TestRefuseUnsafeConfigFileNamesWhatMustNotBeRead` (which uses `/etc/hosts`) both run on Linux, so macOS is not the only leg exercising the owner refusal. What is macOS-only is `TestLoadConfigRefusesAConfigFileOwnedByAnotherAccount`, the end-to-end case. Separately, `config/permissions_unix_test.go:312` evaluates `os.Link` before `os.Geteuid() == 0`, so as root it creates the link and then skips.

**12. `CLAUDE.md:224`—one bullet is 795 words.** That is 17% of a 4,716-word file in a single item, and CLAUDE.md is loaded into context every session. The detail belongs in a package comment at the top of `config/permissions.go`, where it is read when the file is opened; three or four sentences in CLAUDE.md saying where things live and what is refused would carry the rest.

---

### Requested topics

**Go conventions.** The repository's rules are followed closely: the `defer func() { _ = f.Close() }()` pattern, lowercase error strings without trailing punctuation, `const → var → type → func` ordering, `t.Run` per table case, and serial tests wherever process-wide state is touched (`setupTestConfig`'s `t.Setenv`, the `warnedUnrestricted` reassignment). The two departures are suggestion 8 (mixed error classification) and suggestion 10 (naming). `config/permissions.go` also mixes `runtime.GOOS` branching with build-tag separation; the comparison is a compile-time constant so nothing is paid for it, but this package is already split into `_unix.go`/`_windows.go`, and suggestion 6 would settle it on one method.

**Duplicate and unused code.** Removing `restrictDeviceIDFileMode` and `deviceIDFileExcessPerm` is right. Three things remain: `config.go:180-186` duplicates the first half of `refuseUnsafeConfigFile` (the same place as critical 1), `restrictOpenFileMode`'s Windows branch is unreachable (6), and `refuseUnsafeConfigFile` stats the same handle twice (7). The stubs in `permissions_windows.go` are needed to compile and the comment above them says so, so they are fine.

**Performance.** Small in absolute terms, but multiplied. One `LoadConfig` used to be a single open; it is now two directory opens, two file opens, and roughly five `fstat` calls. A command typically calls `LoadConfig` two to five times (21 call sites), so that is a few dozen extra syscalls per command. Microseconds, and not visible in use—but suggestions 5 and 7 together cut it by more than half and close the TOCTOU window as a side effect. `GetOrCreateDeviceID` calling `restrictConfigDirectoryMode` twice is deliberate, documented, and happens once per installation. `warnedUnrestricted` grows only with the number of distinct (subject, cause) pairs, so it is not a leak.

---

### Highlights

- Every `chmod` is bound to an open handle, and the two path-based reads that remain each carry the reason they are safe.
- Reading the mode back after a successful `chmod` catches a filesystem without permission bits, while still letting a surviving owner execute bit pass, because macOS reports every file on a FAT volume as `0700`.
- The symlink errno is declared per OS (`ELOOP`/`EMLINK`/`EFTYPE`) and the wiring is pinned by a test, rather than a `!windows` build tag standing in for a platform.
- Two hangs on a named pipe are caught by tests that actually time out instead of asserting on a return value.
- Every new test was seen RED once, and the two places that stay uncovered (the Linux owner case, the pre-macOS 13 `EINVAL` branch) are named in the description rather than left to be found.
- The macOS runner and the `darwin` lint leg mean these files no longer ship unread.

---

**Verdict**: CHANGES REQUESTED—for critical 1. The rest can land afterwards, though warnings 2 and 4 are worth at least a documentation fix in this PR.

**Checklist**
- [x] Code follows stack conventions (two exceptions: 8, 10)
- [x] Tests included; CI 16/16 pass, macOS leg included
- [ ] No security vulnerabilities—critical 1 is open
- [x] Documentation updated—warning 4 needs its claim narrowed

config/config.go
jisung-02: Right — a symbolic link skipped the owner refusal entirely, and `/etc/hosts` reached the decoder in a RED run that proved it:

```
Error: Target error should be in err chain:
  expected: "unsafe config file"
  in chain: "failed to decode config file: json: cannot unmarshal number into Go value of type config.Config"
```

Fixed in c624314, close to the suggested diff but reaching it through `openCheckedConfigFile` rather than an inline `refuseUnsafeConfigFile` call: `LoadConfig` now keeps the handle the narrowing already opened and only reopens the path for a symbolic link, running the same refusal on what it resolves to. That also closes suggestion 5's double-open in the same change, since both were the same seam.

`TestLoadConfigRefusesASymlinkToANamedPipe` and `TestLoadConfigStillReadsThroughASymlinkedConfigFile` still pass unchanged, and `TestLoadConfigRefusesASymlinkToAFileOwnedByAnotherAccount` is the new regression guard.

config/permissions_unix.go
jisung-02: Agreed on both counts — the message named only the file, and the README's remedy deletes working credentials for someone on sudo or in a container. Fixed in 345545d with the suggested diff (uid on both sides of the message) plus a README paragraph covering `sudo -E` and a bind-mounted `~/.alpacon`, and telling the reader to check the uids before deleting anything.

config/device.go
jisung-02: Confirmed the vfat behavior is the real gap: `fat_setattr` answers `EPERM` for any mode outside the mount's `fmask`, even one that already grants no other account access, so the chmod fails outright and used to be treated as exposure by default. Fixed in bf5480a — a failed chmod is now read back the same way a successful one already was, so only a kept mode that hands out group/other access counts as a failure. A vfat mount at `fmask=0077` keeps its device id; one with a wider fmask still loses it, and that remaining trade-off is named in `createDeviceIDFileWithoutLink`'s comment and in the README rather than left implicit.

config/config.go
jisung-02: Right, and worse than I'd realized while writing the claim — `DeleteConfig` breaks it too, not just `saveConfig`. Fixed in 6a382ef: both now resolve the link first (`resolveConfigWritePath`) and act on what it names, so a token refresh replaces the target and `alpacon logout` empties the tree instead of unlinking the pointer to it. A link left dangling by logout is written through on the next login rather than replaced. The resolved target gets the same owner/regular-file refusal a read applies — added as `TestSaveConfigRefusesASymlinkToAFileOwnedByAnotherAccount` — since writing the refresh token into a file another account owns is exactly as bad as reading one.

All three cases were seen RED with the resolution stubbed out to return the link path unchanged:
```
"...\"token\": \"first\"..." does not contain "second"
Expected error with "file does not exist" in chain but got nil.
Expected error with "unsafe config file" in chain but got nil.
```

config/config.go
jisung-02: Applied — folded into the same fix as the critical issue (c624314) since both were the same double-open. `openCheckedConfigFile` reads the handle `openNarrowedConfigFile` returned and only reopens the path when that failed as a symlink, so the window and the extra syscalls are both gone together.

config/permissions.go
jisung-02: Applied in cccbca8 — added the defense-in-depth note so the next reader isn't hunting for a Windows path that reaches this line.

config/permissions.go
jisung-02: Applied in 581a682, close to what you described: the three guards now take the `os.FileInfo` the caller already read (`restrictOpenFileMode`, `refuseUnsafeConfigFile`, `refuseForeignOwner` all changed signature), so `openNarrowedConfigFile` stats once and passes it down. The chmod-callback stat (`refuseHardLinked`) and the post-chmod read-back kept their own, as you said they should.

config/permissions.go
jisung-02: Applied in ce08209, all three sites plus the one line 114 already covered — replaced with `errors.Is(err, fs.ErrNotExist)` / `errors.Is(err, fs.ErrPermission)` throughout the file.

Turned out this was live, not just latent: `config.GetActiveWorkSession` had the exact same `os.IsNotExist` mistake on `LoadConfig`'s wrapped error, so a machine with no config file got an error instead of the `""` its own doc comment promises. Fixed alongside in be3220d with a RED-verified regression test.

config/permissions_unix.go
jisung-02: Applied in 59bdc18, as a refusal rather than a comment — both guards now go through a shared `statT` helper that returns an error when the assertion fails, instead of `nil`. Nothing on Unix reaches that branch (`os.File` always carries a `syscall.Stat_t`), so this changes no working setup; it changes what happens if that ever stops holding.

config/permissions.go
jisung-02: Applied in 654df42 — renamed to `wrapSymlinkErr`.

.github/workflows/build-and-test.yaml
jisung-02: Applied in 08d71f5 — narrowed to name `TestLoadConfigRefusesAConfigFileOwnedByAnotherAccount` as the macOS-only case and say why (planting a foreign-owned file needs a hard link, which Linux refuses), rather than claiming the whole owner-refusal path is macOS-exclusive.

config/permissions_unix_test.go
jisung-02: Applied in b9031a3 with the suggested diff — root check moved ahead of the link.

CLAUDE.md
jisung-02: Applied in b65b5a0 — the detail moved to the package comment on `config/permissions.go`, and the CLAUDE.md bullet is now 191 words: which file decides what, which two refusals are fatal (`errUnsafeConfigFile`), that everything else warns, that `device_id` fails closed, and that none of it applies on Windows.

config/config.go
```diff
@@ -169,6 +185,108 @@ func LoadConfig() (Config, error) {
...
+	}
+	if errors.Is(err, errUnsafeConfigFile) {
+		return nil, err
+	}
+	if errors.Is(err, fs.ErrNotExist) {
+		// Wrap with %w so callers can detect the missing-config case
+		// via errors.Is(err, os.ErrNotExist).
+		return nil, fmt.Errorf("config file does not exist: %s: %w", configFile, err)
+	}
+	warnUnrestricted("config file", err)
```
geunwoonoh: **Warning:** this line warns on every command for a setup the README now calls supported.

A symlinked `config.json` leaves `openNarrowedConfigFile` with `errSymlink` and lands here. `warnedUnrestricted` is per process, so each `alpacon` invocation prints "could not restrict config file permissions: ... is a symbolic link ... Another local account may be able to read your credentials while that stands."

The claim can also be false. The link target's mode is never read, so when the dotfiles copy is already `0600`, no other account can read anything and the warning says otherwise. A symlinked `~/.alpacon` has the same shape.

`resolveConfigWritePath` already shows the safe way to look: open the resolved target with `O_NOFOLLOW`, check the handle, act on that. Reading the target's mode here and warning only when it is actually wide would make the sentence true and leave the dotfiles setup quiet. Missed in the first review rather than introduced in this round, so a follow-up is fine.

config/permissions_test.go
```diff
@@ -0,0 +1,240 @@
...
+	}{
+		{"other accounts keep access", 0644, true},
+		{"only the owner execute bit survives", 0700, false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			file, path := openKeptMode(t, tc.mode)
+			info, err := file.Stat()
+			require.NoError(t, err)
+			err = restrictOpenFileMode(file, info, 0600, func(os.FileMode) error { return nil })
```
geunwoonoh: **Suggestion:** the refused-`chmod` branch this round added (`bf5480a`) never runs in a test.

This table only ever injects `func(os.FileMode) error { return nil }`, so it covers the kept-mode-after-success path. The branch that judges a *failed* `chmod` by the mode it left behind—the one that decides whether a Linux vfat user gets a device id at all—has no case. Of the eleven fixes in this round it is the only one that was never seen RED.

`openKeptMode` already hands out a handle at an arbitrary mode, so a `chmodErr` field on the same table covers it: a failing callback at `0700` expecting no error, and a failing callback at `0644` expecting one.

geunwoonoh (dismissed):
## Summary

All 13 points from the previous review are closed in the code, verified by reading the files at `be3220d` rather than by the commit messages. Three of them went further than what was raised, and generalizing one of them turned up a real bug that was already misbehaving.

| # | Finding | Commit | Regression guard |
|---|---|---|---|
| C1 | A symbolic link bypassed the owner refusal | `c624314` | `TestLoadConfigRefusesASymlinkToAFileOwnedByAnotherAccount` |
| W2 | sudo/Docker got the wrong remedy | `345545d` | Message names both uids, plus a README paragraph |
| W3 | Device id lost on Linux vfat | `bf5480a`, `3280cde` | None—see suggestion 2 below |
| W4 | A write replaced a symlinked `config.json` | `6a382ef` | Three tests |
| S5 | Narrow and read opened the file twice | `c624314` | Folded into the C1 fix |
| S6 | Unreachable Windows branch | `cccbca8` | Comment states it |
| S7 | Three `fstat` calls on one handle | `581a682` | — |
| S8 | `os.IsNotExist` on wrapped errors | `ce08209` | Plus a real bug found |
| S9 | Guards failed open | `59bdc18` | `statT` refuses instead |
| S10 | `asSymlink` naming | `654df42` | `wrapSymlinkErr` |
| S11 | CI comment; root check after `os.Link` | `08d71f5`, `b9031a3` | — |
| S12 | 795-word CLAUDE.md bullet | `b65b5a0` | 191 words plus a package comment |

Three went past what was asked:

- **W4** named only `saveConfig`; `DeleteConfig` had the same problem and both are fixed, with the dangling link a logout leaves behind written through by the next login and covered by a test. Creating the temp file beside the resolved target so the rename stays on one filesystem is the right call.
- The same refusals now run on the **write** path (`TestSaveConfigRefusesASymlinkToAFileOwnedByAnotherAccount`). Writing the refresh token into another account's file hands it over as surely as reading one does, which the original review did not raise.
- **S8** was three latent sites; applying it surfaced `GetActiveWorkSession` returning an error instead of the empty session its doc comment promises, on any machine with no config file (`be3220d`), with a RED run recorded.

The new test for C1 uses a symbolic link rather than a hard link, so it runs on Linux too. The end-to-end owner refusal is no longer covered on macOS alone.

---

### Warnings (1)

**1. `config/config.go:213`—a setup this PR documents as supported warns on every command**

A symlinked `config.json` leaves `openNarrowedConfigFile` with `errSymlink`, and the next line warns. `warnedUnrestricted` is per process, so every `alpacon` invocation prints:

```
Warning: could not restrict config file permissions: .../config.json is a symbolic link,
so its target is left alone: ... Another local account may be able to read your
credentials while that stands.
```

Two things are off. The README now says keeping one `config.json` in a dotfiles tree works, and that setup draws a warning on every command. And the claim can be false: the link target's mode is never read, so when it is already `0600` no other account can read anything.

`resolveConfigWritePath` already shows the safe shape—open the resolved target with `O_NOFOLLOW`, check it, act on the handle. Reading the target's mode in the fallback and warning only when it is actually wide would make the sentence true and leave the dotfiles setup quiet. This was missed in the first review rather than introduced in this round.

---

### Suggestions (1)

**2. `config/permissions.go:138-150`—the W3 fix has no regression guard**

The branch that judges a refused `chmod` by the mode it left behind decides whether a vfat user gets a device id at all, and `TestRestrictOpenFileModeJudgesAModeAChmodKept` only ever injects `func(os.FileMode) error { return nil }`. It covers the kept-mode-after-success path; the new branch does not execute. Of the eleven fixes in this round it is the only one that was never seen RED.

Adding a `chmodErr` field to the same table covers it: a failing callback with `0700` expecting no error, and a failing callback with `0644` expecting one. `openKeptMode` already hands out a handle at an arbitrary mode.

Minor, not worth its own commit: `resolveConfigWritePath` flattens its open errors with `%v`, which drops `errSymlink` and the errno. Nothing sorts on them today, and the `refuseUnsafeConfigFile` error is returned unwrapped, so `errors.Is(err, errUnsafeConfigFile)` still holds.

---

### Highlights

- C1 and S5 became one design change rather than two patches. The narrowed handle is the one that gets read, and reopening the path is reached only for a link, so the window between the checks and the decode and the second open went away together.
- `resolveConfigWritePath` handles the dangling link too: logout removes the target and leaves the link, and the next login writes through it. The round trip is pinned by a test.
- `statT` turns "could not look" into a refusal, so no guard whose job is to turn things away defaults to allowing.
- The W3 fix says the same thing in the code, the `createDeviceIDFileWithoutLink` comment, and the README, and names `fmask=0077` as the way out.
- The CLAUDE.md bullet went from 795 words to 191, with the detail moved to a package comment where the person changing the file will read it.

---

**Verdict**: APPROVED

Everything raised is closed, most of it with a test, and the generalization of one point caught a bug that was already affecting users. CI is green on `be3220d` across all 16 checks. The two items above are follow-ups, not blockers.

**Checklist**
- [x] Tests included; CI 16/16 pass
- [x] No security vulnerabilities—the critical issue is closed and guarded
- [x] Documentation updated—README, CLAUDE.md, and the package comment agree

config/config.go
jisung-02: Right on both counts, and the second is the one that mattered: the warning was making a claim it had not checked. `warnedUnrestricted` is per process, so a `config.json` linked into a dotfiles tree printed it on every single invocation, and when that copy was already `0600` the sentence was simply false.

Fixed in f8c6956. `openCheckedConfigFile` holds the warning back on `errSymlink` and speaks only once it has read the target's mode. It reads that mode off the handle the fallback already opened rather than resolving the path again—`openConfigForRead` has the target open and `refuseUnsafeConfigFile` has already stat'ed it, so a second open would cost syscalls and reintroduce the window c624314 closed for this same read.

A symlinked `~/.alpacon` had the same shape, as you noted, so `restrictConfigDirectoryMode` gets the same treatment through `symlinkTargetGrantsOtherAccounts`: `EvalSymlinks`, then `openToNarrow` on the resolved path (`O_NOFOLLOW`), then the mode off that handle. It returns `nil` when nothing is exposed; all four callers pass the return straight to `warnUnrestricted`, so no control flow moved. It answers true when it cannot look—the link will not resolve, the target will not open, the handle will not stat—since a mode nothing could read is not one anything may vouch for.

The judgment itself is now `grantsOtherAccounts`, which the two read-backs inside `restrictOpenFileMode` and the directory pre-check also go through; without it this change would have been a fourth copy of `&0077 != 0`.

`TestLoadConfigWarnsOnASymlinkedConfigFileOnlyWhenItsTargetIsWide` and `TestRestrictConfigDirectoryModeReportsASymlinkOnlyWhenItsTargetIsWide` were both seen RED first. The file one failed with the sentence you quoted:

```
Error: "...config.json is a symbolic link, so its target is left alone: ... too many
levels of symbolic links. Another local account may be able to read your credentials
while that stands.\n" should not contain "could not restrict"
```

Both also assert the link's target keeps the mode it had, since it is never chmod'ed either way.

The `%v` on `resolveConfigWritePath` and `readlinkTarget` is b4aa2eb. README, `CLAUDE.md`, and the package comment moved in the same commit as the behavior.

config/permissions_test.go
jisung-02: Correct, and the callback shape is exactly why it slipped: a table that only ever injects `func(os.FileMode) error { return nil }` cannot reach a branch that starts at `if err := chmod(perm); err != nil`.

Fixed in 56f1771, as you suggested—a `chmodErr` field on the same table, using the handle `openKeptMode` already hands out at an arbitrary mode. Four cases now, and the two new ones were seen RED separately rather than together, since they guard opposite halves of the branch.

Returning `err` unconditionally, which is what the code did before bf5480a:

```
--- FAIL: TestRestrictOpenFileModeJudgesAModeAChmodKept/the_chmod_was_refused_over_a_mode_no_other_account_can_use
    Error: Received unexpected error:
           operation not permitted
```

Returning `nil` unconditionally, the mirror mistake of trusting every refusal:

```
--- FAIL: TestRestrictOpenFileModeJudgesAModeAChmodKept/the_chmod_was_refused_and_other_accounts_keep_access
```

The failing case asserts with `ErrorIs` rather than on the message, since the refusal is handed back as it came and an errno is what a caller would sort on.

geunwoonoh (approved):
## Summary

The three follow-ups from the previous approval are closed, verified by reading the files at `b4aa2eb`.

| Item | Commit | Regression guard |
|---|---|---|
| A symlinked config warned on every command | `f8c6956` | `TestLoadConfigWarnsOnASymlinkedConfigFileOnlyWhenItsTargetIsWide`, `TestRestrictConfigDirectoryModeReportsASymlinkOnlyWhenItsTargetIsWide` |
| The refused-`chmod` branch had no test | `56f1771` | Two `chmodErr` cases on the same table |
| `%v` on the config write path | `b4aa2eb` | — |

The warning fix went past what was raised. The file side answers from the handle `openCheckedConfigFile` already holds, so no open was added; the directory side—which the review only mentioned in passing—got `symlinkTargetGrantsOtherAccounts` applying the same rule one level up, and it answers `true` when it cannot look, so a mode nothing could read still warns.

Narrowing the warning condition loses no coverage. Checked by mode:

| Situation | Result |
|---|---|
| Regular file `0644` on a read-only mount | Warns, immediately—not a link |
| Link to `0600` | Silent; nothing is exposed |
| Link to `0640` / `0644` | Warns |
| Symlinked `~/.alpacon` to `0700` | Silent |
| Symlinked `~/.alpacon` to `0755` | Warns |

The `chmodErr` cases reach the branch that decides whether a vfat user gets a device id at all, and `assert.ErrorIs(t, err, tc.chmodErr)` pins that the errno survives—which is what lets a caller sort on it. The three `%w` conversions restore `errSymlink` and the errno on the write path, and the `refuseUnsafeConfigFile` error is still returned unwrapped, so `errors.Is(err, errUnsafeConfigFile)` holds.

### Findings

None. The new code was read as new code and nothing is left open. Confirmed along the way:

- The four sites folded into `grantsOtherAccounts` are equivalent to the bit tests they replace. `openCheckedConfigFile` testing `& 0077` without `&^ 0600` gives the same answer, since `allowed` is `0600`.
- Both new tests get the serial/parallel call right: the one that swaps stderr and reassigns `warnedUnrestricted` is serial with the reason stated, and the directory one touches no package state and runs parallel.
- `symlinkTargetGrantsOtherAccounts` costs about four extra syscalls per `LoadConfig`, and only for a symlinked config directory.
- CLAUDE.md and the README match the behavior that shipped.

### Highlights

- Moving the warning's question from "is this a link" to "does what it names actually grant access" is the right one to ask, and it is answered from a handle that was already open.
- `symlinkTargetGrantsOtherAccounts` says in its own comment why it fails toward warning, so the next reader does not have to infer the direction.
- A repeated `&0077` test became one named function, which makes it visible that four sites were asking the same question.
- The test names carry the judgment rather than the input: "the chmod was refused over a mode no other account can use" says what the case is for.

---

**Verdict**: APPROVED

All 13 points from the first review and both follow-ups from the second are closed. CI is green on `b4aa2eb` across all 16 checks.

**Checklist**
- [x] Tests included; CI 16/16 pass
- [x] Documentation updated—README, CLAUDE.md, and the package comment agree

PR #443, opened 2026-09-09
feat(exec): submit a script as a verified file execution (ADR 0053)

### LINKED ISSUE #405: feat(exec): submit a script as a verified file execution instead of a shell line (ADR 0053)
Client half of ADR 0053 for the CLI. The server, agent, and assessor sides are open (alpacax/alpacon-server#3158, alpacax/alpamon#431, alpacax/alpacon-ai-server#406); nothing here can reach the lane yet.

## The problem this closes for a CLI user

Running a script today means `alpacon exec <server> "bash /opt/deploy.sh"`. That line names a path. Neither the assessor nor the approver can see what the path holds, so the assessor's confidence is structurally low and the baseline routes it to a human — every run, including the tenth identical one.

The new lane asks for the script as structure. The platform hashes the bytes the reviewer sees, and the agent proves at execution time that the file it runs is those bytes. An approver can then mark it standing, and an unchanged re-run stops paging anyone.

## What the server accepts

`POST /api/events/commands/` takes a nested `file` object instead of `line`:

```json
{
  "server": "<uuid>",
  "username": "root",
  "groupname": "root",
  "file": {
    "path": "/opt/deploy.sh",
    "interpreter": "/bin/bash",
    "args": ["--fast"],
    "content": "#!/bin/bash\nset -euo pipefail\n..."
  }
}
```

Rules worth knowing before writing the client:

- **`line` and `data` must not be sent.** The server derives both and refuses a request that carries either — `FILE_EXEC_LINE_NOT_ALLOWED` / `FILE_EXEC_DATA_NOT_ALLOWED`. Key presence is what it checks, so an empty string is still a violation.
- **`path` and `interpreter` must be POSIX-absolute.** `bash script.sh` would let host `PATH` decide what actually runs, which is the one gap that could split what was approved from what executes.
- **`content` is submitted verbatim, byte for byte.** The agent hashes the file on disk with no normalization, so trimming a trailing newline makes every execution fail closed. Do not strip, re-encode, or normalize line endings.
- **64 KB cap** (`FILE_EXEC_CONTENT_TOO_LARGE`), and an empty body is refused (`FILE_EXEC_EMPTY_CONTENT`).
- The content is the bytes that get hashed, but the agent reads the file **from the host's disk**. The submitted content is what the reviewer and the assessor judge; it never travels to the agent. So the file has to already exist at `path` on the target.

## Refusals the CLI should render as guidance, not stack traces

| code | what to tell the user |
|---|---|
| `FILE_EXEC_UNSUPPORTED_AGENT` | the target's alpamon is too old to verify a digest; name the required version |
| `FILE_EXEC_ASSESSOR_DISABLED` | the lane needs the assessor, and this deployment has it off |
| `FILE_EXEC_INVALID_PATH` | path and interpreter must be absolute |
| `FILE_EXEC_CONTENT_TOO_LARGE` / `FILE_EXEC_EMPTY_CONTENT` | size bounds |
| `FILE_EXEC_LINE_NOT_ALLOWED` / `FILE_EXEC_DATA_NOT_ALLOWED` | a client bug, not a user error |

## Shape of the work

`api/event/types.go`'s `CommandRequest` is flat today (`shell`, `line`, `data`, …). It needs a nested optional `File` struct, omitted for every existing call so no current request changes shape.

The UX decision is what the surface looks like. A flag on `exec` that reads a local file is the obvious candidate, but the file has to exist on the **target** as well, so a flag that only reads locally would mislead. Worth settling before implementing: does the CLI read the bytes from a local copy the user names, or require the user to pass content another way? The first is convenient and is the only shape that lets the user see what they are submitting; it also quietly assumes the local copy matches the remote one, and the digest check is exactly what catches that assumption being wrong.

## Not in scope

Approving is console work (alpacax/alpacon-web#1943). The CLI submits; it does not grow an approval surface.

## Reference

ADR 0053, and the server tracking issue alpacax/alpacon-server#3088.

junho226:
Closes #405

## Summary

`alpacon exec --file` submits a script as a verified file execution (ADR 0053) instead of a shell line. The CLI reads the script from a local path, sends the bytes unchanged, and the platform hashes what the reviewer sees; the agent proves at execution time that the file it runs on the server is those bytes. An approver can then mark the run standing, and an unchanged re-run stops paging anyone.

```
alpacon exec --file /opt/deploy.sh root@prod-web -- --fast
alpacon exec --file /opt/deploy.sh --file-from ./deploy.sh --interpreter /bin/sh prod-web
```

## Changes

- `--file PATH` runs the script at PATH on the target. The content is read from the same path locally, or from `--file-from`, byte for byte: no trimming, no newline normalization. `--interpreter` defaults to `/bin/bash`; both must be absolute. Script arguments go after `--`. A command line or `--env` does not combine with `--file`.
- A missing, empty, oversized (64 KB) or non-UTF-8 local file is refused before anything reaches the network. Invalid UTF-8 matters because `encoding/json` would substitute U+FFFD and the digest could never match.
- `CommandRequest` gains an optional nested `File`. A file-lane body carries no `line`, `data` or `env` key at all, because the server refuses those by key presence, so it marshals through a lane-specific shape rather than relying on `omitempty`. A reflection test keeps that shape from drifting from `CommandRequest`.
- The retry, presence step-up and approval-wait layers take a run closure, so the file lane goes through the same layers as a shell line; the generic lane's behavior is unchanged and pinned by tests.
- File-lane refusals from the server render as guidance: an old agent or a disabled assessor names the ordinary command to fall back to; `line_not_allowed` and `data_not_allowed` are reported as CLI bugs.
- The re-run hint quotes file-lane values with a dedicated quoter. Those values were argv on the first run and reached the server as a JSON list, so a hint that left `a;b` or `$(...)` unquoted would have let the metacharacter execute locally when pasted.
- Help and README state what an approval binds (exactly those bytes, on that server, as that account, with those arguments; not "the file is safe") and where a secret goes (a file or the environment the script reads at run time, never an argument, since arguments are recorded and shown to the reviewer).

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` pass; `GOOS=windows go build ./...` passes; golangci-lint v2.9.0 (the CI pin) reports 0 issues.
- New tests cover flag parsing and its refusals, local file checks (empty, at and over the ceiling, non-UTF-8, missing, directory), the file-lane body with the refused keys absent, nil args sent as `[]`, refusal rendering in table and JSON mode, the re-run hint with metacharacters, and two helper-process runs of the real command against a fake server.
- Adversarial review, two rounds (security, correctness, simplicity, tests): 0 critical. Confirmed warnings are fixed above.

## Left for follow-up

- Six of the nine refusal entries are unreachable from this binary because local guards fire first. They are kept because the codes are the wire contract and the server's limits can change independently of a deployed CLI.
- Server ceilings on `args` (64 entries, 1024 chars each), `path` (1024) and `interpreter` (256) are neither pre-checked nor documented; they answer a bare DRF code.
- The approved-then-re-run branch of the file lane is exercised only through the helper-process tests.


cmd/exec/file_test.go
```diff
@@ -0,0 +1,602 @@
...
+			message, hint, ok := fileExecRefusal(tt.err, "prod-web")
+			assert.Equal(t, tt.wantOK, ok, "message %q", message)
+			assert.Equal(t, tt.wantMessage, message)
+			if tt.wantNoHint {
+				assert.Empty(t, hint)
+			}
+			if tt.wantHintPart != "" {
+				assert.Contains(t, hint, "Hint:")
+				assert.Contains(t, hint, tt.wantHintPart)
+				assert.Equal(t, "\n", hint[len(hint)-1:], "hint must end with a newline: %q", hint)
```
jisung-02: Warning: this indexes an empty string when the hint regresses to "". `hint[len(hint)-1:]` with `len(hint) == 0` is `hint[-1:]`, a runtime slice-bounds panic, and the `assert.Contains` above it reports without stopping—so a dropped hint costs the case its readable failure and takes the rest of the table's output with it.

`CLAUDE.md` names the shape for a suffix check under "Assertion conventions": there is no dedicated helper, so `True`/`False` carry the operand in the message. `strings` is already imported here.

```suggestion
				assert.True(t, strings.HasSuffix(hint, "\n"), "hint must end with a newline: %q", hint)
```

jisung-02 (commented):
## Summary

`alpacon exec --file` submits a script as a verified file execution (ADR 0053) instead of a shell line: the CLI reads the bytes from a local path, sends them unchanged, the platform hashes what the reviewer sees, and the agent proves at execution time that the file on the target is those bytes. The generic lane keeps its wire shape, and the retry, presence step-up and approval-wait layers were lifted to take a run closure so both lanes go through the same layers.

I read the wire contract, the byte-for-byte read path, the local-shell exposure in the re-run hint, and the refactor's equivalence against the generic lane. I did not read the server or Alpamon side, and I did not run the suite—CI is 14/14 green.

The design holds up where it matters. `MarshalJSON` switching to `fileLaneRequest` is the right answer to a server that refuses by key presence, and `TestFileLaneRequestMirrorsCommandRequest` is what keeps a hand-copied field list from drifting silently. `argvQuote` is a real finding by the author: those values were argv, so reusing the whitespace-only `shellQuote` in the hint would have let `$(...)` run locally when pasted. Refusing invalid UTF-8 locally, with the reason stated (`encoding/json` substitutes U+FFFD and the digest can never match), is the kind of guard that is usually missing. And the two `clientBug` codes are only ever printed after matching a literal in the table, so a server-sent code cannot smuggle escapes into a hint.

Nothing here blocks the merge. What follows is three small warnings and four suggestions.

### Critical issues (0)

None.

### Warnings (3)

1. **[cmd/exec/file_test.go:370]** `hint[len(hint)-1:]` panics with a slice-bounds error when `hint` is empty, so a regression that drops the hint gets a panic instead of a readable failure—the preceding `assert.Contains` reports but does not stop. `CLAUDE.md` names the shape for a suffix check: `assert.True` with the operand in the message.
2. **[cmd/exec/file_test.go:474]** The helper marker is spelled out as `"exec-worksession-helper"` while `worksession_command_test.go:95` already holds it as `execWorkSessionHelperMarker`. A second copy of a constant is on the recurring-review list. Separately, `runFileLaneHelper` forks `runHelperProcess` (`pending_approval_test.go:107`) to get past its `require.Error`—loosening that one so callers assert their own exit code would remove the fork.
3. **[api/event/types.go:142]** The comment says `MarshalJSON` leaves `line`, `env` and `data` out, but it also drops `shell`—`fileLaneRequest`'s own comment says so and `TestCommandRequestMarshal_FileLaneOmitsRefusedKeys` asserts it. A comment that lists three of the four omitted keys is the one a future reader trusts.

### Suggestions (4)

1. **[cmd/exec/file.go:243]** `HandleFileExecRefusal`'s tail is the credential-denial branch of `HandleCommandResult` (`run.go:851-859`) line for line: the JSON envelope, the red `Error` line, the hint, `os.Exit(1)`. Extracting one `reportCodedRefusal(message, hint string, err error)` would leave each site with its own table and one shared way out.
2. **[cmd/exec/run.go:523]** The file lane calls `runWithPresenceStepUp` directly, while `RunExecWithApprovalWait` goes through the `runPresenceStepUp` package seam. The behavior is the same, but every approval-wait test in `run_test.go` replaces that seam, so none of them reach the file lane—which is the coverage gap the PR body already records as a follow-up. A seam over `commandRun` would let both lanes share those cases.
3. **[cmd/exec/file_test.go:382]** `TestReRunHint` already lives in `sudo_hint_test.go:432`, and `parse_test.go` owns the `ParseRemoteExecArgs` tables. Moving `TestReRunHint_File` and the three `TestParseRemoteExecArgs_File*` tests to those files would leave `file_test.go` holding the file lane's own subjects—`loadFileExecution`, `fileExecRefusal`, and the helper-process runs. Worth weighing against keeping the lane in one place; either answer is defensible, but the split is currently invisible from `parse_test.go`.
4. **[cmd/exec/file.go:211]** `strings.Contains(message, "%s")` decides whether an entry needs the server name, twice. A `needsServer bool` on the table entry states it instead of sniffing for it, and a future message with a literal `%` in prose cannot surprise it.

### Highlights

- `fileLaneRequest` plus the reflection mirror test: the hand-copied field list is the fragile part of a second wire shape, and the test fails the build rather than letting a new `CommandRequest` field silently never travel.
- `argvQuote` with the reason written down. The generic lane wants metacharacters to reach the remote shell; the file lane's values were argv and must not reach any shell. Two quoters is the correct answer here, not duplication.
- Local guards before the network—empty, at and over the 64 KB ceiling, non-UTF-8, missing, a directory—each with its own test, and `TestExecFileLocalRefusalNeverReachesServer` asserting the server saw nothing.
- The README and `--help` say what an approval does *not* bind, and where a secret goes. That is the sentence most feature docs leave out.

### Checklist

- [x] Tests included and CI green (14/14)
- [x] No security vulnerabilities found
- [x] Documentation updated (README, `--help`)

Verdict: COMMENTED—no blockers, seven items above.

cmd/exec/file_test.go
```diff
@@ -0,0 +1,602 @@
...
+// fake server, with the script at scriptPath, and returns what it exited with.
+func runFileLaneHelper(t *testing.T, serverURL string, args ...string) (exitCode int, stdout, stderr string) {
+	t.Helper()
+	home := t.TempDir()
+	writeExecCommandTestConfig(t, home, serverURL)
+
+	helperArgs := append([]string{
+		"-test.run=^TestExecCommandWorkSessionGateHelperProcess$",
+		"--",
+		"exec-worksession-helper",
```
jisung-02: Warning: this is the second copy of a constant. `worksession_command_test.go:95` declares `execWorkSessionHelperMarker = "exec-worksession-helper"`, and the marker has to agree with what `TestExecCommandWorkSessionGateHelperProcess` checks for—a rename there leaves this literal pointing at nothing, and the failure surfaces as a helper process that exits without running the command.

```suggestion
		execWorkSessionHelperMarker,
```

Related, and larger than one line: `runFileLaneHelper` reimplements `runHelperProcess` (`pending_approval_test.go:107`). The one thing that stops you reusing it is its `require.Error(t, err, "helper exited 0")`, which `TestExecFileDetachSubmitsFileLane` cannot satisfy because it wants exit 0. Dropping that `require` and letting each caller assert its own exit code would fold this helper back into the shared one—the existing callers all assert an exit code already, so none of them lose anything.

api/event/types.go
```diff
@@ -134,6 +135,41 @@ type CommandRequest struct {
 	// because every exec path surfaces the demand to its caller instead of
 	// stalling on it.
 	PurposeDemandSupported bool `json:"purpose_demand_supported,omitempty"`
+	// File selects the verified file lane (ADR 0053). Its presence is what the
+	// server switches on, so it is omitted from every generic-lane request and
+	// no existing request changes shape. When set, MarshalJSON leaves line, env
+	// and data out of the body entirely—see fileLaneRequest.
+	File *FileExecution `json:"file,omitempty"`
```
jisung-02, line 142: Warning: the comment lists three of the four keys `MarshalJSON` drops. `shell` goes too—`fileLaneRequest`'s own doc comment says "minus shell, line, env and data", and `TestCommandRequestMarshal_FileLaneOmitsRefusedKeys` asserts `shell` is absent alongside the other three.

This is the comment a reader consults before adding a field, so the omission is the kind that gets trusted. Reading "MarshalJSON leaves shell, line, env and data out of the body entirely" would match the code and the test.

cmd/exec/file.go
```diff
@@ -0,0 +1,249 @@
...
+func HandleFileExecRefusal(err error, serverName string) bool {
+	message, hint, ok := fileExecRefusal(err, serverName)
+	if !ok {
+		return false
+	}
+	if utils.OutputFormat == utils.OutputFormatJSON {
+		utils.CliErrorEnvelopeWithExit("command", err, "%s.", message)
+		return true
+	}
+	fmt.Fprintf(os.Stderr, "%s: %s.\n", utils.Red("Error"), message)
```
jisung-02: Suggestion: this tail is the credential-denial branch of `HandleCommandResult` (`run.go:851-859`) line for line—the JSON-mode envelope through `CliErrorEnvelopeWithExit`, the red `Error` line, the hint, `os.Exit(1)`. Two copies of that sequence is where they start to drift: a change to how a coded refusal renders (a `--output json` field, an exit code other than 1) now has to be made twice, and nothing fails if only one site gets it.

A shared `reportCodedRefusal(operation string, err error, message, hint string)` in `run.go` would leave each site with the part that is genuinely its own—the code table and the guidance—and one way out. Not a merge blocker; worth doing while both copies are still identical enough to diff by eye.

cmd/exec/run.go
```diff
@@ -495,7 +509,29 @@ func pollApprovalOnce(ac *client.AlpaconClient, cmdID string) (approvalOutcome,
...
+	}, waitTimeout, out)
+}
+
+// RunFileExecWithApprovalWait is RunExecWithApprovalWait for the verified file
+// lane (ADR 0053). Only the submission differs—the file object in place of a
+// command line—so the presence step-up, the approval wait and the re-run after a
+// grant are the same layers the generic lane runs through.
+func RunFileExecWithApprovalWait(ac *client.AlpaconClient, serverName string, file event.FileExecution, username, groupname, workSessionID, purpose string, waitTimeout time.Duration, out io.Writer) error {
+	return runWithApprovalWait(ac, func() error {
+		return runWithPresenceStepUp(ac, serverName, func() error {
```
jisung-02, line 523: Suggestion: the file lane calls `runWithPresenceStepUp` directly, while `RunExecWithApprovalWait` (line 512) reaches the same code through the `runPresenceStepUp` package seam. Behaviorally the two are identical, so this is not a bug—but every approval-wait test in `run_test.go` works by replacing that seam, which means none of them can reach the file lane. That is the same gap the PR body records ("the approved-then-re-run branch of the file lane is exercised only through the helper-process tests"), and this line is where it comes from.

Since `runWithApprovalWait` already takes a `commandRun`, a seam of that type—`var runCommandRun = func(run commandRun) error { return run() }`, or simply making the two public wrappers both go through one seam—would let the existing cases cover both lanes instead of one. The helper-process tests are the right end-to-end check, but they cannot drive a rejected-mid-wait or grant-went-elsewhere branch as cheaply as those cases do.

cmd/exec/file_test.go
```diff
@@ -0,0 +1,602 @@
...
+				assert.Empty(t, hint)
+			}
+		})
+	}
+}
+
+// TestReRunHint_File pins the re-run a pending approval prints on the file
+// lane: only the flags the user gave, the arguments after --, and a quoted
+// argument that carries a space.
+func TestReRunHint_File(t *testing.T) {
```
jisung-02, line 381: Suggestion on where these tests live, offered as a question rather than a request.

`TestReRunHint` is already in `sudo_hint_test.go:432`, and `parse_test.go` holds the `ParseRemoteExecArgs` tables including its error table. So `reRunHint` and `ParseRemoteExecArgs` each now have cases in two files, and from `parse_test.go` there is nothing that says a `--file` table exists elsewhere—which is how a later edit to the parser updates one table and misses the other.

Moving `TestReRunHint_File` beside `TestReRunHint`, and `TestParseRemoteExecArgs_File`/`_FileErrors`/`_GenericLaneHasNoFile` into `parse_test.go`, would leave `file_test.go` holding what only the file lane has: `loadFileExecution`, `fileExecRefusal`, and the helper-process runs. The counter-argument—keeping one lane readable in one file—is real, so if you would rather keep them here, a line in each existing file pointing at `file_test.go` does most of the same work.

cmd/exec/file.go
```diff
@@ -0,0 +1,249 @@
...
+	code, _ := utils.ParseErrorResponse(err)
+	if code == "" {
+		return "", "", false
+	}
+	for _, r := range fileExecRefusals {
+		if r.code != code {
+			continue
+		}
+		message = r.message
+		if strings.Contains(message, "%s") {
```
jisung-02: Suggestion: `strings.Contains(message, "%s")` decides whether an entry takes the server name, and line 216 does it again for the hint. It works today because the two entries that need it are the only ones carrying a verb, but it infers a table entry's intent from its text instead of reading it.

A `needsServer bool` on the entry states it, and then `fmt.Sprintf` runs unconditionally for those entries and never for the rest. It also removes the one way this can go wrong quietly: a future message with `%s` in prose gets the server name spliced into it, and a message that needs the name but is written without a verb silently prints without it—`Sprintf` would at least append `%!(EXTRA string=...)` and make it visible.

cmd/exec/file_test.go
junho226: Fixed. `assert.True(t, strings.HasSuffix(hint, "\n"), ...)`, so an empty hint reads as a failed assertion with the operand in the message rather than a slice-bounds panic.

cmd/exec/file_test.go
junho226: Fixed, both halves. The literal is `execWorkSessionHelperMarker`, and `runFileLaneHelper` is gone: `runHelperProcess` now returns exit 0 instead of requiring an error, so the file-lane tests call `runExecHelper` like the others. The existing callers all asserted an exit code already, so none of them changed.

api/event/types.go
junho226: Fixed. The comment now reads "MarshalJSON leaves shell, line, env and data out of the body entirely".

cmd/exec/file.go
junho226: Done. `reportCodedRefusal(operation string, err error, message, hint string)` in run.go is the one way out; `HandleFileExecRefusal` and the inline-credential branch of `HandleCommandResult` both call it and keep only their own table and guidance.

cmd/exec/run.go
junho226: Left as is for this PR, recorded as a follow-up. Moving both lanes onto one `commandRun` seam means changing the signature `run_test.go` replaces in about twenty places, and I would rather do that as its own change with the file-lane approval-wait cases added at the same time, than fold a test-infrastructure rewrite into this one. The behavior is identical today, as you note.

cmd/exec/file_test.go
junho226: I kept the file lane in one file and took your second option: `parse_test.go` and `sudo_hint_test.go` now each carry a line pointing at the `--file` tables in `file_test.go`, so an edit to the parser or the hint finds the other table from where it starts.

cmd/exec/file.go
junho226: Done. Each entry carries `needsServer bool`; `Sprintf` runs unconditionally for those and never for the rest. Your second failure mode showed up immediately: the assessor-disabled message had no placeholder while its hint did, so it printed `%!(EXTRA string=...)` under the new rule. The message now names the server too.

cmd/exec/pending_approval_test.go
```diff
@@ -119,7 +119,9 @@ func runHelperProcess(t *testing.T, workspaceURL, testName, marker string, env,
 	helper.Stderr = &errBuf
 
 	err := helper.Run()
-	require.Error(t, err, "helper exited 0; stderr: %s", errBuf.String())
+	if err == nil {
+		return outBuf.String(), errBuf.String(), 0
```
jisung-02, line 123: Warning: this line changes `runHelperProcess`'s contract, and the doc comment on `runExecHelper` (lines 79-80) still states the old one:

> and returns its stdout, stderr, and exit code. Every caller asserts a refusal,
> so a clean exit fails the test rather than returning 0.

Both clauses are now false. A clean exit returns 0 here instead of failing, and `TestExecFileDetachSubmitsFileLane` is a caller that asserts exit 0 rather than a refusal. That comment is the only place the guard's absence was written down, so a later test that forgets to assert an exit code will read it and believe it is covered.

I checked every call site before saying the removal is safe: all twenty bind `exitCode` and assert on it, so nothing loses a check today. Only the comment needs to catch up — something like "returns its stdout, stderr, and exit code; the caller asserts the exit code it expects."

GitHub will not let me anchor a comment on line 79, which is why this one sits here.

cmd/exec/file.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	for _, r := range fileExecRefusals {
+		if r.code != code {
+			continue
+		}
+		message, hint = r.message, r.hint
+		if r.needsServer {
+			message = fmt.Sprintf(message, serverName)
+			hint = fmt.Sprintf(hint, serverName)
+		}
```
jisung-02, line 219: Suggestion, and it is the mirror of the failure mode you just fixed on the assessor-disabled entry.

`needsServer` now governs both format strings, so an entry that needs the server name in its message but carries no hint runs `fmt.Sprintf("", serverName)`. That does not return an empty string—it returns `%!(EXTRA string=prod-web)`:

```
fmt.Sprintf("", "prod-web")  =>  "%!(EXTRA string=prod-web)"
```

Which is non-empty, so `case hint != ""` below is taken and the user reads `Hint: %!(EXTRA string=prod-web)`. Today no entry hits it—both `needsServer` entries have a hint—but the coupling that produced the bug you found (message without a placeholder, hint with one) is still there in the other direction.

Guarding the hint costs one line and removes the whole class:

```suggestion
		if r.needsServer {
			message = fmt.Sprintf(message, serverName)
			if hint != "" {
				hint = fmt.Sprintf(hint, serverName)
			}
		}
```

cmd/exec/run.go
```diff
@@ -807,18 +849,27 @@ func HandleCommandResult(err error, invokedAs Invocation) {
...
 		if isCommandInlineCredentialError(err) {
-			if utils.OutputFormat == utils.OutputFormatJSON {
-				utils.CliErrorEnvelopeWithExit("command", err, "%s.", commandInlineCredentialMessage)
-				return
-			}
-			fmt.Fprintf(os.Stderr, "%s: %s.\n", utils.Red("Error"), commandInlineCredentialMessage)
-			fmt.Fprint(os.Stderr, credentialInlineHint(invokedAs))
-			os.Exit(1)
+			reportCodedRefusal("command", err, commandInlineCredentialMessage, credentialInlineHint(invokedAs))
 		}
```
jisung-02, line 854: Suggestion: the branch lost its `return` in the move to `reportCodedRefusal`.

The behavior is unchanged—`reportCodedRefusal` exits on both paths, `os.Exit(1)` in table mode and `CliErrorEnvelopeWithExit` in JSON mode—so nothing is broken. What changed is that the control flow is no longer local: before, the `return` said here that the credential case is terminal, and now you have to open another function to know that `utils.CliErrorWithExit("%s", err)` on the next line is unreachable. If `reportCodedRefusal` ever grows a path that returns, this silently starts printing the raw error underneath the credential guidance.

```suggestion
		if isCommandInlineCredentialError(err) {
			reportCodedRefusal("command", err, commandInlineCredentialMessage, credentialInlineHint(invokedAs))
			return
		}
```

jisung-02 (dismissed):
## Re-review at c191b9f

Six of the seven items from the first round are fixed in the code, not just in the replies—I checked each one against the tree rather than taking the commit message for it. The seventh is deferred with a reason I agree with.

**Verified fixed**

- `cmd/exec/file_test.go:366` — `assert.True(t, strings.HasSuffix(hint, "\n"), ...)`. An empty hint now fails with the operand printed instead of panicking on the slice bound.
- `api/event/types.go:140` — the comment names all four keys `MarshalJSON` drops, `shell` included, and now agrees with `fileLaneRequest`'s own comment and with the test.
- `cmd/exec/run.go:861` — `reportCodedRefusal` is the single way out, and both the file-lane refusal and the inline-credential branch call it. Each site keeps only its table and its guidance.
- `cmd/exec/file.go:31` — `needsServer bool` per entry replaces the `%s` sniff, and `Sprintf` now runs because the entry says so rather than because its text happens to contain a verb.
- `cmd/exec/pending_approval_test.go:122` — `runFileLaneHelper` is gone and the four file-lane tests drive `runExecHelper` with the shared marker constant. I checked all twenty call sites of the three helper wrappers before agreeing the `require.Error` guard could go: every one binds `exitCode` and asserts on it, so no existing test loses a check.
- `cmd/exec/parse_test.go:10` / `cmd/exec/sudo_hint_test.go:432` — the pointer lines to `file_test.go` are the option I offered as the alternative, and they do the job: an edit that starts from either table finds the other one.

**Deferred, and correctly so**

`cmd/exec/run.go:523` — the file lane still bypasses the `runPresenceStepUp` seam. Your answer is the right call: that change rewrites a signature `run_test.go` replaces in about twenty places, and it belongs in a commit that adds the file-lane approval-wait cases at the same time rather than riding along here. The behavior is identical today, the PR body already records the coverage gap, and folding test-infrastructure work into a feature commit is how both halves get reviewed badly.

Worth noting one thing the `needsServer` fix produced: the assessor-disabled message is now server-specific ("this deployment has the command assessor disabled, so 'prod-web' cannot run a verified file"). That reads correctly—the deployment-wide cause with the server the user actually asked about—and the test moved with it.

**New, from this round (3, none blocking)**

1. **[cmd/exec/pending_approval_test.go:79-80]** `runExecHelper`'s doc comment still says a clean exit fails the test and that every caller asserts a refusal. This commit made both false. It is one comment, but it is the only written record of the guard that was just removed.
2. **[cmd/exec/file.go:216]** `needsServer` governs both format strings, so a future entry with a message placeholder and no hint gets `Sprintf("", serverName)` — which returns `%!(EXTRA string=...)`, not "", and then prints as a hint. Same coupling that produced the bug you caught, in the other direction; one line closes it.
3. **[cmd/exec/run.go:852]** The credential branch lost its explicit `return`. Terminal today because `reportCodedRefusal` exits on both paths, but the reader now has to leave the function to know that.

**Verification**

`go build ./...` and `go vet ./...` clean locally at c191b9f. CI is green on the new head. I did not run the suite.

Verdict: APPROVE. The three items above are a stale comment and two latent traps; none of them should hold the merge.

cmd/exec/pending_approval_test.go
junho226: Fixed. The comment now reads "returns its stdout, stderr, and exit code; the caller asserts the exit code it expects."

cmd/exec/file.go
junho226: Fixed, as suggested: the hint is formatted only when present, so a needsServer entry without a hint cannot print EXTRA as one.

cmd/exec/run.go
junho226: Fixed. The explicit `return` is back, so the branch reads as terminal without opening `reportCodedRefusal`.

jisung-02 (approved):
## Re-review at 0c0e3ca

All three items from the previous round are fixed, and the fix is three lines with no collateral. Re-approving, since the new commit dismissed the earlier approval.

**Verified in the tree**

- `cmd/exec/pending_approval_test.go:78-80` — the comment now says the caller asserts the exit code it expects, which is what `runHelperProcess` actually does since it started returning 0 on a clean exit. I grepped the package for any other place still describing the removed guard; there is none, and the two sibling wrappers inherit the corrected sentence rather than restating it.
- `cmd/exec/file.go:216` — `hint` is formatted only when present, so a `needsServer` entry with no hint can no longer turn `Sprintf("", serverName)` into `Hint: %!(EXTRA string=...)`.
- `cmd/exec/run.go:853` — the explicit `return` is back, and the credential branch reads as terminal without opening `reportCodedRefusal`.

**One thing left, and it is about the PR body**

The follow-up list still carries "The approved-then-re-run branch of the file lane is exercised only through the helper-process tests." That states the symptom, but the cause and the agreed remedy live only in the thread on `run.go:523`: the file lane calls `runWithPresenceStepUp` directly instead of going through the `runPresenceStepUp` seam, so the twenty-odd approval-wait cases in `run_test.go` cannot reach it, and the fix is to move both lanes onto one `commandRun` seam and add the file-lane cases in that same change.

Threads are the first thing that gets hard to find after a merge. A line naming the seam and the remedy, written as work still to do, would carry the decision to whoever picks it up:

> - The file lane bypasses the `runPresenceStepUp` seam, so `run_test.go`'s approval-wait cases cover the generic lane only. Moving both lanes onto one `commandRun` seam, and adding the file-lane cases with it, is deferred to its own change.

Not a blocker, and not a reason to hold the merge if you would rather open an issue for it instead.

**Verification**

`go build ./...` and `go vet ./...` clean locally at `0c0e3ca`. All 14 CI checks pass on this head. I did not run the suite.

Verdict: APPROVE.

PR #445, opened 2026-09-10
ci(release): demote a hand-published release to a draft before the tests run

jisung-02:
## Summary

The v1.13.1 release run failed in `install-script / end-to-end` with a 404 for `alpacon-1.13.1-windows-amd64.zip` (run 34449641934). The release had been written in the GitHub UI and published from there: Publish created the tag, the tag started the workflow, and by then v1.13.1 was a public release with no assets and the one `releases/latest` resolved to. The end-to-end job installs from latest, so it tried to install this very tag before goreleaser had built anything, and every `alpacon update` and Windows one-liner in that window hit the same 404. The releases before it (v1.11.1, v1.11.2, v1.12.0) went through the UI the same way and passed only because the end-to-end job did not exist yet (#419).

Fixing that exposed a second problem, which the review found: the release `create-release` makes has never been the one that goes public. goreleaser v1 looks a release up by tag, that lookup never returns a draft, so on every tag since v1.6.0 goreleaser created its own release and published it itself, leaving the workflow's draft behind (eight orphans by `github-actions[bot]`), and `publish-release` has been a no-op. Demoting a hand-published release to a draft runs into the same lookup: the v1.13.1 recovery left the UI notes in an orphan draft and published goreleaser's commit list instead.

## Changes

- f157040, a003a3a `create-release` first demoted a public release with no assets, with the lookup assigned to a variable so a transient `gh release view` failure fails the step. 0af543e widens that: any release the tag already has is edited with `--draft --title "$TAG"`, so the step always leaves exactly one candidate for goreleaser, a draft titled with the tag. A UI-published release with a hand-typed title and a finished release being re-run both used to miss the draft search; the edit is a no-op on the draft the job made itself. The jq shape check is gone with the distinction it made.
- ca16df3, 65419d3 The goreleaser config moves to v2: `version: 2`, `format_overrides.formats` in place of the singular `format`, the action pinned to `~> v2`. The Pester assertion and `TestAssetNamesStillMatchTheGoreleaserTemplates`, the two readers of that key, follow the rename; the Go test was what turned `build-and-test` red on ab13010.
- ab13010 `release.use_existing_draft: true` makes goreleaser search the draft list for a release titled with the tag. `mode: keep-existing` keeps that draft's notes and `draft: true` leaves it a draft so `publish-release` is what makes it public after packagecloud has its packages.
- 0af543e `release.replace_existing_artifacts: true`: a re-run of a finished release meets its own assets on the demoted draft, and goreleaser's upload otherwise stops at the first one that already exists.
- 192f03e Whether a tag is a prerelease is decided once, by `prerelease: auto`. goreleaser sets the flag from its own semver judgment whenever it creates or reuses the draft, so the `--prerelease` `create-release` passed was overwritten before anything read it, and `publish-release` now reads `isPrerelease` back from the release to choose `--latest`, instead of judging the tag a second time with a hyphen glob that disagrees with semver on build metadata.

After this, a release started either way ends as one release: the draft `create-release` made, or the UI release demoted to a draft, carrying its own notes, with goreleaser's assets attached, made public by `publish-release`. A whole-workflow re-run of a finished release takes the same path, and hides the release until `publish-release` runs again.

## Testing

- `go test ./pkg/selfupdate` passes on 192f03e. The rename was seen red by mutating the override to `formats: [tar.gz]`: `expected: []string{"formats: [zip]"} actual: []string{"formats: [tar.gz]"}`.
- `actionlint .github/workflows/release.yaml` reports nothing; `bash -n` on each of the three `run:` blocks passes.
- goreleaser v2.18.1, installed locally: `goreleaser check` reports the configuration valid and warns that `brews` and `dockers` are deprecated, which it also does on main. `goreleaser release --snapshot --clean --skip=docker,homebrew,publish` was run on ab13010 and exits 0 with `alpacon-1.13.1-SNAPSHOT-<sha>-windows-amd64.zip`, the matching `-checksums.sha256` and `alpacon_1.13.1-SNAPSHOT-<sha>_linux_amd64.deb`; the three commits since touch neither the archives nor the packages.
- Read in goreleaser v2.18.1's `internal/client/github.go`: `findRelease` calls only `findDraftRelease` when `use_existing_draft` is set, which matches `r.GetDraft() && r.GetName() == name` with the tag as `name`; `CreateRelease` builds the request with `Prerelease: &ctx.PreRelease` unconditionally and `createOrUpdateRelease` sends it to the matched draft; `Upload` returns an unrecoverable error on a 422 unless `replace_existing_artifacts` is set. In gh, `FetchRelease` looks up the published release and the draft in parallel, so `gh release view` and `gh release edit` find the draft by its tag.
- `gh api repos/alpacax/alpacon-cli/releases/generate-notes -f tag_name=v1.13.2 -f previous_tag_name=v1.13.1` returns `"name":"v1.13.2"`, so `gh release create --generate-notes` titles the draft with the tag, and `gh release list` shows every release and draft in the repository titled that way.
- The Pester suite was not run here; this machine has no `pwsh`. The `install-script / unit` jobs on this PR are the real run.
- The workflow steps themselves run only on a tag push, so the next release is the first real run of the whole path, and a whole-workflow re-run of a finished release the first run of `replace_existing_artifacts`.

## Left for follow-up

- `brews` and `dockers` are deprecated in goreleaser v2 in favor of `homebrew_casks` and `dockers_v2`. Both still run and both predate this PR; the migrations change what the tap publishes and how the image is built, so they belong to their own change. No issue is filed yet.
- The nine orphan drafts already in the repository (v1.6.0 through v1.13.0 by `github-actions[bot]`, plus the v1.13.1 UI notes) stay as they are.


choongbum (commented):
### Summary

The demotion works for its stated goal: `install-script / end-to-end` runs before `goreleaser`, so a hand-published tag stops being `releases/latest` before the installer looks, and goreleaser's own release is a draft until its uploads finish, so no 404 window is left. Two things to fix or track.

### Warnings (2)

**1. `release.yaml:31` — demoting hides the release from goreleaser v1, so goreleaser creates a second release and the notes written in the UI are lost.**

goreleaser v1.26.2 (what `~> v1` resolves to) looks the release up with `GetReleaseByTag`, which never returns drafts, and on 404 creates its own (`internal/client/github.go`, `createOrUpdateRelease`). The repo already shows the result:

| tag | release id | author | draft | assets | body |
|---|---|---|---|---|---|
| v1.12.0 | 382466520 | jisung-02 | no | 18 | 1914 chars — UI-published, not demoted, goreleaser attached to it |
| v1.13.1 | 386075698 | jisung-02 | yes (demoted) | 0 | 898 chars, the UI "What's Changed" notes, now at `releases/tag/untagged-…` |
| v1.13.1 | 386084201 | alpacax-release-publisher[bot] | no | 19 | 500 chars, goreleaser's `## Changelog` |

The hand recovery of v1.13.1 is exactly what this step will do to every UI-published release: the public release carries goreleaser's commit list, the UI notes end up in an orphan draft, and `publish-release`'s `gh release edit --draft=false` resolves to the public one and is a no-op. The same lookup explains the 8 orphan drafts by `github-actions[bot]` since v1.6.0: the draft `create-release` makes has never been the release that goes public, so "publish-release is the one step that makes it public" (PR body, and the comment on line 27) does not describe what runs.

Worth merging anyway — the 404 is real and this closes it — but:

- (a) reword line 27 and the PR body so they stop restating the draft→publish-release model, and say that UI notes do not survive the demotion;
- (b) a follow-up issue for the actual fix: goreleaser v2 has `release.use_existing_draft: true`, which makes it reuse a draft whose name matches `name_template` (`findDraftRelease`) instead of creating one, so `mode: keep-existing` would then preserve the create-release notes or the UI notes. The demotion here becomes the prerequisite that makes the UI-published case work end to end.

**2. `release.yaml:29` — a failed metadata lookup takes the skip branch.**

Same as Copilot's suppressed comment: `$(…)` inside `[ … ]` does not trip `bash -e`, so a transient `gh release view` failure yields `""`, the `else` branch prints "already exists — skipping", and the step exits 0 with the assetless release still public. Assign first so the failure is fatal (suggestion inline).

### Highlights

- The root-cause write-up in the PR body is exact, and the retry path (public with assets → untouched) is kept.
- The jq predicate was checked against the three shapes the branch can see.

**Verdict**: COMMENTED — will approve once (2) is in and (1a) is reworded; (1b) as a separate issue.

.github/workflows/release.yaml
```diff
@@ -22,7 +22,16 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           if gh release view "${GITHUB_REF_NAME}" > /dev/null 2>&1; then
-            echo "Release ${GITHUB_REF_NAME} already exists — skipping."
+            # A release published from the GitHub UI is already public when its
+            # tag starts this run, so it is releases/latest with no assets yet.
+            # Hide it as a draft until publish-release; one with assets is a
+            # finished release being re-run and stays as it is.
+            if [ "$(gh release view "${GITHUB_REF_NAME}" --json isDraft,assets --jq '(.isDraft | not) and (.assets | length == 0)')" = "true" ]; then
```
choongbum: Warning: `$(…)` inside `[ … ]` does not trip `bash -e`, so a transient `gh release view` failure yields `""`, the `else` branch prints "already exists — skipping", and the step exits 0 with the assetless release still public. Assign first so the lookup failure fails the step:

```suggestion
            state="$(gh release view "${GITHUB_REF_NAME}" --json isDraft,assets --jq '(.isDraft | not) and (.assets | length == 0)')"
            if [ "${state}" = "true" ]; then
```

.github/workflows/release.yaml
```diff
@@ -22,7 +22,16 @@ jobs:
...
+              echo "Release ${GITHUB_REF_NAME} is public with no assets — hiding it as a draft until goreleaser uploads them."
+              gh release edit "${GITHUB_REF_NAME}" --draft
```
choongbum: Warning: once this release is a draft, goreleaser v1.26.2 cannot see it (`GetReleaseByTag` never returns drafts) and creates its own release for the tag. v1.13.1 shows the outcome: release 386075698 (your UI notes, 0 assets) is now an orphan draft at `releases/tag/untagged-…`, and the public v1.13.1 is 386084201 with goreleaser's `## Changelog`. Compare v1.12.0 (UI-published, not demoted): one release, 18 assets, UI notes kept. So this step trades the 404 for losing the notes written in the UI on every UI-published release. Details and the goreleaser v2 `use_existing_draft` follow-up in the review summary; at minimum the comment on line 27 ("until publish-release") should stop describing the draft→publish-release path, since the public release has been goreleaser's own for every tag since v1.6.0.

.github/workflows/release.yaml
jisung-02: Applied as suggested in a003a3a. You were right that the substitution inside `[ ... ]` is invisible to `set -e`: a failed `gh release view` became an empty string, the comparison was false, and the step exited 0 on the skip branch with the assetless release still public. The lookup is now assigned to `public_without_assets` first, so its failure fails the step.

.github/workflows/release.yaml
jisung-02: Confirmed against the repository and fixed here rather than deferred. The release list shows exactly what you described: v1.13.1 exists twice (386075698, the demoted UI draft with the What's Changed notes and no assets, now at `untagged-…`; 386084201, goreleaser's public release with 19 assets and its `## Changelog`), v1.13.0 has the same pair, and v1.12.0, never demoted, is one release with its UI notes. v1.26.2's `createOrUpdateRelease` goes through `GetReleaseByTag`, which cannot return a draft.

ca16df3 moves the config to goreleaser v2 (`version: 2`, `formats`, action `~> v2`), and ab13010 sets `use_existing_draft: true`, `draft: true` and `prerelease: auto` next to the existing `mode: keep-existing`. In v2.18.1 `findRelease` then calls `findDraftRelease`, which matches a draft by `GetName() == tag`; every draft here is titled with its tag. On a hit the draft keeps its body and its draft state, and `publish-release` is what flips it public, so the comment on the create-release step and the one above `install-script` now describe what runs. Both comments and the PR body are rewritten around that.

Two things this leaves: `brews` and `dockers` are deprecated in v2 (`goreleaser check` says so; both still run), and a whole-workflow re-run of a finished release creates a second draft because the draft search cannot see the public release. Both are in the PR body under follow-up. The nine orphan drafts stay as they are.

.goreleaser.yaml
```diff
@@ -34,7 +36,7 @@ archives:
   - name_template: '{{ .ProjectName }}-{{ trimprefix .Version "v" }}-{{ .Os }}-{{ .Arch }}'
     format_overrides:
       - goos: windows
-        format: zip
+        formats: [zip]
```
choongbum, line 39: Blocker: `pkg/selfupdate/asset_test.go:79-84` (`TestAssetNamesStillMatchTheGoreleaserTemplates`) reads this file, collects the `archives:` lines prefixed `format:`, and expects `["format: zip"]`. After this rename the list is empty, which is the `test (ubuntu-latest)` / `test (windows-latest)` failure on this PR. Same rename on the Go side; `go test ./pkg/selfupdate` passes on ab13010 with it:

```go
		if strings.HasPrefix(line, "formats:") {
			formats = append(formats, line)
		}
	}
	assert.Equal(t, []string{"formats: [zip]"}, formats,
		"ArchiveName builds .tar.gz everywhere but windows; an archives-level formats would change that too")
```

choongbum (changes requested):
### Summary

Both warnings from yesterday are fixed the way the replies describe, and I re-read goreleaser v2.18.1 and gh to confirm the new path end to end. What holds the approve is the red `build-and-test` on ab13010: the `format` → `formats` rename in ca16df3 has a second reader besides the Pester test, and it was not updated.

### Blocker (1)

**`pkg/selfupdate/asset_test.go:79-84` — `TestAssetNamesStillMatchTheGoreleaserTemplates` still looks for `format: zip`.**

It reads `.goreleaser.yaml`, collects the `archives:` lines prefixed `format:`, and expects exactly `["format: zip"]`. On this branch the list is empty, so `test (ubuntu-latest)` and `test (windows-latest)` fail with `expected: []string{"format: zip"} actual: []string(nil)`. Renaming the prefix to `formats:` and the expected line to `formats: [zip]` keeps the guard (an archives-level `formats:` is still caught); `go test ./pkg/selfupdate` passes on ab13010 with that change. Snippet inline on the line that broke it.

### Verified

- a003a3a: the lookup is assigned before the test, so a transient `gh release view` failure fails the step. (2) closed.
- v2.18.1 `internal/client/github.go`: `createOrUpdateRelease` → `findRelease(tag)` → `findDraftRelease`, matching `GetDraft() && GetName() == tag`. On a hit `data.Draft = release.Draft` and the body goes through `getReleaseNotes(…, keep-existing)`; `PublishRelease` returns early when `draft: true`. The matched draft keeps its notes and stays a draft. (1) closed.
- gh's `FetchRelease` runs the by-tag lookup and the draft lookup (GraphQL `release(tagName:)`) in parallel, so `publish-release`'s `gh release edit --draft=false` now resolves to the one draft — the first time that step does what its comment says.
- Release list: all 9 drafts and every UI-published release are titled with the tag, so the name match holds for the history we have.
- `goreleaser check` (v2.18.1) on this config: valid, deprecation warnings for `brews` and `dockers` only, as stated.
- Comments in both files and the PR body describe the path that runs. (1a) done; (1b) done here rather than as a follow-up.

### Note (not blocking)

The name match depends on the UI title being left as the tag (GitHub fills it in with "Generate release notes", and every release so far has it). A hand-typed title misses `findDraftRelease`, goreleaser makes its own draft, and two drafts share the tag — the pre-PR outcome for the notes, not a 404. Adding `--title "${GITHUB_REF_NAME}"` to the `gh release edit --draft` in create-release would remove the dependency, and goreleaser sets that same title on the matched draft anyway (`data.Name` in `updateRelease`). Your call whether it belongs in this PR.

**Verdict**: REQUEST_CHANGES — will approve once the Go test rename is in and `build-and-test` is green.

.goreleaser.yaml
jisung-02: Applied in 65419d3. You were right that the rename had a second reader: the Go guard collected `format:` lines from the `archives:` section and got an empty list against the renamed key, which is the `expected: []string{"format: zip"} actual: []string(nil)` on ab13010. It now collects `formats:` and expects `formats: [zip]`, so an archives-level `formats:` is still caught. Seen red first by mutating the override to `formats: [tar.gz]`: `expected: []string{"formats: [zip]"} actual: []string{"formats: [tar.gz]"}`. `go test ./pkg/selfupdate` passes on 192f03e.

choongbum (approved):
### Summary

All three items are in, and I re-checked the new commits against goreleaser v2.18.1 and gh. `build-and-test` is green on 192f03e. Approving.

### Verified

- 65419d3: the Go guard collects `formats:` and expects `formats: [zip]`, identical to what I ran on ab13010; `test (ubuntu-latest)` and `test (windows-latest)` pass.
- 0af543e: `gh release edit --draft --title "$TAG"` on any existing release leaves exactly one candidate for `findDraftRelease`, and is a no-op on the draft the job made itself. `replace_existing_artifacts`: `Upload` deletes the asset on a 422 and returns `retryx.Retriable`, so a re-run's upload no longer stops at the first existing asset.
- 192f03e: `prerelease: auto` sets `ctx.PreRelease` from `ctx.Semver.Prerelease`, and `CreateRelease` sends `Prerelease: &ctx.PreRelease` on the reuse path, so the draft's flag is goreleaser's judgment by the time `publish-release` reads `isPrerelease` back. That lookup is assigned before the test.
- `goreleaser check` (v2.18.1) on 192f03e: valid, `brews`/`dockers` deprecations only. actionlint: only the two pre-existing SC2086 infos on the `>> $GITHUB_ENV` lines, which this PR does not touch.

### Note (not blocking)

The widened demotion adds one failure path the PR body does not state yet: a whole-workflow re-run of a *finished* release demotes the public release first, so if any later job fails, the release stays a draft and `releases/latest` stays on the previous version until someone runs `gh release edit "$TAG" --draft=false --latest`. Before 0af543e the finished release was left alone and the re-run simply went red. It is a rare, deliberate action with the operator watching the run, so I am fine with it — one line next to the `releases/latest` window in the PR body would tell whoever re-runs what the recovery is. If you would rather not have that path at all, the old "public with assets → skip" branch is the alternative.

**Verdict**: APPROVE.

PR #446, opened 2026-09-14
fix: stop truncating certs, CSRs, and the private key before the write succeeds

### LINKED ISSUE #439: SaveFile truncates the destination before the write, so a failed write destroys the previous file
## Description

`SaveFile` (`utils/file_save.go:28`) writes through `saveStream`, which truncates the destination before it has any of the new content:

```go
file, err := os.Create(fileName)
if err != nil {
    return 0, fmt.Errorf("failed to create file: %w", err)
}

written, copyErr := io.Copy(file, r)
```

`os.Create` opens with `O_TRUNC`. If `io.Copy` then fails partway, the previous file is already gone and what remains on disk is a prefix of the new content. The caller gets an error, but the file it was replacing is not recoverable from anything the CLI holds.

`saveStream` is always handed a `bytes.Reader`, so the read side cannot fail. The write side can: a full disk, a quota, or an I/O error on the destination all surface through `io.Copy`.

The callers write certificates and keys, which is where losing the previous copy costs the most:

- `api/cert/cert.go:255,270,325` write certificate text
- `api/cert/cert.go:310` writes a CRL
- `pkg/cert/cert.go:73` writes a CSR

The same file already has the answer. `SaveStreamAtomic` (`utils/file_save.go:61`) writes to a temp file in the same directory and renames it into place, and it is what `api/ftp`, `api/packages`, and `pkg/selfupdate` use. `config.saveConfig` follows the same write-then-rename shape for the same reason, documented in `CLAUDE.md`: "two alpacon processes can be saving a renewed access token at the same moment, and a plain truncate-and-write leaves a config the next command cannot parse."

## Steps to reproduce

Not reproduced end to end. Triggering it through `SaveFile` needs a real write failure (a full filesystem, a quota, or an I/O error), which is why this is filed from reading `saveStream` rather than from a failing test. The truncate-before-write shape is visible in the four lines quoted above.

An in-package test can pin the behavior directly by handing `saveStream` a reader that returns data and then an error, and asserting the destination still holds its previous content.

## Expected behavior

A failed write leaves the previous file intact. A destination that has no previous file is left absent rather than holding a partial one.

## Actual behavior

The previous file is truncated at open time, and a failed `io.Copy` leaves the destination holding whatever bytes made it through.

## Suggested fix

Route `SaveFile` through `SaveStreamAtomic`, picking the mode the certificate paths should land at. Add a test that a write failure leaves the existing file untouched.

---

Found by a Codex audit pass. Reported against a copy whose `utils/file_save.go` hash does not match `main`; the description above is written against `main` at 694e9e9, where the exported `SaveStream` the report named is the unexported `saveStream` and its only caller is `SaveFile`.

jisung-02:
## Summary
SaveFile opened its destination with os.Create, which carries O_TRUNC—the previous file was gone before a single byte of the replacement was known to be writable. A write that then failed partway left the destination holding whatever bytes had made it through, with no way back. `pkg/cert`'s `savePrivateKey` had the identical shape on the private-key path. Both now write through the existing `SaveStreamAtomic` helper, which stages the write in a temp file in the same directory and renames it into place, so a failed write never touches what was already on disk.

## Changes
- `utils/file_save.go`: `SaveFile` delegates to `SaveStreamAtomic` instead of `os.Create` + `io.Copy`. The new-file mode is pinned at 0666 rather than something narrower—0666 is what `os.Create` requested too, so the umask still decides the final mode and every caller (certs, CRLs, CSRs) lands at the same permissions as before this fix.
- `pkg/cert/cert.go`: `savePrivateKey` now PEM-encodes into memory first (matching the pattern `CreateCSR` already uses) and saves through `SaveStreamAtomic` at its existing 0600. The key's directory keeps its existing 0700—`savePrivateKey` still creates it before the save call, so `SaveStreamAtomic`'s own directory creation is a no-op.
- Tests: `utils/utils_test.go`, `utils/file_save_unix_test.go` (SaveFile), `pkg/cert/cert_test.go` (new file—the package had no tests before).

Neither fix has a test that reproduces the exact mid-`io.Copy` failure #439 reports: `SaveFile`'s reader is always a `bytes.Reader`, which cannot error, and `savePrivateKey` only ever writes a brand-new file, so the one failure that would force it (no directory write permission) blocks the old code identically to the new one. `TestSaveStreamAtomic_RetainsExistingFileOnReadError` (pre-existing, unmodified) already covers that exact path one level down. The new tests instead pin what each migration must preserve—existing content on a blocked write, and the resulting file/directory modes.

## Testing
- `go build ./...`, `go vet ./...`, `golangci-lint run ./...`—clean
- `go test -race -shuffle=on ./...`—all packages pass
- Every new/changed test confirmed RED against a reverted value before being verified GREEN

## Review
Addressed Copilot review feedback: restored `SaveStreamAtomic`'s Go doc comment, which a prior commit had collapsed into a trailing inline comment and so dropped from `go doc` output. The `unix` build-tag comment on `TestSaveFile_AppliesUmaskToNewFile` was clarified rather than changed—`unix` is a Go-recognized meta build constraint since Go 1.19 and needs no `-tags unix` flag, so the test already runs in CI.

Addressed @geunwoonoh's review (approved, with three suggestions on `pkg/cert/cert_test.go`):
- `TestGenerateKey_WritesOwnerOnlyKeyFile` and `TestGenerateKey_ReusesExistingKey` now compare keys with `rsa.PrivateKey.Equal` instead of only `key.N`—the full-key check the modulus-only comparison was missing.
- The `0600`/`0700` exact-mode assertions—already flagged as a suppressed, low-confidence comment by Copilot—now carry a one-line comment explaining why they are umask-safe: both modes carry zero group/other bits, and an ordinary umask only clears bits, so it cannot widen them.

Closes #439


utils/file_save_unix_test.go
```diff
@@ -0,0 +1,29 @@
+//go:build unix
```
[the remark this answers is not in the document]
jisung-02: `unix` is a Go-recognized meta build constraint since Go 1.19—it is satisfied automatically for unix-like GOOS values (linux, darwin, etc.) and needs no `-tags unix` flag. The CI command `go test -race -v -shuffle=on ./...` already runs it on the Linux and macOS jobs.

Verified locally:
```
$ go test -run TestSaveFile_AppliesUmaskToNewFile -v ./utils/...
=== RUN   TestSaveFile_AppliesUmaskToNewFile
--- PASS: TestSaveFile_AppliesUmaskToNewFile (0.00s)
```

No change needed.

utils/file_save.go
```diff
@@ -26,39 +26,11 @@ func IsReplacementTempName(name string) bool {
...
-	return written, nil
-}
-
-// SaveStreamAtomic writes r to fileName through a temp file in the same
-// directory, then renames it into place. The mode is decided when the write
-// starts: a destination already on disk keeps its own mode, so a write does not
-// re-permission a file the user set up, and one that is not there yet gets
-// newFilePerm minus the umask.
-func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (int64, error) {
+func SaveStreamAtomic(fileName string, r io.Reader, newFilePerm os.FileMode) (int64, error) { // an existing file keeps its own mode; a new one gets newFilePerm minus the umask
```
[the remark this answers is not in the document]
jisung-02: Fixed. Restored the Go doc comment immediately above the declaration so `go doc` describes the atomic-write and mode-preservation contract again.

pkg/cert/cert_test.go
```diff
@@ -0,0 +1,55 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, os.FileMode(0600), info.Mode().Perm())
+
+	dirInfo, err := os.Stat(filepath.Dir(keyPath))
+	require.NoError(t, err)
+	assert.Equal(t, os.FileMode(0700), dirInfo.Mode().Perm())
+
+	reread, err := readPrivateKey(keyPath)
+	require.NoError(t, err)
+	assert.Equal(t, key.N, reread.N)
```
geunwoonoh: Suggestion: use the full-key equality check instead of comparing only the modulus.

`crypto/rsa.PrivateKey` has had `Equal(x crypto.PrivateKey) bool` since Go 1.15 (this module pins `go 1.25.13`)—it compares the whole key rather than one field, so it's a stronger and more idiomatic check here (same suggestion applies to line 54 below).

```suggestion
	assert.True(t, key.Equal(reread), "generateKey should round-trip the same key")
```

pkg/cert/cert_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	keyPath := filepath.Join(t.TempDir(), "server.key")
+
+	first, err := generateKey(keyPath)
+	require.NoError(t, err)
+
+	second, err := generateKey(keyPath)
+	require.NoError(t, err)
+
+	assert.Equal(t, first.N, second.N)
```
geunwoonoh: Suggestion: use the full-key equality check instead of comparing only the modulus (see the note on line 38 in `TestGenerateKey_WritesOwnerOnlyKeyFile`).

```suggestion
	assert.True(t, first.Equal(second), "generateKey should reuse the existing key")
```

pkg/cert/cert_test.go
```diff
@@ -0,0 +1,55 @@
...
+	dir := t.TempDir()
+	keyPath := filepath.Join(dir, "nested", "server.key")
+
+	key, err := generateKey(keyPath)
+	require.NoError(t, err)
+	require.NotNil(t, key)
+
+	info, err := os.Stat(keyPath)
+	require.NoError(t, err)
+	assert.Equal(t, os.FileMode(0600), info.Mode().Perm())
```
geunwoonoh, line 31: Suggestion: document why the exact-mode assertion is umask-safe here.

Copilot's review flagged this (and the `0700` check on line 34) as a suppressed/low-confidence comment: "these exact mode assertions depend on the process umask." In practice it's safe—`0600` and `0700` carry zero group/other bits, so no ordinary umask (which only clears bits, never sets them) can change the result. That's different from `utils/file_save_unix_test.go`'s `TestSaveFile_AppliesUmaskToNewFile`, which genuinely needs `syscall.Umask(0)` because `0666` does have those bits.

A short comment stating that (mirroring the one on `TestSaveStreamAtomic_AppliesRequestedModeToNewFile` in utils_test.go) would close the loop on Copilot's comment and preempt future reviewers re-raising it. Applies to both the file-mode check here and the directory-mode check on line 34.

```suggestion
	info, err := os.Stat(keyPath)
	require.NoError(t, err)
	// 0600 has no group/other bits, so no ordinary umask can widen it—unlike
	// SaveFile's 0666 case in file_save_unix_test.go, which pins umask 0.
	assert.Equal(t, os.FileMode(0600), info.Mode().Perm())
```

geunwoonoh (dismissed):
### Summary
`SaveFile` (utils/file_save.go) and `savePrivateKey` (pkg/cert/cert.go) both previously opened their destination with `O_TRUNC`, so a write that failed midway left a corrupted or truncated file with no way back. This PR routes both through the existing `SaveStreamAtomic` primitive—write to a temp file in the same directory, then rename—so a failed write never touches what was already on disk. It also removes the now-redundant private `saveStream` helper and adds targeted tests for permission preservation and failure-path content retention.

Verified locally: `go build ./...`, `go vet ./...`, `gofmt -l` (all clean), `go test -race -shuffle=on ./utils/... ./pkg/cert/...` (pass).

### Critical Issues (0)

None.

### Warnings (0)

None.

### Suggestions (2)

1. **[pkg/cert/cert_test.go:38, 54]** Both new tests compare only `key.N` to prove two `*rsa.PrivateKey` values are "the same key." `crypto/rsa.PrivateKey` has had an `Equal(x crypto.PrivateKey) bool` method since Go 1.15 (this module pins `go 1.25.13`), which compares the full key rather than one field—more complete and idiomatic at no added cost.
2. **[pkg/cert/cert_test.go:30,34]** Copilot's review flagged (as a suppressed/low-confidence comment) that the exact-mode assertions here (`0600`/`0700`) are umask-dependent. In practice this is safe: both requested modes carry zero group/other bits, so no ordinary umask (which only clears bits, never sets them) can change the result—unlike `utils/file_save_unix_test.go`'s `TestSaveFile_AppliesUmaskToNewFile`, which genuinely needs `syscall.Umask(0)` because `0666` does have those bits. A one-line comment stating that reasoning (mirroring the one already on `TestSaveStreamAtomic_AppliesRequestedModeToNewFile` in utils_test.go) would close the loop and preempt future reviewers re-raising it.

### Highlights

- The migration eliminates real duplication: `SaveFile` and `savePrivateKey` each carried their own truncate-then-write logic; both now share one atomic-write path, and the now-dead `saveStream` function was cleanly removed (no dangling references found).
- Permission semantics are preserved exactly through the refactor: `saveFilePerm = 0666` matches `os.Create`'s implicit mode (so certs/CRLs/CSRs don't get narrowed), and the private key keeps 0600.
- The PR description is transparent about a real test-coverage limitation (no test reproduces the exact mid-`io.Copy` failure, since `SaveFile`'s reader is always a `bytes.Reader`) and correctly identifies the pre-existing test one layer down that already covers that path.
- As a side effect, this closes part of `generateKey`'s pre-existing TOCTOU window (two concurrent first-time key generations): previously a losing writer's truncate could leave a corrupted key on disk; now the rename is atomic, so the worst case is a benign overwrite by a second valid key, not corruption. Pre-existing race, not something this PR needs to fully resolve—noted for awareness only.
- Build/vet/format/tests all verified clean locally, and every new test's failure mode is real (e.g., directory `chmod 0555` genuinely blocks temp-file creation rather than faking the error).

---

**Verdict**: APPROVED

**Checklist**:
- [x] Code follows stack conventions (golang.md: error wrapping, decl order, errcheck/testifylint-clean)
- [x] Tests included and passing (`go test -race -shuffle=on ./utils/... ./pkg/cert/...`)
- [x] No security vulnerabilities (key material never exposed at a wider mode during staging)
- [ ] Documentation updated if needed (n/a—no user-facing docs affected)

pkg/cert/cert_test.go
jisung-02: Fixed in bc32d16—asserts `key.Equal(reread)` instead of comparing `key.N`. You're right that the modulus-only check only proves the keys share `N`, not that they're the same key.

pkg/cert/cert_test.go
jisung-02: Fixed in e874750—asserts `first.Equal(second)` here too, same change as the round-trip test above.

pkg/cert/cert_test.go
jisung-02: Fixed in 0579582—added a one-line comment above each assertion (0600 at line 30, 0700 at line 34) stating why they're umask-safe, matching the comment style already on `TestSaveStreamAtomic_AppliesRequestedModeToNewFile` in utils_test.go.

geunwoonoh (approved):
### Summary
Follow-up on the previous review (all three findings addressed):

1. `key.N`/`reread.N` comparison → `key.Equal(reread)` (bc32d16)
2. `first.N`/`second.N` comparison → `first.Equal(second)` (e874750)
3. Umask-safety of the `0600`/`0700` exact-mode assertions → one-line comments added above each (0579582)

No new code paths were touched—the diff since the last review is scoped entirely to `pkg/cert/cert_test.go` (4 lines changed). `go build ./...`, `go vet ./...`, `gofmt -l`, and `go test -race -shuffle=on ./utils/... ./pkg/cert/...` all pass locally, and CI (lint, tests on macOS/Linux/Windows, CodeQL, govulncheck) is green.

### Critical Issues (0)

None.

### Warnings (0)

None.

### Suggestions (0)

None—both prior suggestions are resolved as suggested, verified against the diff.

### Highlights

- Each fix was applied exactly as suggested rather than reworded or partially applied, and the author replied on each thread naming the fixing commit.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (CI green across macOS/Linux/Windows)
- [x] Documentation updated if needed (n/a)

Issue #436, opened 2026-09-08
[BUG] token acl server add --server bypasses the input normalization every other server-name path applies


jisung-02:
## Description

Three separate paths turn a user-supplied server name into a server ID, and they were added in three commits a month apart with no reference to one another. The single-value path applies no input normalization at all, so the same server name succeeds through `--servers` and fails through `--server`.

- `api/server.ResolveServerNames` (`api/server/server.go:124`) trims each name, skips empty entries, resolves them one at a time, and stops at the first failure. Its callers are `cmd/worksession/worksession_create.go:175` and `cmd/worksession/worksession_update.go:141`.
- `cmd/token.resolveServerIDs` (`cmd/token/acl_server.go:31`) spawns one goroutine per name and keeps positions by index. Its caller normalizes the `--servers` value with `utils.SplitAndTrim` before handing it over.
- Direct single-value calls pass the raw value straight through. `cmd/token/acl_server_add.go:55` passes the `--server` flag value, and `cmd/authority/authority_create.go:51` passes the prompt input.

`GetServerIDByName` (`api/server/server.go:100`) sends whatever string it receives as the `?name=` query parameter. It neither trims nor rejects an empty name, so an unnormalized value becomes a server-side lookup miss instead of a client-side validation error.

Two further problems sit in the same code.

**Nondeterministic error selection in the bulk resolver.** `resolveServerIDs` guards `firstErr` with a mutex and sets it only when it is still nil, so the reported name is whichever failing goroutine acquires the lock first, not the earliest one in the input. The same input can name a different server in the error message across runs. The goroutine count is unbounded, so a request goes out for every name at once, and there is no context or cancellation, so a failure does not stop the rest. Service tokens carry an hourly request quota, which is the reason to bound the fan-out.

**Double normalization on the worksession path.** Both worksession callers run `utils.CompactStrings` over the names, and `ResolveServerNames` then trims and drops empty entries again. Behavior is unaffected, but it leaves the ownership of normalization unclear.

**No test coverage.** Only `utils.SplitAndTrim` has a table test (`utils/utils_test.go:150`). `utils.CompactStrings`, `ResolveServerNames`, and `resolveServerIDs` have none, and `cmd/token` has no test file at all. Nothing pins the handling of whitespace, empty entries, duplicates, or error-selection order.

## Steps to reproduce

```bash
alpacon token acl server add TOKEN --server " "
alpacon token acl server add TOKEN --server " my-server "
[...]

PR #447, opened 2026-09-14
fix(api): reject a blank name before it becomes an unfiltered lookup

jisung-02:
## Summary

`alpacon token acl server add TOKEN --server " "` did not fail. It succeeded, and granted the token access to whichever server the API happened to list first. The workspace API trims the `name` filter and drops it when nothing is left, so a whitespace-only name comes back as the full, unfiltered list. Six name-to-ID resolvers read a non-empty page as proof the named object was found and returned its first element, which turns a name that carries no filtering power into an arbitrary match. ServerACL is a deny-by-default boundary, so this is an authorization fault, not a lookup annoyance.

The same decision was reachable from the token, user, group, webhook, and role resolvers. A blank token name resolved to an arbitrary API token, so the ACL change could also land on the wrong token, and a blank role name resolved to an arbitrary role, which `alpacon user role grant` would then have granted.

Review found that the package resolver was a different and worse fault, that the guard was being copied per site rather than shared, and that four of the new tests could not have caught the bug they were written for. All three are fixed here. Review also found that the shared helper took five strings in a row, that the blank refusal in the package resolver did not carry the sentinel, and that three of the fixes above were asserted only for an empty name and never for the whitespace-only one the bug was reported with. Those are fixed here too.

### The blank-name fault

`api.ResolveByName` in `api/resolve.go` is now the single place that resolves a name to one object. It trims, refuses a name that trims to empty before any request goes out, fetches one page under the caller's filter key, and decides existence on `len(results)`. `GetServerIDByName`, `GetAPITokenIDByName`, `GetUserIDByName` (filter key `username`), `GetGroupIDByName`, `GetWebhookIDByName`, and `rbac.ResolveRole` all call it, each keeping its own message text. The guard used to be copied into seven resolvers, which is how `ResolveRole` was missed by the first sweep on this branch.

The old resolvers decided on `response.Count` and then read `Results[0]`. A response carrying a positive `Count` with an empty `Results` would panic there, so the helper decides on `len(results)` instead.

`api.ErrBlankName` marks the refusal, so a caller can tell it apart from an ordinary not-found with `errors.Is` rather than by matching message text. `api/rbac` re-exports it.

The helper takes `api.ResolveByNameOptions` rather than five strings in a row. Swapping `blankMsg` with `notFoundMsg`, or `filterKey` with `name`, used to compile and pass the suite—the only message assertion in the tree is rbac's `case-sensitive`—and now changes a field name the compiler checks.

`api.RequireName` is the trim-and-refuse half on its own, for the one resolver that cannot use the helper.

One message changed shape. A lookup that fails on the HTTP call now comes back wrapped by `FetchPagesUpTo` as `fetching page 1 from /api/servers/servers/: <original>`, where the resolver used to return the bare error. Nothing branches on the error value itself, and both `ParseErrorResponse` and `HTTPStatusCode` walk the `Unwrap` chain, so the MFA retry and the work session denial still take the same paths.

### The package resolver

The package entry endpoints have no `name` filter at all. Neither `SystemPackageEntryViewSet` nor `PythonPackageEntryViewSet` lists `name` in `filterset_fields`; both carry only `package__name`, and DjangoFilterBackend drops a parameter it does not know. So `?name=anything.whl` came back as the whole list and `Results[0]` was the highest-version entry, whatever was asked for, and `DownloadPackage` wrote that content under the requested file name. Asking for `requests-2.32.0-py3-none-any.whl` could save a different package under that name.

This fault fires for every input, not only a blank one, so the blank-name guard did not touch it. `GetPackageIDByName` now narrows with `search` and picks the entry whose `Name` equals the requested name exactly, which also removes the duplicated python and system branches. The bug predates this branch. Its blank refusal goes through `api.RequireName`, so it carries `api.ErrBlankName` like every other resolver's, which the sentinel's own doc comment already claimed of it.

### Input handling that disagreed with itself

- `DownloadPackage` built the save path from the untrimmed argument, so `"  numpy.whl  "` resolved and then landed on disk under a name carrying the spaces. It trims once, up front.
- `ResolveTokenID` ran `utils.IsUUID` on the untrimmed value while `ResolveRole` trimmed first, so a UUID with surrounding spaces took the id branch for a role and the name branch for a token. Both trim first now.
- `cmd/token/acl_server_add.go` trimmed `--server` but not `--servers`, so `--server " " --servers web-01` passed silently while `--server web --servers " "` was refused as "not both". Both are trimmed, and the exactly-one-of check moved into `validateServerAclFlags`, which has a table test. The trim happens inside that helper and the trimmed pair comes back to the caller, so the table can carry the input this branch exists for—a whitespace-only flag reaching the validator untrimmed—and the caller resolves the same name the validator accepted.
- `cmd/iam/user_role_grant.go` ran the transposed-argument check even after a blank role name was refused. When the username matched a role name, the operator was told `No role named " ", but "admin" is one`, pointing at a mistake they did not make. That check is now skipped on a blank-name refusal, which also drops a request that had nothing to find.

### Tests

One table test per resolver, merged into the tables that already covered them rather than added beside. Each carries a `wantCall` column, so a guard added at a single call site cannot pass, and a `wantErr` column instead of `wantID == ""` standing in for the error expectation. The blank cases assert `errors.Is` against the shared sentinel rather than a message, and they do it for a name that trims to empty rather than for an empty string alone. `"   "` is the input the bug was reported with; keying the assertion on the literal empty string left it proving only that no request went out.

`TestResolveTokenID` carries a padded UUID row, the case the `ResolveTokenID` trim fixed and left untested.

Every handler now puts a decoy ahead of the target when the filter is absent. The previous handlers in `auth`, `webhook`, `iam`, and `packages` answered any request with the target, so the exact-name case passed even if the resolver sent the wrong query key. The `iam` handler never distinguished `username` from `name`, so `GetUserIDByName` sending `name` would have looked correct.

`api/cert/cert.go` `GetAuthorityIDByName` and `api/server/server.go` `GetRegistrationTokenByName` were checked and left alone: both compare `Name == name` over a fetched list, so a blank name matches nothing. The membership lookup in `api/iam/iam.go` filters on already-validated UUIDs.

An earlier version of this description claimed a re-sweep over every name and search filter in `api/` had found no further case. That was wrong. It read the filter keys the CLI sends without checking what the server filters on, which is the only way the package entry endpoint would have shown up.

## Verification

17 files changed, 545 insertions, 212 deletions against `origin/main`.

`go test -race -shuffle=on -count=1 ./...`: 44 packages ok, 0 FAIL, out of 58 (14 carry no test files). `golangci-lint run ./...` reports `0 issues.`

Renaming a resolver's filter key fails the lookups, which is what the decoys buy:

```
--- FAIL: TestGetAPITokenIDByName/found
--- FAIL: TestGetWebhookIDByName/found
--- FAIL: TestGetUserIDByName/found
--- FAIL: TestGetGroupIDByName/found
--- FAIL: TestGetGroupIDByName/unknown_name
```

Removing the exact compare in `GetPackageIDByName` and taking the first entry again fails on the decoy:

```
--- FAIL: TestGetPackageIDByName/system/found
    Error:      	Not equal:
    expected: "pkg-uuid-abc"
    actual  : "id-decoy"
```

Inverting the not-both condition in `validateServerAclFlags` fails two of its four cases:

```
--- FAIL: TestValidateServerAclFlags/both_flags_set
    Error:      	An error is expected but got nil.
--- FAIL: TestValidateServerAclFlags/only_servers_set
    Error:      	Received unexpected error: use either --server or --servers, not both
```

Dropping the two trim lines from inside `validateServerAclFlags` fails exactly the whitespace cases and nothing else:

```
--- FAIL: TestValidateServerAclFlags/whitespace_only_server_set
--- FAIL: TestValidateServerAclFlags/whitespace_only_servers_set
--- FAIL: TestValidateServerAclFlags/server_with_surrounding_whitespace
--- FAIL: TestValidateServerAclFlags/whitespace_only_server_with_servers_set
```

Returning a plain `errors.New` from `api.RequireName` instead of the sentinel-bearing error fails the blank cases in all six tables, the whitespace-only ones included:

```
--- FAIL: TestGetServerIDByName/empty_name
--- FAIL: TestGetServerIDByName/whitespace_only_name
--- FAIL: TestGetAPITokenIDByName/empty_name
--- FAIL: TestGetAPITokenIDByName/whitespace_only_name
--- FAIL: TestGetUserIDByName/empty_name
--- FAIL: TestGetUserIDByName/whitespace_only_name
--- FAIL: TestGetGroupIDByName/empty_name
--- FAIL: TestGetGroupIDByName/whitespace_only_name
--- FAIL: TestGetWebhookIDByName/empty_name
--- FAIL: TestGetWebhookIDByName/whitespace_only_name
--- FAIL: TestResolveRole_ByName/empty_name
--- FAIL: TestResolveRole_ByName/whitespace_only_name
```

Moving the `utils.IsUUID` check back ahead of the trim in `ResolveTokenID` fails the padded row alone:

```
--- FAIL: TestResolveTokenID/padded_uuid_fast-path_—_no_HTTP_request
```

The live-workspace evidence below was captured on 2e94759b and has not been re-run since; the guards those commands exercise are unchanged by the commits that follow. Before the fix:

```
$ alpacon token acl server add TOKEN --server " "
Success: Server ACL added: token TOKEN can access

$ alpacon user role describe " "
Permissions
RESOURCE   ACTIONS       ACL
proxy      read, write
```

The audit log records the write the first one caused (`action: create`, `status_code: 201`, `model: ServerACL`). The entry was deleted afterwards.

After the fix:

```
$ alpacon token acl server add TOKEN --server " "
Error: one of --server or --servers is required.

$ alpacon log " "
Error: Failed to get logs: server name is required.

$ alpacon token acl server ls " "
Error: Failed to resolve token: token name is required.

$ alpacon user role describe " "
Error: Failed to resolve the role: role name is required.

$ alpacon log "web-editor" --tail 1
PROGRAM   LEVEL   MESSAGE                                            DATE
alpamon   INFO    [] Completed system information synchronization.   1 hours ago

$ alpacon user role describe "admin"
Permissions
RESOURCE                    ACTIONS
access_policy               create, delete, read, update
```

The package download path has no live-workspace evidence. It is covered by the table test alone, and confirming it against a real registry waits on someone with a workspace that carries two entries whose names share a prefix.

## Notes on the issue

Refs #436. The issue is not fully covered by this PR, so it stays open.

- The reported symptom was a lookup failure. The actual behavior is worse: the command succeeds and grants access. The issue's second reproduction line, `--server " my-server "`, does not reproduce, because the server trims the filter and a padded name resolves correctly.
- The issue's scope item 1 also named the `authority create` prompt. Nothing is needed there: `utils.PromptForInput` already trims and `PromptForRequiredInput` already refuses an empty value, so there is no second site for a shared helper.
- Still open from the issue and deliberately out of scope here: the nondeterministic error selection and unbounded fan-out in `cmd/token.resolveServerIDs`, the double normalization on the worksession path, and table tests for `CompactStrings`, `ResolveServerNames`, and `resolveServerIDs`. None of them cause this symptom.
- Adding a `name` filter to the package entry endpoints on alpacon-server would let the CLI stop fetching and comparing client-side. Filed as alpacax/alpacon-server#3560.

## Stack

common (no `alpacax` frontmatter in `CLAUDE.md`)

## Checklist

- [x] Code follows project style guidelines
- [x] Self-reviewed the code
- [x] No debug print statements left
- [x] Tests pass locally
- [x] Updated relevant documentation (`CLAUDE.md` records the name-resolution helper; no other doc or comment stated the old behavior, checked by grep)
- [x] Commit messages follow conventional format
- [x] Branch is up to date with `main`


api/server/server.go
```diff
@@ -99,6 +99,14 @@ func RequestServerAction(ac *client.AlpaconClient, serverName, action string, fo
 }
 
 func GetServerIDByName(ac *client.AlpaconClient, serverName string) (string, error) {
+	// The API drops a blank name filter and answers with the whole server list, so a
+	// name that survives trimming as empty must never reach it: Count > 0 would then
+	// mean "no filter was applied", not "found", and the first server would be returned.
+	serverName = strings.TrimSpace(serverName)
```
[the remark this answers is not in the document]
jisung-02: Confirmed and fixed in 2e94759.

Reproduced against a live workspace before the fix: `alpacon user role describe " "` answered with an arbitrary role rather than failing, so the roles endpoint does drop a filter that trims to empty, exactly as this comment says. `ResolveRole` now trims `nameOrID` and refuses an empty one before the request, and `api/rbac/rbac_test.go` carries a table test covering exact, padded, unknown, whitespace-only, and empty names, asserting both the error and whether the name reached the API. It was seen red first: only the two blank cases failed.

The blast radius was larger than role lookup. `cmd/iam/user_role_grant.go:101` picks the role through this resolver, so a blank argument would have granted an arbitrary role, and the swap-detection branch at `:106` read a blank user argument as a valid role name.

The earlier sweep grepped for `Results[0]` and missed this one because it indexes a `FetchPagesUpTo` slice instead. A re-sweep over every name and search filter in `api/` found no other case: the remaining one, `GetRegistrationTokenByName`, compares `t.Name == name` over the fetched list, and the membership lookup in `api/iam/iam.go:212` filters on already-validated UUIDs.

geunwoonoh (commented):
## PR review: #447 - fix(api): reject a blank name before it becomes an unfiltered lookup

**Author**: @jisung-02
**Branch**: 436-normalize-server-name-input → main
**Stack**: golang
**Changes**: +375 -4 in 13 files (all 12 CI checks pass)

---

### Summary

The direction and the implementation are right. For the six resolvers whose server-side filter is an exact match, the guard alone closes the fault. The problem is the package entry API: it has no `name` filter at all, so **every** name picks an arbitrary entry, with or without this PR.

| Resolver | Query key | Server filter (alpacon-server main) | Result |
|---|---|---|---|
| `GetServerIDByName` | `name` | `ServerFilter.Meta.fields` (exact) | OK |
| `GetUserIDByName` | `username` | `UserFilter` `username: exact` | OK |
| `GetGroupIDByName` | `name` | `GroupViewSet.filterset_fields` | OK |
| `GetAPITokenIDByName` | `name` | `APITokenViewSet.filterset_fields` | OK |
| `GetWebhookIDByName` | `name` | `WebhookViewSet.filterset_fields` | OK |
| `ResolveRole` | `name` | `RoleFilter.Meta.fields` | OK |
| `GetPackageIDByName` | `name` | **None** (only `package__name`) | Every lookup goes out unfiltered |

### Critical issues (0)

### Warnings (2)

**1. [api/packages/packages.go:76] The package entry API does not filter by name (pre-existing bug, high severity)**

- **Cause**: `name` is not in either entry ViewSet's `filterset_fields`. DjangoFilterBackend ignores parameters it does not know, so `?name=anything.whl` returns the whole list and `Results[0]`—the highest-version entry—is picked.
- **Impact**: `DownloadPackage` saves the file under the name the user asked for (`packages.go:175`). `alpacon package python download requests-2.32.0-py3-none-any.whl .` therefore saves another package's content as `requests-...whl`.
- **Relation to this PR**:
  - This PR did not introduce the bug.
  - The new guard blocks nothing in this function.
  - The new test (`packages_test.go:492`) emulates a filter the server does not have, which pins the wrong contract.
  - The PR description's "re-sweep ... found no further case" does not hold.
- **Recommendation**: fix it with an exact compare. This also removes the duplicated python/system branches.

```go
entries, err := api.FetchAllPages[packageEntryRef](ac, packageEntryURL(packageType), map[string]string{"search": fileName})
if err != nil {
	return "", err
}
for _, e := range entries {
	if e.Name == fileName {
		return e.ID, nil
	}
}
return "", errors.New("no package found with the given name")
```

`packageEntryRef` is a type carrying only `ID` and `Name`, placed in the type block. Adding a `name` filter on the server is the other option.

**2. [api/auth/auth_test.go:593, api/webhook/webhook_test.go:175, api/packages/packages_test.go:492, api/iam/iam_test.go:478] Four tests still pass when the filter is dropped**

- **What happens**: when the filter parameter is missing, these handlers return only the target itself, and the iam handler always returns one result.
- **Result**: the `exact name` case passes even if the resolver sends the wrong key or the server ignores it—which is exactly the fault this PR closes, and exactly warning 1.
- **Contrast**: `server_test.go:598` and `rbac_test.go:555` put a decoy (`first-server`, `auditor`) first when unfiltered, so those tests catch it.
- **Recommendation**: use the same pattern in all four. For iam, filter users on `username` and groups on `name`.
- **PR description**: "Each handler emulates the real filter" does not match the iam test either.

### Suggestions (7)

1. **[api/server/server.go:103 and 6 more] Fold the duplicated resolvers into a shared helper**
   - The same four-line guard appears in 7 places, and 6 `GetXIDByName` bodies are nearly identical.
   - One helper in the `api` package can own trimming, the blank-name refusal, and the not-found check. `api` already imports `utils` and `client`, so no import cycle follows.
   - The PR description notes the grep sweep missed `ResolveRole`. As long as the guard is copied per site, the next resolver can miss it too.
2. **[api/server/server.go:123 and 5 more] Decide on `len(Results)` rather than `Count`**
   - The code checks only `response.Count == 0` and then reads `Results[0]`.
   - A response with a positive `Count` and empty `Results` panics. The code predates this PR, and the helper above fixes it in one place.
3. **[api/server/server_test.go:596 and 5 more] Merge into the existing table tests**
   - The same files already have `TestGetServerIDByName` (:99), `TestGetAPITokenIDByName` (auth_test.go:194), `TestGetUserIDByName`/`TestGetGroupIDByName` (iam_test.go:154/199), `TestGetWebhookIDByName` (webhook_test.go:74), and `TestResolveRole_ByName` (rbac_test.go:103).
   - The new cases are not identical because the handlers differ, but they walk the same path. Adding the blank cases and a `wantCall` column to the existing tables is cleaner.
4. **Test readability**
   - `wantID == ""` doubles as the expect-an-error flag. Use a separate `wantErr bool`, as the existing tables do.
   - `TestGetIDByName_...` names a function that does not exist.
   - The `count`/`results` assembly in the packages test can be as short as the webhook test's, and the double loop can become a nested `t.Run(packageType)`.
5. **[api/packages/packages.go:175] The trimmed value does not reach the caller**
   - `DownloadPackage` builds the save path from `fileName` before trimming.
   - `"  numpy.whl  "` resolves, but is saved under a file name with the surrounding spaces.
6. **[api/auth/auth.go:199] `ResolveTokenID` and `ResolveRole` now order their checks differently**
   - This PR makes `ResolveRole` trim before `IsUUID`, but `ResolveTokenID` checks `IsUUID` on the untrimmed value.
   - A UUID with surrounding spaces resolves by ID for a role, while for a token it falls through to a name lookup and fails.
7. **Asymmetric input handling**
   - **[cmd/iam/user_role_grant.go:101-106]** When the role argument is blank, the transposed-argument check calls `ResolveRole` again with the user argument. If the username matches a role name such as `admin`, it prints the misleading hint `No role named " ", but "admin" is one`. Skip that check on a blank-name error.
   - **[cmd/token/acl_server_add.go:34-35]** Reassigning in place, as the resolvers do, reads more consistently than `rawServerName`. `--servers` is also not trimmed, so `--server " " --servers web-01` passes silently while `--server web --servers " "` fails with "not both". Moving flag validation into a helper would also make it unit-testable, per CLAUDE.md.

### Additional review

**Go conventions**
- **Error strings**: lowercase with no trailing punctuation (ST1005).
- **Imports and declaration order**: import sorting and the `const → var → type → func` order are unaffected.
- **Test rules**: every case runs under `t.Run` with `t.Parallel()`, case names carry no regex metacharacters, and no handler calls `require`. The iam `run` closure is called on the test goroutine, so it is fine.
- **Concurrency**: `atomic.Bool` records the call without a race. The shared `servers`/`roles` slices are only read, and `append` starts from nil, so they are safe.
- **Nit**: the comment explaining why the guard exists appears only on `GetServerIDByName` and `ResolveRole`. With the helper from suggestion 1, it would live in one place.

**Duplicated or unused code**
- **Unused code**: none. Every new import is used, and lint passes.
- **Duplication**:
  - The guard in 7 places and the resolver bodies in 6 (suggestion 1)
  - The python/system branches in `GetPackageIDByName`, which read only the ID and differ only in type (warning 1)
  - Six nearly identical test bodies, which also overlap the existing table tests (suggestion 3)

**Performance**
- **Improvement**: a blank name fails immediately without a network request. The effect is negligible; the point is correctness.
- **Regression**: none. `strings.TrimSpace` returns a substring of the original with no allocation, and trimming twice (in `cmd` and in the resolver) costs nothing measurable.
- **Notes**:
  - The fix proposed in warning 1 uses `FetchAllPages`, but `search` narrows the list, so it usually ends after one page.
  - `resolveRoleForBinding` sends an unneeded second request for a blank role (suggestion 7).

### Highlights
- **Real behavior verified**: the PR confirmed against a live workspace and the audit log that the actual behavior (the ACL grant succeeding) is worse than the reported symptom (a lookup failure).
- **Guard placement**: putting the guard inside the resolvers protects about 50 call sites at once. It also closes the `GetMFALinkByServerName` path, which built an MFA link for an arbitrary server from a blank name.
- **Test design**: the tests check not only the error but whether the request reached the API (`wantCall`), so a guard added at a single call site cannot pass them.
- **Change scope**: `GetAuthorityIDByName` and `GetRegistrationTokenByName`, which already compare names exactly, were left alone.

---

**Verdict**: COMMENTED

The six resolver fixes match the server's contract and are fine to merge as they are. Warning 1, however, means the same fault remains for every input in a function this PR changes and tests. I recommend fixing it here or splitting it into an issue, and correcting the test and the sweep claim in the PR description.

**Checklist**:
- [x] Code follows stack conventions (minor test style aside)
- [x] Tests included and passing in CI
- [ ] No security vulnerabilities (the package resolver's unfiltered lookup remains)
- [x] Documentation update not needed

api/packages/packages.go
```diff
@@ -68,6 +68,11 @@ func GetPythonPackageEntry(ac *client.AlpaconClient) ([]PythonPackage, error) {
 }
 
 func GetPackageIDByName(ac *client.AlpaconClient, fileName string, packageType string) (string, error) {
+	fileName = strings.TrimSpace(fileName)
+	if fileName == "" {
+		return "", errors.New("package name is required")
+	}
+
 	params := map[string]string{"name": fileName}
```
geunwoonoh: Warning: The package entry endpoints have no `name` filter (pre-existing, high severity).

On alpacon-server main, neither `SystemPackageEntryViewSet` nor `PythonPackageEntryViewSet` lists `name` in `filterset_fields`; both have only `package__name`. DjangoFilterBackend ignores parameters it does not know, so `?name=anything.whl` returns the whole list and `Results[0]` is the highest-version entry, whatever name was asked for. `DownloadPackage` then saves that content under the requested name (line 175).

The blank-name guard therefore blocks nothing here, and the new test emulates a filter the server does not have. An exact compare fixes it and removes the python/system branch duplication:

```go
entries, err := api.FetchAllPages[packageEntryRef](ac, packageEntryURL(packageType), map[string]string{"search": fileName})
if err != nil {
	return "", err
}
for _, e := range entries {
	if e.Name == fileName {
		return e.ID, nil
	}
}
return "", errors.New("no package found with the given name")
```

`packageEntryRef` would be a type with only `ID` and `Name`. Adding a `name` filter on the server is the other option.

api/iam/iam_test.go
```diff
@@ -454,3 +454,54 @@ func TestDeleteMember(t *testing.T) {
...
+		{"padded name", "  " + stubUserName + "  ", stubUserID, true},
+		{"whitespace only name", "   ", "", false},
+		{"empty name", "", "", false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			run := func(param, wantID string, resolve func(*client.AlpaconClient, string) (string, error)) {
```
geunwoonoh: Warning: This handler returns a result for any request, so the test cannot tell a dropped filter from a working one.

`exact name` and `padded name` pass even if `GetUserIDByName` sends `name` instead of `username`. `UserFilter` does not define `name`, so that would be the same unfiltered lookup this PR closes. `TestGetServerIDByName_BlankNameResolvesNothing` and `TestResolveRole_BlankNameResolvesNothing` put a decoy first when unfiltered, which catches it. Suggest the same here: filter users on `username` and groups on `name`, and add an `unknown name` case.

api/auth/auth_test.go
```diff
@@ -563,3 +565,50 @@ func TestGetWhoamiEmptyPrincipalTypeFailsClosed(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			var called atomic.Bool
+
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				called.Store(true)
+				results := []APITokenResponse{{ID: "token-uuid-abc", Name: "deploy-token"}}
+				if filter := strings.TrimSpace(r.URL.Query().Get("name")); filter != "" && filter != "deploy-token" {
```
geunwoonoh: Warning: Without a `name` parameter this handler returns the target itself, so the `exact name` case passes even if the resolver sends the wrong key or the server ignores it. Putting a decoy first when unfiltered, as `TestGetServerIDByName_BlankNameResolvesNothing` does, would catch that.

api/webhook/webhook_test.go
```diff
@@ -145,3 +147,50 @@ func TestDeleteWebhook(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			var called atomic.Bool
+
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				called.Store(true)
+				results := []WebhookResponse{{ID: "hook-uuid-abc", Name: "deploy-hook"}}
+				if filter := strings.TrimSpace(r.URL.Query().Get("name")); filter != "" && filter != "deploy-hook" {
```
geunwoonoh: Warning: Without a `name` parameter this handler returns the target itself, so the `exact name` case passes even if the resolver sends the wrong key or the server ignores it. Putting a decoy first when unfiltered, as `TestGetServerIDByName_BlankNameResolvesNothing` does, would catch that.

api/packages/packages_test.go
```diff
@@ -460,3 +463,56 @@ func TestGetPythonPackageEntry_Pagination(t *testing.T) {
...
+	for _, packageType := range []string{"python", "system"} {
+		for _, tt := range tests {
+			t.Run(packageType+" "+tt.name, func(t *testing.T) {
+				t.Parallel()
+				var called atomic.Bool
+
+				ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					called.Store(true)
+					count := 1
+					if filter := strings.TrimSpace(r.URL.Query().Get("name")); filter != "" && filter != "numpy-2.0.0.whl" {
```
geunwoonoh: Warning: Without a `name` parameter this handler returns the target itself, so the `exact name` case passes even if the resolver sends the wrong key or the server ignores it. Here the server really does ignore `name` (see the comment on `packages.go`), so this handler pins a contract the API does not have.

api/server/server.go
```diff
@@ -98,7 +98,14 @@ func RequestServerAction(ac *client.AlpaconClient, serverName, action string, fo
 	return err
 }
 
+// GetServerIDByName refuses a name that trims to empty: the API drops a blank filter and
+// answers with the whole list, so Count > 0 would mean "unfiltered", not "found".
 func GetServerIDByName(ac *client.AlpaconClient, serverName string) (string, error) {
```
geunwoonoh, line 101: Suggestion: The same four-line guard now appears in 7 resolvers, and 6 `GetXIDByName` bodies are nearly identical. One helper in `api`, which already imports `client` and `utils`, could own trimming, the blank-name refusal, and the not-found check. It could also check `len(Results) == 0` instead of `Count == 0`, which today panics on a response with a positive `Count` and empty `Results`.

The PR description notes the grep sweep missed `ResolveRole`. As long as the guard is copied per site, the next resolver can miss it too.

api/server/server_test.go
```diff
@@ -588,3 +590,65 @@ func TestRequestServerAction(t *testing.T) {
 		})
 	}
 }
+
+// TestGetServerIDByName_BlankNameResolvesNothing pins that a blank name never reaches
+// the API, which drops the filter and would answer with the whole server list.
+func TestGetServerIDByName_BlankNameResolvesNothing(t *testing.T) {
```
geunwoonoh: Suggestion: `TestGetServerIDByName` (line 99) already covers `found` and `not found` for this function, and the same holds for the token, user, group, webhook, and role resolvers. The new cases are not identical because the handler differs, but they walk the same path. Adding the blank cases and a `wantCall` column to the existing tables would keep one table per resolver.

api/iam/iam_test.go
```diff
@@ -454,3 +454,54 @@ func TestDeleteMember(t *testing.T) {
 		})
 	}
 }
+
+// TestGetIDByName_BlankNameResolvesNothing pins that a blank name never reaches the API
+// for either resolver, which would otherwise hand back an arbitrary user or group.
+func TestGetIDByName_BlankNameResolvesNothing(t *testing.T) {
```
geunwoonoh: Suggestion: `TestGetIDByName` names a function that does not exist; cases in `TestGetUserIDByName` and `TestGetGroupIDByName` would read better. Also, `wantID == ""` doubles as the expect-an-error flag here and in the other new tests. A separate `wantErr bool`, as the existing tables use, says it directly.

api/packages/packages.go
```diff
@@ -68,6 +68,11 @@ func GetPythonPackageEntry(ac *client.AlpaconClient) ([]PythonPackage, error) {
 }
 
 func GetPackageIDByName(ac *client.AlpaconClient, fileName string, packageType string) (string, error) {
+	fileName = strings.TrimSpace(fileName)
```
geunwoonoh: Suggestion: The trimmed name stays local to this function. `DownloadPackage` builds the save path from the untrimmed `fileName` (line 175), so `"  numpy.whl  "` resolves but is saved under a file name with the surrounding spaces.

api/auth/auth.go
```diff
@@ -170,6 +170,11 @@ func GetAPITokenList(ac *client.AlpaconClient) ([]APITokenAttributes, error) {
 }
 
 func GetAPITokenIDByName(ac *client.AlpaconClient, tokenName string) (string, error) {
+	tokenName = strings.TrimSpace(tokenName)
```
geunwoonoh: Suggestion: `ResolveRole` now trims before `IsUUID`, but `ResolveTokenID` (line 199) checks `IsUUID` on the untrimmed value before calling this. A UUID with surrounding spaces resolves by ID for a role, while for a token it falls through to a name lookup and fails. Trimming in `ResolveTokenID` too would make the two agree.

api/rbac/rbac.go
```diff
@@ -44,10 +45,14 @@ func GetRoleCatalog(ac *client.AlpaconClient, autoAssigned *bool) ([]RoleRespons
...
 
-// The name filter is exact and case-sensitive. Role lists are narrowed to what the caller
-// may see rather than refused, so an invisible role yields an empty page and never a 403—the
-// not-found message covers both readings.
+// ResolveRole refuses a blank name: the API drops an empty filter and would answer with the
+// whole list. The name filter is exact and case-sensitive; an invisible role reads as not found.
 func ResolveRole(ac *client.AlpaconClient, nameOrID string) (*RoleResponse, error) {
+	nameOrID = strings.TrimSpace(nameOrID)
+	if nameOrID == "" {
+		return nil, errors.New("role name is required")
```
geunwoonoh: Suggestion: When this error fires, `resolveRoleForBinding` (`cmd/iam/user_role_grant.go:106`) still calls `ResolveRole` with the user argument to check for transposed arguments. If the username matches a role name such as `admin`, the user sees `No role named " ", but "admin" is one`. Skipping that check on a blank role also saves the extra request.

cmd/token/acl_server_add.go
```diff
@@ -31,7 +31,8 @@ func init() {
 
 func runServerAclAdd(cmd *cobra.Command, args []string) {
 	tokenArg := args[0]
-	serverName, _ := cmd.Flags().GetString("server")
+	rawServerName, _ := cmd.Flags().GetString("server")
+	serverName := strings.TrimSpace(rawServerName)
```
geunwoonoh: Suggestion: Reassigning in place matches the resolvers.

```suggestion
	serverName, _ := cmd.Flags().GetString("server")
	serverName = strings.TrimSpace(serverName)
```

`--servers` is not trimmed, so `--server " " --servers web-01` passes silently while `--server web --servers " "` fails with "not both". Moving flag validation into a helper would also make it unit-testable.

api/packages/packages.go
jisung-02: Fixed in df27428c. You were right that the guard blocks nothing here: the entries endpoints carry only `package__name`, so the `name` query was dropped and every lookup, blank or not, took the highest-version entry. I confirmed your reading against alpacon-server main before changing anything.

`GetPackageIDByName` now narrows with `search` and picks the entry whose `Name` equals `fileName` exactly, over `packageEntryRef` as you sketched. The python and system branches collapsed into one. It does not go through the new `api.ResolveByName` helper, because that helper is for a server-side exact filter and this lookup has to compare client-side.

The test moved with it in the same commit, since the old handler was emulating the filter that does not exist. Removing the exact compare and taking `entries[0]` again fails four cases:

```
--- FAIL: TestGetPackageIDByName/system/found
    Error:      	Not equal:
    expected: "pkg-uuid-abc"
    actual  : "id-decoy"
```

api/packages/packages_test.go
jisung-02: Fixed in df27428c, together with the resolver. You were right that this handler pinned a contract the API does not have.

It now behaves like the real endpoint: it honors `search` as substring narrowing and returns `numpy-2.0.0.whl.sig` ahead of `numpy-2.0.0.whl`, so the decoy is the one a first-result pick would take. Both package types run as a nested `t.Run(packageType)` rather than the double loop.

api/server/server.go
jisung-02: Applied in b066c11a. Your point about copying is what the branch had already demonstrated once: the first sweep grepped for `Results[0]` and missed `ResolveRole`, which indexes a `FetchPagesUpTo` slice.

`api.ResolveByName[T]` in the new `api/resolve.go` owns the trim, the blank refusal before any request, the single-page fetch under the caller filter key, and the not-found check. Six resolvers call it and keep their own message text, so no user-facing wording changed. `api/packages` is the one that does not, for the reason in the thread on `packages.go`.

The `Count` versus `len(Results)` half is in the same commit: the helper decides on `len(results)`, so a positive `Count` with an empty `Results` returns not-found instead of panicking.

It also carries `api.ErrBlankName`, so the refusal is distinguishable with `errors.Is` rather than by message text. That is what the `cmd/iam` fix in your other thread needed.

api/iam/iam_test.go
jisung-02: Fixed in e5894cf5. You were right on both counts: the handler answered every request with the target, and it keyed on neither field consistently, so `GetUserIDByName` sending `name` would have looked correct.

Users now filter on `username` with `first-user` as the decoy, groups on `name` with `first-group`, and both tables gained an `unknown name` case. Renaming either key fails the lookups:

```
--- FAIL: TestGetUserIDByName/found
--- FAIL: TestGetUserIDByName/unknown_name
--- FAIL: TestGetUserIDByName/padded_name
--- FAIL: TestGetGroupIDByName/found
--- FAIL: TestGetGroupIDByName/unknown_name
--- FAIL: TestGetGroupIDByName/padded_name
```

api/auth/auth_test.go
jisung-02: Fixed in e5894cf5. The handler answered any request with the target, so the test proved nothing about the filter.

`first-token` now comes back first when the filter is absent. Renaming the key to anything the handler does not know fails three cases:

```
--- FAIL: TestGetAPITokenIDByName/found
--- FAIL: TestGetAPITokenIDByName/padded_name
--- FAIL: TestGetAPITokenIDByName/not_found
```

api/webhook/webhook_test.go
jisung-02: Fixed in e5894cf5, same treatment as the token test. `first-webhook` is returned first when the filter is absent:

```
--- FAIL: TestGetWebhookIDByName/found
--- FAIL: TestGetWebhookIDByName/padded_name
--- FAIL: TestGetWebhookIDByName/not_found
```

api/server/server_test.go
jisung-02: Applied in e5894cf5. There is one table per resolver now: the blank and padded cases moved into `TestGetServerIDByName`, `TestGetAPITokenIDByName`, `TestGetUserIDByName`, `TestGetGroupIDByName`, `TestGetWebhookIDByName`, and `TestResolveRole_ByName`, and the separate `_BlankNameResolvesNothing` tables are gone. `wantCall` came along, so a guard added at a single call site still fails them.

api/iam/iam_test.go
jisung-02: Applied in e5894cf5. `TestGetIDByName_BlankNameResolvesNothing` is gone and its cases sit in `TestGetUserIDByName` and `TestGetGroupIDByName`, which are functions that exist.

`wantErr bool` replaced `wantID == ""` across all six tables, matching what the pre-existing tables already did. The blank cases also assert `require.ErrorIs` against the shared sentinel instead of matching a message.

api/packages/packages.go
jisung-02: Fixed in 6eb438f9. The trimmed name stayed inside the resolver, so a padded argument resolved and downloaded and then landed on disk under a file name carrying the spaces. `DownloadPackage` trims once at the top, so the name it resolves and the name it writes are the same one.

api/auth/auth.go
jisung-02: Fixed in 1da6d266. `ResolveTokenID` trims before `utils.IsUUID` now, so both resolvers take the id branch for a padded UUID. You were right that the disagreement came from this branch, not from the original code.

api/rbac/rbac.go
jisung-02: Fixed in 3b3d78bc. The hint exists for `grant admin john`, and firing it on an empty argument points the operator at a mistake they did not make.

The transposed-argument check is skipped when the failure was the blank-name refusal, read as `errors.Is(err, rbac.ErrBlankName)`. The sentinel comes from `api.ErrBlankName` in b066c11a and `api/rbac` re-exports it. That also drops the second request.

cmd/token/acl_server_add.go
jisung-02: Applied in 3a1a9179. `serverName` is reassigned in place as you wrote it.

The asymmetry you found behind it is fixed too: `--servers` is trimmed now, so a whitespace-only value counts as absent on either side rather than passing silently on one and being refused as "not both" on the other. The exactly-one-of check moved into `validateServerAclFlags`, with a table test. Inverting its not-both condition fails two of the four cases:

```
--- FAIL: TestValidateServerAclFlags/both_flags_set
    Error:      	An error is expected but got nil.
--- FAIL: TestValidateServerAclFlags/only_servers_set
    Error:      	Received unexpected error: use either --server or --servers, not both
```

api/server/server_test.go
```diff
@@ -124,17 +147,16 @@ func TestGetServerIDByName(t *testing.T) {
 			id, err := GetServerIDByName(ac, tt.serverName)
 
 			if tt.wantErr {
-				if err == nil {
-					t.Errorf("expected error, got nil")
+				require.Error(t, err)
+				assert.Empty(t, id)
+				if tt.serverName == "" {
```
geunwoonoh: Suggestion: Only the empty case asserts the sentinel. `"   "`, the input the bug was reported with, never checks which error came back—`wantCall=false` still proves no request went out, so it cannot slip through, but the sentinel check is what `errors.Is` callers rely on. The same condition is in `auth_test.go:247`, `iam_test.go:208` and `:273`, `rbac_test.go:154`, and `webhook_test.go:127`.

```suggestion
				if strings.TrimSpace(tt.serverName) == "" {
```

cmd/token/acl_server_add.go
```diff
@@ -84,3 +84,16 @@ func runServerAclAdd(cmd *cobra.Command, args []string) {
 	}
 	utils.CliSuccess("Server ACLs added: token %s can access [%s]", tokenArg, strings.Join(names, ", "))
 }
+
+// validateServerAclFlags takes both values already trimmed, so a blank one means the flag was
+// absent or whitespace-only.
+func validateServerAclFlags(serverName, serversCSV string) error {
```
geunwoonoh: Suggestion: The trim that makes `--server " "` count as absent happens in the caller (lines 36-38), so `TestValidateServerAclFlags` only ever sees values that are already trimmed and cannot cover the input this change was made for. Trimming inside the helper would let the table carry a `"   "` case for each flag.

api/packages/packages.go
```diff
@@ -67,32 +72,26 @@ func GetPythonPackageEntry(ac *client.AlpaconClient) ([]PythonPackage, error) {
...
 }
 
+// GetPackageIDByName matches client-side because the entries endpoints filter on package__name,
+// never name, so an unknown name query is dropped and the whole list comes back.
 func GetPackageIDByName(ac *client.AlpaconClient, fileName string, packageType string) (string, error) {
-	params := map[string]string{"name": fileName}
-	body, err := ac.SendGetRequest(utils.BuildURL(packageEntryURL(packageType), "", params))
+	fileName = strings.TrimSpace(fileName)
+	if fileName == "" {
+		return "", errors.New("package name is required")
```
geunwoonoh: Suggestion: This refusal does not satisfy `api.ErrBlankName`, although the sentinel's comment says it marks a resolver's refusal of a blank name. No caller tells this error apart today, so nothing breaks. A small `api` helper that only trims and refuses (say `api.RequireName`) could serve both `ResolveByName` and this function, which would also remove the last copy of the guard.

api/resolve.go
```diff
@@ -0,0 +1,38 @@
...
+// blankNameError carries a caller-chosen message while still satisfying errors.Is(err,
+// ErrBlankName), so each resolver keeps its own wording without losing the shared sentinel.
+type blankNameError struct{ msg string }
+
+func (e *blankNameError) Error() string        { return e.msg }
+func (e *blankNameError) Is(target error) bool { return target == ErrBlankName }
+
+// ResolveByName looks up a single resource through a server-side, exact-match name filter.
+// A blank name is refused up front—an empty filter comes back as the whole list, not "found".
+func ResolveByName[T any](ac *client.AlpaconClient, endpoint, filterKey, name, blankMsg, notFoundMsg string) (*T, error) {
```
geunwoonoh: Suggestion: Five string parameters in a row means swapping `blankMsg` and `notFoundMsg`, or `filterKey` and `name`, compiles and passes the tests—the only message assertion is rbac's `case-sensitive`. Grouping them in a small struct is worth considering.

api/auth/auth.go
```diff
@@ -170,28 +170,17 @@ func GetAPITokenList(ac *client.AlpaconClient) ([]APITokenAttributes, error) {
...
-	if response.Count == 0 {
-		return "", errors.New("no token found with the given name")
-	}
-
-	return response.Results[0].ID, nil
+	return result.ID, nil
 }
 
 func ResolveTokenID(ac *client.AlpaconClient, nameOrID string) (string, error) {
+	nameOrID = strings.TrimSpace(nameOrID)
```
geunwoonoh, line 188: Suggestion: This fix has no regression test. A UUID with surrounding spaces is one more row in the `TestResolveTokenID` table, expecting the UUID back with no HTTP call.

geunwoonoh (dismissed):
## PR review: #447 - fix(api): reject a blank name before it becomes an unfiltered lookup

**Author**: @jisung-02
**Branch**: 436-normalize-server-name-input → main
**Stack**: golang
**Changes**: +459 -212 in 16 files (head `e5894cf`, all 12 CI checks pass)

---

### Summary

Every finding from the previous review is in the code. The refactor also raises two possible regressions, and neither happens:
- **Paging parameters**: `ResolveByName` sends `page`/`page_size`, which all five endpoints already accept through `FetchAllPages`.
- **Error detection**: HTTP errors now come back wrapped as `fetching page 1 from ...: %w`. `ParseErrorResponse` and `HTTPStatusCode` walk the `Unwrap` chain, and nothing type-asserts an error directly, so the MFA retry and the work session denial (exit 3) still branch the same way.

### Previous findings

| Finding | Commit | Result |
|---|---|---|
| W1 package entry API has no `name` filter | df27428 | Narrows with `search`, then compares `packageEntryRef.Name` exactly; the python/system branches are merged |
| W2 tests pass with the filter dropped | df27428, e5894cf | All six tables put a decoy first; iam filters on `username` and `name` separately |
| S1 shared helper | b066c11 | Six resolvers call `api.ResolveByName`; packages is left out because the server has no filter there, which is right |
| S2 decide on `len(Results)` | b066c11 | The helper decides on `len(results)` |
| S3 merge into existing tables | e5894cf | The six separate tests are gone; their cases live in the existing tables |
| S4 `wantErr`, test name | e5894cf | Done; one gap in the blank-name check (suggestion 1) |
| S5 `DownloadPackage` trim | 6eb438f | Done; no test |
| S6 `ResolveTokenID` trim | 1da6d26 | Done; no test |
| S7 transposed-argument hint | 3b3d78b | Skipped when `errors.Is(err, rbac.ErrBlankName)` |
| S7 `--servers` asymmetry | 3a1a917 | Done, with a validation helper and a test; one gap (suggestion 2) |
| Sweep claim in the description | description | Corrected in the description |

### Critical issues (0)

### Warnings (0)

### Suggestions (7, none blocking)

1. **[api/server/server_test.go:152 and 5 more] A whitespace-only name never checks `ErrBlankName`**
   - The condition is `tt.serverName == ""`, so only the empty case asserts the sentinel.
   - `"   "`, the input the bug was reported with, does not check which error came back. `wantCall=false` still proves no request went out, so the input cannot slip through.
   - `strings.TrimSpace(tt.serverName) == ""` covers both.
2. **[cmd/token/acl_server_add.go:90] The input behind this change is untested**
   - `validateServerAclFlags` takes values already trimmed, and the trim happens in the caller (:36-38).
   - So the behavior that motivated it, `--server " "` counting as absent, is not in the table.
   - Trimming inside the helper would let the table carry a `"   "` case.
3. **[api/packages/packages.go:80] The package resolver's blank refusal does not satisfy `api.ErrBlankName`**
   - The `ErrBlankName` comment says it marks a resolver's refusal of a blank name, and this one does not carry it.
   - No caller tells this error apart today, so nothing breaks.
   - A small helper that only trims and refuses (say `api.RequireName`) could serve both `ResolveByName` and this function.
4. **[api/resolve.go:23] Five string parameters in a row**
   - Swapping `blankMsg` and `notFoundMsg` would pass both the compiler and the tests; the only message assertion is rbac's `case-sensitive`.
   - Grouping them in a struct is worth considering.
5. **Missing regression tests**
   - A UUID with surrounding spaces (1da6d26) is one more row in the `TestResolveTokenID` table.
   - The skipped hint (3b3d78b) runs through `CliErrorWithExit`, so leaving it untested is reasonable.
6. **[CLAUDE.md] Add `api.ResolveByName` to Key patterns**
   - The helper exists so the next resolver cannot miss the guard.
   - If the docs do not mention it, the next author may copy the logic again without knowing it is there.
7. **Stale output in the PR description**
   - Line 94's `Error: One of --server or --servers is required.` now reads `one of ...` in lowercase since 3a1a917.
   - The description also does not mention the `fetching page 1 from /api/...:` prefix that lookup HTTP errors now carry.

### Additional review

**Go conventions**
- **Declaration order**: `resolve.go`, `packages.go`, and `rbac.go` all follow `const → var → type → func`.
- **Error strings**: lowercase with no trailing punctuation (ST1005).
- **Test rules**: every case runs under `t.Run` with `t.Parallel()`, and no handler calls `require`. On Go 1.25, loop variable capture is not a concern.
- **`Is` method on `blankNameError`**: `fmt.Errorf("%s: %w")` would append `: blank name` to every message, so this is the right way to keep the existing wording.
- **Re-exporting `rbac.ErrBlankName`**: keeps `cmd/iam` from importing `api` directly; a common pattern.

**Duplicated or unused code**
- **Unused code**: none.
  - `PythonPackageDetail` and `SystemPackageDetail` are still used by the list functions.
  - Imports that became unnecessary are removed, and lint passes.
- **Duplication**:
  - **Resolver bodies**: six are now a single helper call each.
  - **Test handlers**: the filter emulation is nearly the same across six tables, but each package has its own response type, so sharing it gains little.
  - **Trim and blank refusal**: still in two places, the helper and the package resolver (suggestion 3).

**Performance**
- **`page_size=1` in the helper**: an exact filter returns a single row anyway, so there is effectively no difference.
- **Package lookup**: one request before, now every page of the `search` results. Requests only grow past 100 matches, and a file name search usually ends on the first page. A fair cost for correctness.
- **Negligible costs**: `ResolveRole` builds its not-found message with `fmt.Sprintf` even on success, and `DownloadPackage` trims twice.

### Highlights
- **Commits and verification**: each finding got its own commit, and each fix was reverted to watch the test fail, with the output recorded.
- **Helper scope**: packages stays out of the helper because the server has no filter there, a decision that follows the server's contract.
- **Error branching**: `ErrBlankName` with `errors.Is` instead of comparing message text.
- **PR description**: the wrong sweep claim is corrected in the open rather than removed.

---

**Verdict**: APPROVED

The remaining seven items are suggestions and do not block the merge.

**Checklist**:
- [x] Tests included and passing in CI
- [x] No security vulnerabilities (the package resolver's unfiltered lookup is closed)
- [ ] Documentation updated if needed (a CLAUDE.md note on the helper is recommended, not required)

api/server/server_test.go
jisung-02: Fixed in 3a9ee8b5. You were right that `wantCall=false` only proves no request went out—the sentinel is what an `errors.Is` caller reads, and `"   "` never reached that assertion. All six tables now key on `strings.TrimSpace(...) == ""`.

Returning a plain `errors.New` from `api.RequireName` instead of the sentinel-bearing error now fails the whitespace rows alongside the empty ones:

```
--- FAIL: TestGetServerIDByName/whitespace_only_name
--- FAIL: TestGetAPITokenIDByName/whitespace_only_name
--- FAIL: TestGetUserIDByName/whitespace_only_name
--- FAIL: TestGetGroupIDByName/whitespace_only_name
--- FAIL: TestGetWebhookIDByName/whitespace_only_name
--- FAIL: TestResolveRole_ByName/whitespace_only_name
```

cmd/token/acl_server_add.go
jisung-02: Fixed in aee7786e. The helper only ever saw trimmed values, so the input this branch exists for was outside its table.

`validateServerAclFlags` trims both flags itself now and hands the trimmed pair back, so the caller resolves the same name the validator accepted. The table carries a whitespace-only case for each flag. Dropping the two trim lines fails exactly those and nothing else:

```
--- FAIL: TestValidateServerAclFlags/whitespace_only_server_set
--- FAIL: TestValidateServerAclFlags/whitespace_only_servers_set
--- FAIL: TestValidateServerAclFlags/server_with_surrounding_whitespace
--- FAIL: TestValidateServerAclFlags/whitespace_only_server_with_servers_set
```

api/packages/packages.go
jisung-02: Fixed in e2aec4c7, as `api.RequireName`—the name you suggested. You were right that the sentinel comment claimed something this refusal did not do.

`RequireName` is the trim-and-refuse half split out of `ResolveByName`, which now calls it, so this is the last copy of the guard gone. `GetPackageIDByName` still cannot use the full helper: the entries endpoints filter on `package__name`, so it narrows with `search` and compares client-side.

api/resolve.go
jisung-02: Applied in e2aec4c7. `ResolveByName` takes `api.ResolveByNameOptions` with `Endpoint`, `FilterKey`, `Name`, `BlankMsg` and `NotFoundMsg`, so swapping two same-typed arguments changes a field name the compiler checks rather than passing silently. All six call sites pass the struct.

api/auth/auth.go
jisung-02: Fixed in 3a9ee8b5. `TestResolveTokenID` carries a padded UUID row expecting the UUID back with no HTTP call. Moving the `utils.IsUUID` check back ahead of the trim fails that row alone:

```
--- FAIL: TestResolveTokenID/padded_uuid_fast-path_—_no_HTTP_request
```

api/rbac/rbac_test.go
```diff
@@ -98,41 +100,67 @@ func TestScopeLabel(t *testing.T) {
...
+			role, err := ResolveRole(newTestClient(ts), tt.roleName)
+
 			if tt.wantErr {
 				require.Error(t, err)
-				assert.Contains(t, err.Error(), "case-sensitive")
-				return
+				assert.Nil(t, role)
+				if strings.TrimSpace(tt.roleName) == "" {
+					require.ErrorIs(t, err, ErrBlankName)
+				} else if strings.TrimSpace(tt.roleName) != "" {
```
geunwoonoh: Suggestion: Now that the branch above tests `strings.TrimSpace(tt.roleName) == ""`, this condition holds on every path that reaches it.

```suggestion
				} else {
```

geunwoonoh (dismissed):
## PR review: #447 - fix(api): reject a blank name before it becomes an unfiltered lookup

**Author**: @jisung-02
**Branch**: 436-normalize-server-name-input → main
**Stack**: golang
**Changes**: +545 -212 in 17 files (head `a11dd1b`, all 12 CI checks pass)

---

### Summary

All seven suggestions from the previous round are in, and the only new finding is one dead condition in a test.

| Suggestion | Commit | Result |
|---|---|---|
| S1 `ErrBlankName` on a whitespace-only name | 3a9ee8b | All six tables now key off `strings.TrimSpace(...) == ""` |
| S2 validator saw only trimmed values | aee7786 | The validator trims and hands the trimmed pair back, so the caller resolves the name it validated; four whitespace cases added |
| S3 package resolver missed the sentinel | e2aec4c | `api.RequireName` splits the trim-and-refuse half out, and packages carries `ErrBlankName` too |
| S4 five string parameters | e2aec4c | `ResolveByNameOptions` gives every argument a field name |
| S5 missing regression test | 3a9ee8b | `TestResolveTokenID` covers the padded UUID, including that no HTTP call goes out |
| S6 CLAUDE.md note | a11dd1b | Added to Key patterns; the helper's reach, the `len(results)` decision, the sentinel, and the packages exception all match the code |
| S7 stale PR description | description | The output reads `one of ...` in lowercase, and the `fetching page 1 from ...` wrapping is written up |

### Critical issues (0)

### Warnings (0)

### Suggestions (1)

1. **[api/rbac/rbac_test.go:156] A condition that is now always true**
   - With the branch above changed to `strings.TrimSpace(tt.roleName) == ""`, the `else if strings.TrimSpace(tt.roleName) != ""` holds on every path that reaches it.
   - `} else {` says the same thing.

### Additional review

**Go conventions**
- **Declaration order**: `resolve.go` runs `var → type → type → func`, which follows `const → var → type → func`. `ResolveByNameOptions` does sit between `blankNameError` and its methods, which is a matter of reading preference, not a rule.
- **Error handling**: `validateServerAclFlags` returns values alongside its error. The caller exits on the error and the comment says why the values come back, so nothing is left ambiguous.
- **Tests**: the new cases keep `t.Run` with `t.Parallel()`, and no case name carries a regex metacharacter.

**Duplicated or unused code**
- **Unused code**: none. `errors` and `strings` are still used in every file that keeps them, and lint passes.
- **Duplication**: the trim and the blank refusal now live in `RequireName` alone, which removes the last copy.

**Performance**
- The options struct copies five strings per lookup, which is not measurable.
- The request shape and the paging parameters are unchanged from the previous round.

### Highlights
- **How the suggestions were taken**: rather than applying the suggestion as written, `RequireName` splits the guard so even the package resolver, which cannot use the helper, carries the same sentinel.
- **The validator**: returning the trimmed values removes any chance of validating one string and resolving another.
- **Documentation**: the CLAUDE.md entry records the helper's reach, its one exception, and the reason (#436), so the next resolver is unlikely to repeat this.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing in CI
- [x] Documentation updated (CLAUDE.md)

api/rbac/rbac_test.go
jisung-02: Applied in 772416de. The `else if` restated the branch above it: once `strings.TrimSpace(tt.roleName) == ""` has been taken, the negation holds on every path that reaches the second arm, so the condition decided nothing. A plain `else` is what it meant.

The arm still has teeth—`not found or not visible` is the one row that lands there, and it asserts the `case-sensitive` message:

```
ok  	github.com/alpacax/alpacon-cli/api/rbac	1.509s
```

This was the last `else if` of that shape; the other five tables key on the trim in a single `if`/`else`.

geunwoonoh (approved):
## PR review: #447 - fix(api): reject a blank name before it becomes an unfiltered lookup

**Author**: @jisung-02
**Branch**: 436-normalize-server-name-input → main
**Stack**: golang
**Changes**: +545 -212 in 17 files (head `772416d`, all 12 CI checks pass)

---

### Summary

The one suggestion from the previous round is in. With `strings.TrimSpace(tt.roleName) == ""` already taken above, the `else if` that followed held on every path, and it is now a plain `else`. The assertions behave the same: a blank name is checked against `ErrBlankName`, and every other failure against the `case-sensitive` message. Nothing else changed.

### Critical issues (0)

### Warnings (0)

### Suggestions (0)

### Highlights
- **A one-line change**: the finding was fixed where it was raised, and the commit message records why the condition always held.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing in CI
- [x] Documentation updated (CLAUDE.md, in the previous round)

PR #453, opened 2026-09-17
refactor(ftp): drop the webftp_transfer_in_progress status branch

### LINKED ISSUE #450: Drop the dead webftp_transfer_in_progress string branch and TransferErrorResponse
## Background

`PollTransferStatus` (`api/ftp/ftp.go:105`) detects an in-progress transfer by substring-matching the flattened error text:

```go
if !strings.Contains(err.Error(), "webftp_transfer_in_progress") {
    return false, "", fmt.Errorf("failed to check transfer status: %w", err)
}
```

alpacax/alpacon-server#3404 stops the server from ever producing that response. `GET /api/webftp/{uploads,downloads}/<id>/status/` now answers 200 with `{"success": null, "message": null}` while a transfer is running, instead of 400 with `{"code": "webftp_transfer_in_progress"}`.

Nothing breaks. The 200 path is already the normal one here: `roundTrip` accepts it, the body parses, `if statusResp.Success != nil` (`api/ftp/ftp.go:137`) declines to return, and the loop sleeps and polls again. `TestPollTransferStatus_TimesOut` (`api/ftp/ftp_test.go:1442`) already drives a server that only ever answers `success: null`. Released tags v1.13.1, v1.11.2, v1.10.0 and v1.9.0 carry the same function, and even v1.5.0 falls through to its sleep.

So this is cleanup, not a fix.

## What to remove

1. The string branch and its comment at `api/ftp/ftp.go:125-131`. Once the server no longer sends that payload, a non-2xx on this endpoint is a real failure and should be returned as one.
2. `TransferErrorResponse` in `api/ftp/types.go:62`. It declares a `Code string` field and has zero references anywhere in the repo—nothing ever read a structured error code from this endpoint.
3. The test fixture that mocks the old response, `api/ftp/ftp_test.go:1395-1403`. Worth noting it mocks the in-progress answer as 422, while the server actually sent 400. The mismatch never surfaced because the poller reads the text and ignores the status code—which is the fragility the server issue set out to remove.

## Ordering

Only after the server change has shipped to every region. Until then the branch is what keeps a poll against an older server from failing. It costs nothing to keep, so there is no hurry.

jisung-02:
## Summary

`PollTransferStatus` detected an in-progress transfer by substring-matching the flattened error text for `webftp_transfer_in_progress`. alpacon-server 2.36.0 (alpacax/alpacon-server#3404, shipped 2026-09-16) stopped producing that response: `GET /api/webftp/{uploads,downloads}/<id>/status/` now answers 200 with `{"success": null, "message": null}` while a transfer runs, instead of 400 with `{"code": "webftp_transfer_in_progress"}`.

The 200 path is the one the poller already took for every other answer: the body parses, `Success` is nil, and the loop sleeps and polls again. So the error branch has nothing left to catch.

## Changes

- `api/ftp/ftp.go` — remove the string branch and its comment. A non-2xx on this endpoint is a real failure now and is returned as one. The `if err != nil` arm always returns, so the `else` becomes an early return and the success block unindents. The `json.Unmarshal` failure path also stops returning `statusResp.Message`, which is always `""` there because `statusResp` is still the zero value.
- `api/ftp/types.go` — remove `TransferErrorResponse`. It declared a `Code` field and had zero references in the repo; nothing ever read a structured error code from this endpoint.
- `api/ftp/ftp_test.go` — remove `TestPollTransferStatus_RetriesWhileInProgress`, which exercised only the removed branch. Drop the neighbouring test's comment rather than reword it: `TestPollTransferStatus_FatalErrorNoRetry` and its `"fatal error must not be retried"` assertion already say what it said. Cut `TestPollTransferStatus_Timeout`'s budget from 3s to 1s, since `TestPollTransferStatus_TimesOut` covers the same wait in 0.00s and additionally pins the poll count.
- `README.md` — state the minimum server version `alpacon cp` needs, and what breaks below it.

## The assumption this change rests on, and where it now lives

The branch was not dead before 2.36.0, contrary to what #450 states. `parseAPIErrorPayload` (`client/client.go`) folds every top-level string field into the message when the body carries no `detail`, so the old body `{"code": "webftp_transfer_in_progress"}` reached the poller as the error text `code: webftp_transfer_in_progress`, and the substring matched. Verified by driving the real client against an httptest server returning that body.

Removing the branch therefore assumes every server this CLI talks to is at 2.36.0 or newer. Nothing in the code enforces that—there is no minimum-version gate, no API version header, no negotiation—and self-hosted workspaces upgrade on their own schedule. Keeping a version-tolerant guard was the other option considered, in the form @geunwoonoh proposed: read the structured code through `utils.ParseErrorResponse` rather than the message text, which costs three lines and matches no prose at all. It was rejected on the maintainer's confirmation that every deployed server is already at 2.36.0 or newer.

What the earlier revision of this body got wrong is where that assumption was written down. A PR body stops being findable the day after merge, so the assumption now lives in `README.md` beside the `cp` documentation: `alpacon cp` requires server 2.36.0 or later, and below it the command fails on its first transfer-status poll instead of waiting for the transfer to finish.

## On the deleted test

It mocked the in-progress answer as 422 while the server sent 400. The mismatch never surfaced because the poller read the text and ignored the status code, which is the fragility the server change set out to remove.

Coverage does not drop: `TestPollTransferStatus_TimesOut` already drives a server that only ever answers `success: null`, `TestPollTransferStatus_BacksOffThenSucceeds` covers polling through to success, and `TestPollTransferStatus_FatalErrorNoRetry` covers an error ending the poll at once.

## Not in this PR

`PollTransferStatus` keeps its own pacing (`backoffDelay`, `nextPollInterval`, `alignedPollDelay`) while `CLAUDE.md` names `utils/poll.go` as the one place that paces the CLI's wait loops. With the error arm now unconditional, a single 429 mid-transfer ends the wait, where `utils.ThrottleBudget` and `utils.WarnThrottled` exist for that case. Filed as #458.

## Testing

Run on the pushed head, `go version go1.26.6 darwin/arm64`:

```
$ go build ./...
$ gofmt -l api/ftp
$ go vet ./...
(all three clean, no output)

ok  	github.com/alpacax/alpacon-cli/api/ftp	2.380s
...
ok  	github.com/alpacax/alpacon-cli/utils	1.798s
(44 packages, all ok)
```

The `api/ftp` package went from 3.478s to 2.380s with the timeout budget cut.

Closes #450


api/ftp/ftp.go
```diff
@@ -122,21 +122,14 @@ func PollTransferStatus(ac *client.AlpaconClient, transferType, id string, timeo
...
-			}
-		} else {
-			var statusResp TransferStatusResponse
-			if err := json.Unmarshal(respBody, &statusResp); err != nil {
-				return false, statusResp.Message, fmt.Errorf("failed to parse transfer status response: %w", err)
-			}
-			if statusResp.Success != nil {
-				return *statusResp.Success, statusResp.Message, nil
-			}
+			return false, "", fmt.Errorf("failed to check transfer status: %w", err)
```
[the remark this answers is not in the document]
jisung-02: The mechanism you describe is correct, and it is sharper than #450 assumed. The issue called this branch dead code; it was not. `parseAPIErrorPayload` folds every top-level string field into the message when the body carries no `detail`, so the pre-2.36.0 body `{"code": "webftp_transfer_in_progress"}` reached the poller as the error text `code: webftp_transfer_in_progress` and the substring matched. I verified that by driving the real client against an httptest server returning that body.

You are also right that nothing in this repo enforces a floor: there is no minimum server version, no API version header, and no negotiation.

Removing it anyway is a deliberate call, on the maintainer's confirmation that every deployed server is already at 2.36.0 or newer. That is an operational fact the code cannot state, so it is written into the PR body and the commit message under "Assumption this change rests on", together with what breaks if it stops holding: a server left on an older release fails its first status poll instead of waiting for the transfer to finish.

Keeping the branch was the other option considered. It was rejected because the substring match is the fragility alpacax/alpacon-server#3404 set out to remove, and retaining it keeps the CLI reading error prose. If a compatibility floor is ever needed, the right shape is a version gate plus a check on the structured error code the client already parses, not this text match.

api/ftp/ftp.go
```diff
@@ -122,21 +122,14 @@ func PollTransferStatus(ac *client.AlpaconClient, transferType, id string, timeo
...
+		}
+		var statusResp TransferStatusResponse
+		if err := json.Unmarshal(respBody, &statusResp); err != nil {
+			return false, statusResp.Message, fmt.Errorf("failed to parse transfer status response: %w", err)
+		}
+		if statusResp.Success != nil {
+			return *statusResp.Success, statusResp.Message, nil
 		}
```
geunwoonoh, line 133: Warning: the reason given for dropping compatibility does not hold—a structured-code fallback is already available.

The reply to Copilot rejects keeping the fallback because "retaining it keeps the CLI reading error prose". That is true of the *old* shape, but the client already parses the structured code and `utils` already exports a reader for it. Verified by driving the real `client.AlpaconClient` against an httptest server returning the pre-2.36.0 body:

```
err="code: webftp_transfer_in_progress"  code="webftp_transfer_in_progress"  status=400
```

`utils.ParseErrorResponse(err)` returns the code (`client.parseAPIErrorPayload` -> `newAPIError` -> `codedError`), and `api/ftp` already imports `utils`, so the guard matches no text at all. The loop's deadline still bounds it, so a genuinely stuck old server times out as before.

The suggestion restores the `else` this PR unindented. That is what keeping the fallback actually costs, and showing it is the honest version of the trade:

```suggestion
		respBody, err := ac.SendGetRequest(statusURL)
		if err != nil {
			// A pre-2.36.0 server answers a running transfer with a coded 400
			// instead of 200 {"success": null}. Read the code the client already
			// parsed—not the message text—and keep polling; the deadline bounds it.
			if code, _ := utils.ParseErrorResponse(err); code != "webftp_transfer_in_progress" {
				return false, "", fmt.Errorf("failed to check transfer status: %w", err)
			}
		} else {
			var statusResp TransferStatusResponse
			if err := json.Unmarshal(respBody, &statusResp); err != nil {
				return false, statusResp.Message, fmt.Errorf("failed to parse transfer status response: %w", err)
			}
			if statusResp.Success != nil {
				return *statusResp.Success, statusResp.Message, nil
			}
		}
```

This is the maintainer's call, not mine—the fleet-upgraded confirmation is an operational fact I cannot check. But the trade is no longer "fragile prose match vs. clean removal", it is "three lines with no operational assumption vs. clean removal plus an assumption nothing enforces". If the removal stands, the assumption deserves a home the code can point at (a minimum server version in `README.md`, or the `## Important notes` block in `CLAUDE.md`)—a PR body stops being findable the day after merge, and the failure it predicts is `alpacon cp` failing its first status poll on a live transfer.

api/ftp/ftp.go
```diff
@@ -122,21 +122,14 @@ func PollTransferStatus(ac *client.AlpaconClient, transferType, id string, timeo
...
-				return false, statusResp.Message, fmt.Errorf("failed to parse transfer status response: %w", err)
-			}
-			if statusResp.Success != nil {
-				return *statusResp.Success, statusResp.Message, nil
-			}
+			return false, "", fmt.Errorf("failed to check transfer status: %w", err)
+		}
+		var statusResp TransferStatusResponse
+		if err := json.Unmarshal(respBody, &statusResp); err != nil {
+			return false, statusResp.Message, fmt.Errorf("failed to parse transfer status response: %w", err)
```
geunwoonoh: Suggestion: this message is always empty (pre-existing; the line only moved here).

When `json.Unmarshal` fails, `statusResp` is the zero value, so the caller receives `""`. Returning `""` outright would say what actually happens instead of implying a partial parse is being salvaged.

```suggestion
			return false, "", fmt.Errorf("failed to parse transfer status response: %w", err)
```

api/ftp/ftp_test.go
```diff
@@ -1390,37 +1390,10 @@ func TestPollTransferStatus_BacksOffThenSucceeds(t *testing.T) {
...
-	assert.Equal(t, "done", message)
-	assert.Equal(t, int32(3), calls.Load())
-}
-
 func TestPollTransferStatus_FatalErrorNoRetry(t *testing.T) {
 	t.Parallel()
-	// A non-in-progress error (e.g. 403) is fatal: return immediately without
-	// polling again.
+	// Any error from the status endpoint is fatal: return immediately
+	// without polling again.
```
geunwoonoh: Suggestion: this comment restates the identifiers—delete it.

The test is named `TestPollTransferStatus_FatalErrorNoRetry` and its assertion already carries `"fatal error must not be retried"`. Three sayings of one fact. The old comment earned its place by naming a distinction (403 vs. in-progress) that no longer exists; the new one names nothing the identifiers do not.

```suggestion
```

If something should stay here, it is the thing the code cannot say—that after 2.36.0 there is no exempt error class left, because an in-progress transfer is not an error at all.

geunwoonoh (commented):
### Summary

The poller no longer keys retry off error prose. With server 2.36.0 an in-progress transfer arrives as `200 {"success": null}`, which is the path the loop already took, so the error arm collapses to a plain fatal return and the success block unindents. The unreferenced `TransferErrorResponse` type and the test that only drove the removed branch go with it.

I verified the PR body's central claim rather than taking it: driving the real `client.AlpaconClient` against an httptest server returning the pre-2.36.0 body produced `err = "code: webftp_transfer_in_progress"`, status `400`. The branch was live, not dead—#450's premise was wrong and this PR's is right.

### Critical Issues (0)

None. The change is correct against a 2.36.0+ server, builds, vets, and passes with `-race -shuffle=on`.

### Warnings (1)

**1. [`api/ftp/ftp.go:123-125`] The reason given for dropping compatibility does not hold—a structured-code fallback is already available**

The reply to Copilot rejects keeping the fallback because "retaining it keeps the CLI reading error prose". That is true of the *old* shape, but the client already parses the structured code and `utils` already exports a reader for it. Verified in this repo:

```
err="code: webftp_transfer_in_progress"  code="webftp_transfer_in_progress"  status=400
```

`utils.ParseErrorResponse(err)` returns the code (`client.parseAPIErrorPayload` -> `newAPIError` -> `codedError`), and `api/ftp` already imports `utils`, so the version-tolerant guard costs three lines and matches no text at all. The loop's deadline still bounds it, so a genuinely stuck old server times out as before. Inline comment carries the applicable diff.

This is the maintainer's call, not mine—the fleet-upgraded confirmation is an operational fact I cannot check. But it changes the trade: the choice is no longer "fragile prose match vs. clean removal", it is "three lines with no operational assumption vs. clean removal plus an assumption nothing enforces". If the removal stands, the assumption deserves a home the code can point at (a minimum server version in `README.md`, or the `## Important notes` block in `CLAUDE.md`)—a PR body stops being findable the day after merge, and the failure it predicts is `alpacon cp` failing its first status poll on a live transfer.

### Suggestions (3)

**1. [`api/ftp/ftp_test.go:1395-1396`] The rewritten comment is a restatement—delete it**

The test is named `TestPollTransferStatus_FatalErrorNoRetry` and its assertion already carries `"fatal error must not be retried"`. Three sayings of one fact. The old comment earned its place by naming a distinction (403 vs. in-progress) that no longer exists; the new one names nothing the identifiers do not.

**2. [`api/ftp/ftp_test.go:779`] `TestPollTransferStatus_Timeout` and `TestPollTransferStatus_TimesOut` overlap, and the slow one is the package's slowest test** (pre-existing, no inline comment—the line is outside this diff)

Measured on this branch:

```
--- PASS: TestPollTransferStatus_FatalErrorNoRetry    (0.00s)
--- PASS: TestPollTransferStatus_TimesOut             (0.00s)
--- PASS: TestPollTransferStatus                      (0.00s)
--- PASS: TestPollTransferStatus_BacksOffThenSucceeds (0.75s)
--- PASS: TestPollTransferStatus_Timeout              (2.00s)
```

Both drive a server that only ever answers `success: null` and assert a timeout error; `_TimesOut` does it in 0.00s and additionally pins the poll count. The only thing `_Timeout` covers on its own is "timeout after several polls", which a ~1s budget keeps. Two seconds of a 3.478s package, returned to every CI run—and it is the same cleanup this PR is already doing one file over.

**3. [`api/ftp/ftp.go:129`] The unmarshal failure returns a message that is always empty** (pre-existing; the line only moved)

When `json.Unmarshal` fails, `statusResp` is the zero value, so the caller receives `""`. Returning `""` outright would say what actually happens instead of implying a partial parse is being salvaged. Not introduced here; worth a line while the hunk is open.

### Non-blocking

`PollTransferStatus` carries its own pacing (`backoffDelay`, `nextPollInterval`, `alignedPollDelay`) while `CLAUDE.md` states `utils/poll.go` is "the one place that paces the CLI's wait loops" with exactly one carve-out—the two MFA step-up polls—and this poller is not on that list. The practical consequence is visible right in this diff: with the error arm now unconditional, a single 429 mid-transfer ends the wait, where `utils.ThrottleBudget` / `utils.WarnThrottled` exist precisely for that. `api/ftp` already uses `utils.IsRetryLaterStatus` in the download path (`ftp.go:580`) but not here. Entirely out of this PR's scope—worth an issue.

### Highlights

- The PR did not accept #450's "dead code" framing; it traced `parseAPIErrorPayload`, reproduced the old body against the real client, and corrected the issue. That is the difference between deleting a branch and knowing what you deleted.
- The assumption the change rests on is written down along with what breaks if it stops holding, instead of being left implicit.
- Cleanup is complete: `TransferErrorResponse` and `webftp_transfer_in_progress` have zero references repo-wide afterwards, and `strings` is still used at seven other sites so the import correctly stays.
- Idiomatic Go after the change—error returns early, no `else` after a returning `if`.
- The coverage argument checks out: the deleted test's only unique leg was the removed branch, and success / multi-poll backoff / fatal error / timeout all remain covered.

---

**Verdict**: COMMENTED

**Checklist**:
- [x] Code follows stack conventions—`gofmt -l api/ftp` clean, `go vet ./api/ftp/...` clean
- [x] Tests included and passing—`go test -race -shuffle=on ./api/ftp/` -> `ok 3.478s`
- [x] No security vulnerabilities—no credentials, no injection surface; the change removes a string comparison
- [ ] Documentation updated if needed—see Warning 1: no minimum server version is recorded anywhere in the repo

api/ftp/ftp.go
jisung-02: You are right that the reason I gave Copilot does not hold. `utils.ParseErrorResponse` reads the code the client already parsed, so a version-tolerant guard would match no prose at all, and "retaining it keeps the CLI reading error prose" was true of the old shape only. Thanks for driving the real client to show it rather than arguing it.

The removal stands anyway, on the maintainer's confirmation that every deployed server is at 2.36.0 or newer—so this is the second path you named, not the first. Your condition on that path is met in 7823b382: `README.md` now states the minimum version beside the `cp` documentation, together with the failure it predicts.

> `alpacon cp` requires Alpacon Server 2.36.0 or later, both for Alpacon Cloud workspaces and self-hosted workspaces. Below that version, it fails on its first transfer-status poll instead of waiting for the transfer to finish.

The PR body no longer claims the assumption lives in the PR body. It says where it lives and records your guard as the option that was considered and why it was not taken.

api/ftp/ftp.go
jisung-02: Applied in d849c529. You are right that `statusResp` is the zero value on an unmarshal failure, so the caller always received `""` while the code implied a partial parse was being salvaged. It now returns `""` outright.

api/ftp/ftp_test.go
jisung-02: Applied in 2c5e896f, and deleted rather than reworded. You are right that the old comment earned its place by naming the 403 versus in-progress distinction, and that distinction is gone—the test name and the `"fatal error must not be retried"` assertion already carry the whole fact.

jisung-02:
Thanks for reproducing the pre-2.36.0 body against the real client instead of taking the PR body's word for it—that is what settled whether #450's premise or this PR's was the right one.

| Item | Where it landed |
|---|---|
| Warning 1—the stated reason for dropping compatibility does not hold | Accepted. The removal stands on the fleet-upgraded confirmation, and the minimum version is now in `README.md` (7823b382). The PR body records your `utils.ParseErrorResponse` guard as the option considered and why it was not taken. |
| Suggestion 1—the rewritten test comment restates its identifiers | 2c5e896f, deleted rather than reworded |
| Suggestion 2—`_Timeout` and `_TimesOut` overlap, and the slow one costs 2s | c7284f3a, budget cut from 3s to 1s. `api/ftp` went from 3.478s to 2.380s |
| Suggestion 3—the unmarshal failure returns an always-empty message | d849c529, returns `""` |
| Non-blocking—`PollTransferStatus` paces itself, and one 429 now ends the wait | #458 |

On the documentation checkbox you left unticked: `README.md` now carries the sentence below beside the `cp` documentation, which is the home the assumption lacked.

> `alpacon cp` requires Alpacon Server 2.36.0 or later, both for Alpacon Cloud workspaces and self-hosted workspaces. Below that version, it fails on its first transfer-status poll instead of waiting for the transfer to finish.

Verification on the pushed head, `go version go1.26.6 darwin/arm64`:

```
$ go build ./...
$ gofmt -l api/ftp
$ go vet ./...
(all three clean, no output)

ok  	github.com/alpacax/alpacon-cli/api/ftp	2.380s
...
ok  	github.com/alpacax/alpacon-cli/utils	1.798s
(44 packages, all ok)
```

PR #457, opened 2026-09-18
fix(test): validate sample CLI script syntax

### LINKED ISSUE #455: sample_test_cli.sh cannot run because cleanup command has an unterminated quote
## Summary

`sample_test_cli.sh` has an unterminated double quote in its cleanup section, so Bash rejects the script before any live end-to-end test can run.

## Reproduction

From the repository root:

```bash
bash -n sample_test_cli.sh
```

Actual output:

```
sample_test_cli.sh: line 369: syntax error near unexpected token `('
sample_test_cli.sh: line 369: `echo "4. MFA authentication (if configured):"'
```

The command exits with status `2`.

## Root cause

At `sample_test_cli.sh:340`, the cleanup command starts a double-quoted path but never closes it:

```bash
rm -f "$LOCAL_PATH/$TEST_FILE
```

Bash continues parsing the following lines as part of that string and reports the later `echo` at line 369 as the unexpected token.

## Impact

The README directs users to copy, configure, and run `sample_test_cli.sh` for live workspace end-to-end testing. Because the script fails shell parsing, that documented test path cannot start. CI does not run a `bash -n sample_test_cli.sh` check, so this syntax error can remain undetected.

## Expected behavior

`sample_test_cli.sh` passes `bash -n` and can be run after the documented workspace configuration.

## Acceptance criteria

- [ ] Close the quote in the cleanup command at `sample_test_cli.sh:340`.
- [ ] `bash -n sample_test_cli.sh` exits with status `0`.
- [ ] CI runs `bash -n sample_test_cli.sh` so future shell syntax errors fail the check.

jisung-02:
## Summary

`sample_test_cli.sh` could not run at all: the cleanup command at line 340 opened a double quote it never closed, so Bash swallowed the rest of the file as one string and the script failed to parse. Closing the quote restores it.

Nothing caught this, because no job ever parsed the script. `build-and-test.yaml` now runs `bash -n sample_test_cli.sh` right after checkout, which parses without executing, so the next unterminated quote fails the check in seconds instead of reaching a user.

## Changes

- `sample_test_cli.sh`: close the quote in `rm -f "$LOCAL_PATH/$TEST_FILE"`.
- `sample_test_cli.sh`: quote the 8 unquoted `$SERVER_NAME` expansions in `cleanup()` and the section 8 remote cleanup (shellcheck SC2086)—a `SERVER_NAME` containing a space would word-split there. The expansions inside `run_test`'s command strings are unaffected; they are double-quoted at the outer level and re-split by `bash -c` on purpose.
- `.github/workflows/build-and-test.yaml`: add a `Verify the sample CLI script parses as Bash` step directly after `Check out code`, guarded by `runner.os == 'Linux'`. It needs no Go toolchain, so it runs before `Set up Go` and `go mod tidy` and fails a broken script before paying for toolchain setup. The matrix runs three legs (ubuntu, macos, windows) and the parse result is the same on each, so one leg is enough.

The gate stays bound to `sample_test_cli.sh` rather than generalizing to every `*.sh` file—it is the only shell script in the repo today, and a second one can widen the gate when it lands.

Left two pre-existing issues in `sample_test_cli.sh` untouched, out of scope for this PR: section 8's local/remote cleanup duplicates the `EXIT` trap, and the trap reads `$TEST_FOLDER` before it is assigned if the script exits early.

## Verification

```
$ bash -n sample_test_cli.sh; echo "exit=$?"
exit=0
```

Closes #455


.github/workflows/build-and-test.yaml
```diff
@@ -65,6 +65,10 @@ jobs:
           go mod tidy
           git diff --exit-code go.mod go.sum
 
+      - name: Sample CLI script parses as Bash
+        if: runner.os == 'Linux'
+        run: bash -n sample_test_cli.sh
```
geunwoonoh: Suggestion: generalize the gate, and match the sibling step names.

Two small things on this step.

**The gate is bound to one filename.** `git ls-files '*.sh'` returns exactly one file today, so this is correct right now—but the PR's own framing is "nothing caught this, because no job ever parsed the script," and the second script to land escapes the gate in exactly the same way. The generalized form costs the same milliseconds.

**The step name is declarative where its siblings are imperative.** `Check out code`, `Set up Go`, `Verify go.mod and go.sum are tidy`, `Build`—all verb-first. Folded into the suggestion below.

```suggestion
      - name: Verify shell scripts parse as Bash
        if: runner.os == 'Linux'
        run: git ls-files -z '*.sh' | xargs -0 -r -n1 bash -n
```

Separately, and not expressible as a suggestion block: this step needs no Go toolchain, but sits after `Set up Go` and `go mod tidy`. Moving it directly after `Check out code` fails a broken script in seconds rather than after toolchain setup—no cost on the happy path, it only shortens the failure path.

sample_test_cli.sh
```diff
@@ -337,7 +337,7 @@ echo "         8. CLEANUP AND SUMMARY"
 echo "=========================================="
 
 # Clean up additional test files
-rm -f "$LOCAL_PATH/$TEST_FILE
+rm -f "$LOCAL_PATH/$TEST_FILE"
```
geunwoonoh, line 340: Suggestion: the fix is right—notes on what this gate still cannot see.

I reproduced the pre-fix failure against `c750cfc^` and got the same `line 369` error the issue reports, with `exit=2`; after the fix, `exit=0`. Closing the quote is unambiguously the intended line, given lines 341-342 share its shape.

Three follow-ups for this file, none of them blocking this PR:

**`shellcheck` catches the class `bash -n` cannot.** It is pre-installed on `ubuntu-latest` and would flag the 8 genuinely unquoted `$SERVER_NAME` expansions at lines 69-70, 73-74, and 348-351 (SC2086)—a `SERVER_NAME` containing a space word-splits there. The expansions inside `run_test` string arguments are not in this class: they are double-quoted at the outer level and re-split by `bash -c` at runtime, which is the function's intent.

**This section duplicates the EXIT trap.** `trap cleanup EXIT` is armed at line 78 and always fires. Line 342 is byte-identical to `cleanup()` line 62, line 345 to line 65, and the remote cleanup at 348-351 overlaps lines 69-74. Every one of those commands runs twice on a successful run; the local and remote cleanup here could be deleted and the trap would still cover it.

**The trap reads `TEST_FOLDER` before it exists.** `cleanup()` uses `$TEST_FOLDER` at lines 65, 73, and 74, the trap is armed at line 78, but `TEST_FOLDER` is not assigned until line 127. Any exit before then—the `command -v alpacon` guard, the `WORKSPACE_URL` guard, a failed login—runs `rm -rf "$LOCAL_PATH/downloaded_user_"`. Harmless with the placeholder path, but it is the same class of latent breakage this PR is about, and `bash -n` will never see it.

geunwoonoh (dismissed):
## Summary

A one-character fix with a CI gate behind it. The quote at `sample_test_cli.sh:340` is closed, and a `bash -n` step is added so the same class of error cannot land silently again.

I reproduced the failure against `c750cfc^`:

```
$ bash -n before.sh
before.sh: line 369: syntax error near unexpected token `('
before.sh: line 369: `echo "4. MFA authentication (if configured):"'
exit=2

$ bash -n sample_test_cli.sh    # after
exit=0
```

Identical to the error in #455, down to the line number. The fix is correct and the gate would have caught it.

Gating verified: the step lives in the `test` matrix job, and `build-and-test` requires `needs.test.result == 'success'`, so a parse failure on the Linux leg fails the required check. `fail-fast: false` keeps the other legs reporting. The workflow's `paths-ignore` only excludes `**.md` and `LICENSE`, so `.sh` edits do trigger it. I also confirmed the script parses under bash 3.2 (macOS system bash), so the Linux-only guard is not hiding a portability gap today.

## Critical issues (0)

None.

## Warnings (0)

None blocking.

## Suggestions (4)

1. **[.github/workflows/build-and-test.yaml:68-70]** The gate is bound to one filename. `git ls-files '*.sh'` returns exactly one file today, so this is correct now—but the PR's own framing is "nothing caught this, because no job ever parsed the script," and the second script to land escapes the gate the same way. The generalized form costs the same milliseconds:

   ```yaml
   - name: Shell scripts parse as Bash
     if: runner.os == 'Linux'
     run: git ls-files -z '*.sh' | xargs -0 -r -n1 bash -n
   ```

2. **[.github/workflows/build-and-test.yaml:68]** The step needs no Go toolchain, but sits after `Set up Go` and `go mod tidy`. Moving it directly after `Check out code` fails a broken script in seconds rather than after toolchain setup. No cost on the happy path; it only shortens the failure path.

3. **[sample_test_cli.sh]** Follow-up, not this PR: `bash -n` proves the file parses and nothing more. `shellcheck` is pre-installed on `ubuntu-latest` and would flag the 8 genuinely unquoted `$SERVER_NAME` expansions at lines 69-70, 73-74, and 348-351 (SC2086)—a `SERVER_NAME` containing a space word-splits there. The expansions inside `run_test` string arguments are not in this class; they are double-quoted at the outer level and re-split by `bash -c` at runtime, which is the function's intent.

4. **[.github/workflows/build-and-test.yaml:68]** Nit: sibling step names are verb-first imperative (`Check out code`, `Set up Go`, `Verify go.mod and go.sum are tidy`). `Sample CLI script parses as Bash` is declarative. `Verify the sample CLI script parses as Bash` matches the file's own convention.

## Non-blocking observations (pre-existing, outside the diff)

These are real but predate the PR and should not hold it up.

- **Section 8 duplicates the EXIT trap.** `trap cleanup EXIT` is armed at line 78 and always fires. Line 342 is byte-identical to `cleanup()` line 62, line 345 to line 65, and the remote cleanup at 348-351 overlaps lines 69-74. Every one of those commands runs twice on a successful run. Section 8's local and remote cleanup could be deleted outright and the trap would cover it.
- **The trap reads `TEST_FOLDER` before it exists.** `cleanup()` uses `$TEST_FOLDER` at lines 65, 73, and 74; the trap is armed at line 78; `TEST_FOLDER` is not assigned until line 127. Any exit before then—the `command -v alpacon` guard, the `WORKSPACE_URL` guard, a failed login—runs `rm -rf "$LOCAL_PATH/downloaded_user_"`. Harmless with the placeholder path, but it is the same class of latent breakage this PR is about, and `bash -n` will never see it.
- **`README.md:433`** says the script exercises "server lookup, exec, websh, cp, tunnel". There is no tunnel test in the script.

## Go conventions, duplication, performance

- **Go conventions**: no Go code in this diff. `go vet`, golangci-lint, declaration order, and the errcheck patterns are all unaffected. Null result, not an oversight.
- **Duplication in the diff**: none. The duplication is the pre-existing cleanup overlap noted above.
- **Performance**: `bash -n` is a parse-only pass over a 370-line file on one matrix leg—unmeasurable against the Go build and `-race` test run. Suggestion 2 is the only timing change, and only on the failure path.

## Highlights

- The fix is minimal and matches the root cause exactly, with no incidental reformatting of a file that was ripe for it.
- Shipping the CI gate in the same PR as the fix addresses why the bug survived, not just the bug. That is the right instinct.
- The `if: runner.os == 'Linux'` guard follows the existing `go mod tidy` step's convention, and the PR body states the reasoning for one leg rather than leaving it to be inferred.
- The PR body carries the verification command and its output.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing—the CI gate is the test; I reproduced its pass and fail
- [x] Documentation updated if needed—none required

.github/workflows/build-and-test.yaml
jisung-02: Moved the step right after checkout in 5108dc0—agreed, no reason to pay for toolchain setup before finding out the script doesn't parse. Also renamed it to `Verify the sample CLI script parses as Bash`, matching the imperative sibling names.

Left the filename-generalization part as-is: `sample_test_cli.sh` is the only shell script in the repo today, and generalizing now would be speculative. Happy to widen it with `git ls-files -z '*.sh' | xargs -0 -r -n1 bash -n` the day a second script lands.

sample_test_cli.sh
jisung-02: Quoted the 8 `$SERVER_NAME` expansions in `cleanup()` and the section 8 remote cleanup in 609f746—confirmed the SC2086 read: they're unquoted and would word-split a `SERVER_NAME` with a space. Left the `run_test` command-string ones alone, since those are correctly double-quoted at the outer level and meant to be re-split by `bash -c`.

Left the EXIT-trap duplication and the trap reading `$TEST_FOLDER` before it's assigned untouched—both predate this branch and are out of scope here. No follow-up issue filed yet.

jisung-02:
Both inline items from geunwoonoh's review are addressed:

| Item | Commit | Note |
|---|---|---|
| Move the parse step before toolchain setup, rename to imperative | 5108dc0 | Applied as suggested |
| Generalize the gate to every `*.sh` file | — | Not applied—only one shell script exists today; see reply |
| Quote the 8 unquoted `$SERVER_NAME` expansions (SC2086) | 609f746 | Applied as suggested |
| EXIT-trap duplication in section 8 | — | Not applied, pre-existing, out of scope |
| Trap reads `$TEST_FOLDER` before assignment | — | Not applied, pre-existing, out of scope |

Verification, re-run on the pushed head:

```
$ bash -n sample_test_cli.sh; echo "exit=$?"
exit=0
```

.github/workflows/build-and-test.yaml
```diff
@@ -51,6 +51,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      - name: Verify the sample CLI script parses as Bash
+        if: runner.os == 'Linux'
+        run: bash -n sample_test_cli.sh
```
geunwoonoh, line 56: Verified: the step sits directly after `Check out code` and ahead of `Set up Go`, and `grep -c 'bash -n sample_test_cli.sh'` over the file returns `1`—no leftover at the old position. On the ubuntu leg of `609f7467` it ran as step 3 and succeeded, so moving it earlier did not cost it the gate: the job and the `needs.test.result == 'success'` check are unchanged, so a parse failure still fails the required `build-and-test`.

The name now matches the imperative siblings.

On leaving the gate bound to this one filename—agreed, and no need to revisit. One `.sh` file in the repo makes generalizing speculative, and `git ls-files -z '*.sh' | xargs -0 -r -n1 bash -n` is recorded in both your reply and the PR body for whoever adds the second script.

sample_test_cli.sh
```diff
@@ -337,18 +337,18 @@ echo "         8. CLEANUP AND SUMMARY"
...
```
geunwoonoh, line 340: Verified: all 8 unquoted `$SERVER_NAME` expansions are quoted (lines 69-70, 73-74, 348-351), and none remain at command position. The quote fix on this line still holds—`bash -n` exits `0` under both bash 5 and bash 3.2.

Good call leaving the `run_test` command strings alone. Those are double-quoted at the outer level and meant to be re-split by `bash -c`, so quoting them would have broken the function. Applying an SC2086 sweep without that distinction is the usual way this fix goes wrong.

One follow-up: I agree the `EXIT` trap duplication and the `$TEST_FOLDER` ordering are out of scope here, but your reply notes no issue is filed yet. Once this merges, the PR body is the only place either is written down. Worth opening one issue to hold them.

geunwoonoh (approved):
Re-review at `609f7467`. Three of the four suggestions are applied, one is declined with sound reasoning. I verified the applied three against the files and the CI logs rather than taking the replies at face value.

## Summary

| # | Suggestion | Status | Verification |
|---|---|---|---|
| 1 | Generalize the gate to every `*.sh` | **Declined** | Reasoning sound, decision recorded in the PR body |
| 2 | Move ahead of `Set up Go` | Applied (`5108dc0`) | Directly after `Check out code`, zero leftovers at the old position |
| 3 | Quote the 8 SC2086 `$SERVER_NAME` expansions | Applied (`609f746`) | No unquoted ones remain; `run_test` strings untouched |
| 4 | Imperative step name | Applied (`5108dc0`) | `Verify the sample CLI script parses as Bash` |

Evidence:

```
$ bash -n new.sh            # bash 5
exit=0
$ /bin/bash -n new.sh       # bash 3.2, macOS system bash
exit=0
$ grep -cE '^\s*alpacon (exec|websh)( -u root)? \$SERVER_NAME' new.sh
none remaining
$ grep -c 'bash -n sample_test_cli.sh' new.yaml
1
```

CI is green on `609f7467`, and the gate actually ran—step 3 on the ubuntu leg, `Verify the sample CLI script parses as Bash -> success`. Moving it ahead of `Set up Go` did not cost it the gate: `build-and-test` still fails on it, since the job and the `needs.test.result` check are unchanged.

The delta since my first review is 8 line pairs in the script plus the workflow move. No drive-by changes.

## Critical issues (0)

None.

## Warnings (0)

None.

## Suggestions (2)

1. **[sample_test_cli.sh]** I agree with leaving the two pre-existing problems out of scope—section 8 duplicating the `EXIT` trap, and the trap reading `$TEST_FOLDER` before it is assigned. But the reply says "No follow-up issue filed yet," and once this merges the PR body is the only record of them. Worth opening one issue so they do not evaporate.

2. **[README.md:433]** The one item from my first review that drew no response either way. The README says the script exercises "server lookup, exec, websh, cp, tunnel", but `grep -c 'alpacon tunnel'` on the script at this head returns `0`. Pre-existing and outside the diff, so it does not block—either add a tunnel test or drop the word.

## On the declined suggestion

Declining the glob generalization is the right call. With one `.sh` file in the repo, generalizing now is speculative, and the widening command is written down in both the reply and the PR body for whoever adds the second script. The decision is recorded rather than silently dropped, which is what matters. Not raising it again.

## Highlights

- Three suggestions split across two commits, each holding one concern—the CI change and the script change did not get mixed together.
- The SC2086 fix was applied with judgment, not mechanically: the expansions inside `run_test` command strings were correctly left alone, because they are double-quoted at the outer level and meant to be re-split by `bash -c`. That distinction was read, not copied.
- The PR body was updated to match the code. The step's position and name are described accurately, and the declined suggestion and the deferred items are written down, so the description has not gone stale.
- What was declined was declined with a reason.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing—CI green, gate step confirmed to have executed
- [x] Documentation updated if needed—PR body updated; the README drift is pre-existing

.github/workflows/build-and-test.yaml
jisung-02: Thanks for verifying. Agreed on keeping the gate bound to the one file for now—will widen it with the `git ls-files -z '*.sh' | xargs -0 -r -n1 bash -n` form the day a second script lands.

sample_test_cli.sh
jisung-02: Filed #459 for the EXIT-trap duplication and the TEST_FOLDER-before-assignment read, so both are tracked past this PR.

